PR #6, opened 2017-09-26
Documents API Fields & Filter/Sort URL Parameters

anglinb:
:wave: This PR documents all the fields coming back from the API and adds docs on how to use the URL parameters to limit the results returned from the API.


owocki:
🙌

PR #9, opened 2017-09-28
Fix instruction for running locally

cifvts:
Adjust some paths according to actual source tree. Remove `sync_db` command,
deprecated from Django 1.9. Add comand to create Django cache table in Postgres


owocki:
🙌 thanks @cifvts

PR #11, opened 2017-10-01
Fix space, typo and rename

gamwe6:


owocki:
is `LISCENSE` the standard, instead of `LISCENSE.txt`?

gamwe6:
You can use both https://www.apache.org/dev/apply-license.html#license-file-name.
Just to be consistent with README.md, I've uppercased it. If you want, we can let it lowercased.
Which leads to another question? What do you think of a CONTRIBUTING.md file (for coding style, naming conventions, etc.)?

gamwe6:
Do you need these extra spaces?

owocki:
> What do you think of a CONTRIBUTING.md file (for coding style, naming conventions, etc.)?

All for it.  Want to PR up a quick draft and I'll contribute in with anything I want to specify?

>Do you need these extra spaces?
No, thanks.

PR #39, opened 2017-10-30
Standard pseudo-random generators are not suitable for security/cryptographic purposes.

sergio-alonso:
class random.SystemRandom([seed])

Class that uses the os.urandom() function for generating random numbers from sources provided by the operating system.

#38

### DIFF
--- a/app/marketing/management/commands/post_quote.py
+++ b/app/marketing/management/commands/post_quote.py
@@ -42,7 +42,7 @@ class Command(BaseCommand):
         "Software Dev QOTD",
         "Software Development Quote",
         ]
-        intro = random.SystemRandom(intros)
+        intro = random.SystemRandom().choice(intros)
 
 
         #setup
--- a/app/marketing/management/commands/post_quote.py
+++ b/app/marketing/management/commands/post_quote.py
@@ -42,7 +42,7 @@ class Command(BaseCommand):
         "Software Dev QOTD",
         "Software Development Quote",
         ]
-        intro = random.choice(intros)
+        intro = random.SystemRandom(intros)
 
 
         #setup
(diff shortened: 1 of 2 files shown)


owocki:
Thanks @sergio-alonso :) -- none of the PRed files are actually involved in security/cryptographic purposes, but I can see why it's important to have discussion on this point.  where do you think this issue falls on the OWASP model?

sergio-alonso:
This PR is related to security by default and with establishing minimum levels of security, if there is a safer alternative you should opt for it.

* The OWASP includes it as a vulnerability

[OWASP Insecure Randomness](https://www.owasp.org/index.php/Insecure_Randomness)

> Many PRNGs are completely predictable, though their outputs are statistically random. A pseudo-random number generator may be operated in a secure way if it is regularly "seeded" with enough unpredictable entropy. Most popular pseudo-random number generators are not secure.

* The SANS Institute recently listed "Weak random number generators" as one of the 7 most dangerous security techniques to watch in 2017:

[SANS Weak Random Number Generators](https://www.sans.org/the-seven-most-dangerous-new-attack-techniques)

* The CWE list it as a common software security weaknesses:

[CWE-338: Use of Cryptographically Weak Pseudo-Random Number Generator (PRNG)](https://cwe.mitre.org/data/definitions/338.html)

app/marketing/management/commands/post_quote.py
```diff
@@ -42,7 +42,7 @@ def handle(self, *args, **options):
         "Software Dev QOTD",
         "Software Development Quote",
         ]
-        intro = random.choice(intros)
+        intro = random.SystemRandom(intros)
```
owocki: this won't work.  `intros` is of type `list`

```
>>> random.choice(['a','b','c'])
'c'
>>> random.SystemRandom(['a','b','c'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

owocki:
thanks, thats good context

PR #41, opened 2017-11-03
HTTP Strict Transport Security

sergio-alonso:
HTTP Strict Transport Security (HSTS) is an opt-in security enhancement that is specified by a web application through the use of a special response header. Once a supported browser receives this header that browser will prevent any communications from being sent over HTTP to the specified domain and will instead send all communications over HTTPS. It also prevents HTTPS click through prompts on browsers.

[OWASP HSTS](https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet)

[OWASP HSTS test](https://www.owasp.org/index.php/Test_HTTP_Strict_Transport_Security_(OTG-CONFIG-007))

[Django HSTS](https://docs.djangoproject.com/en/1.11/ref/middleware/#http-strict-transport-security)

[Django HSTS settings](https://docs.djangoproject.com/en/1.11/ref/settings/#secure-hsts-include-subdomains)

#38

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -179,6 +179,14 @@ CACHES = {
     }
 }
 
+# HTTP Strict Transport Security
+# https://docs.djangoproject.com/en/1.11/ref/middleware/#http-strict-transport-security
+SECURE_HSTS_INCLUDE_SUBDOMAINS = True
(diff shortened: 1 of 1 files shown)


owocki:
Thank you!

PR #83, opened 2017-12-05
Fix grammer in alert on tip page

edkek:
# Changes

* Modified alert to display proper grammer when user enters a value below the minimum on tip page

# Files Changed

* `send.js`


owocki:
thanks @edkek !

Issue #73, opened 2017-12-04
support ERC20 token conversion rates on tips/bounties funding


owocki:
A little while ago i [wrote some code](https://github.com/gitcoinco/web/commit/96c3613907d67f01104c37039fc292c6f8f63811) that would estimate the USD amount of a new bounty as the user filled out the form.

here is an example of what it looks like 

![screen recording 2017-11-29 at 12 21 pm](https://user-images.githubusercontent.com/513929/33394650-41a2b9a8-d500-11e7-9883-6de55428c239.gif)

The scope of this task is to update gitcoin.co/tip and gitcoin.co/funding/new to support this estimation ability for all ERC20 tokens for which a conversion rate to USD is available from the backend.

gitcoinbot:
__This issue now has a funding of 0.05 ETH (23.0 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/73).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/73)

rajatkapoor:
On it

owocki:
great @rajatkapoor !

owocki:
heres an example => https://gitcoin.co/sync/get_amount?amount=100&denomination=GRID

some places to start in coddebase:

https://github.com/gitcoinco/web/blob/master/app/dashboard/helpers.py#L39-L57

https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/shared.js#L243-L280

rajatkapoor:
Also, I see that this is already present on the new funding flow, that is why I left it off as it is. 
Please have a look whenever you get time and confirm 👍

owocki:
@rajatkapoor the estimations only work for ETH on the funding flow and don't work for ERC20 tokens.

the last comment i see on https://github.com/gitcoinco/web/pull/84 is you saying "will make the requested changes" but i dont see any more commits on the PR.  am i missing them?

tra38:
I'm willing to be on the waitlist for this project. If @rajatkapoor cannot finish this, I'm willing to take it over from where he stopped.

rajatkapoor:
I will push in the changes very soon. Sorry for the delay

owocki:
🎉 shipped.  thanks @rajatkapoor

gitcoinbot:
[...]

PR #84, opened 2017-12-05
Tips now show USD estimates

rajatkapoor:
implements #73

### DIFF
--- a/app/assets/v2/js/amounts.js
+++ b/app/assets/v2/js/amounts.js
@@ -1,0 +1,31 @@
+var getUSDEstimate = function (amount, denomination) {
+    if (document.conversion_rates && document.conversion_rates[denomination]) {
+        var usd_amount = amount / document.conversion_rates[denomination];
+    } else {
+        var request_url = '/sync/get_amount?amount=' + amount + '&denomination=' + denomination;
+        jQuery.get(request_url, function (result) {
+            var usd_amount = result['usdt'];
+            var conv_rate = amount / usd_amount;
+            //store conv rate for later in cache
+            if (typeof document.conversion_rates == 'undefined') {
+                document.conversion_rates = {}
+            }
+            document.conversion_rates[denomination] = conv_rate;
+        }).fail(function () {
+            return None
+        });
+    }
+    if (usd_amount) {
+        usd_amount = Math.round(usd_amount * 100) / 100;
+        if (usd_amount > 1000000) {
+            usd_amount = Math.round(usd_amount / 100000) / 10 + "m"
+        } else if (usd_amount > 1000) {
+            usd_amount = Math.round(usd_amount / 100) / 10 + "k"
+        }
+        return ('Approx: ' + usd_amount + ' USD');
+    } else {
+        return null;
+    }
+
+
+};
--- a/app/assets/v2/js/amounts.js
+++ b/app/assets/v2/js/amounts.js
@@ -7,40 +7,39 @@ var estimate= function(amount, conv_rate) {
         } else if (estimateAmount > 1000) {
             estimateAmount = Math.round(estimateAmount / 100) / 10 + "k"
         }
-        return ('Approx: ' + estimateAmount + ' USD');
+        return 'Approx: ' + estimateAmount + ' USD';
     } else {
-        return null;
+        return 'Approx: Unknown amount';
     }
 };
 
-var getUSDEstimate = function (amount, denomination) {
+var getUSDEstimate = function (amount, denomination, callback) {
     var conv_rate;
-    var usd_amount;
+    var eth_usd;
     var eth_amount;
     try {
         amount = parseFloat(amount);
     } catch (e) {
-        return null;
+        return "Incorrect amount";
     }
 
     if (document.conversion_rates && document.conversion_rates[denomination]) {
         conv_rate = document.conversion_rates[denomination]
-        return estimate(amount, conv_rate)
+        return callback(estimate(amount, conv_rate))
     } else {
         var request_url = '/sync/get_amount?amount=' + amount + '&denomination=' + denomination;
         jQuery.get(request_url, function (result) {
-            usd_amount = result['usdt'];
+            eth_usd = result['usdt'];
             eth_amount = parseFloat(result['eth']);
-            conv_rate = (amount/eth_amount)*usd_amount;
+            conv_rate = (amount/eth_amount)*eth_usd;
             //store conv rate for later in cache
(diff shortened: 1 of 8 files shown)


app/assets/v2/js/amounts.js
```diff
@@ -0,0 +1,31 @@
+var getUSDEstimate = function (amount, denomination) {
+    if (document.conversion_rates && document.conversion_rates[denomination]) {
+        var usd_amount = amount / document.conversion_rates[denomination];
+    } else {
+        var request_url = '/sync/get_amount?amount=' + amount + '&denomination=' + denomination;
+        jQuery.get(request_url, function (result) {
+            var usd_amount = result['usdt'];
+            var conv_rate = amount / usd_amount;
```
owocki: i think there was a mix up about what this endpoint returns.  it returns `usd_amount` which is the eth/usd conv rate, and `eth_amount` which is the denomination/eth conv rate

this should be `conv_rate = amount / usd_amount / eth_amount`

owocki:
per  https://github.com/gitcoinco/web/issues/73 `The scope of this task is to update gitcoin.co/tip and gitcoin.co/funding/new`

can you please update the `gitcoin.co/funding/new` portion too?

owocki:
left two comments. otherwise looks really good!

app/dashboard/templates/yge/send2.html
```diff
@@ -1,7 +1,9 @@
 {% extends 'yge/base.html' %}
-
+{% load staticfiles %}
 {% block 'scripts' %}
 
+{#<script src="{% static 'v2/js/jquery.js' %}"></script>#}
```
owocki: should this be commented out?

app/assets/v2/js/amounts.js
thelostone-mc: @rajatkapoor Could we replace `var` with `let` / `const` from `ES6` ?

rajatkapoor: Hi.
No we can't do that as this is pure client side JS with no babel transpilers in between. If we use ES6/7 then we will mis function on a lot of browsers which do not have es6/7 implemented

rajatkapoor:
@owocki will make the requested changes

owocki:
thanks :) looking fwd to getting this live

app/assets/v2/js/amounts.js
thelostone-mc: Ah I assumed that there was basic support for `let`, `const` & `arrow functions` in almost all the browsers. I usually follow [this](http://kangax.github.io/compat-table/es5/).

Bad idea to assume developers who visit this site -> don't use legacy browsers 😓 

Also is it worth considering integrating babel transpilers ?

rajatkapoor: @owocki I see this endpoint returns two things: 
```
            'eth': amount_in_eth,
            'usdt': amount_in_usdt
```
According to which this code looks fine. Let me know if I am missing something somewhere

rajatkapoor: @thelostone-mc We will be missing a lot on a lot of users on android/webview users. 
For such compatibility check, I use: https://caniuse.com/
Also, moving to es6 will be good idea if we plan to move full fledged. Then, I would also like to pitch in to use typescript in order to make the development process even better :)

app/dashboard/templates/yge/send2.html
rajatkapoor: yes, it is commented out intentionally. I will probably remove it out in the next commit

owocki:
the endpoint still shows the ETH conversion rate even when "EDG" or any ERC20 token is selected in the dropdown.  http://bits.owocki.com/3v1r2I2Q0m2m/Screen%20Shot%202017-12-11%20at%209.29.24%20AM.png

this is incorrect

app/assets/v2/js/amounts.js
owocki: @rajatkapoor 

the UI still shows the ETH conversion rate even when "EDG" or any ERC20 token is selected in the dropdown.  http://bits.owocki.com/3v1r2I2Q0m2m/Screen%20Shot%202017-12-11%20at%209.29.24%20AM.png

this is incorrect.  you need use the returned responses from the endpoint as i described in https://github.com/gitcoinco/web/pull/84/files#r155141576 to make it work

owocki:
> can you please update the `gitcoin.co/funding/new` portion too?

This still needs to be turned around too

owocki:
lmk if rdy to test!

rajatkapoor:
Sure. Just integrating it in the new bounty also

rajatkapoor:
You can check it now

owocki:
seems to be working on the new bounty form but not on the 'new tip' form http://bits.owocki.com/0Y3k2q0p0O2R/Screen%20Shot%202017-12-13%20at%2010.41.47%20AM.png

rajatkapoor:
Fixed now

owocki:
looks good now..

for bot the bounty/tip pages, can we make it so that if the selected currency changes, then the estimate updates?  right now you have to click in and out of the `amount` textbox to get the change to trigger

owocki:
looking at the code now

app/assets/v2/js/amounts.js
```diff
@@ -0,0 +1,48 @@
...
+            conv_rate = (amount/eth_amount)*usd_amount;
+            //store conv rate for later in cache
+            if (typeof document.conversion_rates == 'undefined') {
+                document.conversion_rates = {}
+            }
+            document.conversion_rates[denomination] = conv_rate;
+            return estimate(amount, conv_rate)
+
+        }).fail(function () {
+            return null
```
owocki: itd be great if, on error, the UI was smart enough to update to say "uknown amount" instead of just keeping the last value.

app/assets/v2/js/amounts.js
```diff
@@ -0,0 +1,48 @@
...
+    var eth_amount;
+    try {
+        amount = parseFloat(amount);
+    } catch (e) {
+        return null;
+    }
+
+    if (document.conversion_rates && document.conversion_rates[denomination]) {
+        conv_rate = document.conversion_rates[denomination]
+        return estimate(amount, conv_rate)
```
owocki: itd be great if, on error or missing data, the UI was smart enough to update to say "uknown amount" instead of just keeping the last value.

owocki:
left a few code comments

rajatkapoor:
On it

owocki:
this is awesome.  thank you @rajatkapoor

PR #85, opened 2017-12-05
PostgreSQL alpine and docker compose restart policies

mbeacom:
The goal of this PR is to modify the docker compose configuration to use the PostgreSQL `10.1` Alpine image opposed to the standard PostgreSQL `latest` image.

This decreases the on disk size of the PostgreSQL image from around 117 MB to 14 MB before migrations are executed.
I have pinned the postgresql version to 10.1 to ensure a future upgrade can be vetted before potentially breaking the dev environment.

Additionally, I have modified the `restart` policy for both web and db to `unless-stopped` which will instruct both the postgresql and app containers to restart on failure, but remain stopped if the container is manually stopped for whatever reason.  This can be helpful during local development.

I removed the `ENV` declaration from the Dockerfile since we are assigning the environment variables when we launch the stack.


owocki:
i dont know enough about docker to properly comment here.  anyone else have feedback? 

@mbeacom lets chat again if we don't hear anything in the next day

Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM python:2.7-slim
-ENV PYTHONUNBUFFERED 1
```
abitrolly: Still needed if somebody needs to up container manually for debug.

abitrolly:
Looks good to me, except for unbuffered thing.

@owocki basically, without this commit, if you stop container manually, it will start again on reboot. Usually it is not that you want for development. YMMV.

mbeacom:
@abitrolly  While I agree with you, if they're starting it manually, they could `docker run -e PYTHONUNBUFFERED=1 ` and we'd avoid the layer.  I'll re-add it.

Dockerfile
mbeacom: 👍 That's reasonable!

jasonrhaas:
The `unless-stopped` keyword must be a new thing!  Haven't seen that one before.  I like it for the `postgres` container, but maybe not for the `web` container.  What happens if I'm doing development, write some bad code, and it crashes?  Will it just keep restarting constantly?  I would rather it just crash and then show me the trace back so I can fix it and then restart the container.

mbeacom:
@jasonrhaas It'll still restart the django app as you'd normally expect.  The behavior now is that django will halt restarting the app and log the stack trace (available via `docker-compose logs`).  Once you save the fixing change, it'll automatically restart the django app.  The behavior for that hasn't changed with this update.

The only thing this will change is say for instance, you know something you're going to do is going to indefinitely break something system-level.  You can `docker-compose stop web` and it won't attempt to start it until you manually start it.  The docker-level restart policy will only kick in if a non-zero exit code is encountered (so if the django app legitimately crashes), it'll restart the container.

jasonrhaas:
@mbeacom so I have a syntax error or something in the Django code, it will show the traceback, and then once I fix it, restart right?  Ok that works, I think Flask does that as well.  

I just want to make sure that the automatic restart on the `web` container doesn't hide error messages.  We used to have automatic restarts on `supervisord` processes, but the side effect to that was that it was hiding error messages that we should have been catching.

abitrolly:
`unless` you run `docker stop`, container will be continuously restarted. It doesn't care about your exit status. With `always` you need to additionally `docker rm` container, or it will pop up again on system/docker reboot.

mbeacom:
@abitrolly Is this good to go for you or would you like me to make adjustments?
Thanks for your involvement!

abitrolly:
It is good, but only @owocki can merge it.

mbeacom:
Thanks! I just wanted to make sure you were good with the changes.

owocki:
> only @owocki can merge it.

this will change soon :)

merging now!

PR #88, opened 2017-12-05
Fix display of plurals for some models in admin

mbeacom:
The goal of this PR is to properly display:
- `Matches`
- `Access codes`
- `Bounties`
- `Whitepaper access`
while navigating the django admin.


owocki:
👍

PR #93, opened 2017-12-06
Simplify bounty management command

mbeacom:
The goal of this PR is to simplify the refresh_bounties management command by eliminating the second queryset and save while adding docstrings.


app/economy/management/commands/refresh_bounties.py
```diff
@@ -31,20 +35,22 @@ def add_arguments(self, parser):
...
 
+        """
         all_bounties = Bounty.objects.all()
-        for b in all_bounties:
-            if not b.avatar_url:
-                b.avatar_url = b.get_avatar_url()
-            b.save()
-            print('2/ refreshed {}'.format(b.pk))
+        fetch_remote = options['remote']
+        for bounty in all_bounties:
```
owocki: unless im mistaken, i think we only want to iterate `current_bounties` here?

mbeacom: I'm performing the iteration only once. If it's a current bounty **and** remote is set to true, we fetch the issue.  It performs the same as previously, but only uses one QS versus iterating two and saving twice.

owocki: ah good call. carry on

owocki:
this LGTM.  if anyone else has comments let me know.  if not will likely merge tonight or next time someone pings me

Issue #103, opened 2017-12-09
mail: tip amount rounded off to 2 decimals showing cause incorrect amount on mail


thelostone-mc:
The tip amount is rounded off to two decimal places in the mail template when a tip is rewarded to a user. 

I was a given a tip of **0.075** 

_The email : (rounded off tip -> was a tad bit bummed out  )_ 

<img width="1151" alt="screen shot 2017-12-09 at 2 51 01 am" src="https://user-images.githubusercontent.com/5358146/33793482-a86c6676-dcde-11e7-9316-f96f1df07898.png">

_The redeem page shows the right amount : (realized it was just a template issue 😛 )_ 

<img width="585" alt="screen shot 2017-12-09 at 12 37 45 pm" src="https://user-images.githubusercontent.com/5358146/33793504-ad01c550-dcde-11e7-9896-87d0ffa40f23.png">

**Fix**

round the amount to 3 decimal places (I'm going ahead and assuming the smallest tip we do is **0.001** ) that way avoiding inconsistent data as well making sure we don't mess up the UI template of the mail

owocki:
thanks again aditya :)

PR #104, opened 2017-12-09
fix: rounded tip amount to 3 decimal places in mail

thelostone-mc:
This change to ensure that the tip amount  displayed in the mail is consistent with tip
I'm assuming the smallest tip which can be assigned to anyone is `0.001`

**File(s) modified**
- mails.py

**Module(s) changed**
- template

**Fixes**: https://github.com/gitcoinco/web/issues/103

**Checklist**
- [x] rebased with master
- [x] runs without breaking after changes


owocki:
Thanks :)

Issue #65, opened 2017-12-02
core : contributing.md guideline + linter


thelostone-mc:
Folks have different editor setting based on personal preference & since gitcoin looks promising  maybe we should consider having a style guide to ensure everyone follows a particular design style
Example:
- Lotta trailing whitespace -> me editing one line results in a lot of changes
- restrict tabspace to 4 (there is a mixture of 2 and 4 though the latter is more prominent )
the list goes on and on

Reason I suggest this:  I had changed a line in css and git diff showed 18 changes ( 17 were removal of trailing whitespace )

owocki:
i was talking to someone on slack the other day who suggested this too.  idea must be in the 'ether' right now so to speak!

im game to 

- try to get rid of trailing whitespace
- restrict tabspace to 4

beyond that, should we adopt the pep8 standard?  what are all the cool kids doing these days?

adityaanandmc:
few folks in my circle ( not sure if we could label ourselves cool just yet ) use google python style guide but `pep8` still stands tall.
It's recommended by folks at [django](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/). We could always update it suit our needs like maybe avoid breaking lines at 79 chars to make code look less clunkier!

For CSS/SCSS the options are plenty! 
I came across this neat little style guide  by [airbnb](https://github.com/airbnb/css)
As long as it's uniform everyone is happy. ^_^

gamwe6:
When I saw this issue I thought "but there is already a CONTRIBUTING.md file", but when I checked the repo ... nothing ... I forgot to push it XD. If you want to do it @thelostone-mc go ahead!

thelostone-mc:
I'll get to it, if @owocki hasn't already have one ready ^_^

owocki:
go for it @thelostone-mc !  

btw, send me a message on slack pls.  have a quick question for you

ethikz:
I'll implement a linter and fix all lint issues for scss.  After that I'll add in eslint

thelostone-mc:
@ethikz I've sort of added a css linter. Could you take of the JS lint bit ?

ethikz:
@thelostone-mc haha just updated my comment to make that change 😄

thelostone-mc:
@ethikz Yup def comparing ^_^ 
PS:  are we planning on using ESlint or something else ?

ethikz:
[...]

PR #106, opened 2017-12-12
WIP - chore(javascript): convert all non-vendor javascript to es6 and make sure it passes linting

ethikz:
This is a work in progress to convert all js to es6.  This will need to be thoroughly tested to ensure no javascript is broken.

This is a portion of #65 

Part of this or a subsequent task should be to create unit tests.

@thelostone-mc here is what I have so far.

### DIFF
--- a/app/marketing/views.py
+++ b/app/marketing/views.py
@@ -155,7 +155,7 @@ def email_settings(request, key):
             validation_passed = False
             msg = 'Invalid Email'
 
-        if level not in ['lite', 'regular', 'nothing']:
+        if level not in ['lite', 'lite1', 'regular', 'nothing']:
             validation_passed = False
             msg = 'Invalid Level'
         if validation_passed:


thelostone-mc:
Ah that's heavy re-write. will have to review this carefully.
Quick questions / points : 
- The commit with id `3fe8a91`, is this supposed to be a part of this PR ? Me and @tra38 had got into the code base.
- we should ensure browser compatibility not being an issue , might have to set up babel so that nothing breaks on browsers which don't have `es6` support
- @owocki this totally slipped my mind, we'd have to ask the current contributors if they are okay with the linting style ? 
_example :_
`const foo = ( bar ) =>  {
  return "foobar";
}`
_to ( what folks usually use )_
`const foo = (bar) =>  {
  return "foobar";
}`

PS:  👍  on removing line ending in library files.

ethikz:
@thelostone-mc 

- No `3fe8a91` shouldn't be part of it, thanks for letting me know.  My branches must be messed up, I'll fix it.
- This does have babel and renders everything for ES2015 **babel-env** which was added to replace **babel-es2015**
- I agree, I didn't mention the linting style but didn't mention the committing style.  I'll post in Slack and get a consensus.

Thanks for pointing it out!

owocki:
looks like theres some conflicts already :/  we should try to get this merged soon so there aren't a bunch of ongoing  conflcits.

meanwhile, i need to find a linter for sublime text

owocki:
testing new tip GH PR comment above ^^ @thelostone-mc you're next !

ethikz:
If we are going to merge this then I would start with a partial merge since this one includes all js.

I can create a new branch for the basic implementation and 1 file that can be merged instead if you'd prefer @owocki 

or I can change all the js back except 1 file and push that up

thelostone-mc:
New branch would probably be neater 👍

ethikz:
Was just thinking same thing

owocki:
works4me

ethikz:
Now being tracked here #111

thelostone-mc:
@ethikz I'm actually okay with either ^_^ we haven't enforced it on our guidelines yet

PR #109, opened 2017-12-13
css: fixed overflow scroll on tip recieve page

thelostone-mc:
**Problem:**

 On this tip, the background image overflows (as the position keeps changing ) cause the screen to be scroll-able as the screen bg changes position, bad user experience.

overflow-scroll

**Fix:** A single line of css

@owocki Might need an extra pair of eyes to look into this , to make sure nothing else is broken.

### DIFF
--- a/app/assets/yge/css/main.css
+++ b/app/assets/yge/css/main.css
@@ -10,6 +10,10 @@ body {
   width: 100%;
 }
 
+#yge {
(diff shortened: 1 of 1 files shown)


owocki:
tested LGTM

owocki:
just sent ya a tip :) appreciate it

Issue #81, opened 2017-12-04
Missleading button "Accept/reject claim" for non  founder of the issue


ernaneluis:
After claim to a issue and later check the issue page. The website shows a button "accept/reject" issue. However, is not the owner of the issue. Besides, if you click the button, it shows tabs for creating a issue, which is meant for the founder of the issue to accept or reject the claim.

<img width="803" alt="screen shot 2017-12-04 at 21 06 21" src="https://user-images.githubusercontent.com/1398769/33581841-c84882ca-d952-11e7-8eac-7efa9779abc9.png">

<img width="841" alt="screen shot 2017-12-04 at 21 06 28" src="https://user-images.githubusercontent.com/1398769/33581845-cd986c90-d952-11e7-926e-fbb4bc4c157d.png">

owocki:
more generally, i feel strongly that the funded issue detail page should only present to users actions that are enabled for them specifically

ernaneluis:
Yes I agree with you.

owocki:
posting a bounty for this. requirements
* no button shall appear under the 'funding amount: xxx ETH yyy USD' area unless it is actoinable be the user
* includes but this is not an exhaustive list:
* clawback funds
* accept/reject
* claim issue
etc

this is where the buttons are generated on the JS https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/pages/bounty_details.js#L317-L376

tra38:
I would like to take this bounty once it gets posted live, @owocki.

owocki:
@tra38 am waiting for this issue to be turned around https://github.com/MetaMask/metamask-extension/issues/2736 -- the service relies upon metamask/infura, so i can't reliably post a bounty until thats fixed.

you can go ahead and get started @tra38 -- i will just tip you directly when its done.  how does `0.05 ETH` sound?  can you submit a PR back to the repo in some reasonable amount of time, say the next 3 days?

tra38:
```0.05 ETH``` sounds good to me. I'll submit a PR soon.

gitcoinbot:
⚡️ A tip worth 0.06  ETH ($41.03) has been granted to @tra38 for this issue. ⚡️ 

The sender had the following public comments: 
> this is a good fix for the ui of bounty details page!

Nice work @tra38, check your email for further instructions. 

 * $10396.19 in Funded OSS Work Available at: https://gitcoin.co/explorer
[...]

PR #112, opened 2017-12-14
Remove misleading buttons on gitcoin UI

tra38:
PR is being submitted for #81.

We are currently generating some buttons on the gitcoin UI
by first querying information about the bounty and then deciding
what buttons to add on the UI based on that info.

However, we are only generating these buttons based on the
status of the bounty on the blockchain, and we do not take into account
whether the user is supposed to see those buttons. (For example, it makes
little sense for a person who just created a bounty to then claim their
own bounty.)

To prevent users from seeing buttons that they are not meant to see,
we will use the 'bounty_owner_address' to decide whether somebody
is a "Bounty Owner". If a user is currently logged into Metamask
and have the same address as the 'bounty_owner_address',
then they are likely the Bounty Owner. We then use this knowledge (in addition
to the current status of the bounty) to decide what buttons we should generate.

Bounty owners cannot watch or claim their own issues.

Non-"bounty owners" cannot accept/reject issues or clawback funds.

**WARNING** - Though this change will remove the "Claim" button, it does not stop "bounty owners" from actually claiming their own issue, simply by visiting '/funding/claim' and then typing in the URL of the GitHub issue.

I know we have some code in ```clawback_expired_bounty.js``` and ```process_bounty.js``` that checks if an action is valid, and I am tempted to use that same code in ```claim_bounty.js```, but I want to get your confirmation first. There might be some rare instances where a "bounty owner" might want to claim their own bounty (example: to indicate that the bounty is unavailable).

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -174,9 +174,13 @@ var callbacks = {
         }
         return [ label , response];
     },
-    
+
 }
 
+var isBountyOwner = function(result) {
+    var bountyAddress = result['bounty_owner_address']
+    return (web3 && (web3.eth.coinbase == bountyAddress))
+}
 
 var pendingChangesWarning = function(issueURL, last_modified_time_remote, now){
         console.log("checking this issue for updates:");
@@ -236,7 +240,7 @@ var pendingChangesWarning = function(issueURL, last_modified_time_remote, now){
                 this_transaction = "<a target=new href='"+link_url+"'>"+this_transaction+"</a>"
                 title = "Your transaction has been posted to web3.";
             }
-            var msg = `<br>This funded issue has recently been updated and while the blockchain syncs it has `+pendingchanges+`.  
+            var msg = `<br>This funded issue has recently been updated and while the blockchain syncs it has `+pendingchanges+`.
             Please wait a minute or two for web3 to sync `+this_transaction+`.
             <br>(You can close the browser tab.  If not, this page will automatically refresh as soon as web3 is updated.)`
             _alert({ title: title, message: msg},'info');
@@ -249,7 +253,7 @@ var pendingChangesWarning = function(issueURL, last_modified_time_remote, now){
         var local_delta = parseInt(timestamp() - localStorage[issueURL]);
         var is_changing_local_recent = local_delta < (60 * 60); // less than one hour
 
-        //remote warning 
+        //remote warning
         var remote_delta = (new Date(now) - new Date(last_modified_time_remote)) / 1000;
         var is_changing_remote_recent = remote_delta < (60 * 60); // less than one minute
 
@@ -263,7 +267,6 @@ var pendingChangesWarning = function(issueURL, last_modified_time_remote, now){
     }
     return should_display_warning;
 };
-
 
 window.addEventListener('load', function() {
     setTimeout(function(){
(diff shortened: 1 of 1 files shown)


tra38:
It seems that the build failed because it could not load the plugin 'ethereum'. I'm not familiar with the Ether ecosystem to know why that might happen, and I don't know how to kick off another build of Travis CI to see if this is just a temporary issue.

owocki:
> It seems that the build failed because it could not load the plugin 'ethereum'. I'm not familiar with the Ether ecosystem to know why that might happen in a Travis CI test (and if my changes may be responsible for that), and I don't know how to kick off another build of Travis CI to see if this is just a temporary issue.


there is some chatter in the dev-python room about this

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -174,9 +174,13 @@ var callbacks = {
         }
         return [ label , response];
     },
-    
+
 }
 
+var isBountyOwner = function(result) {
+    var bountyAddress = result['bounty_owner_address']
+    return (web3 && (web3.eth.coinbase == bountyAddress))
```
owocki: make sure you make this comparison case insensitive!

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -358,15 +361,18 @@ window.addEventListener('load', function() {
...
+                    if (is_on_watch_list(result['github_url']) && canWatch) {
                         var entry = {
                             href: '/unwatch',
                             text: 'Unwatch',
                             parent: 'left_actions',
                             color: 'darkGrey'
                         }
                         actions.push(entry);
-                    } else {
+                    } else if (canWatch) {
```
owocki: seems to me that anyone shoudl be able to watch an issue.  even bountie owners.. no?

tra38: The problem with that is that it doesn't really matter if a bounty owner is watching an issue or not. When I view the issue in the ```/explore``` dashboard, the bounty owner's issue is automatically marked with the word "Mine", so it is impossible to ever see "Watched".

See [this JavaScript snippet](https://github.com/gitcoinco/web/blob/32d8f4519e7b5b9d4133d6f1766794ba41329c4b/app/assets/v2/js/pages/dashboard.js#L174-L181) where we decide whether to generate what button to place next to a bounty on the dashboard ("Mine", "Watched", etc.). It seems this is the only place where we actually use the Watch list.

app/assets/v2/js/pages/bounty_details.js
tra38: There are other places in the codebase where we aren't doing case-insensitive comparison checks (for example, [here](https://github.com/gitcoinco/web/blob/32d8f4519e7b5b9d4133d6f1766794ba41329c4b/app/assets/v2/js/pages/dashboard.js#L175)). If you want, I can probably find all these instances and replace them with case-insensitive comparison cases...I just want to confirm first to make sure that case-insensitive comparison cases is indeed what we want.

owocki: ah shoot.. i didnt realize this. 

i dont think making these updates as part of this PR makes sense. thanks for the offer tho.

app/assets/v2/js/pages/bounty_details.js
owocki: > the bounty owner's issue is automatically marked with the word "Mine", so it is impossible to ever see "Watched".

got it; this is a UI issue that we should fix.

note that in the sidebar filters it is possible to filter only by your 'watched' issues... so there is some functional difference here though.

we should def fix the UI thing you noticed above eventually

owocki: http://bits.owocki.com/312J1709150k/Screen%20Shot%202017-12-14%20at%208.35.22%20AM.png

tra38:
I undid the change for watching, and will leave the case-insensitive issue as another issue.

owocki:
🎉 thanks @tra38 !

PR #129, opened 2017-12-15
Values on Landing Page

owocki:
# Why

1. I always get the reaction "oh no, another ICO token" when people see the Gitcoin brand.
2. I think that if we're going to build a distribued movement we have to very explicitly design the culture, and what binds us together as a community.

# What

A statement of the networks Mission, Vision, Values, and Cultural attributes, front and center, on the landing page

screencapture-localhost-8080-1513361191970

# Debate welcome

Please comment on the _aesthetic / design feedback on this PR.

Please comment on the _CONTENT_  at https://github.com/gitcoinco/gitcoinco/issues/6


owocki:
clearly its a bit crowded right now… so im thinking i might move a lot of it to another page (or make it hidden until you click ‘see more’)

owocki:
screencapture-localhost-8080-1513361810803

maybe less offensive to the eyes if grayscale until hover

owocki:
from slack 

> i suggest you use the color ones @owocki but suggest you use more spacing in between the partitions
> if that doesnt work for you, you could use another page, or the see more option..

Coinguybri:
IMO too much going on. May want to reduce # of values/icons to 1 or 2 for each category.

ethikz:
I would agree with @Coinguybri. It is way to busy and someone might get lost or even overwhelmed which might cause them to leave.

Also I think maybe changing the headline for that section.  Most might not know what ERC20 Token is.  For instance people are aware of what Bitcoin is, so maybe some way to tie it to that as a reference?

PixelantDesign:
I agree with @ethikz and @coinguybri on the number of principles and ERC20 heading. I recommend aligning on 5 principles or combining somehow.

Monochrome green could look interesting on the normal state, and then full color on hover.

:)

PixelantDesign:
I think the most powerful ideas are at the top. Breaking up the content like this might help to provide a some hierarchy.

gitcoin_page

Coinguybri:
I think it would be extremely effective if you could somehow say “it’s Uber but for devlopers”. Or something similar. I am not sure if you can use the Uber name or not. However, maybe you could have it as a testimonial from a user on the landing page?  I just think that association clearly explains the use case for the project.

owocki:
> I think it would be extremely effective if you could somehow say “it’s Uber but for devlopers”. Or something similar. I am not sure if you can use the Uber name or not. However, maybe you could have it as a testimonial from a user on the landing page? I just think that association clearly explains the use case for the project.

not sure if that referene can be made.. but even if it could, i'm not sure we should.  i'm concerned about conferring a reference to a sharing economy app that has controversial ethical practices... or is considerd by some to be exploitative

owocki:
@PixelantDesign

that is hott!  thank you.

some random feedback: 
1. can the mission section say something about the mission to push open source forward?
2. is there any way to collapse the 'mission' and 'world were building' sections to be side by side?
3. not sure if i'm crazy about the round black cicrcles in thie mission / world were building section 

just thinking out loud; i wonder if it's worth designing a model that brings up a detail view on any item  thats clicked.

Coinguybri:
@owocki  
I understand your concern. I was just trying to think of a way to make it instantly understandable and recognizable. Looks like it is already mentioned under “the world we want to build”

PixelantDesign:
Hi All! Here are a few revised designs of the val prop page. My preference is probably #1 or #2 since I think the mission is the star and there is hierarchy between the information. 
 
V1
gitcoin_valprop

V2
gitcoin_valprop2

V3
gitcoin_valprop3

Issue #32, opened 2017-10-13
waiting room entertainment while tx mines


owocki:
idea from @cmditch : 

“
I was thinking of some web3 UX ideas regarding the screen where you wait for the tx to be mined…

waiting room entertainment”, some kind of interesting read - an article, a quote, a tweet - or some interesting video
"

gitcoinbot:
__This issue now has a funding of 0.1 ETH (33.79 USDT) attached to it.__  To view or claim this funding, [click here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/32).

owocki:
cc @cmditch

owocki:
yes still looking to do it

> still am unsure what should be the "theme" of the quotes we want to display. Do we want them to be broadly focused on technological growth

@cmditch may have other ideas... but i think that quotes about open source software or freelancing would be cool.  

i'm all for a pre-screened list of interesting quotes (maybe stored in django, or just in the code base is fine to start...).

tra38:
Cool! I'll start work on this project, and will see if I can incorporate any of @cmditch's suggestions.

owocki:
hey @tra38 -- any interest in working off the design that @pixelant came up with ?

https://github.com/gitcoinco/web/issues/181#issuecomment-356364885

tra38:
Sure, I have some interest. I'll come back to this ASAP and start the CSS work. :+1:

PR #158, opened 2017-12-21
Display random quote when a web3 transaction takes place

tra38:
##### Description

From the commit message itself:

>Display random quote when a web3 transaction takes place  …

>The hope is that this random quote will entertain users
while they are waiting for the transaction to be processed.
If the quotes themselves are too short, we could increase their
length, or display multiple quotes.

> Ref: #32 

##### Checklist

- [ ] linter status: 100% pass (This is unclear to me, as we don't have a linter yet so I can't check right now to make sure if my PR is fine. I think it's okay though.)
- [X] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/contributing/CONTRIBUTING.md#step-4-commit)

Only affects the text displayed on the bounty_details page when an ETH transaction is in progress. In the future, we may need to move the random quote functionality to other parts of the site (transactions with EC20 tokens and tipping operations).

##### Refers/Fixes

Refers to #32. I do hope it fixes the issue, but the problem with entertainment is that it's hard to know whether people like it until you ship it.


app/assets/v2/js/pages/bounty_details.js
```diff
@@ -243,7 +243,8 @@ var pendingChangesWarning = function(issueURL, last_modified_time_remote, now){
             }
             var msg = `<br>This funded issue has recently been updated and while the blockchain syncs it has `+pendingchanges+`.
             Please wait a minute or two for web3 to sync `+this_transaction+`.
-            <br>(Please DO NOT close the browser tab.  This page will automatically refresh as soon as the blockchain is updated.)`
+            <br>(Please DO NOT close the browser tab.  This page will automatically refresh as soon as the blockchain is updated.)
+            <p>QUOTE OF THE DAY: ` + waitingRoomEntertainment() + `</p>`
```
owocki: the info message is big enough already.. 

might be good to get a designer into this and show off the quote in a beautifully desiigned way on the page itself... 

what does everyone think?

owocki:
good first pass.  thanks @tra38 !

tra38:
Moved the quotes over the JavaScript layer and cause the quote section to refresh every 30 seconds. This definitely will make waiting more entertaining. (That being said, if the site gets used over and over, it is possible that users will begin to exhaust the random quotes. So we need to periodically add more and more quotes to keep the user's attention.)

I also no longer use the (incredibly inaccurate) term "QUOTE OF THE DAY". Instead, I use the words  "INTERESTING QUOTE".

>the info message is big enough already..
>might be good to get a designer into this and show off the quote in a beautifully desiigned way on the page itself...

>what does everyone think?

I agree with this. To help this future designer out, I wrote the JavaScript quote layer in such a way that we place our random quote in any div with the ```waiting_room_entertainment``` class. So a future designer can write a module to display the "waiting room entertainment" quote, and simply add ```.waiting_room_entertainment```, and the code should still work. This approach also means that we can reuse this "waiting room entertainment" on other parts of the site, if we need to.

owocki:
@tra38 mind posting a screenshot of what the current `waiting_room_entertainment` div looks like?

i will post a small bounty to put a nice design around it

tra38:
Sure @owocki.

HTML Markup (displayed onto the page using javascript):
```javascript
var msg = `<br>This funded issue has recently been
updated and while the blockchain syncs it has `+pendingchanges+`.
Please wait a minute or two for web3 to sync `+this_transaction+`.
<br>(Please DO NOT close the browser tab.  This page will automatically
refresh as soon as the blockchain is updated.)
<br>INTERESTING QUOTE: <span class='waiting_room_entertainment'></span>`
```

Good screenshot of the output:
screen shot 2017-12-28 at 12 15 31 am

(Sidenote: the pending transaction, in this case, was to create a new bounty on an issue that previously had a bounty on it successfully fulfilled. That's why the status is marked as being fulfilled -- its status is being updated to "active".)

owocki:
https://github.com/gitcoinco/web/issues/181

owocki:
@tra38 can you implement the design posted at https://github.com/gitcoinco/web/issues/181 ? if not, can you pair with someone to make it happen?

tra38:
Got it. Let me see if I can try to do it on my own, and if not, I'm finding a pair to help me with this.

tra38:
@owocki, I'm handing the project off to @eswarasai, but we're having difficulty setting up the project. Specifically, we need to update the current exchange rates on the web-app so that we can gain access to the ```/funding/new``` and ```/funding/claim``` endpoints, but whenever @eswarasai is running the "get_prices" command, he is getting a 520 error from etherdelta and a 403 error from Poloniex. Can you please help @eswarasai out in the Slack chat?

owocki:
> Specifically, we need to update the current exchange rates on the web-app so that we can gain access to the /funding/new and /funding/claim endpoints

you shouldnt need to run these commands to use  /funding/new and /funding/claim 

but nonetheless feel free to post in the slack channels if youre having problems @eswarasai .. im in meetings a lot of the week but if you post in the public channels thats better than DMing me

eswarasai:
@owocki -- So I was trying to find out the root cause for the `/funding/new` and `/funding/claim` not working. I've pulled the code from `gitcoinco/web:master` and was trying it out locally. I can see that the changes that were done by you in [this commit](https://github.com/gitcoinco/web/commit/d8498273ab45cc1279b88ecafeb7830eed911ecb#diff-f41d183716f64bdb8600ffb3df9aacdb) are causing these errors.

The function `eth_usd_conv_rate()` in `gas/utils.py` is calling `convert_amount()` in `economy/utils.py`. Inside the `convert_amount()`, we're fetching the `ConversionRate` data from the DB which is not populated in the local database. This is the reason the above pages are broken locally. On server, I guess the data is already populated, hence no issues.

One more thing to note, `./manage.py gas_prices` command won't populate data now because the URL -- https://poloniex.com/public?command=returnTicker is now using Google Captcha which cannot be bypassed programatically I guess.

Let me know the resolution here. Thanks.

owocki:
please just comment out the calls to `eth_usd_conv_rate()` (or whatever function is reliant upon having `ConversationRate` rate data as a means of working around this... 

i will file another ticket for this:

> One more thing to note, ./manage.py gas_prices command won't populate data now because the URL -- https://poloniex.com/public?command=returnTicker is now using Google Captcha which cannot be bypassed programatically I guess.

it's curious that this runs okay locally for me though.

eswarasai:
@owocki -- I've done some little research on the Poloniex Captcha issue. There's a security layer added which doesn't allow Public API calls if the Request IP Address falls out of some specific range/region. That could be the reason why it is working in US and not in India.

eswarasai:
@tra38 @owocki -- I've pushed the Design changes for the waiting room quote to match that of the spec. Let me know your feedback. Thanks.

owocki:
this is dope! 
screencapture-localhost-8080-funding-details-1516139031669

owocki:
can we resolve the merge conflicts?

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -184,72 +184,85 @@ var isBountyOwner = function(result) {
...
-                var bounty = web3.eth.contract(bounty_abi).at(bounty_address());
-                setTimeout(function(){
-                    bounty.bountydetails.call(issueURL, function(error, result){
-                        if(error){
-                            setTimeout(check_for_bounty_changed_updates_web3, 1000);
-                            console.error(error);
+    console.log("checking this issue for updates:");
+    console.log(issueURL);
+
+    //setup callbacks
```
owocki: its not quite clear to me what changed here.. spacing?

app/dashboard/templates/bounty_details.html
```diff
@@ -20,77 +20,102 @@
...
                 </div>
-                <div class="row nonefound" style="display:none;">
-                    <img src="/static/v2/images/shrug_bot.png">
-                    <h3>No funded issue found.</h3>
-                    <p>Please go back to the <a href="/explorer">Issue Explorer</a> and try again.</p>
-                    <br>
-                    <h6>Be the OSS Funding you wish to see in the world.</h6>
-                    <p id='submitsolicitation'>Looking to fund some work?  You can submit a new Funded Issue <a href="/funding/new">here</a>.</p>
+            </div>
+
```
owocki: are these all just indentation changes?

app/dashboard/templates/shared/rain.html
```diff
@@ -0,0 +1,32 @@
+{% comment %}
```
owocki: theres already a rain.html template.. can we just reuse that one?

owocki:
UI looks good to me.  had some trouble doing code review due to the volume of the changes...

@mbeacom wanna take a 👀 ?

app/dashboard/templates/shared/rain.html
thelostone-mc: @eswarasai [here](https://github.com/gitcoinco/web/blob/master/app/retail/templates/shared/rain.html) it is ^_^

owocki:
im good with keeping the quotes in something static... like the js file.. i dont think we'll be adding / removing quotes all that often.

thelostone-mc:
This looks solid!! ^_^ 
Just two suggestion : 
- does it look better bringing down the size of the text on hand held devices ( **18px** like in the gif )
- we need to fix that scroll with `overflow: hidden`, it's annoying to see that scroll cause of the background. Just bothers me :P

git

other than that LGTM 🎉 

@tra38 thanks for helping out  with the setup 👍

app/assets/v2/js/pages/bounty_details.js
eswarasai: Yeah @owocki, they're just spacing issues here

app/dashboard/templates/bounty_details.html
eswarasai: Not all of them are indentation changes. There are some additional HTML DIV sections and CSS classes to get the Design match that of the spec

eswarasai:
@thelostone-mc : Thanks for the above suggestions.
@owocki -- I have an interesting question here. Can someone try to claim or fund the issue on Mobile devices? I don't think we can have a Metamask extension on Mobile devices right?

owocki:
>  I have an interesting question here. Can someone try to claim or fund the issue on Mobile devices? I don't think we can have a Metamask extension on Mobile devices right?

Not at the moment, but I bet ya dollars to donuts there will be a mobile browser released soon.

Plus responsive design is good for desktop at lower resolutions too :)

owocki:
looking fwd to merging after this https://github.com/gitcoinco/web/pull/158#issuecomment-358157030

tra38:
Looking forward to the tip distribution. :+1:

owocki:
> Looking forward to the tip distribution. 👍


did i send this yet?  sorry a little behind this week

tra38:
You did sent the tip to me so that's good. :+1:

PR #166, opened 2017-12-23
Generate high-entropy secrets and codes

10a7:
#####  Description

Gitcoin Core currently generates hashes from low-entropy, predictable
data. While they may appear random to outside observers, looking at the
source code quickly shows these as being predictable.

The construct used in both places appears to be a RIPEMD-160 hash
salted with the RIPEMD-160 hash for empty string and then the Django
timezone. An adversary with knowledge of the rough time in which a
secret was generated may use this knowledge to predict the hash used for
the whitepaper access code or the `priv` property.

This pull request removes the `hashlib` from use and generates
hexadecimal secrets from the OS `urandom` facility. These strings are
then truncated to match the same length as previously to avoid any
compatibility issues.

While the construct with `codecs` may appear strange, it will generate
the proper string in both Python 2.x and 3.x.

Testing: No joke, I just tested the construct, linted this and
called it good. Please pull this and try it locally before merge. I
don't have docker / docker-compose working on this workstation yet.

##### Checklist

### DIFF
--- a/app/marketing/models.py
+++ b/app/marketing/models.py
@@ -40,13 +40,11 @@ class EmailSubscriber(SuperModel):
         return self.email
 
     def set_priv(self):
-        import hashlib
+        import codecs
+        import os
         from django.utils import timezone
 
-        h = hashlib.new('ripemd160')
-        h.update("{}-{}-{}".format(h.hexdigest(), timezone.now(), self.email))
-        self.priv = h.hexdigest()[:29]
-
+        self.priv = codecs.getencoder('hex')(os.urandom(16))[0][:29]
 
 class Stat(SuperModel):
     key = models.CharField(max_length=50, db_index=True)
--- a/app/tdi/views.py
+++ b/app/tdi/views.py
@@ -15,7 +15,7 @@
     along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 '''
-import hashlib
+import codecs
 import os
 import StringIO
 from wsgiref.util import FileWrapper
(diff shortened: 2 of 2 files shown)


owocki:
👍  glad to have you here looking out for us @10a7 

@mbeacom what do you think of this

mbeacom:
@10a7 Thanks! This looks fine to me.  The only thing I can see here is we might want to put this into a method and call it with a length.  Beyond that, it seems to replicate similar random strings to what is currently being generated.  We'd be limited by `codecs.getencoder('hex')(os.urandom(16))[0][:length]` to a maximum of `32` characters, though so the change might make much of a difference, beyond removing the repeats.

@owocki The only additional changes we might want to consider moving forward is whether or not we'd like to consider mixed case and/or greater than `32` character random strings with something like:

```python
import random
import string


def generate_random_string(length=29):
    return ''.join(random.choice(string.ascii_uppercase + string.ascii_lowercase + string.digits) for _ in range(length))

key = get_random_string()
```
or django's builtin:

```python
from django.utils.crypto import get_random_string

key = get_random_string(length=29)
```

In any case, I'm good with this and it makes sense! Plus your proposed solution is faster than the hashlib `RIPEMD-160` solution.
Thanks again!

10a7:
:+1: thank you.

@mbeacom `random.choice` will drop you back off of a cryptographically secure random number generator again, [per Python documentation](https://docs.python.org/2/library/random.html). The Django `get_random_string` is supposed to use the `SystemRandom` which should be secure, but it does have a fallback in case that facility isn't there. I can't imagine a situation in which you won't have the `urandom` facility for the application.

Yes, you're right about DRYing the code. It probably should be abstracted into a method, but since `hashlib` wasn't, I just followed the code style that was already present.

mbeacom:
@10a7 I agree with that.  Likewise, I understand not abstracting the logic to a method, given it's previous state.  That was more so a note for future changes to the existing implementation.

Thank you for the contribution!

Issue #181, opened 2017-12-28
design: waiting room quote while page is waiting for web3 updates


owocki:
Over at https://github.com/gitcoinco/web/pull/158 we are putting some code together to provide a quote to a user while they wait for their ethereum transaction to sync.

here is the design the developer has come up with:

<img width="1278" alt="34401982-68e965c8-eb64-11e7-8dbf-b9ffa1c49a5b" src="https://user-images.githubusercontent.com/513929/34415163-60bf129e-ebaa-11e7-9322-97d394e33b4b.png">

i'd like to move the quote OUT of the blue box, and onto the page.
please feel free to use any assets in the press kit https://www.dropbox.com/sh/bsjzbu0li2z0kr1/AACKgnQC3g6m52huYI3Gx3Ega?dl=0

gitcoinbot:
__This issue now has a funding of 0.1 ETH (69.94 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/181).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/181)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $9976.75 more Funded OSS Work Available at: https://gitcoin.co/explorer

usulix:
@owocki .. I mentioned this in the slack channel, but thought I would add here as well.. Looks like a great starter issue... I should have results in just a bit. 
G

olafghanizadeh:
Is coding it a part of the issue too?

thelostone-mc:
@olafghanizadeh This is a design issue. Once a design accepted -> there will an issue created to code it down ^_^

usulix:
@thelostone-mc,
oic.
I guess that explains why the quote is not showing in my branch. But, I am setup locally and have generated html to edit. Altering markup now.

usulix:
While working on this, I discovered that it looks like the alert does not push the #bounty_detail down in mobile view, but instead displays over it.

![image](https://user-images.githubusercontent.com/4955172/34422071-633dc7ba-ebd0-11e7-9824-c197a73c115f.png)

Any guidance on what should be done for mobile view for displaying on the page ?

gitcoinbot:
[...]

Issue #175, opened 2017-12-27
small module design: waiting room quote


owocki:
scope of this task is to submit a nice, on-brand design for the waiting room quote on gitcoin.co (see this comment: https://github.com/gitcoinco/web/pull/158#issuecomment-353760228 )

gitcoinbot:
__This issue now has a funding of 0.05 ETH (37.48 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/175).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/175)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $12451.08 more Funded OSS Work Available at: https://gitcoin.co/explorer

owocki:
style guide https://www.dropbox.com/s/93wxuoeysaamsch/Gitcoin_Brand.pdf?dl=0

gitcoinbot:
__The funding of 0.05 ETH (34.91 USD) attached has been claimed by @PixelantDesign.__ 

 @PixelantDesign, please leave a comment to let the funder (@owocki) and the other parties involved your implementation plan.  If you don't leave a comment, the funder may expire your claim at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/175)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $9959.64 more Funded OSS Work Available at: https://gitcoin.co/explorer

PixelantDesign:
@owocki  Claimed!

owocki:
derp, i am disorganized and posted two bounties for this.  happy to pay both claimees, as it was me who accidently posted twice

https://github.com/gitcoinco/web/issues/181
https://github.com/gitcoinco/web/issues/175

PixelantDesign:
@owocki No worries, we can work together on the other ticket.

owocki:
ok i will reject the claim and pay it out to myself.  thanks!

gitcoinbot:
__The funding of 0.05 ETH (47.32 USD) attached has been claimed by @owocki.__ 

[...]

PR #186, opened 2017-12-28
Waiting room quote while page is waiting

usulix:
##### Description

It seems as far as I could tell #181 and #175 overlap a bit, but this PR is designed to address 181, the moving of the Important Quote from the alert into the main page and addressing the display requirements in that.
This pR also addresses some encountered mobile styling issues and adjust the signup CTA slightly for very small screens

mobile view
image

desktop view
image

##### Checklist

- [X] lint rules followed
- [X] commit message follows [commit guidelines]

##### Refers/Fixes
- adjust media queries to resize icon, input and button in sign-up cta
- create display area with styling for placing interesting quote in page instead of alert
https://github.com/gitcoinco/web/issues/181
may have some cross scope with 
https://github.com/gitcoinco/web/issues/175

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,3 +294,5 @@ Have an ERC20 compatible token that you'ud like to add support for?  Great!
 
 <!-- Google Analytics -->
 <img src='https://ga-beacon.appspot.com/UA-102304388-1/gitcoinco/web' style='width:1px; height:1px;' >
+
+# check commit test comment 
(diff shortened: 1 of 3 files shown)


owocki:
can we make the background of the quote a little less neon.  right now its too attention grabbing, and i think it should be a little more subtle.

seems like the quote should have a little more design elbow grease associated with it too.  right now it's just a long `p` tag.

PR #189, opened 2017-12-29
Add sane fallback for localStorage usage

### LINKED ISSUE #188: Disabled localStorage breaks Issue Explorer
When viewing Gitcoin in a browser with localstorage disabled (or, where `window.localStorage` is undefined) will cause the application to fail to load.

##### Description
If `localStorage` is not an object (or access is denied to items in localStorage), an uncaught exception is thrown on [line 21 of `v2/js/pages/dashboard.js`](https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/pages/dashboard.js#L21), preventing further code from being executed.

To reproduce:
* Disable cookies for `https://gitcoin.co`
* Navigate to `https://gitcoin.co/explorer?q=`
* Check the console for errors

##### Possible Implementation
Provide a sane fallback for lack of or lack of access to `localStorage`

##### Environment

* **Browser**:  Chromium
* **Version**:  63.0.3239.84
* **Subsystem**: Gitcoin UI, dashboard.

eth-button:
##### Description
Fixes #188.  Feel free to add to close this PR and open a new one if this doesn't sufficiently fix the problem. I'm not able to run this locally with the resources at my disposal, so this is a shot in the dark that's sort of been tested in Chrome dev tools.

##### Checklist
- [ ] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/contributing/CONTRIBUTING.md#step-4-commit)

ui

##### Refers/Fixes
Fixes #188

### DIFF
--- a/app/assets/v2/js/pages/dashboard.js
+++ b/app/assets/v2/js/pages/dashboard.js
@@ -1,6 +1,13 @@
 //helper functions
 var sidebar_keys = ['experience_level', 'project_length', 'bounty_type', 'bounty_filter', 'idx_status', 'network'];
 
+var localStorage;
+try {
+    localStorage = window.localStorage;
+} catch (e) {
+    localStorage = {};
+}
+
 //sets search information default
 var save_sidebar_latest = function(){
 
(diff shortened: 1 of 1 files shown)


owocki:
code looks good.  is it worth adding this in a more global context, so that all pages have a localstorage fallback (instead of just dashboard)?

eth-button:
I would assume so.  This only addresses the case that I encountered, since I wasn't able to load much more of the site.

I'll update the original issue with steps to reproduce, as well.

owocki:
thanks again!

PR #195, opened 2017-12-30
Remove all-caps default in /tip (yge)

bumi:
##### Description
Similar to #116 and discussed in #151 this PR removes the global (on body) `text-transform: uppercase` style from the /tip (yge) application. 

As discussed in #151 properties like the bounty status should also be normalized (capitalized in the UI) 
These properties are stored in the DB and we need to decide where to put that presentation logic. 

##### Checklist

- [ ] normalize bounty properties (status, etc.)

##### Affected core subsystem(s)
ui

##### Refers/Fixes
* #151 - "typography cleanup after PR #116"

PR #202, opened 2018-01-02
dos2unix line endings for yge and html cleanup

bumi:
Seems like only yge was using non unix line endings. This PR does a dos2unix on the yge files. (please review).
Additionally I also did some indentation cleanup in the HTML files.
And fix an html error I introduced in #195


owocki:
thanks.. yeah the tests need to be cleaned up.

i should maybe run `dos2unix` on the whole codebase myself.  its very hard to review this PR and make sure there's no vulnerability injected.. since theres so much code taht's changed.  unless thers an easier way to do a diff?  not that i don't trust ya @bumi , but.... its crypto... so the downside is high.

@mbeacom what do you think

bumi:
yeah, not sure how to diff it best or how to make it easier to review.
Such huge PRs suck. I've made two commits, one for the html that also includes some indentation fixes and one for the assets. 
(btw. regarding the HTML: I mostly use 2 spaces indentation for HTML - is it in python projects 4 spaces, like in python files? )

If the reason is "crypto" then I think we should not even "trust" your computer and as a project not even "trust" your account :)  actually guess it is less about trust but more about reviewing changes in general.

owocki:
@bumi appreciate you doing this..  i think in leiu of accepting this Pr, could you just let us know what command you used to migrate everything over?

bumi:
[dos2unix](https://waterlan.home.xs4all.nl/dos2unix.html) is the friend :) 
let me know if I can help something

mbeacom:
Thanks @bumi! We're working to move everything over to a standard format and we'll likely validate syntax/styling via a pre-commit hook.

bumi:
great! standards sound good :) 
I think for the dos file endings it is more a local git setting isn't it?

Issue #177, opened 2017-12-27
code: `/mission` page


owocki:
# Requirements

* new page, at `/mission`
* link this new page in the footer
* code the page to look like the design (linked below in ref section)
* mobile, tablet, and desktop views must all look good.
* 'read about our mission' should link to https://medium.com/gitcoin/pushing-open-source-forward-for-the-web3-generation-26a5726902ea

# Bounty

Please only claim this bounty if you can turnaround the task in the next 2 days.

# Ref

Initial Inspiration PR (now closed) https://github.com/gitcoinco/web/pull/129
Design: https://github.com/gitcoinco/web/issues/139
Creative Assets https://www.dropbox.com/sh/tgiymbpz5cmxjfe/AABpLxtlnhd_7k2QB7KH9gqSa?dl=0

gitcoinbot:
__This issue now has a funding of 0.08 ETH (57.44 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/177).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/177)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $11336.35 more Funded OSS Work Available at: https://gitcoin.co/explorer

owocki:
@thelostone-mc i dont think we should put this on hold due to https://github.com/gitcoinco/web/issues/178.  #178 was just meant to be a note to remember to circle back for a v2.

thelostone-mc:
Ah I'll give out a go later tonight! Not claiming it as I'm not certain if I can commit to it 1 day dealing but if progress is made, thou shall claim it

gitcoinbot:
__The funding of 0.08 ETH (55.85 USD) attached has been claimed by @eswarasai.__ 

 @eswarasai, please leave a comment to let the funder (@owocki) and the other parties involved your implementation plan.  If you don't leave a comment, the funder may expire your claim at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/177)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $9959.64 more Funded OSS Work Available at: https://gitcoin.co/explorer

[...]

PR #203, opened 2018-01-02
Gitcoin's Mission

eswarasai:
##### Description

Moved the Gitcoin's Mission and Values to a separate page (`/mission`) which is now linked in both Top Nav as well as Footer Menu.

##### Checklist

UI

##### Refers/Fixes

Fixes: #177


eswarasai:
TODO:
1) The top banner in the mockup has `Gitcoin's Mission` but the image in the hero banner is just `Gitcoin`
2) Minor fixes in the responsiveness layout for the Mobile devices

owocki:
> The top banner in the mockup has Gitcoin's Mission but the image in the hero banner is just Gitcoin

You could maybe make the positioning of the text relative to the logo configurable.... maybe a boolean template var entitled `position_h1_below_logo` would do the trick..

README.md
```diff
@@ -296,7 +296,7 @@ DATABASES = {
 ```
 virtualenv gcoin
 source gcoin/bin/activate
-pip install -r ../requirements.txt
+pip install -r ../requirements/dev.txt
```
owocki: 👍 good catch

.gitignore
```diff
@@ -13,3 +13,4 @@ app/assets/other/avatars/
 venv/
 .coverage
 node_modules/
+gcoin/
```
owocki: 👍

app/assets/v2/css/gitcoin.css
```diff
@@ -237,6 +237,14 @@ div.button-pink {
   max-height: 150px;
 }
 
+.mission-img-wrapper {
+    height: 250px;
+}
```
owocki: some of the images are almost 2500px x 2500px.  mind scaling them down to be less bandwidth hogs?

owocki:
left a few comments..  looks good (mostly) to me.  

@pixelantdesign here is a screenshot if u want to give ur own feedback

screencapture-localhost-8080-mission-1514934552309

eswarasai:
@owocki: I've fixed the mobile responsiveness issue and also updated the images with much lesser resolution now. I've also added the provision of `position_h1_below_logo` in the `minihero.html` but this logo will be `GITCOIN'S` instead of `GITCOIN`.

Let me know if there's anything else pending. Thanks.

PixelantDesign:
@eswarasai Looking good! Would it be possible to scale down some of the graphics under “We Value” and “We Interact” so that it’s proportionate to the original mock-up?

eswarasai:
@PixelantDesign : The screenshot which you're looking at is not at 100% width of the screen.
@owocki : Correct me if I'm wrong.

So based on the device width, the number of images in a row will be increased/decreased accordingly.

Here's a screenshot on how it looks at full-screen size :

mission

PixelantDesign:
@eswarasai that makes sense. Thank you!

One thing I was wondering is where we are linking “Read the mission of Gitcoin” to (since this is the mission page)? If it’s somewhere internal, maybe the button needs renaming? If external, it will open in a new tab?

eswarasai:
@PixelantDesign: Yes. Currently the `Read About Our Mission` button is pointed to a [Medium Blog post](https://medium.com/gitcoin/pushing-open-source-forward-for-the-web3-generation-26a5726902ea) which opens up in a new tab.

I think you're right. Since this is the Mission page, the text of the button should be updated to something relevant about the blog post.

@owocki : Any suggestions here?

PixelantDesign:
Looking at it again. I think the current language is okay.

It would be great in the future to have clear distinction and set expectations for users whether they are diving into more content on the gitcoin site vs. content that lives outside.

PR #204, opened 2018-01-02
Changed links to MetaMask Chrome extension to MetaMask website

tossj:
##### Description

All links to install MetaMask point to the Chrome extension, but there is a Firefox extension as well. This change makes the links point to the MetaMask website, so the user can get the appropriate MetaMask extension.

`rel="noopener noreferrer"` is included to reduce the risk of the security vulnerability described [here](https://mathiasbynens.github.io/rel-noopener/). The `noopener` is not supported by MS Edge, so `noreferrer` is included. This may interfere with MetaMask's traffic analytics from these links on Gitcoin, so the `utm_source` and `utm_medium` are included as a way for MetaMask to still know that these links are visited from Gitcoin.

##### Checklist


owocki:
LGTM.  @mbeacom any comments?

PR #235, opened 2018-01-13
Changed bounties URL in JS to prevent 301 redirect

tossj:
##### Description

When the Issue Explorer is visited there are XHR requests to an endpoint related to bounties information. Due a missing trailing slash in the XHR's the API server returns with a 301 redirect and duplicate XHR's (now with the trailing slash) are sent out. This fix adds the trailing slashes into the JS files where the requests are sent out, so as to prevent the 301 and reduce loading time and bandwidth, if only by a small amount.
##### Checklist


owocki:
lgtm.  @mbeacom ?

PR #243, opened 2018-01-15
roundup email 20170116

owocki:
what the title says


app/retail/templates/emails/bounty_roundup.html
```diff
@@ -48,16 +48,20 @@ <h1><img src="https://gitcoin.co/static/v2/images/helmet_emoji.png" width=40px h
     <tr>
         <td id="welcome" colspan=6 valign="top" style='padding: 5px 10px 0px;'>
             <p>
-                Howdy doo,
+                Hi folks 👋
             </p>
             <p>
-                Last week, Gitcoin added it's first full time (non-founder) team member.  Please join me in <a href="https://medium.com/gitcoin/growing-gitcoin-c39f8520134d">welcoming Vivek Singh to the project</a>!
+            Gitcoin is thrilled to <a href="https://medium.com/@coderberry/5684c4adf4b4">add Open Source advocate and developer Eric Berry to our ranks! </a>
```
mbeacom: We could template this out at some point, so you don't have to commit the email body.

owocki: its in the templates folder... what else could we do?

Issue #173, opened 2017-12-27
Design: Faucet page


owocki:
Very often, I get people who want to claim an issue on Gitcoin but can't because they don't have ETH.  Either there are no exchanges in their area, or they dont have the mone yto spend on gas.

Pursuant to making ethereum and Gitcoin accessible to everyone as broadly as possible, I would like to design a page entitled 'faucet' in which a user can go to the page and submit an application to receive a small amount of ETH.

The form should contain 

* a Github Oauth button
* a comments field (for explaining why you need the ETH)
* an ETH address input field
* an email address input field
* a TOS acceptance checkbox input field
* a submit button

Functional spec:

* Design will be responsive
* All inputs are required.
* The form should not allow users to submit more than once per github account.
* when the form is submitted a new row will be stored in `FaucetApplication` ( a new model)
* when the form is submitted, an email will be sent to `settings.SERVER_EMAIL` notifying the admin of the faucet application.

# Bounty

Scope of this bounty is to design a faucet input form page.

gitcoinbot:
__This issue now has a funding of 0.1 ETH (74.96 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/173).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/173)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $11963.83 more Funded OSS Work Available at: https://gitcoin.co/explorer

mauri2015:
Are you looking for someone to design the form as well as develop the functionality?

owocki:
style guide https://www.dropbox.com/s/93wxuoeysaamsch/Gitcoin_Brand.pdf?dl=0

owocki:
just design for now @mauri2015 .. will bounty the fucntionality later

gitcoinbot:
__The funding of 0.1 ETH (73.4 USD) attached has been claimed by @PixelantDesign.__ 

 @PixelantDesign, please leave a comment to let the funder (@owocki) and the other parties involved your implementation plan.  If you don't leave a comment, the funder may expire your claim at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/173)
[...]

PR #258, opened 2018-01-19
Faucet page

usulix:
##### Checklist

- [x] commit message follows [commit guidelines]- add option to skip credentials on git search
- add faucet app and faucet_amount to config
- add faucet routes
- add faucet migrations
- add faucet templates
- add faucet model
- add faucet view and admin functions
- add faucet css additions
- add faucet specific js
- hook into existing function for faucet w3 errors
- add image for faucet page
- process faucet request

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/184


usulix:
@owocki 
I still get the BadRequest error on sending the email.
I also tried to send an email from the referenced whitepaper process and got a different error in the send_email function from mail.py, so I think that must be something in my config.

I do not get any token request or errors.. I can only imagine that would be CSRF token... but I did not see that anywhere.

testing boils down to visit localhost:8000/faucet and you should be able to create a request...
and visiting http://localhost:8000/_administration should allow you to click into faucet management and click process_me to pop up the meta_mask dialogue to okay the transfer

owocki:
this still doesnt use the design provided by @pixelantdesign https://github.com/gitcoinco/web/issues/173#issuecomment-354347567

app/faucet/views.py
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    print request.POST.get('destinationAccount')
+    print request.POST
+    if obj.fulfilled:
+        return redirect('/_administrationfaucet/faucetrequest/')
+
+    if request.POST.get('destinationAccount', False):
+        obj.fulfilled = True
+        obj.save()
```
owocki: per the spec at https://github.com/gitcoinco/web/issues/184

> upon success, mark the faucet object as fulfilled and send an email notifying the recipient that their faucet request has been funded.

please make sure you trigger the email to the end user!

usulix:
@owocki ,

re #173 (comment) -- since it was only mentioned under resources and not called out in the description, I was unaware that the mock ups existed.

These call for a slightly different workflow from what I have, but probably a desirable fix as it will require a user to actually sign  into their github - thus verifying that they are connected to the github account they are referencing.

Do we have psd's or similar of the design comps ?

owocki:
> These call for a slightly different workflow from what I have, but probably a desirable fix as it will require a user to actually sign into their github - thus verifying that they are connected to the github account they are referencing.

yes, i agree this is important.. it'll prevent sybill attacks

> Do we have psd's or similar of the design comps ?

If you check https://github.com/gitcoinco/web/issues/173#issuecomment-354969169 @pixelantdesign put a ZIP file in the comments with any assets you'll need for this.

owocki:
per @usulix message on slack.. he is super busy at work and needs some help getting this across the finish line..  i need to confirm with him, but i think yall should split the ETH.  anyone want?  @ethikz @thelostone-mc ?

ethikz:
@owocki I'm looking into #184.  I believe these are both tied together but this one is pretty much finished based on comments on the referenced issue, am I correct in assuming that?

owocki:
just moved over to this PR => https://github.com/gitcoinco/web/pull/407/

Issue #263, opened 2018-01-23
standardbounties:  Check web3 vs. database to see if bounty exists


jasonrhaas:
Currently, if a user tries to Kill/Fulfill/Accept an issue that does not exist, the code queries the API to see if that issue URL exists.  If it doesn't, it will pop up a red error banner to the user.

While this works, it might be more thourough to also check the blockchain if the bounty exists.   In theory, the database should always be syncing with web3 on any changes, so this shouldn't be an issue.

If we were to do this, the flow would have to be:

- Call back-end API using github issue URL to get database information
- Grab the `standard_bounties_id` and use web3 to query `getBounty(_bountyId)`.
- If this call fails or returns nothing, we return an error to the user.

thelostone-mc:
@jasonrhaas 

went through `kill_bounty.js`
I'm guessing the code would sort of look like this as everything else is available
```
 bounty.getBounty(bountyId, (error, result) => {
    if(error || result == null)
        errormsg = "No active funded issue found at this address.  Are you sure this is an active funded issue?";
 });
```
A lil confused here though, under what scenario would the bounty not exist on the blockhain but on our backend API ? (or is it just a safety precaution )

thelostone-mc:
^ @mbeacom / @owocki on the off chance @jasonrhaas is held up

jasonrhaas:
@thelostone-mc The database should always be sync'ed up with the blockchain.  It might be possible for it to get out of sync if the contract was altered somewhere outside of gitcoin.  Also, when #128 is implemented I think that should also help to make sure the DB and the blockchain are always in sync.

owocki:
If a user closes their browser window on the pending confirmation page, it will get out of sync.  This is an especially pervasive issue in times of network slowness.

PR #301, opened 2018-01-28
standardbounties: check web3 to see if bounty exists

thelostone-mc:
updated fulfill_bounty, kill_bounty, process_bounty to check web3 if
API call returns result for the bounty github url.

##### Checklist

- bounty 
##### Refers/Fixes
https://github.com/gitcoinco/web/issues/263

### DIFF
--- a/app/assets/v2/js/pages/fulfill_bounty.js
+++ b/app/assets/v2/js/pages/fulfill_bounty.js
@@ -122,6 +122,13 @@ window.onload = function(){
                             }
 
                             var bountyId = result['standard_bounties_id'];
+                            bounty.getBounty(bountyId, (error, result) => {
+                              if(error || result == null) {
+                                _alert({ message: "No active bounty found on the blockchain for this Github URL." });
+                                unloading_button($('#submitBounty'));
+                                return;
+                              }
+                            });
 
                             bounty.fulfillBounty(bountyId, localStorage['dataHash'], {gasPrice:web3.toHex($("#gasPrice").val()) * 10**9}, web3Callback);
                         })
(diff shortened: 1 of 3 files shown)


owocki:
thjanks for the PR @thelostone-mc :)

https://github.com/gitcoinco/web/issues/263

doesnt this PR check BOTH the api and web3 to see if the bounty exists.. whereas the solution we *want* in #263 would only check web3?

thelostone-mc:
@owocki I let the api check there so as to avoid the web3 call if it's not available within our db!
I sort of assumed that the web3 check would be secondary check if the api returned a null/error response

Is it an overkill ? :P

thelostone-mc:
@jasonrhaas Do we want the extra safety net on checking up the db ?

jasonrhaas:
so I think this one is OBE, given the new way that we are doing bounty syncs on the back-end.  I'll re-post here when I wrote in Slack to @mbeacom:

> The db auto syncs every x seconds with web3 right?  I’m pretty sure the logic of checking both the db and web3 is already accounted for in the python logic.

I know we are still support the "old way" for the legacy bounties, but I'm not sure if we want to update the legacy code at this point.

mbeacom:
For now, we'll be closing this PR.  Thanks for the contribution @thelostone-mc !

darkdarkdragon:
Issue Status: 1. Open 2. **Cancelled** 

<hr>

__This issue now has a funding of 0.001 ETH (0.69 USD @ $692.1/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](http://localhost:8000/issue/gitcoinco/web/301/454).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $0 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

PR #326, opened 2018-01-30
Adjust email handling for tip flow

### LINKED ISSUE #321: tip emails not going out
example tip that didnt go out: https://github.com/gitcoinco/web/issues/278#issuecomment-361445997

upon further inspection, it looks like the `send_tip_2` view is pulling the SENDER's access key, and primary email, and sending the tips to that => https://github.com/gitcoinco/web/blob/master/app/dashboard/views.py#L255-L268  shouldnt it be pulling the receipient's email instead? 

TODOs 
1. shouldn't it be pulling the receivers email?  and using the access key to receive better privileged access to this information?
2. we don't prompt the user to login via github at any point in this flow.  should we do that so taht `request.session['access_key']` is populated?

mbeacom:
##### Description
The goal of this PR is to resolve an issue with `Tip` notifications not going out to the intended recipient email address.

Additional adjustments include fixing the github auth template to display the login link if they are not authenticated with github and adjusted the github scope to request read-only permissions.

##### Checklist

##### Refers/Fixes
Fixes #321

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -34,7 +34,7 @@ from dashboard.helpers import normalizeURL, process_bounty_changes, process_boun
 from dashboard.models import Bounty, BountySyncRequest, Interest, Profile, ProfileSerializer, Subscription, Tip
 from dashboard.notifications import maybe_market_tip_to_email, maybe_market_tip_to_github, maybe_market_tip_to_slack
 from gas.utils import conf_time_spread, eth_usd_conv_rate, recommend_min_gas_price_to_confirm_in_time
-from github.utils import get_auth_url, get_github_emails, is_github_token_valid
+from github.utils import get_auth_url, get_github_emails, get_github_primary_email, is_github_token_valid
 from marketing.models import Keyword
 from ratelimit.decorators import ratelimit
 from retail.helpers import get_ip
@@ -262,6 +262,7 @@ def send_tip_2(request):
     """
     from_username = request.session.get('handle', '')
     primary_from_email = request.session.get('email', '')
+    access_token = request.session.get('access_token')
     to_emails = []
 
     if request.body:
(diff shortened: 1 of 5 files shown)


app/dashboard/views.py
```diff
@@ -250,9 +250,19 @@ def receive_tip(request):
 @csrf_exempt
 @ratelimit(key='ip', rate='1/m', method=ratelimit.UNSAFE, block=True)
 def send_tip_2(request):
-    """Handle the second stage of sending a tip."""
-    username = request.session.get('handle')
-    primary_email = request.session.get('email', '')
+    """Handle the second stage of sending a tip.
+
+    TODO:
+        * Convert this view-based logic to a django form.
```
owocki: i need to learn more about django forms :P

app/dashboard/views.py
```diff
@@ -262,14 +272,26 @@ def send_tip_2(request):
...
         params = json.loads(request.body)
 
-        username = username or params['username']
-        access_token = request.session.get('access_token')
-        emails = get_github_emails(access_token) if access_token else [primary_email]
+        to_username = params['username'].lstrip('@')
+        try:
+            to_profile = Profile.objects.get(handle__iexact=to_username)
+            if to_profile.github_access_token:
+                to_emails = get_github_emails(to_profile.github_access_token)
```
owocki: is it worth checking `to_profile.email` in this block of code?  as opposed to below

app/dashboard/views.py
```diff
@@ -300,8 +322,8 @@ def send_tip_2(request):
         'class': 'send2',
         'title': 'Send Tip',
         'recommend_gas_price': recommend_min_gas_price_to_confirm_in_time(confirm_time_minutes_target),
-        'from_email': primary_email,
-        'from_handle': username
+        'from_email': primary_from_email,
```
owocki: thx for updating these variable names.. i can see why they were confusing in retrospect

app/dashboard/views.py
mbeacom: :P

app/dashboard/views.py
```diff
@@ -250,9 +250,19 @@ def receive_tip(request):
...
+    TODO:
+        * Convert this view-based logic to a django form.
+
+    Returns:
+        JsonResponse: If submitting tip, return response with success state.
+        TemplateResponse: Render the submission form.
+
+    """
+    from_username = request.session.get('handle', '')
+    primary_from_email = request.session.get('email', '')
```
owocki: should this be `''`?  a few of the recent production tips have had empty email strings... and i think the default here might be causing that.

app/dashboard/views.py
mbeacom: 👍

app/dashboard/views.py
mbeacom: I just added the default `''` to `primary_from_email` - The only reason I could see the `from_email` being empty is if it was unable to fetch the email address from the session.  I'll add a check in to try and catch that.

owocki: 👍

PR #327, opened 2018-01-30
adds user action table, stats cleanup

owocki:
##### Description
This PR adds a `UserAction` table for recording user actions (initially just login / logout) and reporting them to the stats table.

##### Checklist

- [X] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#step-4-commit)

analytics


owocki:
not totally related to this PR's specified, but i put in some fixes for google analytics and KPI processing

owocki:
@mbeacom can i get an 👀  on this pls?

mbeacom:
isort is failing, but other than that lgtm.

app/github/views.py
```diff
@@ -105,6 +112,14 @@ def github_logout(request):
         request.session.pop('access_token_last_validated')
         Profile.objects.filter(handle=handle).update(github_access_token='')
 
+        # record a useraction for this
+        if Profile.objects.filter(handle=handle).count():
```
mbeacom: We could modify this to `get` the specific profile first so we only perform one query versus the second below.

app/marketing/management/commands/pull_stats.py
```diff
@@ -139,6 +139,22 @@ def profiles_ingested():
...
+
+    for action_type in UserAction.ACTION_TYPES:
+        action_type = action_type[0]
+
+        val = UserAction.objects.filter(
+            action=action_type,
+            ).count()
+
+        Stat.objects.create(
+            key='user_action_{}'.format(action_type),
```
mbeacom: This will work, but you could use literal string interpolation via: `key=f'user_action_{action_type}',`
Sadly, `string.format()` is slow :( - f-strings are typically as fast as the old style: `key='user_action_%s' % action_type,`

app/github/views.py
owocki: yes... the reason i didnt do that is because i was worried about generating an exception if for some reason that handle didnt exist.

app/marketing/management/commands/pull_stats.py
owocki: thanks.. i need to get in the habit of doing fstrings

PR #363, opened 2018-02-01
Reduce Win10 Docker Friction

razordaze:
##### Description
```docker-compose up -d``` should be all that's needed to get web running; unfortunately, web_web_1 would have errors complaining about ```\r``` unreadable characters in [docker-command.bash](bin/docker-command.bash), which of course breaks everything. 

Following [git documentation](https://git-scm.com/docs/gitattributes) I created a .gitattributes file to include in version control which should help eliminate this as a potential problem for others. 

Running ```git add --renormalize .```, converted some SCSS / html EOFs to LF. Testing various applications of gitattributes, I was able to discern that the files affected were outliers, and are now how they should be.

##### Checklist

None.

##### Testing
Docker is up in Win10; Mac and Linux use LF natively and won't be affected.

##### Refers/Fixes
Fixed prior to issue creation


mbeacom:
@razordaze Thanks for the contribution!

I appreciate you bringing this issue to our attention.  I'm leaning towards the solution in #365 - I'd like to avoid changing all of these files to run the docker image on Windows, if possible.

I tested it on Windows 10 and macOS high sierra without issue.
The PR resolves the issue by using `dos2unix` during the image build to convert `docker-command.bash`.

Feel free to pull down the branch and test it out, but it should be good to go.

razordaze:
@mbeacom TravisCI would beg to differ. 

This affects a bunch of files by applying --renormalize; I can commit this with the addition of the single file by itself, but those files are going to get normalized at some point, so why not just bit the bullet now?

mbeacom:
@razordaze Travis is failing due to dependency conflicts for the package:  `eth-utils` - See: https://github.com/gitcoinco/web/issues/367
The `web` container runs correctly without error with the introduction of https://github.com/gitcoinco/web/pull/365

I'm not sure why we need to normalize these?  If you were to clone the repository like:
`git clone git@github.com:gitcoinco/web.git --config core.autocrlf=input` on Windows, the docker stack would be successful without the changes in #365 

Can you elaborate on why we need to normalize these files?

mbeacom:
Per our previous discussion, closing this for now.  Feel free to ping me if you have any questions, concerns, or would like to implement something similar in the future!

Thanks for the contribution!

Issue #268, opened 2018-01-23
standardbounties: automatically cross-process bounties posted to bounties.network to gitcoin


owocki:
acceptance criteria:  automatically cross-process bounties posted to bounties.network to gitcoin

owocki:
@mbeacom @jasonrhaas either of you want to take this on next?  i feel like it's an important piece of scope that we missed in v1

mbeylin:
The low hanging fruit implementation of this would be allowing the Gitcoin app to support fulfillments which are submitted on the Bounties Network, and allow bounty creators to accept them. This means that non-code bounties can have deliverables which aren't given on a github repo, but through other formats (files, links).

jasonrhaas:
The question I have is:  Do we want all the bounties posted on bounties.network to show up in Gitcoin?  Or should there be some kind of Gitcoin Tag on the bounties.network post that shows it as a Gitcoin task?  We can use the `meta.schema` field that we set up in IPFS for that.

owocki:
maybe initially we just filter through and process ones that adhere to the gitcoin schema...

it it could be built such that it was extensible to other schemas.. thatd be 🌮

mbeylin:
To add, when bounties are created on BN we let users select one of several categories, and the 2 related to gitcoin are `code` and `bugs`. These are also the only bounties on BN which will ask users to input a github url (stored as adheres to the schema). My advice would be to support bounties which include either the `code` or `bugs` keyword, with a github URL which matches Gitcoin's platform (i.e. a url that links to a github issue).

jasonrhaas:
So when a new bounty is created on BN, what will the meta data be?  Will the `platform` still be `bountiesNetwork` or will it be Gitcoin?  In other words, when Gitcoin is looking for bounties, should it look for the keywords as you say?  Or just the `meta.platform`?  If we do the latter than it is on BN to determine that issue applies to Gitcoin.

```
  meta: {
    platform: // a string representing the original posting platform (ie 'gitcoin')
    schemaVersion: // a string representing the version number (ie '0.1')
    schemaName: // a string representing the name of the schema (ie 'standardSchema' or 'gitcoinSchema')
  }
```

In terms of implementation, I suppose we would need some kind of micro service to:

- Listen for new transactions on the Standard Bounties contract
- Determine if it applies to Gitcoin.  Whether by keyword or `meta.platform`.
- Run the `sync_web3()` function on Gitcoin to update the database with the data from the contract/IPFS

[...]

Issue #128, opened 2017-12-15
As a user, I'd like to be able to close my browser window after I submit a web3 tx, so I can do other things without canceling.


owocki:
014429d8ce15a484e59e3ad7760be5475e455cb0

# the issue

Once a user submits a new bounty / claim bounty / accept bounty, their browser is what polls web3 for the tx to be mined.  When the tx is mined, their browser receives the updated info from the blockchain, and submits it to the server.

this is a problem because if the user closers browser window, or if they lose web3 connectivity, the backend won't update.

# workaround

a workaround is if an admin is using the `process_bounty.js` helper script to process the bounty.

# proposed root cause fix

there is a serverside management command called `sync_geth`.  it is partially implemented right now.  when it is fully implemented, it will poll / listen for changes to the jobs on web3 and update the server side immediately

gitcoinbot:
__This issue now has a funding of 0.05 ETH (33.89 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/128).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/128)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $10303.76 more Funded OSS Work Available at: https://gitcoin.co/explorer

wingleung:
Where can I find the partially implemented `sync_get` command?

owocki:
@wingleung https://github.com/gitcoinco/web/blob/master/app/dashboard/management/commands/sync_geth.py

sc0Vu:
@owocki Maybe it's better to run sync_geth as daemon on serverside?

owocki:
sure.  im open to that... we could put it into `supervisorctl` but first we need to get it working

step21:
Is someone working on this? Just curious, as I think I am not that familiar with the codebase to decide which way to handle this.

abitrolly:
Nobody claimed it, so feel free to do it. My question are - how state is stored and what is stored?

Is there something like this stored on server?

    change requested -> transaction pending -> transaction completed
                                            -> transaction cancelled

So that when you reload the page, the state is shown on issue page?

step21:
You can look at /app/dashboard/models.py for example to get an idea of the database layout.
[...]

PR #364, opened 2018-02-02
serverside syncing of bounties

owocki:
# What

Moves the processing of bounties from web3 => database into the server side, as opposed to client side.

For maintability/performance reasons..

# Testing

* Tested end to end submission of bounties
* Tested `./manage.py sync_geth mainnet`

# Refers

https://github.com/gitcoinco/web/issues/128
https://github.com/gitcoinco/web/issues/268


owocki:
@mbeacom @jasonrhaas can i have your review here?

app/dashboard/helpers.py
```diff
@@ -325,7 +345,7 @@ def process_bounty_details(bountydetails, url, contract_address, network):
             expires_date=timezone.make_aware(timezone.datetime.fromtimestamp(bounty.get('deadline')), timezone=UTC),
             standard_bounties_id=bountyId,
             balance=bounty.get('balance'),
-            num_fulfillments=fulfillments.get('total', 0),
+            num_fulfillments=len(fulfillments),
```
jasonrhaas: This will fail if `fulfillments` is None.  What happens if the bounty doesn't have any fulfillments yet?

app/dashboard/management/commands/sync_geth.py
```diff
@@ -15,117 +15,47 @@
...
+        # iterate through all the bounties
+        bounty_enum = 0
+        more_bounties = True
+        while more_bounties:
+            try:
+
+                # pull and process each bounty
+                bounty = get_bounty(bounty_enum, network)
+                print(f"Processing bounty {bounty_enum}")
+                process_bounty(bounty)
```
jasonrhaas: Maybe break this out into an `else` statement.  Which statement are you concerned will fail?  Normally I put that one statement in the `try`, and then the `else` is the code that will run if the code under `try` works.

app/dashboard/management/commands/sync_geth.py
```diff
@@ -15,117 +15,47 @@
...
+
+            except BountyNotFoundException:
+                more_bounties = False
+            except UnsupportedSchemaException as e:
+                print(f"* {e}")
```
jasonrhaas: General comment:  I think we should get in the habit of putting all these statements to a `logger` with the appropriate level (DEBUG, INFO, WARNING, ERROR).  And then handle which logging gets displayed with the Django Config.

owocki:
@mbeacom do we have to do anything to get `ipfs` installed on the docker container?  https://ipfs.io/docs/install/

app/dashboard/utils.py
```diff
@@ -0,0 +1,194 @@
...
+    standardbounties_addr = getStandardBountiesContractAddresss(network)
+    bounty_abi = json.loads(standardbounties_abi)
+    getBountyContract = web3.eth.contract(standardbounties_addr, abi=bounty_abi)
+    return getBountyContract
+
+
+def get_bounty(bounty_enum, network):
+    try:
+        ipfs = getIPFS()
```
jasonrhaas: Same comment with this `try` statement.  I try to make the `try` statement pretty thin to clearly show what is being tested.

app/dashboard/utils.py
```diff
@@ -0,0 +1,194 @@
...
+        issuer, deadline, fulfillmentAmount, paysTokens, bountyStage, balance = standard_bounties.functions.getBounty(bounty_enum).call()
+        bountydata = standard_bounties.functions.getBountyData(bounty_enum).call()
+        arbiter = standard_bounties.functions.getBountyArbiter(bounty_enum).call()
+        token = standard_bounties.functions.getBountyToken(bounty_enum).call()
+        fulfillments = []
+        fulfill_enum = 0
+        more_fulfillments = True
+        while more_fulfillments:
+            try:
+                accepted, fulfiller, data = standard_bounties.functions.getFulfillment(bounty_enum, fulfill_enum).call()
```
jasonrhaas: Instead of just looping through the fulfillments until there is an error, could we use `getNumFulfillments()` first?  And then just iterate N times?

owocki:
hmmm not sure how to fix travis.. 

https://travis-ci.org/gitcoinco/web/builds/336738688?utm_source=email&utm_medium=notification

seems like pytest failed because of a missing dependancy..

app/dashboard/utils.py
```diff
@@ -0,0 +1,194 @@
...
+def has_tx_mined(txid, network):
+    web3 = getWeb3(network)
+    try:
+        transaction = web3.eth.getTransaction(txid)
+        return transaction.blockHash != HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000')
+    except:
+        return False
+
+
+def getBountyID(issueURL, network):
```
jasonrhaas: Looks like this tries the db first, and then web3.  What happens if they get out of sync or something?  Should it try to reconcile the db and web3 before returning anything?

app/dashboard/views.py
```diff
@@ -542,43 +544,51 @@ def save_search(request):
...
-            result['status'] = 'OK'
-            for existing_bsr in BountySyncRequest.objects.filter(github_url=issueURL, processed=False):
-                existing_bsr.processed = True
-                existing_bsr.save()
+    result = {
+        'status': '400',
+        'msg': "bad request"
+    }
+
+    issueURL = request.POST.get('url', False)
```
jasonrhaas: Is there a reason to use `False` here, instead of the default value of `None`?

requirements/base.txt
```diff
@@ -21,7 +21,7 @@ PyPdf2
 python-twitter==3.2
 sendgrid
 slackclient
-web3
+web3==4.0.0b7
```
mbeacom: I think your travis builds are failing due to the dependencies specified for the 4.x beta.  Can you pin to `3.16.5` and see if that resolves the build failures?

owocki: no.. unfortunately i cant because i wanted to use some of the new syntax / features of web3py4... 

are there dependancy conflicts with other packages in the requirements.txt?  

IIRC i had to `pip install --pre web3` to get this to work..

app/dashboard/utils.py
owocki: > Looks like this tries the db first, and then web3.

yeah, for performance reasons this is _way_ faster

>  What happens if they get out of sync or something?

Given that the bounty ids increment on the smart contract... I dont see how they could get out of sync.   

Maybe if there was a new version of standardbounties.. but in that case we'd have to migrate the old bounties over to that contract and we'd incur a migration cost anyway.

app/dashboard/helpers.py
owocki: `fulfillments` will never be `None` because of the way fulfillments are constructed in `dashboard/utils#L91`.  they will also be class `list`.

requirements/base.txt
mbeacom: https://github.com/ethereum/eth-abi/issues/25

owocki: you are a machine.. thanks for reporting that upstream.. and wow, theyve already handled it!! :)

app/dashboard/views.py
owocki: updated

app/dashboard/utils.py
owocki: good idea.. updated

app/dashboard/utils.py
owocki: updated

app/dashboard/management/commands/sync_geth.py
owocki: updated

app/dashboard/management/commands/sync_geth.py
owocki: i dont think i can do this with this one... the exception handler is meant for the entire block

owocki:
closing because its merged into #379

PR #396, opened 2018-02-12
easier tip redemption

owocki:
##### Description

makes it so that users can redeem their tips straightaway from gitcoin, instead of having to check their email.

example:

screen shot 2018-02-12 at 2 15 40 pm

this solves a problem wherein we cannot find the users email address via github api, insofar that users can now redeem tips without having to get the email.

##### Checklist

frontend

##### Testing
- i tested it end to end


owocki:
1e5d9eef211f9ef42bda0c0080bdda05bc828e4a

thx, i made this commit locally but forgot to push it up remote

PR #513, opened 2018-03-01
record user actions

owocki:
##### Description
Records `UserAction` events for Tip, Start/Stop Work, and Bounty related events

##### Checklist

Analytics


app/dashboard/helpers.py
```diff
@@ -467,6 +467,29 @@ def process_bounty_details(bounty_details):
     return (did_change, latest_old_bounty, latest_old_bounty)
 
 
+def record_user_action(event_name, old_bounty, new_bounty):
+    user_profile = None
+    try:
+        if event_name in ['new_bounty', 'killed_bounty', 'work_done']:
+            user_profile = Profile.objects.get(handle=new_bounty.bounty_owner_github_username)
+        if event_name in ['work_submitted']:
+            fulfillment = bounty.fulfillments.order_by('pk').first()
```
mbeacom: `bounty` is undefined.

app/dashboard/helpers.py
```diff
@@ -467,6 +467,29 @@ def process_bounty_details(bounty_details):
...
+            user_profile = Profile.objects.get(handle=fulfillment.fulfiller_github_username)
+    except Exception as e:
+        logging.error(f'{e} during record_user_action for {new_bounty}')
+        # TODO: create a profile if one does not exist already?
```
mbeacom: We could handle this with a `get_or_create` with something like:

```python
def record_user_action(event_name, old_bounty, new_bounty):
    user_profile = None

    try:
        if event_name in ['new_bounty', 'killed_bounty', 'work_done']:
            handle = new_bounty.bounty_owner_github_username
        elif event_name in ['work_submitted']:
            fulfillment = new_bounty.fulfillments.order_by('pk').first()
            handle = fulfillment.fulfiller_github_username

        user_profile, _ = Profile.objects.get_or_create(
            handle=handle,
            defaults={'data': {}}
        )
    except Exception as e:
        logging.error(f'{e} during record_user_action for {new_bounty}')

    if user_profile:
        UserAction.objects.create(
            profile=user_profile,
            action=event_name,
            metadata={
                'new_bounty': new_bounty.pk if new_bounty else None,
                'old_bounty': old_bounty.pk if old_bounty else None,
            })
```

This reminds me...  We could probably update `sync_profile` to use `Profile.objects.update_or_create()` and return the profile.

app/dashboard/helpers.py
```diff
@@ -505,9 +528,13 @@ def process_bounty_changes(old_bounty, new_bounty):
...
         event_name = 'unknown_event'
         logging.error(f'got an unknown event from bounty {old_bounty.pk} => {new_bounty.pk}: {json_diff}')
-    
+
     print(f"- {event_name} event; diff => {json_diff}")
 
+    # record a useraction for this
+    record_user_action(event_name, old_bounty, new_bounty)
+
+
```
mbeacom: Can we drop the second new line here?

app/dashboard/views.py
```diff
@@ -63,6 +63,36 @@ def send_tip(request):
     return TemplateResponse(request, 'yge/send1.html', params)
 
 
+def record_user_tip_action(profile_handle, event_name, tip_pk):
```
mbeacom: Could we get away with something like this?

```python
def record_user_action(profile_handle, event_name, instance):
    instance_class = instance.__class__.__name__.lower()

    try:
        user_profile = Profile.objects.get(handle=profile_handle)
        UserAction.objects.create(
            profile=user_profile,
            action=event_name,
            metadata={
                f'{instance_class}_pk': instance.pk,
            })
    except Exception as e:
        # TODO: sync_profile?
        logging.error(f"error in record_action: {e} - {event_name} - {instance}")
```

app/dashboard/helpers.py
owocki: 75955a499bc59cb67ea7a914736d9678314ae391

app/dashboard/helpers.py
owocki: > We could handle this with a get_or_create with something like:

+1 on this idea.. its a very good one.  i almost feel like we could do one-better and call `sync_profile()` after the get_or_create so that way we have the `data` `repo_data` etc fields available.  

here is a commit that does so.. what do you think? 60799407a849a5ea3ea697e511f85f08eb33172d

> This reminds me... We could probably update sync_profile to use Profile.objects.update_or_create() and return the profile.

doesnt the get_or_create in that method prettymuch accomplish this already though?  or you mean that it should update the `data` `repo_data` etc fields?  seems like this lines alraedy does this.. although maybe less performantly:

```
    org.handle = handle
    org.data = data
    org.repos_data = repos_data
    org.save()
```

app/dashboard/views.py
owocki: hows this? 5a44b0d81dac184b93a731d37c5fc10022fbaa29

app/dashboard/helpers.py
mbeacom: Can you rebase with master? I pushed a change to `sync_profile` last night that uses `update_or_create` and returns the profile.

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -344,9 +344,10 @@ pull_interest_list(result['pk'], function(is_interested){
     }
     actions.push(entry);
 
-    var is_expired = result['status']=='expired' || (new Date(result['now']) > new Date(result['expires_date']));
-    var is_done = result['status']=='done';
-    if(!is_done && !is_expired){
+    var is_date_expired = (new Date(result['now']) > new Date(result['expires_date']));
```
owocki: hmm... this is in master.. should not be in this PR

app/dashboard/helpers.py
owocki: 9284b2fb024e9b312447929ef7c08a170c85ba7a

app/dashboard/views.py
```diff
@@ -63,6 +63,21 @@ def send_tip(request):
     return TemplateResponse(request, 'yge/send1.html', params)
 
 
+def record_user_action(profile_handle, event_name, instance):
+    instance_class = instance.__class__.__name__.lower()
```
mbeacom: 🙌

owocki:
i think this PR needs more testing before merging... was just testing and found some issues

doing dinner now.. can test more later tonight / tomorrow

app/app/utils.py
```diff
@@ -48,32 +48,45 @@ def add_contributors(repo_data):
...
-    is_error = 'name' not in data.keys()
-    if is_error:
-        print("- error main")
-        rollbar.report_message('Failed to fetch github username', 'warning', extra_data=data)
-        return None
-
-    repos_data = get_user(handle, '/repos')
-    repos_data = sorted(repos_data, key=lambda repo: repo['stargazers_count'], reverse=True)
-    repos_data = [add_contributors(repo_data) for repo_data in repos_data]
+def get_profile(handle, sync=True):
```
owocki: i forgot that the `sync_profiles` management command already takes care of syncing (and does it out the context of a request... which is great.

i think this PR would be massively simplified if i were to remove all the changes related to sync_profile/get_profile and just add the `record_user_action` stuff.  planning on doing that soon

PR #559, opened 2018-03-07
css : fixed display for devices below 380px

thelostone-mc:
##### Description

 noticed that a few devices have width 380px which causes the github comment section to break
screen shot 2018-03-07 at 11 50 23 pm

Fixed that up ! It's not needed but meh why not

screen shot 2018-03-07 at 11 50 51 pm

### DIFF
--- a/app/assets/v2/css/dashboard.css
+++ b/app/assets/v2/css/dashboard.css
@@ -108,3 +108,9 @@ body {
     font-size: 10px;
   }
 }
+
+@media (max-width: 380px) {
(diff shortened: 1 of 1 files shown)


mkosowsk:
Finally I can browse Gitcoin on my device of choice:

calculator 2

Great work @thelostone-mc 👍

thelostone-mc:
@mkosowsk  I can't think of a comeback for this. Damn :P

owocki:
epic troll

PR #566, opened 2018-03-08
mocks expiration date on standardbounties

owocki:
##### Description
implements option 3 that @mbeylin suggested here => https://github.com/Bounties-Network/StandardBounties/issues/25

> the deadline you enter on gitcoin is stored off-chain (in the JSON object that's on IPFS), and the on-chain deadline is just max_int or something like that to ensure it never really expires

##### Checklist

- [X] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#step-4-commit)

* sync_geth
* frontend

##### Testing
* i have tested it 

##### Refers/Fixes
* https://github.com/Bounties-Network/StandardBounties/issues/25


owocki:
@mbeacom can i get 👀 on this?

app/dashboard/utils.py
```diff
@@ -172,6 +172,11 @@ def get_bounty(bounty_enum, network):
     if 'Failed to get block' in bounty_data_str:
         raise IPFSCantConnectException("Failed to connect to IPFS")
 
+    # https://github.com/Bounties-Network/StandardBounties/issues/25
+    override_deadline = bounty_data.get('payload', {}).get('expire_date', False)
```
mbeacom: Will the override ever be `False`?

app/assets/v2/js/pages/new_bounty.js
```diff
@@ -264,11 +267,11 @@ $(document).ready(function(){
             var _paysTokens = !isETH;
             var bountyIndex = bounty.issueAndActivateBounty(
                 account,            // _issuer
-                expire_date,        // _deadline
+                mock_expire_date,   // _deadline
```
mbeacom: In this case, we always want to use the `mock_expire_date` here?

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -345,7 +345,7 @@ pull_interest_list(result['pk'], function(is_interested){
         actions.push(entry);
     }
 
-    if(result['status']=='open' || result['status']=='started' || result['status']=='submitted' ){
+    if(result['status']=='open' || result['status']=='started' || result['status']=='submitted' ){ // TODO -- do we want to change how this functions if expired due to https://github.com/gitcoinco/web/pull/566 ?
```
mbeacom: 🤔

app/dashboard/utils.py
owocki: yes.. for bounties that were submitted on the bounties network or on gitcoin, but before this PR is merged

app/assets/v2/js/pages/new_bounty.js
owocki: i believe so... unless we want to provide the user the optional of actually expiring their bounty.. which i think might be too many explanatory overhead.. maybe pixelant will disagree

app/assets/v2/js/pages/bounty_details.js
owocki: i think the answer is that if theres a mock expiration date we want to display all these options even if the bounty is expired... sound okay to you?

PR #579, opened 2018-03-09
removes optional add_bcc option from mailer

owocki:
This can be done straight away in sendgrid using https://app.sendgrid.com/settings/mail_settings


owocki:
🆗 to ship?

PR #582, opened 2018-03-09
General faucet cleanup

mbeacom:
##### Description

The goal of this PR is to cleanup some of the logic and syntax associated with `Faucet` and add docstrings.

##### Checklist

faucet

### DIFF
--- a/app/faucet/views.py
+++ b/app/faucet/views.py
@@ -1,10 +1,29 @@
+# -*- coding: utf-8 -*-
+"""Define faucet views.
 
+Copyright (C) 2018 Gitcoin Core
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


mbeacom:
@owocki When you have a moment, can I get 👀 on this?

app/faucet/views.py
```diff
@@ -24,98 +43,63 @@ def faucet(request):
...
-        'status': 200,
-        'user': False
-      }
-    else:
-       response = {
-        'status': 200,
-        'user': user['items'][0]
-      }
+    response = {'status': 200, 'user': False}
+    if not len(user['items']) == 0 or not user['items'][0]['login'].lower() != profile.lower():
```
owocki: this looks different than what was being done before?

app/faucet/views.py
```diff
@@ -24,98 +43,63 @@ def faucet(request):
...
 
 @csrf_exempt
 def save_faucet(request):
-    try:
-        validate_slug(request.POST.get('githubProfile'))
-    except Exception as e:
-        return JsonResponse({
-          'message': e.messages[0]
-        }, status=400)
+    github_profile = request.POST.get('githubProfile')
```
owocki: did not test here either.. can you do a runthrough of testing?

owocki:
just left a few comments.. LGTM once its tested both success/fail cases

app/faucet/views.py
mbeacom: It is.  I'm checking if not for both cases, defaulting to `{'status': 200, 'user': False}` and only updating `response['user']` if the case is correct ^  - It's performing the same logic, just updating the one value if necessary to be more DRY.

app/faucet/views.py
mbeacom: It's doing the same exact thing, only catching in one place versus three since we're returning on exception anyway.

mbeacom:
@owocki The logic hasn't really changed. As for the python code, it's the same exact thing as before only a little more DRY.

owocki:
🚢 ed-- thanks!

Issue #651, opened 2018-03-19
Where are our users located geographically?


owocki:
per @vs77bb convo today

mkosowsk:
@vs77bb As everyone knows and despite outrageous claims to the contrary, I am a proud resident of NYC 👍

mbeacom:
https://docs.djangoproject.com/en/1.11/ref/contrib/gis/geoip2/

e-n-l:
Tag this on to profile, as a form field?

mbeacom:
User location data is now accessible via `app.utils.get_location_from_ip`.  We haven't added a location field anywhere other than `UserAction.location_data` at the moment.

Via #657 we can retrieve location data from the user's IP in any request: `loc = get_location_from_ip(get_real_ip(request))` and user language code via: `request.LANGUAGE_CODE`.

PR #657, opened 2018-03-19
Add GeoIP2 and lang middleware

mbeacom:
##### Description

The goal of this PR is to add geoip2 lookup based on IP address, store in the session, and track it on `UserAction`.

Added fields:

- Attr: `UserAction.ip_address` - `GenericIPAddress`
- Attr: `UserAction.geolocation_data` - `JSON`

Modified logic:

- Github middleware and view mechanisms to store geolocation and IP data to the session
- `UserAction.create` usage throughout app has implemented the new fields above.
- Added `django.middleware.locale.LocaleMiddleware` django middleware for accessing: `request.LANGUAGE_CODE`

Moving forward:

- Determine a forever home for the geoip2 logic.
- Decide on the best place to store location data longterm. (Possibly the `User` once django auth goes in).

##### Checklist

backend

##### Testing

Tested these changes locally with GH integration.

##### Refers/Fixes

Ref #651


Dockerfile
```diff
@@ -2,15 +2,35 @@ FROM python:3.6-slim-jessie
...
+    ./bootstrap && \
+    ./configure && \
+    make && \
+    make check && \
+    make install && \
+    echo /usr/local/lib  >> /etc/ld.so.conf.d/local.conf && \
+    ldconfig
+
+# GeoIP2 Data Files
+RUN mkdir -p /usr/share/GeoIP/ && \
```
owocki: hopefully these steps dont take too long and/or they're not too brittle!

app/github/middleware.py
```diff
@@ -32,9 +39,15 @@ def process_response(self, request, response):
         token = request.session.get('access_token')
         expiration = request.session.get('access_token_last_validated')
         handle = request.session.get('handle')
+        ip_address = '24.210.224.38' if settings.DEBUG else get_real_ip(request)
```
owocki: note: we are behidn a load balancer and cloudfront in prod.. so we'll need to make sure `get_real_ip()` works with that

owocki:
didnt test.. but code all looks sane to me.  hopefully theres not too much of a performance burden of the middleware

app/github/middleware.py
mbeacom: @owocki As long as nginx is setup to forward the IP addresses appropriately, we'll be fine.  I'll take a look at the nginx config and find out.

Dockerfile
mbeacom: I wouldn't say they're brittle, but it does add a little extra time to the build... but they are in their own cache layer, so the user should only need to pull these files once per build environment (until they clear docker cache/system images).

mbeacom:
@owocki If you'd like, I can modify the geoip logic to only grab the user's location on login/logout and store it as is for the duration of the session to avoid possible performance issues.
In my past experience, I never encountered any performance issues by handling geoip lookups via middleware, but we can definitely make the changes if it's a concern.

mbeacom:
@owocki Per our discussion about performance and impact, I've removed the middleware tracking of geolocation and opted to only record location and IP when the user logs in or logs out.

Additionally, this PR adds the `Locale` django middleware that now provides us with `request.LANGUAGE_CODE` which returns the user's language code, for example: `en-us`

This should be good to merge, unless you have objections.

PR #662, opened 2018-03-20
no faucet spoofing

owocki:
doesnt allow a user to spoof their github username by editing the DOM

also, fixes that `e.messages` does not exist.

```
ipdb> e.messages[0]
*** AttributeError: 'Exception' object has no attribute 'messages'
```

### DIFF
--- a/app/faucet/views.py
+++ b/app/faucet/views.py
@@ -66,8 +66,11 @@ def save_faucet(request):
         validate_slug(github_profile)
         validate_email(email_address)
         validate_slug(eth_address)
+
+        if github_profile != request.session.get('handle'):
+            raise Exception("Could not authenticate your github profile")
     except Exception as e:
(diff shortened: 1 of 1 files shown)


owocki:
i dont seem to be able to test this since login is broken on my local since the .env changes were merged.. i seem to be able to redirect back to my localhost, but the `request.session.get('handle')` is never set.  need to investigate in the morning

owocki:
got it! my local.env `SESSION_COOKIE_SECURE=on` when it should have been `SESSION_COOKIE_SECURE=off`

mbeacom:
Looks like this was already merged in: #661

owocki:
> Looks like this was already merged in: #661

i merged it and reverted it after i realized it was late at night and i should wait for code review before letting it in master.

PR #738, opened 2018-03-29
bounty: display submit only on starting work

thelostone-mc:
##### Description

x

##### Checklist

- bounty

@mbeacom This is a better approach that we discussed yesterday :P 

@PixelantDesign

### DIFF
--- a/app/assets/v2/css/bounty.css
+++ b/app/assets/v2/css/bounty.css
@@ -3,6 +3,10 @@ body {
   background-color: #FFF;
 }
 
+.none {
+  display: none;
+}
+
 .explorer {
   font-family: "Muli-SemiBold";
   margin-top: 2em;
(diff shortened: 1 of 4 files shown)


thelostone-mc:
@mbeacom The green success alert doesn't appear on clicking start work immediately  cause we display the alert only after we update db get a 200 response! See [here](https://github.com/thelostone-mc/web/blob/b8e8e01a78fc87a64f1fc2d27c5803b3b393b85b/app/assets/v2/js/shared.js#L195)

Initially I added the toggle submit class here, but it feels slow on my local, so moved it to [here](https://github.com/thelostone-mc/web/blob/b8e8e01a78fc87a64f1fc2d27c5803b3b393b85b/app/assets/v2/js/shared.js#L190) !
Could you check if it's slow on staging ? If so, we could move the alert also above (the downside being that alert will trigger even if the [POST call](https://github.com/thelostone-mc/web/blob/b8e8e01a78fc87a64f1fc2d27c5803b3b393b85b/app/assets/v2/js/shared.js#L191) fails

PixelantDesign:
on open issues, users should only see the start button. Please remove submit work.
screen shot 2018-03-30 at 12 20 46 am

thelostone-mc:
@mbeacom is it possible that when @PixelantDesign checked out the staging, this commit was not longer deployed there ?

Issue #710, opened 2018-03-26
in my settings, i should be able to disguise myself from the leaderboard/having a profile


owocki:
in my settings, i should be able to disguise myself from the leaderboard/having a profile

mkosowsk:
Big 👍 to this, lots of folks value their privacy in the crypto community and this is especially important with the Facebook disaster currently happening in slow motion.

I think this should be opt-in, wouldn't want to expose someone who doesn't want to put their info out there

PixelantDesign:
@owocki What problem are we solving for? 
A user doesn't want to be displayed on the leaderboard because they want privacy (rank & amount they've made)? Or a user doesn't want to have a Gitcoin profile at all?  How do you see this working?

owocki:
> A user doesn't want to be displayed on the leaderboard because they want privacy (rank & amount they've made)

Right now, this is what I've coded up

> Or a user doesn't want to have a Gitcoin profile at all? 

I think that we could perhaps enable this option.  Or allow the user to just hide their bounties from their public profile too

PixelantDesign:
@owocki Do we need to find a place for this in #656 or does it make sense for this to live in settings? maybe settings makes more sense.

owocki:
@PixelantDesign https://github.com/gitcoinco/web/issues/795 breaks up settings into hte following tabs: /settings/email, /settings/privacy, /settings/matching, 'settings/feedback' -- which i think seperates concerns enough for the time being

PR #745, opened 2018-03-29
in my settings, i should be able to disguise myself from the leaderboard/having a profile

owocki:
##### Description
in my settings, i should be able to disguise myself from the leaderboard/having a profile

also sends an email when a user puts feedback in the comments

In the future, we should maybe break apart the email settings / gitcoin profile settings (as there are two models: EmailSubscriber / Profile models)

##### Checklist

email settings, profile

##### Testing
tested it

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/710


app/external_bounties/management/commands/sync_bounty0x.py
```diff
@@ -43,6 +43,7 @@ def handle(self, *args, **options):
 
         response = requests.get(url, headers=headers)
         bounties = response.json()
+        print(len(bounties))
```
thelostone-mc: @owocki  take this away ?

app/marketing/management/commands/assemble_leaderboards.py
```diff
@@ -33,6 +33,7 @@
 quarterly_cutoff = timezone.now() - timezone.timedelta(days=90)
 yearly_cutoff = timezone.now() - timezone.timedelta(days=365)
 
+
```
thelostone-mc: @owocki  extra line added ?

owocki:
> In the future, we should maybe break apart the email settings / gitcoin profile settings (as there are two models: EmailSubscriber / Profile models)

Any thoughts on this?  I'm gonna take another pass here this afternoon and it'd be good to have this fleshed out.

app/marketing/management/commands/assemble_leaderboards.py
owocki: i think 2 line breaks is the standard before top level methods in pep8

app/external_bounties/management/commands/sync_bounty0x.py
owocki: da73e3cd79cbdc8e7114e8bdde8577571b455a68

owocki:
> In the future, we should maybe break apart the email settings / gitcoin profile settings (as there are two models: EmailSubscriber / Profile models)


follow up ticket for the above

https://github.com/gitcoinco/web/issues/795

owocki:
i went in to resolve some of the merge conflicts and it was borderline impossible to work inside the view as it was.

769033eb1b020a9a98f11b65fff7011da726ea06 <=== ended up just implementing https://github.com/gitcoinco/web/issues/795 so that we could work with the information in a sane way in the future

Issue #506, opened 2018-02-27
Code - /pitch pages


owocki:
# Why 

Because there's good ideas out there, and the community needs to know about them.

Because there's *bad* ideas out there, and the community needs to help the pitcher evolve them.

#  What

## First page

Page that lives at `ideas/new`.  Page has a web form
* Email
* Name
* Github username
* Summarize your idea (short text input)
* Tell us more (long markdown input)
* What are you looking for?  radio input (select all that apply)
** capital
** builders
** designers
** customers
* What are do you have already?  radio input (select all that apply)
** capital
** builders
** designers
** customers

When it's submitted,

* an email goes to settings.CONTACT_EMAIL
* the data is stored in the DB as a new object `Pitch`
* slackbot posts to `community-general` on gitcoin slack with the title of the idea and a link to it

## Second Page

This page lives at `ideas/title-url-encoded` and is the 'view' page for the idea.  It will display the same information as the form, and it will also give the end user the ability to upvote / comment on the idea. one upvote per ip address.

## Third Page

This page is the index page of all of the ideas, and it lives at `ideas/

## Designs

https://github.com/gitcoinco/web/issues/198

[gitcoin-ideas.zip](https://github.com/gitcoinco/web/files/1763493/gitcoin-ideas.zip)

![36632889-41740d76-195a-11e8-9041-9cc5c80abd65](https://user-images.githubusercontent.com/513929/36737526-8094b52e-1b98-11e8-9104-892ba476757f.png)
![36632883-3255e5f8-195a-11e8-8300-b66427fd2d09](https://user-images.githubusercontent.com/513929/36737527-80b2e2a6-1b98-11e8-9c16-c0a1d118b4fc.png)
![36632886-36a49014-195a-11e8-857c-1d05c7a69063](https://user-images.githubusercontent.com/513929/36737528-80cad8a2-1b98-11e8-85b0-e0138ee6b38c.png)

# Bounty

* Code the frontend (HTML/CSS/JS) for the above three pages (must support tablet + mobile + desktop views)
* Code the backend (DB models) for the above three pages too.
* PR back to the web repo.

gitcoinbot:
__This issue now has a funding of 0.22 ETH (193.79 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/gitcoinco/web/issues/506).
[...]

PR #753, opened 2018-03-30
pitch pages

kziemianek:
Refs: https://github.com/gitcoinco/web/issues/506

##### Description
Add pitch pages as described #506 

##### Checklist

- frontend
- backend

##### Testing
Performed manual local tests.
##### Refers/Fixes
Refs #506


app/retail/templates/idea_template.html
```diff
@@ -0,0 +1,51 @@
+<script id="idea_template" type="text/x-jsrender">
```
kziemianek: this file needs to be formated properly ...

app/retail/views.py
```diff
@@ -459,6 +463,88 @@ def browser_extension_chrome(request):
...
+@require_POST
+@csrf_exempt
+def create_idea(request):
+    idea = Idea(
+        full_name=request.POST.get('fullName'),
+        email=request.POST.get('email'),
+        github_username=request.POST.get('githubUsername'),
+        summary=request.POST.get('summary'),
+        more_info=request.POST.get('moreInfo'),
+        looking_for_capital=True, #request.POST.get('lookingForCapital') returns 'fasle' or 'true'
```
kziemianek: I have no idea how to serialize properly javascript's true/false to python's True/False, any ideas?

kziemianek:
Actual implementation fetches disqus thread's data on client side (most recent values, api limits may be exceeded) and there are no likes and posts counts in model so there is no way to perform sort. 

To provide sorting functionality these values should be placed in model but syncing them with disqus may be problematic. 

I can think of two syncing modes:
1. Lazy - values are updated periodically(no risk to exceed api limits, values may be outdated)
2. Eager - values are updated during ideas list/idea fetch(api limits may be exceeded, most recent values, longer request times).

@mbeacom , @owocki does actual implementation or any of syncing modes satisfy You ?

app/retail/views.py
mkosowsk: Per https://stackoverflow.com/questions/10693630/how-to-pass-a-boolean-from-javascript-to-python

mkosowsk: Per https://stackoverflow.com/questions/10693630/how-to-pass-a-boolean-from-javascript-to-python

mkosowsk: Per https://stackoverflow.com/questions/10693630/how-to-pass-a-boolean-from-javascript-to-python

mkosowsk: Per [this](https://stackoverflow.com/questions/10693630/how-to-pass-a-boolean-from-javascript-to-python )

> looking_for_capital = True if request.POST.get('lookingForCapital') == 'true' else False
or
> looking_for_capital = request.POST.get('lookingForCapital') == 'true'

might be a good solution 🤔

mkosowsk: Per [this](https://stackoverflow.com/a/10693668)

> looking_for_capital = True if request.POST.get('lookingForCapital') == 'true' else False

or

> looking_for_capital = request.POST.get('lookingForCaptail') == 'true'

kziemianek: Yeah i saw this solution, i wonder if there is another...

kziemianek: I mean framework should provide proper deserializer, no manual checks spread across the code

mkosowsk:
Looks great!!! :D

I defer to @owocki and @mbeacom but think option number 2 of values are updated during ideas list/idea fetch would be interesting to run with to start. Nice to have those recent values as part of feedback for the user and I don't think we'll be bumping up against API limits to start 🤔

app/retail/views.py
kziemianek: Imma try this tomorrow: https://stackoverflow.com/questions/15476983/deserialize-a-json-string-to-an-object-in-python

mkosowsk: Neat that looks a lot cleaner 👍🏻

kziemianek:
I'll take a look at webhook later, but seems like it's trigerred only by new comments.

kziemianek:
Does the counter next to heart icon in top right corner reflect recommends count in disqus thread ?
If not, then I think we can go with webook...

kziemianek:
So if we want that kind of sorting we need to somehow sync our model with disqus threads data. 

As @mkosowsk suggested, fetchin disquss threads data right before ideas list fetch might not have a big impact at start (disqus api max page size of 100), but as soon as we hit 200, 300, 400+ ideas on list requests times will get longer. I think we can defer this optimization.

So if we stick with this solution then when request hits ideas/fetch following actions will be taken:

1. disquss threads fetch
2. ideas fetch
3. ideas post_count and like_count update (if changed)
4. ideas fetch with sorting applied

owocki:
> So if we want that kind of sorting we need to somehow sync our model with disqus threads data.

hopefully there is an API and we can just refresh the model to the disquis data on the serverside and not have to do it everytime a user hits the page

kziemianek:
syncing will occour every 10 minutes

owocki:
very excited for this.. let me know when ready to test

kziemianek:
@owocki  :checkered_flag: ready to test

kziemianek:
@owocki hold on, I'll add i18n.

kziemianek:
@owocki finally :checkered_flag:

app/retail/models.py
```diff
@@ -0,0 +1,81 @@
...
+from economy.models import SuperModel
+from github.utils import get_user
+from rest_framework import serializers
+
+
+class Idea(SuperModel):
+    """Define the structure of idea"""
+    full_name = models.CharField(max_length=255)
+    email = models.EmailField(max_length=255)
+    github_username = models.CharField(max_length=255)
```
kziemianek: I think it might be better to reference profile here in order to remove redundancy but I see avatar_url spread across models. What do You think? Maybe refactor in another PR ?

app/app/urls.py
```diff
@@ -118,6 +118,12 @@
     url(r'^help/portal?', retail.views.portal, name='portal'),
     url(r'^help/pilot?', retail.views.help_pilot, name='help_pilot'),
     url(r'^help/?', retail.views.help, name='help'),
+    path('ideas/new', retail.views.new_idea, name='new_idea'),
```
mbeacom: 👍 for new `path`

app/dashboard/migrations/0051_merge_20180404_1535.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.0.3 on 2018-04-04 22:35
```
mbeacom: Please delete all three of these migration files and rerun: `make migrations`

app/retail/management/commands/sync_ideas.py
```diff
@@ -0,0 +1,69 @@
...
+import requests
+from retail.models import Idea
+
+logging.basicConfig(level=logging.DEBUG)
+
+logger = logging.getLogger(__name__)
+
+logging.getLogger("requests").setLevel(logging.WARNING)
+logging.getLogger("urllib3").setLevel(logging.WARNING)
+class Command(BaseCommand):
```
mbeacom: Two new lines between `logging.getLogger` and `class Command` - Also, `logger` doesn't appear to be used to pass logging anywhere.

app/retail/migrations/0004_idea_trending_score.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.0.3 on 2018-04-04 11:40
```
mbeacom: Delete all of the `retail` migrations and re-run `make migrations`

app/retail/models.py
```diff
@@ -0,0 +1,81 @@
+from django.db import models
+
+from economy.models import SuperModel
+from github.utils import get_user
+from rest_framework import serializers
+
+
+class Idea(SuperModel):
+    """Define the structure of idea"""
```
mbeacom: New line between docstring and attrs.

app/retail/models.py
mbeacom: @kziemianek Ideally, if we could reference `profile` and add a related name for reverse as: `ideas`, that would be 👍

app/retail/models.py
```diff
@@ -0,0 +1,81 @@
...
+    def load_avatar_url(self):
+        try:            
+            response = get_user(self.github_username)
+            self.avatar_url = response['avatar_url']
+        except Exception as e:
+            print(e)
+
+    @property
+    def thread_ident(self):
+        return 'idea-'+str(self.id)
```
mbeacom: Can we switch this to an f-string? `return f'idea-{self.id}'

app/retail/templates/idea.html
```diff
@@ -0,0 +1,42 @@
+{% comment %} Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
```
mbeacom: 😂 I've seen this happen in VSCode. 
Can we drop the`{% comment %}` block back down the way it normally is?

app/retail/templates/idea.html
```diff
@@ -0,0 +1,42 @@
...
+              <img src="{% static "v2/images/loading_v2.gif" %}">
+            </div>
+          </div>
+          <div id="disqus_thread"></div>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  {% include 'shared/bottom_notification.html' %} {% include 'shared/analytics.html' %} {% include 'shared/footer_scripts.html'%}
```
mbeacom: Can we place each include on an individual line for readability?

app/retail/templates/idea_template.html
```diff
@@ -0,0 +1,57 @@
+{% load i18n static %}
+<script id="idea_template" type="text/x-jsrender">
```
mbeacom: Can we fix the indentation to 2-spaces in this file?

app/retail/templates/ideas_list.html
```diff
@@ -0,0 +1,63 @@
+{% comment %} Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
```
mbeacom: Can we resolve the comment block here and move includes and what not to their own lines?

app/retail/templates/new_idea.html
```diff
@@ -0,0 +1,120 @@
+{% comment %} Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
```
mbeacom: Comment block and individual lines

app/retail/views.py
```diff
@@ -461,6 +468,64 @@ def browser_extension_chrome(request):
 def browser_extension_firefox(request):
     return redirect('https://addons.mozilla.org/en-US/firefox/addon/gitcoin/')
```
mbeacom: Can you place two new lines between module level methods?

app/retail/templates/idea.html
kziemianek: yes, i hit format all the time

app/retail/templates/idea_template.html
kziemianek: yes, format doesn't work here :rofl:

app/retail/templates/idea.html
```diff
@@ -0,0 +1,49 @@
+{% comment %} 
+  Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the
+  License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+  General Public License for more details. You should have received a copy of the GNU Affero General Public License along with
+  this program. If not, see <http://www.gnu.org/licenses/>.
+{% endcomment %} 
+{% load static %}
```
mbeacom: Duplicate load `static`

app/retail/templates/ideas_list.html
```diff
@@ -0,0 +1,70 @@
+{% comment %} 
+  Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the
+  License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+  General Public License for more details. You should have received a copy of the GNU Affero General Public License along with
+  this program. If not, see <http://www.gnu.org/licenses/>.
+{% endcomment %}
+{% load static %}
```
mbeacom: Duplicate load `static`

app/retail/templates/new_idea.html
```diff
@@ -0,0 +1,122 @@
+{% comment %} 
+  Copyright (C) 2017 Gitcoin Core This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the
+  License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+  General Public License for more details. You should have received a copy of the GNU Affero General Public License along with
+  this program. If not, see <http://www.gnu.org/licenses/>.
+{% endcomment %}
+{% load static %}
```
mbeacom: Duplicate load `static`

app/retail/views.py
```diff
@@ -496,6 +504,70 @@ def browser_extension_firefox(request):
...
+def new_idea(request):
+    return TemplateResponse(request, 'new_idea.html', {})
+
+
+@require_POST
+@csrf_exempt
+def create_idea(request):
+    idea = Idea(**json.loads(request.body))
+    if request.user.is_authenticated:
+        idea.profile = Profile.objects.get(pk=request.user.profile.id)
```
mbeacom: You should be able to use `idea.profile = request.user.profile` here.

app/retail/views.py
```diff
@@ -496,6 +504,70 @@ def browser_extension_firefox(request):
...
+
+@csrf_exempt
+def new_idea(request):
+    return TemplateResponse(request, 'new_idea.html', {})
+
+
+@require_POST
+@csrf_exempt
+def create_idea(request):
+    idea = Idea(**json.loads(request.body))
```
mbeacom: Just a note, we'll ultimately want to migrate this to a form to handle cleaning and validation on the posted values.

app/retail/views.py
```diff
@@ -496,6 +504,70 @@ def browser_extension_firefox(request):
     return redirect('https://addons.mozilla.org/en-US/firefox/addon/gitcoin/')
 
 
+def idea_show(request, idea_id):
+    return TemplateResponse(request, 'idea.html', {})
+
+
+@csrf_exempt
+def new_idea(request):
+    return TemplateResponse(request, 'new_idea.html', {})
```
mbeacom: We should use the same view for `new_idea` and `create_idea`, handling create on POST, and render page on GET.

app/retail/views.py
kziemianek: i get it!

app/retail/views.py
kziemianek: ok, i get it!

kziemianek:
Thanks for your review. I'll post changes in a couple of hours

app/retail/views.py
```diff
@@ -538,6 +544,78 @@ def browser_extension_firefox(request):
     return redirect('https://addons.mozilla.org/en-US/firefox/addon/gitcoin/')
 
 
+def idea_show(request, idea_id):
+    return TemplateResponse(request, 'idea.html', {})
+
+
+def new_idea(request):
+    if request.POST:
```
kziemianek: if post create idea and redirect to created idea, otherwise show idea entry form.
Is this what You were asking about ?I can also use ModelForm :)

kziemianek:
@mbeacom I've changed code according to Your comments. Thanks for review:)

kziemianek:
Sounds great. Hit me up if something needs to be done.

app/retail/templates/idea_template.html
```diff
@@ -0,0 +1,55 @@
...
```
mbeacom: Missing i18n load here.  Adding now.

app/retail/models.py
```diff
@@ -0,0 +1,83 @@
...
+    designers_exists = models.BooleanField()
+    customer_exists = models.BooleanField()
+    posts = models.IntegerField(default=0)
+    likes = models.IntegerField(default=0)
+    trending_score = models.IntegerField(default=101)
+
+    @property
+    def local_avatar_url(self):
+        if self.profile:
+            return self.profile.local_avatar_url
```
mbeacom: Bad ref

app/assets/v2/js/pages/idea.js
```diff
@@ -0,0 +1,31 @@
...
+      renderIdea(idea);
+
+      window.disqus_config = function() {
+        this.page.url = window.location.href;
+        this.page.identifier = idea.thread_ident;
+      };
+      // ajax request to load the disqus javascript
+      $.ajax({
+        type: 'GET',
+        url: 'http://' + forumName + '.disqus.com/embed.js',
```
mbeacom: Can we load this over HTTPS?

mbeacom:
This is on https://stage.gitcoin.co

Example:
https://stage.gitcoin.co/ideas/idea/1/show

mbeacom:
@kziemianek Made a few minor tweaks.

It appears the username isn'r rendering on the list view:

screenshot 2018-05-02 13 50 05

scripts/crontab
```diff
@@ -43,5 +43,5 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/us
 1 10 * * 1 cd gitcoin/coin; bash scripts/run_management_command.bash assemble_leaderboards  >> /var/log/gitcoin/assemble_leaderboards.log  2>&1
 10 * * * * cd gitcoin/coin; bash scripts/run_management_command.bash pull_stats  >> /var/log/gitcoin/pull_stats.log  2>&1
 10 3 * * * cd gitcoin/coin; bash scripts/run_management_command.bash pull_github  >> /var/log/gitcoin/pull_github.log  2>&1
+*/10 * * * * cd gitcoin/coin; bash scripts/run_management_command.bash sync_ideas 1  >> /var/log/gitcoin/sync_ideas.log  2>&1
```
mbeacom: `sync_ideas` doesn't take any arguments. This crontab fails.

thelostone-mc:
@sethmcleod mind checking this out since it was your design -> you might have a better grasp over it 
https://stage.gitcoin.co/ideas/list

@kziemianek 

- Nothing happens when I click on the heart icon ? (I'm logged in)
- the username doesn't turn up next to the `submitted by` 
- Could you update the styling of the dropdown (checkout the `bounty/new`) The CSS is already there, you just need to update the html
- Also the page is using a mix of Futura & Muli. Could we stick to Muli ? 
- Rather than defining the font-size , to ensure uniformity could we stick to using the classes defined https://github.com/gitcoinco/web/blob/master/app/assets/v2/css/gitcoin.css#L197-L211
Body -> `font-body`, heading -> `font-title` and so on
- Could we add some breathing space between disqus component and ideas (margin top/bottom of 1em ? )
- Is it just me / does it look odd, when the Has & Needs section are empty. Would it make sense to leave the default entry as `-`

screen shot 2018-05-03 at 12 17 28 am

@PixelantDesign your thoughts ?
Anything obvious I'm missing out @mbeacom @SaptakS ?

kziemianek:
@owocki  @mbeacom @thelostone-mc few fixes in last commits:

> It appears the username isn't rendering on the list or detail views:

fixed

> Recommended doesn't display updated count on list or details view.

the command should update this value, it gets updated after a while, you want it updated right after hitting hearth ?

> ‘github username’ input — shouldnt this be just the ’github login

changed label to github login

> when i submit and im not logged in

fixed

> when i submit and i am logged in

unfortunetly I cannot reproduce this problem

> javascript injection! bwahahhahaha! 

cool! fixed :)

> the ‘ideas/new’ and the ‘ideas show’ page need links back to the main ideas page i think

there is a link in navbar, should i add next?

> URIs - why ideas/list and not ideas/ ?

changed to ideas/

> URIs — why ideas/idea/2/show and not ideas/<pk>-<slug> ?

changed to ideas/ideas/<pk>... what is slug?

owocki:
> changed to ideas/ideas/... what is slug?

can we change to `ideas/<pk>-<slug>` ? i think the double `ideas` is wasteful in the URI

slug == the `slugify` method on this django util https://docs.djangoproject.com/en/2.0/ref/utils/

owocki:
sweet looking fwd to testing now

kziemianek:
@owocki sorry the url is `ideas/idea/<int:idea_id>`

kziemianek:
@thelostone-mc, i made some changes according to Your comments, thanks a lot :)
here is the summary:

> Nothing happens when I click on the heart icon ? (I'm logged in)

i've described it in last comment

> the username doesn't turn up next to the submitted by

i've described it in last comment

> Could you update the styling of the dropdown (checkout the bounty/new) The CSS is already there, you just need to update the html

You mean these dropdowns with search built-in ?

> Also the page is using a mix of Futura & Muli. Could we stick to Muli ?

Changed it to Muli

> Rather than defining the font-size , to ensure uniformity could we stick to using the classes defined 

does this meet your expectations ? 
changed_fonts


> Could we add some breathing space between disqus component and ideas (margin top/bottom of 1em ? )

definetly yes!


> Is it just me / does it look odd, when the Has & Needs section are empty. Would it make sense to leave the default entry as -

added - for empty has and needs

thelostone-mc:
> You mean these dropdowns with search built-in ?

Yup! We'll throw in the dropdown without search later and update this then

@kziemianek looks better ^_^

kziemianek:
@thelostone-mc I also updated select to use jquery select2 with hidden search box.

kziemianek:
@mbeacom are we waiting for something? have you seen my last comments :) ?

app/retail/views.py
```diff
@@ -542,6 +548,82 @@ def browser_extension_firefox(request):
...
+        looking_for_capital = bool(request.POST.get('looking_for_capital'))
+        looking_for_builders = bool(request.POST.get('looking_for_builders'))
+        looking_for_designers = bool(request.POST.get('looking_for_designers'))
+        looking_for_customers = bool(request.POST.get('looking_for_customers'))
+        capital_exists = bool(request.POST.get('capital_exists'))
+        builders_exists = bool(request.POST.get('builders_exists'))
+        designers_exists = bool(request.POST.get('designers_exists'))
+        customer_exists = bool(request.POST.get('customer_exists'))
+        profile = request.user.profile if request.user.is_authenticated else None
+        idea = Idea(full_name=full_name, email=email, github_username=github_username, summary=summary, more_info=more_info,
```
mbeacom: v2 we should probably move this to a form.

mbeacom:
Seeing the same ideas after clicking `Show More`

screenshot 2018-05-17 09 04 40

mbeacom:
Also, I `recommended` the cooking discussion (of course! 😆) and it appears we'll need to add custom caching rules to the list page.

kziemianek:
@mbeacom i bet duplicated idea was dropped to next page after another idea becoming more trending, so it was returned by paginator again. 

As a workaround, I can add check to ensure only one idea with given `id` is displayed in list.

PR #791, opened 2018-04-04
Limit Profile stat results to valid bounties

mbeacom:
##### Description
The goal of this PR is to allow filtering of bounties by a status field versus a property.  This allows us to filter out `unknown` or `cancelled` bounties from the profile stats.

This PR introduces the `Bounty.status` field, renaming `Bounty.status` property to `Bounty.current_status`, migrates existing statuses, adds the `stats_eligible` queryset function, and makes use of it in `Profile.stats`.

##### Checklist

##### Testing
Tested locally and on staging.

You can observe the state of these changes on: https://stage.gitcoin.co/profile/bradford-hamilton and compare against: https://gitcoin.co/profile/bradford-hamilton


mbeacom:
@owocki I've updated the proposed changes to use `idx_status` since we're copying the return from `Bounty.status` property during `pre_save` anyway, so we're not introducing possible breaking changes or modifying all idx/status refs throughout the codebase.

mkosowsk:
Code LGTM tho didn't test 👍🏻 thanks @mbeacom, this will be a nice fix so we don't harm stats for contributors for stuff that's outside of their control :)

owocki:
> @owocki I've updated the proposed changes to use idx_status since we're copying the return from Bounty.status property during pre_save anyway, so we're not introducing possible breaking changes or modifying all idx/status refs throughout the codebase.

okie dokie.. its a little less than elegant, but its still a step forward.  makes sense to me

mbeacom:
@owocki What would your ideal implementation be?

 If you'd prefer, I can refactor the way we're determining the status, rename the field throughout, and modify the signaling. I just figured less changes in this PR was ideal. I am fine with making the changes in this branch, though.

owocki:
> @owocki What would your ideal implementation be?

prettymuch what you describe below, though i agree with your judgement call to not bother in retrospect.

> If you'd prefer, I can refactor the way we're determining the status, rename the field throughout, and modify the signaling. I just figured less changes in this PR was ideal. I am fine with making the changes in this branch, though.

i say lets shippit once tested

PR #861, opened 2018-04-11
docker shell fixes

owocki:
This PR fixes a problem where if I 

`docker-compose up -d` then `docker-compose stop web; docker-compose run --service-ports web` 

as described [in the docs](https://github.com/gitcoinco/web/blob/master/docs/RUNNING_LOCALLY_DOCKER.md#ipdb)

i can no longer `exec` on the docker intance

```docker-compose exec web /bin/bash```

returns

```ERROR: No container found for web_1```

this is what `docker-compose ps` returns

```
Name                   Command                State             Ports
----------------------------------------------------------------------------------
web_db_1        docker-entrypoint.sh postgres    Up         0.0.0.0:5432->5432/tcp
web_node_1      bin/docker-node.sh               Exit 1
web_pgweb_1     /usr/bin/pgweb --bind=0.0. ...   Up         0.0.0.0:8081->8081/tcp
web_testrpc_1   node ./build/cli.node.js - ...   Up         0.0.0.0:8545->8545/tcp
web_web_1       /usr/local/bin/dumb-init - ...   Exit 143
web_web_run_1   /usr/local/bin/dumb-init - ...   Up         0.0.0.0:8000->8000/tcp

```

but none of those container names can be passed into docker-compose exec

Id expecteed that the ‘Up’ container names ( like web_web_run_1 ) is what i shoudl be able to do

# Workaround

This adds another container, called `shell` which is independent of the web server and can be connected to like so:

```docker-compose exec shell /bin/bash```

PR #863, opened 2018-04-11
Add ipdb and django shell access to docker setup

mbeacom:
##### Description

The goal of this PR is to introduce entrypoints for `ipdb` and the django shell via `make` commands.
The method used in this PR allows us to inspect the ipdb shell and the django shells without an additional container in our compose stack.

##### Checklist

##### Testing
Tested with multiple `web` repo directories to ensure the parent directory name (first part of compose image name) is properly detected.

##### Refers/Fixes
Ref #861 - This PR aims to handle the same case as this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,10 @@
 
 .PHONY: help
 
+PROJECT_DIR := $(subst -,, $(shell pwd | xargs basename))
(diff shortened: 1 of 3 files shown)


owocki:
very excited about this :)

PR #880, opened 2018-04-13
view on github button on bounties

owocki:
##### Description

we've gotten some customer feedback that we are not doing a great job of showing issue scope and people would rather read it on github.

i think thats fair, and we shouldn't pretend to be the *collaboration* layer for OSS.  we are the incentivization layer and we good with that.

this PR takes the bounty detail page and removes any issue desc after 300 chars and instead shows a 'view on github' button.

[screenshot](http://bits.owocki.com/2d2i0d2e3A2U/Screen%20Shot%202018-04-12%20at%209.40.01%20PM.png)

##### Checklist

- x] linter status: 100% pass

UI!

##### Testing
i done tested it

##### Refers/Fixes
self

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -120,10 +120,10 @@ var callbacks = {
     val = val.replace(/script/ig, 'scr_i_pt');
     var ui_body = val;
 
-    if(ui_body.length > 300){
-      ui_body = ui_body.substring(0,400) + "... <BR><BR><BR>";
-      ui_body = ui_body + "<a class='button button--primary' href='"+result['github_url']+"'><span class='font-caption'>View issue on github</span></a>";
-      ui_body = ui_body + "<br><br>";
+    if (ui_body.length > 300) {
+      ui_body = ui_body.substring(0, 400) + '... <BR><BR><BR>';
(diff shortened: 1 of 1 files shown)


owocki:
@PixelantDesign this ok with u?

thelostone-mc:
@owocki I had removed this earlier 😂 
Would it make sense if we integrated markup language to render content exactly as on github ?

**Reason** 
To be honest, Github is more comfortable, but when I'm working on a repo with multiple issues
navigating to github and searching it directly is painful. Gitcoins Issue Details /  Explorer makes it easier.
Once I'm there -> I end up feeling lazy for the extra button click to go to github (especially when you have slow internet connection which is true for places outside US)

owocki:
curious what @PixelantDesign thinks is the right move here, per https://github.com/gitcoinco/web/pull/880#issuecomment-381053051

PixelantDesign:
@owocki curious to know where the feedback came from and what the issues were.

In the updated explorer detail page, we decided to show the full github description so that user could see the criteria to make a determination whether they’d like to click start work to take on the project. 

If we cap it at 300 characters ( hardly enough info to understand the ticket), user then needs to go to Github to view the rest of the description and then know to come back to gitcoin to click start work. To avoid the back and forth, should we consider adding start work buttons on Github instead?

Issue #540, opened 2018-03-06
(DESIGN) As a user, I want a Blockchain Job Board so I can find professional work in the space


owocki:
Gitcoin should host a Blockchain Job Board which allows anyone to post a blockchain job for free.

# What

Three pages

1. Index of Jobs
2. Job Detail Page
3. Post Job

# Attributes of Jobs

Title,
Desc (markdown formatted),
where to apply (email or url)
linked to a gitcoin/github profile

# Bounty

I would like to see mockups of what the above 3 pages look like.  Sketch or PSD files are fine.

Feel free to re-use any of the creative assetes in the website or on https://github.com/gitcoinco/creative

gitcoinbot:
__This issue now has a funding of 0.3 ETH (159.6 USD @ $531.99/ETH) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/issue/gitcoinco/web/540).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/issue/gitcoinco/web/540)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $2657.58 more Funded OSS Work Available at: https://gitcoin.co/explorer

vs77bb:
Not that this isn't a good idea to have locally... but resources do exist elsewhere. 

- https://angel.co/talent-hacks/how-to-get-a-job-at-a-crypto-startup
- Cryptocurrency Startup Jobs - AngelList
- https://crypto.jobs/
- http://www.workincrypto.com/

owocki:
endgame: one day gitcoin will be a one stop shop for anyone looking to BUIDL a career in blockchain software.  that means core things we do well (community, bounties, tips, grants, and more) but also commodified services (like job boards, references to bounties in other bounty projects)

gitcoinbot:
__Work has been started on the 0.3 ETH (173.94 USD @ $579.79/ETH) funding by__: 
 1. [@cmac2454](https://gitcoin.co/profile/cmac2454) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

[...]

PR #897, opened 2018-04-16
Github comments when user is warned/removed (in addition to the emails that are sent when this happens)

owocki:
##### Description

Github comments from gitcoinbot when a user is removed from the issue.

[example issue](https://github.com/gitcoinco/web/issues/540)

This gives context to anyone looking back at the thread to try and figure out what is going on with the issue that the github bot has kicked them off of the issue.

In addition, this PR will enable notifications to slack so that we can deal with it internally

##### Checklist

bot

##### Testing
tested

##### Refers/Fixes
self

### DIFF
--- a/app/marketing/management/commands/expiration_start_work.py
+++ b/app/marketing/management/commands/expiration_start_work.py
@@ -27,7 +27,10 @@ from django.utils import timezone
 from dashboard.models import Bounty, Interest
 from dashboard.notifications import maybe_notify_user_removed_github, maybe_warn_user_removed_github
 from github.utils import get_interested_actions
-from marketing.mails import bounty_startwork_expire_warning, bounty_startwork_expired
+from marketing.mails import (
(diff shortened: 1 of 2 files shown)


owocki:
another route we could go is just to disable the auto-kick functinoality.  instead, we could make the bot just warn the user in perpetuity

app/marketing/management/commands/expiration_start_work.py
```diff
@@ -25,8 +25,12 @@
 from django.utils import timezone
 
 from dashboard.models import Bounty, Interest
+from dashboard.notifications import maybe_notify_user_removed_github, maybe_warn_user_removed_github
 from github.utils import get_interested_actions
-from marketing.mails import bounty_startwork_expire_warning, bounty_startwork_expired
+from marketing.mails import (
+    bounty_startwork_expire_warning, bounty_startwork_expired, maybe_notify_bounty_user_removed_to_slack,
```
mbeacom: This import will fail. `maybe_notify_bounty_user_removed_to_slack` and `maybe_notify_bounty_user_warned_removed_to_slack`

Issue #656, opened 2018-03-19
Design: Profile Refresh


owocki:
As a user I'd like my profile to be representative of my status in Gitcoin.

Profile Refresh
Current gitcoin profiles ( https://gitcoin.co/profile/owocki ) 

## Current Implementation
* some stats i just made up when i was coding the page
* a very rudimentary design
* some responsive design issues
* aren't as purdy as they could be
* doesnt really talk about my core skils
* is not edit-able
* what else?

## Bounty
A new profile design with the following items addressed....

## Definition of Done
- Designs are based off of @willsputra's initial concept
- Designs and ux that provides a good overview of who I am and what I've done in the system.
- The profile is designed using the Gitcoin UI Kit.
- Designs include the following info: User name, social media info, issues funded, issues completed, success rate, tips, Issue status, Github repos, activity feed, pending transactions.
- Designs include public profile and private profile.
- Finalized designs of desktop / tablet/ mobile view of the profile.  Pls allow for 2-3 feedback iterations from the core team

## Reviewers
@owocki @pixelantdesign @vs77bb

mkosowsk:
Some quick thoughts:

0. ETH/Fiat buttons should follow consistency from redesigned Explorer
1. Text overflow issue in details, probably shouldn't be too difficult to fix
2. Recommend following general styling from the Explorer: lead with the Logo of the project on the far left. Font also might be slightly different?
3. Might be interesting to include dates for various actions
4. Could be interesting to add Filters for top of columns. If Relationship is clicked, that could be filtered A->Z and if clicked again Z->A. Same for status. Bounty amount could go highest -> lowest and then lowest -> highest.
5. Hair seems to be wrong in your profile photo? Would recommend fixing this ASAP, very important to stay on brand

owocki:
@PixelantDesign what do you think of the above design?  personally i think its a big step up aesthetically from the current profile

skip-carlson:
Is the section of the profile that has the - Level 1 Funder - Level 3 Worker - ultimately going to reflect the reputation a user builds up? When people look at this, are they thinking "oh this guy is 1/5 here and 3/5 there..." Or is it even on a scale of 5, of 10? Do users know the basis? I like the idea of a certain rating there so people know the quality of the user. I imagine once the reputation system defined and up and running it will live in this part of the profile.

owocki:
maybe instead of levels we can just list the number of bounties youve worked on as a funder / worker

mkosowsk:
[...]

PR #918, opened 2018-04-18
make expiration status less inclusive

owocki:
##### Description
This PR makes the expiration status less inclusive.

Before, if a bounty was past its expiration date and had no fulfillments it would be marked as expired

Now, after this change, the bounty will only be status expired if its past its expiration date if
* it was one of the bounties that existed before https://github.com/Bounties-Network/StandardBounties/issues/25
* and it's past its expiration date

otherwise, it will revert to one of the following statuses: open, started, submitted, done.

##### Checklist

db

##### Testing
tested it

##### Refers/Fixes
https://gitcoincommunity.slack.com/archives/G9ZRKFS68/p1524066432000794

### DIFF
--- a/app/dashboard/models.py
+++ b/app/dashboard/models.py
@@ -323,6 +323,19 @@ class Bounty(SuperModel):
         return timezone.now()
 
     @property
+    def past_expiration_date(self):
+        """Return true IFF issue is past expiration date"""
+        return timezone.localtime().replace(tzinfo=None) > self.expires_date.replace(tzinfo=None)
+
+    @property
+    def past_hard_expiration_date(self):
+        """Return true IFF issue is past smart contract expiration date
+        and therefore cannot ever be claimed again"""
+        return self.past_expiration_date and self.can_submit_after_expiration_date
+
+
+
+    @property
     def status(self):
         """Determine the status of the Bounty.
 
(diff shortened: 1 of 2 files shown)


app/dashboard/models.py
```diff
@@ -357,10 +370,10 @@ def status(self):
         else:
             try:
                 if not self.is_open:
-                    if timezone.localtime().replace(tzinfo=None) > self.expires_date.replace(tzinfo=None) and self.num_fulfillments == 0:
-                        return 'expired'
                     if self.accepted:
                         return 'done'
+                    if self.past_hard_expiration_date:
```
mbeacom: Moving this if check above `if self.accepted:` should resolve test failures, I'm thinking 🤔

pelsasser:
So the idea being if a bounty is expired based on time, but still has funding it shows as what state?

My main reason for requesting this was an easy way to know where to claw back funds for issues that have expired, but still having funding attached.

owocki:
> So the idea being if a bounty is expired based on time, but still has funding it shows as what state?

It shows as either open, started, or submitted.

Basically we aren't enforcing the expiration dates... so there isn't a good reason to show it as a status. I think.

> My main reason for requesting this was an easy way to know where to claw back funds for issues that have expired, but still having funding attached.

Got it... so in this case any issues in open, started, or submitted could be eligible for clawback.

pelsasser:
@owocki - perfect, that makes sense. Manually expiring a bounty seems to make more sense for my personal use case than a hard time stop.

That was a pretty quick turn around! thanks guys.

owocki:
@pelsasser word... once we ship this check out your profile and taht should give you what you need 

related: check out the upcoming profile changes => https://github.com/gitcoinco/web/issues/656#issuecomment-382457008 will hopefully improve the profile UX

owocki:
hmm failing test seems unrelated.. im going to merge this, and we should circle back on whats wrong with the leaderboard assembly test

mbeacom:
@owocki I resolved the failing test for `test_bounty` in https://github.com/gitcoinco/web/pull/918/commits/ab6dca1c8a5bcdb91ef65e5d9f956b1423a0dcef -  I'll look into the leaderboard issue now.

Issue #876, opened 2018-04-12
timezone issues with opened date


owocki:
http://bits.owocki.com/2I460x3e1T0Z/Screen%20Shot%202018-04-12%20at%202.30.06%20PM.png

https://gitcoin.co/issue/gitcoinco/web/865

This bounty was really opened 3 hours ago but the 'opened' date says 10 hours ago

kziemianek:
@owocki, @mbeacom seems like switching to UTC timezone solves problem. I agree that dates should be stored in UTC format, rather than other timezone.

What do You think about changing timezone ? Seems to be fundamental change and i'm not aware of all django internals and problems it might rise with current implementation.

mkosowsk:
Time zone funkiness is why I changed my life to be running on UTC, just one less thing to worry about :)

- Mitch Kosowski, Posted at 11:02 AM UTC

mbeacom:
@kziemianek Agreed.

We should ensure we're storing UTC and implement: https://docs.djangoproject.com/en/2.0/topics/i18n/timezones/

owocki:
@mbeacom what do you think about changing settings.py:163 to set the default as UTC as a means of solving the original issue?

owocki:
fixed in https://github.com/gitcoinco/web/pull/919#issuecomment-382477494

owocki:
thanks @kziemianek

gitcoinbot:
⚡️ A tip worth 0.025  ETH (12.85 USD @ $513.94/ETH) has been granted to @kziemianek for this issue from Kevin. ⚡️ 

The sender had the following public comments: 
> Thanks for the tip about the django setting

Nice work @kziemianek! To redeem your tip, login to Gitcoin at https://gitcoin.co/explorer and select 'Claim Tip' from dropdown menu in the top right, or check your email for a link to the tip redemption page. 

 * $4578.35 in Funded OSS Work Available at: https://gitcoin.co/explorer
 * Incentivize contributions to your repo: <a href='https://gitcoin.co/tip'>Send a Tip</a> or <a href='https://gitcoin.co/funding/new'>Fund a PR</a>
 * No Email? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>

PR #919, opened 2018-04-18
fixes https://github.com/gitcoinco/web/issues/876

owocki:
https://github.com/gitcoinco/web/issues/876

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -168,7 +168,7 @@ LANGUAGE_CODE = env('LANGUAGE_CODE', default='en-us')
 USE_I18N = env.bool('USE_I18N', default=True)
 USE_L10N = env.bool('USE_L10N', default=True)
 USE_TZ = env.bool('USE_TZ', default=True)
-TIME_ZONE = env.str('TIME_ZONE', default='MST')
+TIME_ZONE = env.str('TIME_ZONE', default='UTC')
 
 if not ENV in ['local', 'test']:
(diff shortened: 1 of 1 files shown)


mbeacom:
I was intending on tackling the whole lot of this in a PR aimed at adding timezone support to the platform, but this should work fine for serving everything as UTC for the moment.  The postgresql dbs are set to UTC for timezone right now, but I'm not sure what the implications are pertaining to DTs that have been using `SuperModel` timezone specific DTs versus `auto_add`.

owocki:
most of the times on the frontend of the app are relative to current time anyway.. so i dont know if whole timezone support is necessary... what were you hoping to accomplis htere?

owocki:
http://bits.owocki.com/1Y1z3q3t2R0U/Screen%20Shot%202018-04-18%20at%2012.06.01%20PM.png

yay this fixes the issue in #876

mbeacom:
Adding fully baked timezone support, so if a user is viewing details with datetimes, it displays appropriately for their TZ.  My main goal with implementing proper TZ support is to keep aligned with our other initiatives for localization and internationalization support on the platform... seems like it'd make sense to be all in on `i18n` - 🤷‍♂️  than to half implement.  This makes sense as a stopgap for the moment, but it's probably a good idea revisit the way we're pulling `now` and handling DTs throughout the platform, whether we stick with everything displaying in UTC or not.

PR #931, opened 2018-04-21
verboseness of gitcoinbot reminders

owocki:
adds the following copy to gitcoin bot warnings about removal for non-action on an issue:

* [x] warning 1 (3 days)
* [x] warning 2 (6 days)
* [x] auto removal (10 days)

### DIFF
--- a/app/marketing/management/commands/expiration_start_work.py
+++ b/app/marketing/management/commands/expiration_start_work.py
@@ -46,6 +46,7 @@ class Command(BaseCommand):
             print('not running start work expiration because DEBUG is on')
             return
 
+        #todo: DRY with dashboard/notifications.py
         num_days_back_to_warn = 3
         num_days_back_to_delete_interest = 10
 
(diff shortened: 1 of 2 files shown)


app/marketing/management/commands/expiration_start_work.py
```diff
@@ -78,6 +79,7 @@ def handle(self, *args, **options):
                         if not actions:
                             should_warn_user = True
                             should_delete_interest = False
+                            last_heard_from_user_days = (datetime.now() - interest.created.replace(tzinfo=None)).days
```
mbeacom: Can we use `tzinfo=pytz.UTC` here?

owocki: yes we can make that change.. but the problem is that the rest of the datetime's in the `expiration_start_work` flow are not timezone aware..   let me see if it's an easy lift to do this everywhere

owocki: 965505b67c1b6c8035e276db61b814375ed14cef

mbeacom:
This PR had failing tests :(

owocki:
just put test fixes in

PR #936, opened 2018-04-21
sets default search project status to 'any'

owocki:
like the title says ^^

i think this makes it easier to do a search with default settings.

esp when someone passes you a link like https://gitcoin.co/explorer?q=augur


mbeacom:
@owocki This is good to merge. My only concern is defaulting to `any` might cause performance problems for new users browsing the site for the first time or anyone with the default filter settings.  I'm still seeing 5 second load times when hitting `any` on the dashboard.

owocki:
@mbeacom here are two github issues designed to mitigate this performance burden:

https://github.com/gitcoinco/web/issues/951
https://github.com/gitcoinco/web/issues/950

do they look okay?  if so ill have some of the community work on them and that'll hopefully unblock us here

owocki:
ok to 🚢 now that https://github.com/gitcoinco/web/pull/1026 and #1019 are live?

PR #943, opened 2018-04-23
fix wallpaper category

kziemianek:
##### Description
*it should be listed in for_fun category


app/dashboard/views.py
```diff
@@ -871,6 +871,7 @@ def toolbox(request):
     profile_up_votes_tool_ids = ','.join(str(x) for x in ups)
     downs = list(ToolVote.objects.filter(profile_id = profile_id, value = -1).values_list('tool', flat=True))
     profile_down_votes_tool_ids = ','.join(str(x) for x in downs)
+    print(profile_up_votes_tool_ids)
```
owocki: did you mean to keep this in there?

owocki:
migration LGTM

app/dashboard/views.py
kziemianek: no! :)

mbeacom:
@kziemianek Thanks for this! - I was in the middle of updating some of the `Tool` related models/views when this came in. I opted to include the migration in the other PR (#944) to avoid an additional `migration` file.  If you're strongly opposed to that, I can merge this first... But I figured it would make sense to consolidate them.

Issue #259, opened 2018-01-20
Add slack bot integration


pelsasser:
##### Description
It would be nice to be able to have a slack integration that would notify our team when bounties have been posted to one of our repositories. 

# What

* An area in my settings where I can specify a slack API key (or key pair)
* an area in the settings where I can define what repos I want slack updates on
* An area in my settings where I can define what channel I want slack updates in 

then

Any updates to github that hit `maybe_market_to_slack()` will also hit my slack channel

mkosowsk:
Interesting! Want to post some quick thoughts on this altho I won't have time to tackle the issue. And truth be told, this functionality may already exist! But I don't know if it does

Changes to bounties are posted to the notif-gitcoin channel on the Gitcoin slack

https://gitcoincommunity.slack.com/messages/C8DCUQU91/convo/C7A4F4UHK-1516388972.000266/

An example message:

<img width="851" alt="screen shot 2018-01-20 at 5 35 02 pm" src="https://user-images.githubusercontent.com/7970539/35188690-608ab52e-fe08-11e7-8098-b0fe0c3f79fa.png">

So looks like it wouldn't be a big ask to get this functionality as this info is already being used by the Slack API Tester. Just do a quick check on the Github url for projects that want to be notified and blast them out.

May be a silly question, but I imagine you would want the Slack notifications to be on your team's Slack and not on the Gitcoin Slack?

pelsasser:
@mkosowsk - a way to get those same notifications into our team's slack is exactly what I am after.  Not urgent or needed ASAP, but I think a cool enhancement in general that would probably provide utility for a lot of teams.

mkosowsk:
Cool, that makes sense to me. Want to ping @owocki or @vs77bb, does this already exist? 

If not, I may fund it myself with something like .05 ether come Tuesday... seems like some cool functionality 👍

owocki:
this does not exist already yet..

i think we need some way for github repo admins to be able to auth to github (to prove they own a repo) and save a SLACK API key (and preferred channel for their slack) in the system...

from there, we can have slackbot post each time there is an issue update

mkosowsk:
Great! Think it's worth adding funding to the issue now or should we wait for a bit more of the groundwork to be built up?

owocki:
@mbeacom and i are working on getting github integration built in this month.. after that i think this would be worth bounty-ing

mkosowsk:
Cool, will keep my eyes peeled for that 👍

[...]

PR #955, opened 2018-04-24
Add slack bot integration 

darkdarkdragon:
- add slack settings tab on user's settings page
- send same messages to user's slack as to gitcoin's slack

##### Checklist

- backend
- frontend

##### Testing
Tested manually

##### Refers/Fixes
Fixes: #259


darkdarkdragon:
@owocki Now this should work for your case, but you need to specify repo with full name, like `gitcoin/web`

app/marketing/views.py
```diff
@@ -604,18 +604,67 @@ def email_settings(request, key):
...
+
+    msg = ''
+
+    if request.POST and request.POST.get('submit'):
+        token = request.POST.get('token', '')
+        repos = request.POST.get('repos').split(',')
+        channel = request.POST.get('channel', '')
+        no_access = [repo_name for repo_name in repos if repo_name != '' and not profile.has_repo(repo_name)]
+        if len(no_access):
+            msg = _('You don\'t have access to {}').format(no_access)
```
mbeacom: When possible, try to use literal string interpolation / f-strings since they're significantly faster than `format()`

app/retail/templates/settings/slack.html
```diff
@@ -0,0 +1,22 @@
...
+  <div class="form-group">
+    <label class="form__label" for="repos">{% trans "Github repos to monitor (use full name, like 'gitcoin/web')" %}</label>
+    <input type="text" name="repos" class="form__input" placeholder='comma, separated, list (gitcoin/web)' value="{{ repos }}">
+  </div>
+  <div class="form-group">
+    <label class="form__label" for="channel">{% trans "Slack channel to post to" %}</label>
+    <input type="text" name="channel" class="form__input" value="{{ profile.slack_channel }}">
+  </div>
+  {% csrf_token %}
+  <input class='button button--primary' type='submit' name='submit' value="Go">
```
mbeacom: We might want to `trans` the input button value.

app/marketing/views.py
```diff
@@ -636,8 +636,8 @@ def slack_settings(request):
         repos = request.POST.get('repos').split(',')
         channel = request.POST.get('channel', '')
         no_access = [repo_name for repo_name in repos if repo_name != '' and not profile.has_repo(repo_name)]
-        if len(no_access):
-            msg = _('You don\'t have access to {}').format(no_access)
+        if no_access:
+            msg = _(f'You don\'t have access to {no_access}')
```
darkdarkdragon: Isn't in this case gettext_lazy will receive string with variable substituted for the value received from user,
like
`''You don\'t have access to gitcoin/web'`
?

darkdarkdragon:
@owocki sorry, missed one comment
does this looks ok?
screen shot 2018-04-24 at 21 28 07

darkdarkdragon:
@owocki 
> s there validation on the slack channel to post to field? do i include a # or not?
for me it works without #

mbeacom:
I agree with adding a placeholder.  One additional thing...  maybe add a `slack_channel.lstrip('#')` so it supports one or the other?

darkdarkdragon:
Added placeholder
I've checked with slack's api, it accepts channel name with # and without without problem

darkdarkdragon:
@owocki any more feedback or considered done?

owocki:
testing now

owocki:
testing notes:
* can we resolve merge conflicts?
* i get [this](http://bits.owocki.com/3b0T2i0c3Q18/Screen%20Shot%202018-04-25%20at%204.54.06%20PM.png) 500 error when i visit the /slack locally - does not matter if i am logged in or not (though it does seem to matter if im logged in as one of these [oddity profiles](https://github.com/gitcoinco/web/issues/975) or not)
* i get [this message] when i try to connect to `gitcoinco/web` => http://bits.owocki.com/3a1s3a082o3W/Screen%20Shot%202018-04-25%20at%204.59.02%20PM.png -- why are we limiting users to only use repos that are on their profile?

app/marketing/views.py
```diff
@@ -604,18 +604,67 @@ def email_settings(request, key):
...
+        login_redirect = redirect('/login/github?next=' + request.get_full_path())
+        return login_redirect
+
+    msg = ''
+
+    if request.POST and request.POST.get('submit'):
+        token = request.POST.get('token', '')
+        repos = request.POST.get('repos').split(',')
+        channel = request.POST.get('channel', '')
+        no_access = [repo_name for repo_name in repos if repo_name != '' and not profile.has_repo(repo_name)]
```
owocki: when i test.. it seems to work if i remove this line.

darkdarkdragon:
@owocki 
Conflicts resolved, also I removed my generated migration, but didn't generated new one - needs to be regenerated after final merge

> i get this 500 error when i visit the /slack locally - does not matter if i am logged in or not (though it does seem to matter if im logged in as one of these oddity profiles or not)

you shouldn't see this page if not logged in: 
```python
    profile, es, user, is_logged_in = settings_helper_get_auth(request)
    if not es:
        login_redirect = redirect('/login/github?next=' + request.get_full_path())
        return login_redirect
```
maybe we should check here `profile` for `None` too

> why are we limiting users to only use repos that are on their profile?

from discussion:

>i think we need some way for github repo admins to be able to auth to github (to prove they own a repo)

if want to allow anyone to hook up to any repo, then we should just remove this check

owocki:
> if want to allow anyone to hook up to any repo, then we should just remove this check

i would prefer to do this.  i dont see any good reason why someone needs to own a github repo in order to watch slack notifications for it

any objetions @mbeacom ?

mbeacom:
@owocki Agreed! I don't see any harm in allowing anyone to follow a repo with the bot.

owocki:
word. @darkdarkdragon mind removing that line?  i think were in a good place here and hoping to merge today

darkdarkdragon:
@owocki done

owocki:
lgtm based upon my last test... @mbeacom ?

PR #998, opened 2018-04-26
adds rates estimate on bounty details page

owocki:
[screenshot](http://bits.owocki.com/3E312o0V3a1a/Screen%20Shot%202018-04-26%20at%205.05.27%20PM.png)

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -157,6 +157,9 @@ var callbacks = {
     if (val === null) {
       return [ null, null ];
     }
+    var rates_estimate = get_rates_estimate(val);
+    $('#value_in_usdt_wrapper').attr('title', rates_estimate);
+
     return [ 'Amount_usd', val ];
   },
   'fulfillment_accepted_on': function(key, val, result) {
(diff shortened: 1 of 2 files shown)


owocki:
not sure if a tooltip is the right place for it.. but thought it was nice context to add on hover

@PixelantDesign @vs77bb @mbeacom

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -157,6 +157,10 @@ var callbacks = {
     if (val === null) {
       return [ null, null ];
     }
+    var rates_estimate = get_rates_estimate(val);
+
+    $('#value_in_usdt_wrapper').attr('title', rates_estimate);
```
thelostone-mc: Could we wrap it under a div to have the same effect as the other tooltips ?
` '<div class="tooltip-info tooltip-sm">' +rates_estimate + '</div>'` like [here](https://github.com/gitcoinco/web/blob/400c2ae6596320c6db1306347b450a78e422ae44/app/dashboard/templates/bounty_details.html#L192)

owocki: good call.. updating

PR #1018, opened 2018-04-30
issue 115 duplicated emails

kylewu:
`web3_process_bounty` where email, as well slack etc, are sent, Is used in two places

* `sync_web3` in dashboard/views
* `sync_geth` in command

with opened browser to access `sync_web3`, it's possible there is a *race*

Since in `sync_web3`, it simply requires two fields (`did_change` and `url`). We could have `dry_run` parameter to control if `web3_process_bounty` should handle changed bounty, or return information only.

### DIFF
--- a/app/dashboard/utils.py
+++ b/app/dashboard/utils.py
@@ -249,10 +249,10 @@ def get_bounty(bounty_enum, network):
 
 
 # processes a bounty returned by get_bounty
-def web3_process_bounty(bounty_data):
+def web3_process_bounty(bounty_data, dry_run=False):
     did_change, old_bounty, new_bounty = process_bounty_details(bounty_data)
 
-    if did_change and new_bounty:
+    if did_change and new_bounty and not dry_run:
         _from = old_bounty.pk if old_bounty else None
         print(f"- processing changes, {_from} => {new_bounty.pk}")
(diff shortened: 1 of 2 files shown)


app/dashboard/utils.py
```diff
@@ -249,10 +249,10 @@ def get_bounty(bounty_enum, network):
 
 
 # processes a bounty returned by get_bounty
-def web3_process_bounty(bounty_data):
+def web3_process_bounty(bounty_data, dry_run=False):
     did_change, old_bounty, new_bounty = process_bounty_details(bounty_data)
 
-    if did_change and new_bounty:
+    if did_change and new_bounty and not dry_run:
```
owocki: the problem here is that `process_bounty_changes()` does a lot more than just send the email..

kylewu: correct, the idea is to use `sync_geth` command to call `process_bounty_changes()` only.

to make code clear, i could update `sync_web3()` using `process_bounty_details()` instead of `web3_process_bounty()`

kylewu:
Checked feedback from jakerockland  that he received email twice. But in slack, i see only one message. So this PR probably is wrong. I'm keep checking

owocki:
thanks for trying!

Issue #950, opened 2018-04-23
API Index Performance Improvements: Dynamic Calculated Information


owocki:
related to https://github.com/gitcoinco/web/pull/936#issuecomment-383610724 : 

The API takes about 6s to return when 'any' is selected from https://gitcoin.co/explorer

Some quick profiling shows that when you disable the dynamic lookups in the API ( 6bd14d66b2cc3a61ff23dade68bded857a53da0c ), then the performance for the same query is down to 1.5s.

For this reason, I believe we should take each of the values that are currently dynamically calculated on API request and store them as part of the model.  This information should then be refreshed in the `refresh_bounties` command which runs every 15 minutes or so.

This will drastically improve API performance

owocki:
yes. since i think this is (1) different scope (2) gonna be done in models.py not the api

mbeacom:
While I agree it's different scope, I just figured this applies to some degree.

I suppose my question was more about whether or not the scope of this issue might be defined differently.

If we're considering migrating those property methods over to actual fields or using some stored value, I don't think it pertains to the API specifically since it's more of a performance burden platform wide to access any of those properties and carry a heavy DB footprint when considered at scale. One solution would be to eliminate the property fields that aren't in use throughout JS files from `dashboard/api.py@BountySerializer.fields` and migrate properties that are used over to something less hindering to performance **OR** cache the offending querysets with a defined TTL, but I don't think caching or property modifications would be considered API centric changes, though.

In any case, it would appear that we're targeting an issue that's independent from the API.

owocki:
while i agree that the changes that need to be made are not API-centric, the only place where this issue manifests is within the API.  

happy to have a convo about tradeoffs of the diff approaches next chance we get

kuhnchris:
This was superseeded by (at least) #3628

PR #1026, opened 2018-04-30
drastically improve api performance

owocki:
##### Description
This stores data that was previously calculated on read to be calculated on write. 

On my local, this takes the response time of an 'Any' API request (which fetches about 400 issues) down from 3.5 seconds to 1.2 seconds.

@mbeacom i know you and i don't quite see eye to eye on whether this is the right final move or not.. but i was on a call today and passed the time by coding this up / testing real quick.  figured it might be an adequate stopgap

##### Checklist

API

##### Testing
tested it

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/950


app/dashboard/models.py
```diff
@@ -392,11 +399,11 @@ def status(self):
                 return 'unknown'
 
     @property
-    def value_true(self):
+    def _value_true(self):
```
mbeacom: Can we move away from using `_` for public methods?

owocki:
> if possible, we should consider moving away from prefixing methods with _ or __ unless the method is intended to be private.

good point.. ill update these methods to be `get_` instead of `_`

fixing tests now too

app/dashboard/models.py
owocki: done

owocki:
@mbeacom just put up two commits to address your concerns around naming conventions and tests

Issue #1013, opened 2018-04-27
List used contracts under metamask icon in the footer


abitrolly:
### User Story

As a ETH enthusiast, I want to see what contracts belong to gitcoin or what contracts it uses, so that I can ensure that I am using right interface. I also want a quick way to see if Metamask is on or off in the footer.

### Why Is this Needed?

This can serve as onboarding experience for new contributors. Also as an indicator that Gitcoin is DAPP, and also while site is visible, it is not clear what public code in blockchain this deployment uses.

### Description

*Type*: Feature

### Current Behavior

I can't inspect actual contract versions that are used, can't check that they are security audited and by whom.

### Expected Behavior / Solution

Metamask icon with dropdown arrow in the footer. Icon is colored if Metamask is active. On arrow click, the control expands with a list of addresses and their links on Etherscan (alternatives?).

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started on the 17.0 DAI (17.0 USD @ $1.0/DAI) funding by__: 
 1. [@norhh](https://gitcoin.co/profile/norhh)  

 On the above list? Please leave a comment to let the funder (@owocki) know your  plan. If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/1013/320)
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $6,020.75 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

thelostone-mc:
@abitrolly

>As a ETH enthusiast, I want to see what contracts belong to gitcoin or what contracts it uses, so that I can ensure that I am using right interface

Noob question:  Could you elaborate on this? You mean you'd like to see the contract (bounties network / gitcoin)? That's something which would be a set of generic ones which is what would be used by every bounty put on gitcoin. Should this be something if needed, we put in the `onboarding` flow rather than across the site? Or am I entirely wrong in my understanding  😂 

#### Metamask status

[...]

PR #1062, opened 2018-05-03
WIP: List Used Contracts Under Metamask Icon In The Footer

dolebas:
##### Description

As Metamask/Web3 status is already present in the upper left corner, would this be sufficient?

image

##### Checklist

##### Testing

In progress

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/1013

### DIFF
--- a/app/retail/templates/shared/footer.html
+++ b/app/retail/templates/shared/footer.html
@@ -28,6 +28,22 @@
                 <a class="nav-link" href="{% url "terms" %}" target="_blank" rel="noopener noreferrer">{% trans "Terms" %}</a>
                 <a class="nav-link" href="{% url "privacy" %}" target="_blank" rel="noopener noreferrer">{% trans "Privacy" %}</a>
                 <a class="nav-link" href="{% url "help" %}">{% trans "Help" %}</a>
+                <div class="nav-item dropdown">
+                  <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button"
+                  data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                  Contracts
+                  </a>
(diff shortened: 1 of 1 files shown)


thelostone-mc:
Still in two minds whether this belongs on the footer as against in the about page 
This is not something which the user would want to see everytime. 

@vs77bb  @PixelantDesign What do you think?

mbeacom:
I agree.  It seems out of place.  Then again, I feel like anyone that would be interested in viewing the contract would be able to easily find it via a tx.  I think it would make more sense to keep the list somewhere they would intentionally visit to find out more about the platform (about?) versus easy access to it at all times.

dolebas:
One aspect of this growing pattern that makes sense could be that it is easily findable and verifiable at the time of contract execution. In the screenshot below, the contract address in the bottom left corner matches the contract address in MetaMask.

image

abitrolly:
1. Where do names of smart contracts come from?
2. I was thinking about specific addresses
3. I was thinking about them to be always visible, perhaps with a horizontal paging in the footer, no dropdowns

abitrolly:
https://github.com/gitcoinco/web/blob/f1d7fbc3e7fd751466359060df4570e5e02d3b35/app/assets/v2/js/abi.js - is that all contracts that are used?

owocki:
yes that is one of them.

`app/assets/yge/youvegoteth/index.js` is another

abitrolly:
https://github.com/gitcoinco/web/blob/f1d7fbc3e7fd751466359060df4570e5e02d3b35/app/assets/yge/youvegoteth/index.js#L32

Is it possible to move into some global configuration file that is easy to find?

Issue #1011, opened 2018-04-27
Explorer multiselect "Any" option doesn't behave as expected


joshmorel:
### User Story

As an explorer, I want a multiselect "Any" control that selects all.

### Why Is this Needed?

This UI is confusing if intended, if not intended it's a bug
### Description

Bug

### Current Behavior

When clicking Any on the gitcoin explorer multiselect options, the only thing that happens is the page refresh.

### Expected Behavior

"Any" will be selected, page will refresh with all in that search category.

### Reproduction

1. Navigate to [gitcoin explorer(https://gitcoin.co/explorer) in Chrome or Firefox
2. Expand "NETWORK" option menu
3. Click "Any" button

### Solution

* [ ] Create design for Any button (*not* additional checkbox - more like a [button](https://docs.webix.com/desktop__multicombo.html))
* [ ] Update JS code so that clicking that button will select all options
* [ ] Add ability to clear select "all" with respect to above 2
* [ ] Test
* [ ] Gain acceptance

### Definition of Done
Multiselects in Explorer have All/Clear button which responds in an intuitive fashion

### Additional Information
I'm willing to take this on.

thelostone-mc:
@joshmobley You mean you'd like a button where on click all the filters are removed? 

We've got a reset  -> which selects `Any` aka removes all filters 

<img width="244" alt="screen shot 2018-04-29 at 2 29 53 pm" src="https://user-images.githubusercontent.com/5358146/39404970-fd387330-4bb9-11e8-9848-911aae8ba22a.png">

PixelantDesign:
Any should show all. Is this a bug @mbeacom?

joshmorel:
Well, really a new design isn't needed, to be honest. The multiselect options just need to work like "PROJECT STATUS"

For example in "EXPERIENCE LEVEL" clicking "Any" does nothing except refresh the search result with the exact same filter as before.

joshmorel:
![peek 2018-04-29 16-35](https://user-images.githubusercontent.com/10546720/39410776-6c12ef7c-4bcb-11e8-82c8-8f3f4f05a6dc.gif)

Issue #1064, opened 2018-05-03
Local storage of explorer filter does not consistently load


joshmorel:
### User Story

As a web bounty explorer using Chrome or Firefox, I want selected filters to persist *consistently* so that I can use my preferred searching method. 

### Why Is this Needed?

*Summary*: Local storage of explorer filter is not consistent.

### Description

*Type*: Bug

### Current Behavior

1. Select 3 options in NETWORK group
2. Refresh browser
3. Only 2 options selected

### Expected Behavior

The same options would persist between browser refresh in Chrome.

### Reproduction

1. In Chrome navigate to https://gitcoin.co/explorer
2. Select 3 option within `NETWORK` group - `Ropsten`, `Rinkeby`, `Mainnet`
3. Refresh browser (hold `SHIFT` if necessary to bypass cache) 
4. Inspect `NETWORK` group - only `Ropsten` and `Mainnet` are selected

### Solution
*Summary*: TBD - willing to take a look

### Definition of Done

Selected filters = loaded filters

joshmorel:
Just an update looking into this more for all the different filters:

* Status (Radio): This is the only one that works as expected
* Experience Level, Project Length, Type, Network: Only once option is kept in localStorage, when refreshing however, sometimes two are selected
* Tech Stack: Multiple comma-delimited options are kept in localStorage, but on refresh the selected options are not populated

thelostone-mc:
@joshmorel Ah thanks for pointing that out! ^_^
Can't believe we didn't see that ! -.-

PR #1065, opened 2018-05-03
fix toggle any filter

joshmorel:
@thelostone-mc - please take a look, I found passing the event to the now renamed `toggleAny` function made it easier to reason about and achieve the expected behaviour.

##### Description
Fix toggle any filter - could not previously click "Any" 

##### Checklist

- [x ] linter status: 100% pass
- [ x] changes don't break existing behavior
- [ x] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#step-4-commit)

```   
    for multiselect checkbox groups:
    if all selected --> others deselected
    if other selected --> any deselected
    if none selected --> any selected
```

UI

##### Testing

Confirmed XHR request matches to existing app for:

1. Initial load
1. Select another option
1. Deselect all options within group (any)
1. Select other option from any
1. Select from techstack 

Note that I noticed a weird issue with local storage that is also in production for [which an issue was created](https://github.com/gitcoinco/web/issues/1064).

##### Refers/Fixes
Fixes: #1011

### DIFF
--- a/app/assets/v2/js/pages/dashboard.js
+++ b/app/assets/v2/js/pages/dashboard.js
@@ -119,9 +119,11 @@ var set_filter_header = function() {
 };
 
 var toggleAny = function(event) {
-  if (!event) return;
+  if (!event)
+    return;
   var key = event.target.name;
   var anyOption = $('input[name=' + key + '][value=any]');
+
   // Selects option 'any' when no filter is applied
   if ($('input[name=' + key + ']:checked').length === 0) {
     anyOption.prop('checked', true);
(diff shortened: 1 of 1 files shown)


app/assets/v2/js/pages/dashboard.js
```diff
@@ -118,28 +118,23 @@ var set_filter_header = function() {
...
-      $('input[name=' + key + '][value=any]').prop('checked', true);
-    }
+  // Selects option 'any' when no filter is applied
+  if ($('input[name=' + key + ']:checked').length === 0) {
+    anyOption.prop('checked', true);
+    return;
+  }
+  if (event.target.value === 'any') {
+    // Deselect other filters when 'any' is selected
+    $('input[name=' + key + '][value!=any]').prop('checked', false);
```
thelostone-mc: 👍

Issue #1091, opened 2018-05-07
As a member of the community, I want Profile Privacy Features, so I can be more private


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)

As a member of the community, I want Profile Privacy Features, so I can be more private
### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

1. provide the ability for users to hide their contact info on bounties.
1. write a migration such that anyone who's never logged into gitcoin before will not have a profile
1. Provide the ability for users to hide their profile in their settings
1. When someone goes to gitcoin.co/profile/xxxxx and xxxxx does not exist in our system.. they shoudl be given an oppy to invite them (instead of just us ingesting that info now)

mbeacom:
Just a note for those of you coming from into this for the first time...  We don't indiscriminately scrape Github data...  Currently, when you hit `https://gitcoin.co/profile/<github_user_handle>` you will be show a Gitcoin profile card for the corresponding Github user...  This does not automatically opt a user into Gitcoin.  It's rendering data you already have made publicly available and is already accessible for Anyone on the internet.

For example:

https://api.github.com/users/mbeacom

owocki:
> It's rendering data you already have made publicly available and is already accessible for Anyone on the internet.

That said, we can respect and understand why people are discomforted by the card being presented as Gitcoin branded when they have not explicitly signed up before.  

I just put in the first fix for this https://github.com/gitcoinco/web/pull/1095

PR #1095, opened 2018-05-07
As a member of the community, I want these Profile Privacy Features, so I can be more private

owocki:
##### Description

As a member of the community, I want these Profile Privacy Features, so I can be more private

1. set the default profile hidden = False, but profile visible = True when the user logs in
1. write a migration such that anyone who's never logged into gitcoin before will not have a profile unless they explicity sign up.

##### Checklist

profiles

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/1091


app/dashboard/migrations/0069_auto_20180507_1841.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.0.4 on 2018-05-07 18:41
+
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    Profile = apps.get_model('dashboard', 'Profile')
+    profiles = Profile.objects.filter(github_access_token='').all()
```
mbeacom: We'll want to make sure we're not marking valid profiles to hidden if they have a `user` as well.  Also, we should be able to just do a `Profiles.objects.filter(github_access_token='').update(hide_profile=True)`

mbeacom:
Also, looks like we'll need to update the Leaderboard tests.

app/dashboard/migrations/0069_auto_20180507_1841.py
owocki: 3d7e03ad8996c3057490b56d8846313f09af67a8

owocki:
> Also, looks like we'll need to update the Leaderboard tests.

yes just updated

Issue #1030, opened 2018-04-30
Funder Dropdown Options


vs77bb:
### User Story

As a Funder, I want to know the top funding options (ETH, DAI) so that I can fund my bounties in the currency which is most easily consumed by the developer community.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: Make ETH and DAI the top two currencies on the Funder drop-down, and then roll alphabetically from there with all other accepted ERC-20 tokens. 

### Actual behavior
Currently, fully alphabetical order. 

### Expected behavior
ETH, DAI, then alphabetical order.

thelostone-mc:
We could default the value to ETH so that that gets populated as the default value !

I agree that the ETH and DAI are the most popular -> but is it okay to mix up the order ?
(Our dropdown have a built in search filter so prevents them from aimlessly scrolling )

That said if users feels that they'd like to see the popular tokens on the top of the list -> we could do that ^_^

owocki:
> We could default the value to ETH so that that gets populated as the default value !

this happens already :)

> I agree that the ETH and DAI are the most popular -> but is it okay to mix up the order ?

we could make it so that they appear twice!

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.015 ETH (11.12 USD @ $741.02/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/1030/395).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $8,694.40 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.

 1. [@iamonuwa](https://gitcoin.co/profile/iamonuwa) 

 has committed to working on this project to be completed 4 weeks, 1 day from now.

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

[...]

PR #1115, opened 2018-05-09
Fixes: https://github.com/gitcoinco/web/issues/1030

bakaoh:
##### Description
Add priority to sort token list

##### Checklist

Asset js

##### Testing
Yes

##### Refers/Fixes
Fixes: #1030


owocki:
ux questions:

do we want to move them to the top or just duplicate them at the top? 
should there be a 'bar' between the priority sorted ones and the alphabetical ones, like so?

```
ETH
DAI
==========
A
B
C
```

owocki:
code looks good and sane to me!

thelostone-mc:
@PixelantDesign do you think we should repeat it ?
I'm leaning towards it cause it's makes sense as I scroll through the list but then again user might notice it being redundant

PixelantDesign:
@owocki @thelostone-mc 
It's okay to repeat because the list is very long.

I like this approach:

ETH
DAI
==========
A
B
C

thelostone-mc:
@bakaoh Ah sorry about this ! Could we update the code to let the list unchanged and just use the priority to duplicate the tokens & display them on top of the list ?

bakaoh:
Ok, i will do that on monday

bakaoh:
@thelostone-mc -- i've duplicated the tokens & display them on top of the list as your request ^^

app/assets/v2/js/tokens.js
```diff
@@ -168,40 +168,42 @@ var tokens = function(network_id) {
       {'addr': '0xb581e3a7db80fbaa821ab39342e9cbfd2ce33c23', 'name': 'ARCD', 'decimals': 18},
       {'addr': '0x255aa6df07540cb5d3d297f0d0d4d84cb52bc8e6', 'name': 'RDN', 'decimals': 18},
       {'addr': '0x056017c55aE7AE32d12AeF7C679dF83A85ca75Ff', 'name': 'WYV', 'decimals': 18},
-      {'addr': '0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359', 'name': 'DAI', 'decimals': 18},
+      {'addr': '0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359', 'name': 'DAI', 'decimals': 18, 'priority': 100}
```
thelostone-mc: `, 'priority': 100},`

thelostone-mc:
@PixelantDesign this is how it looks !

It looks off as the when I open the dropdown -> I have to scroll to the top to see DAI and ETH.
Still on 2 minds if it's needed ! @owocki 

Also I believe @PixelantDesign wanted a line separating the two ( a`border-bottom maybe` ? )  as mentioned [above](https://github.com/gitcoinco/web/pull/1115#issuecomment-388233248)

x1

cc @mbeacom

app/assets/v2/js/tokens.js
owocki: tokens.js:172 Uncaught SyntaxError: Unexpected token {

bakaoh:
@owocki --- i've done as your requests

thelostone-mc:
@bakaoh could you resolve the conflicts ?
cc @PixelantDesign

PR #1076, opened 2018-05-05
Dynamic widget documentation

matthewlilley:
- [x] linter status: 100% pass
- [x] changes don't break existing behavior
- [x] commit message follows [commit guidelines]
##### Affected core subsystem(s)
doc
##### Refers/Fixes
https://github.com/gitcoinco/web/issues/354

PR #1126, opened 2018-05-10
Adds cors headers, and conditionally allow all for GET requests of bounties.

matthewlilley:
##### Description
Adds cors headers, and conditionally allow all for GET requests of bounties.

##### Checklist

##### Refers/Fixes
https://github.com/gitcoinco/web/pull/1076

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -56,6 +56,7 @@ ENABLE_NOTIFICATIONS_ON_NETWORK = env(
 
 # Application definition
 INSTALLED_APPS = [
+    'corsheaders',
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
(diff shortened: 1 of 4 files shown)


thelostone-mc:
Code looks alright 👍 

@matthewlilley @mbeacom haven't worked with a large scale project like this before up close & personal , any security implications of enabling CORS for`GET`calls ?

matthewlilley:
@thelostone-mc It wouldn't hurt to get a second opinion here, though I believe the cors headers should only be sent for that single endpoint "/api/v0/bounties". I was unsure if POST methods are used on this endpoint, as it seems to be a resource, so to be cautious I limited the method to GET only.

owocki:
>  It wouldn't hurt to get a second opinion here, though I believe the cors headers should only be sent for that single endpoint "/api/v0/bounties".

yes tahts right

owocki:
i didnt test.. but code looks sane. very happy if this enables https://github.com/gitcoinco/web/pull/1076 !
@mbeacom what do you think of this?

PR #1170, opened 2018-05-16
Add fulfillment endpoint for each entry by std bounty id

mbeacom:
##### Description

The goal of this PR is to split off some view context generation and add a url route for bounty fulfillment lookup by `network` and `standard_bounties_id` to be used by BN and @mbeylin.

This is a quick patch to provide quick fulfillment linking for bounties network.

Example Usage:

`https://gitcoin.co/issue/fulfill?standard_bounties_id=111&network=rinkeby`

The above would yield the fulfillment page for the bounty associated with `standard_bounties_id=111` on the Rinkeby network.  `network` defaults to `mainnet` if not provided.

##### Checklist

backend

##### Testing
Tested locally


app/dashboard/models.py
```diff
@@ -658,12 +658,30 @@ def action_urls(self):
 
         """
         return {
-            'fulfill': f"/issue/fulfill/{self.pk}",
+            'fulfill': f"/issue/fulfill?pk={self.pk}&network={self.network}",
             'increase': f"/issue/increase/{self.pk}",
```
owocki: is it worth making all of these URLs follow the same format?  i know its a little bit of scope creep but the consistency will be worth it i think

PR #1171, opened 2018-05-16
removes get-prices from docker command

owocki:
* its not actually needed to run the app
* it makes restarting your docker container 4 minutes longer than it needs to be


mbeacom:
@owocki Get prices only happens once, even if you `docker-compose restart web`.  It will only occur again if you completely destroy the container via `docker-compose down -v` . If we're not going to run this on initial provisioning as it is now, we might want to replace it with ConversionRate fixtures.

mbeacom:
@owocki We can totally go this route! I'm good with yanking it out, I just figured we might want to include an entry or two as a general base.  Want me to add them and we'll merge this?

PR #1175, opened 2018-05-16
Add CR fixtures and disable get_prices by default

mbeacom:
##### Description

The goal of this PR is to introduce `ConversionRate` initial fixture data for local development.  Additionally, this PR disabled `get_prices` from running during provisioning, but allows users to pass `FORCE_GET_PRICES` via env var to force loading fresh conversionrates on container restart.

##### Checklist

docker, backend

##### Testing
Tested locally

### DIFF
--- a/app/economy/fixtures/initial.json
+++ b/app/economy/fixtures/initial.json
@@ -1,0 +1,18900 @@
+[{
+  "model": "economy.conversionrate",
+  "pk": 1,
+  "fields": {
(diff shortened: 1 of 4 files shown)


mbeacom:
The only gripe I have about this is the size of the fixture file... but this is expected, since the initial dataset is so large.  This is still much faster than running `get_prices` since we're dumping from json.

Issue #1007, opened 2018-04-27
as a user, i want one click approval of erc20 token bounties, because its faster and easier


owocki:
follow up from https://github.com/gitcoinco/web/pull/1006

as a user, i want one click approval of erc20 token bounties, because its faster and easier

right now, there are two confirmations (one for `.approval()` and AFTER THE BLOCK CONFIRMS FOR THE FIRST the next to issue the bounty).  Itd be an ideal if the user only had to do 1 tx in order to do an erc20 bounty

Possible implementation paths
1. In the future, we could batch up the transactions using web3js batch ( which i haven't had any luck with thus far)
2) we could also create an atomic contract which manages the `approval()` and the `issueAndActivateBounty` all in one call, and just use that in web3js

Definition of Done:
1. user only has to click the metamask approval once when submitting an ERC20 bounty.

owocki:
https://github.com/gitcoinco/web/pull/1006

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.222 ETH (143.65 USD @ $647.08/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/1007/347).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $8,228.21 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.

 1. [@bakaoh](https://gitcoin.co/profile/bakaoh) 

 has committed to working on this project to be completed 4 weeks, 1 day from now.


__Please answer following questions/comments__ @owocki:

I think i will go the second path, which use an contract to do that. Correct me if i'm wrong but i don't think batch transaction can resolve this issue

bakaoh:
It looks like i picked the wrong path. Atomic contract will not work except the token implement approveAndCall (https://github.com/ethereum/EIPs/issues/827), which is not erc20 standard
[...]

PR #1190, opened 2018-05-20
READY FOR QA : one time approvals for each erc20 token used on gitcoin

owocki:
##### Description
one time approvals for each erc20 token used on gitcoin per token (instead of one approval per bounty posted)

this makes it much easier to use erc20 tokens on the platform since you only have to approve() once and then you can use the tokens hundreds of time after that.

this is similar to (and inspired by) how radar relay does this.  they do one click `.approve()`s for each ERC20 token you use.. then you never need to go through that process again with that token

[screencast](http://bits.owocki.com/3v2G3A0q2J2H/Screen%20Recording%202018-06-08%20at%2002.38%20PM.gif)

##### Checklist

New funding, increase funding, tip

##### Testing
Custom network

##### Refers/Fixes
Fixes: #1007


bakaoh:
@owocki --- I think this route give a better UX, but we need to convince funders that ours smart contract is safe. I will never accept `approve` request from radar relay without carefully read their contract. Even if i do read it, i'll till hesitant. 

And we will need a way for the funders to remove the approval.

owocki:
> @owocki --- I think this route give a better UX, but we need to convince funders that ours smart contract is safe. I will never accept approve request from radar relay without carefully read their contract. Even if i do read it, i'll till hesitant.

the standardbounties contract is open source and if you read it, you will see that this approval() cannot be manipulated in any way. -- @mbeylin @villanuevawill could we maybe upload the source of it to etherscan?  better yet, we'd love to have a security audit posted.

> And we will need a way for the funders to remove the approval.

do we need this for MVP? i guess we could create a `settings/` page for this

app/assets/v2/js/pages/tokens_settings.js
```diff
@@ -0,0 +1,66 @@
...
+
+    e.preventDefault();
+    //actual approval
+    var token_contract = web3.eth.contract(token_abi).at(token_address);
+    var from = web3.eth.coinbase;
+    var to = contract_address;
+    token_contract.allowance.call(from, to, function(error, result){
+    if(error || result.toNumber() == 0){
+      var amount = (2**256)-1; // uint256
+      var amount = 10 * 18 * 9999999999999999999999999999999999999999999999999999; // uint256
```
mbeacom: Which of these `amount` variables do we want to keep? @owocki

Issue #1241, opened 2018-05-23
GDPR: Update Email Preferences


PixelantDesign:
*Due: 5.25.18*

Allow users to select or unselect which email they would like to receive

Examples in mockup are listed, please use full list available in database.

Add ability to save.

![Screen Shot 2018-05-23 at 3.09.06 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5a83295aa5cc6700136b4bc8/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b77266456f06b7191a37bf112224f5f595ecc3c4a4356bcbbd671f3e57862804e0ebcd73c5c22953910afb9864aee01ba24df4fd4b0e9821ad2f42f67b6d5db1063474f4dcbed5727a19e41c95df44ce04d57a21b3567ee772431.png)

owocki:
PR is in ^^

PR #1247, opened 2018-05-23
Email Preferences at a much more granular level

owocki:
##### Description

Email Preferences at a much more granular level

Removes (or rather, migrates) email 'level's and adds a much more granular email preference data structure.

screen recording 2018-05-23 at 04 46 pm

##### Checklist

email settings

##### Refers/Fixes

Fixes https://github.com/gitcoinco/web/issues/1241


app/marketing/models.py
```diff
@@ -64,6 +65,54 @@ def __str__(self):
     def set_priv(self):
         self.priv = token_hex(16)[:29]
 
+    def should_send_email_type_to(self, _type):
+        is_suppressed = es.preferences.get('suppression_preferences', {}).get(_type, False)
```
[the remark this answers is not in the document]
SaptakS: This should be `self` instead of `es` right?

app/retail/templates/settings/email.html
```diff
@@ -2,28 +2,39 @@
...
-          </select>
-          <div class="form__select__icon">
-            {% include 'svgs/arrow-down.svg' %}
-          </div>
+        <a href="#" id="select_all" data-direction="+" >[Select All]</a>
+        {% for email_type in email_types %}
+        <div class="email_pref_container">
+            <input class='priv_checkbox' type=checkbox name="{{email_type.0}}" id="{{email_type.0}}" value="1">
+            <label class="form__label" for="{{email_type.0}}">{% trans "Don't send me" %} <strong>{{email_type.1}}</strong></label>
+            <small id="{{email_type.0}}" class="form__input-help">{% trans "Frequency" %}: {{email_type.2}}</small>
```
SaptakS: Since in both line 30 and 32 we have `id="{{email_type.0}}"`won't this create duplicate ids? id is unique in a particular html and hence may cause problems if tried to access from JS. Maybe we can use something like `id="{{email_type.0}}-checkbox"` in the checkbox?

app/retail/templates/settings/email.html
```diff
@@ -41,13 +52,35 @@ <h5>{% trans "Basic Preferences" %}</h5>
...
   </div>
 {% endblock %}
 
 {% block scripts %}
 <script>
+    $("document").ready(function(){
+        // select / deselect all
+        $("#select_all").click(function(e){
+            e.preventDefault();
+            if($(this).data('direction') == '+'){
```
SaptakS: Do we want to use `===` maybe?

app/retail/templates/settings/email.html
```diff
@@ -41,13 +52,35 @@ <h5>{% trans "Basic Preferences" %}</h5>
...
+                $(this).data('direction','-');
+            } else {
+                $(".priv_checkbox").prop('checked', false);
+                $(this).text('[Select All]');
+                $(this).data('direction','+');
+            }
+        });
+
+        // suppression preferences form
+        var suppression_preferences = {{suppression_preferences|safe}};
```
SaptakS: Format in `{{ var }}`

app/retail/templates/settings/email.html
```diff
@@ -41,13 +52,35 @@ <h5>{% trans "Basic Preferences" %}</h5>
...
       {% csrf_token %}
-      <input class='button button--primary' type='submit' name='submit' value="Go">
+      <input class='button button--primary' type='submit' name='submit' value="Save">
     </form>
   </div>
 {% endblock %}
 
 {% block scripts %}
 <script>
+    $("document").ready(function(){
```
SaptakS: I think `$(document)` (the gobal variable) instead of `$("document")` (the string) ?

app/retail/templates/settings/email.html
SaptakS: Also, here and in the for loop and callback function, I think we should have a space before `{` for better indentation. And also after `if` and `for`

mbeacom: Fixing this and the double quotes in the JS blocks.

mbeacom:
@SaptakS @thelostone-mc updates are in for this.

app/marketing/management/commands/roundup.py
```diff
@@ -58,7 +58,7 @@ def handle(self, *args, **options):
         exclude_startswith = options['exclude_startswith']
         filter_startswith = options['filter_startswith']
 
-        queryset = EmailSubscriber.objects.filter(newsletter=True)
+        queryset = EmailSubscriber.objects.all()
```
mbeacom: @owocki We're good to iterate over *every* use for this, right?

PR #1279, opened 2018-05-25
records form submission records for GDPR purposes + mailchimp permi-delete

owocki:
##### Description

records form submission records for GDPR purposes
also does a permi-delete in mailchimp when users account is closed

##### Checklist

settings/profiles

##### Testing

tested it

##### Refers/Fixes

Self, Emails


owocki:
@SaptakS this is a starter PR to register the form submission in which we contain / modify consent from our users.  

we could either
1. merge this now and add the cookie modal records to another PR 
2. add in in the cookie modal records (from email with carl in consensys legal) to this PR

SaptakS:
I will add the cookie consent to this PR if that's okay.

owocki:
@SaptakS sounds great

app/marketing/views.py
```diff
@@ -407,15 +412,30 @@ def account_settings(request):
...
+
+            # remove email
+            try:
+                client = MailChimp(mc_user=settings.MAILCHIMP_USER, mc_api=settings.MAILCHIMP_API_KEY)
+                result = client.search_members.get(query=es.email)
+                subscriber_hash = result['exact_matches']['members'][0]['id']
+                client.lists.members.delete(
+                    list_id=settings.MAILCHIMP_LIST_ID,
+                    subscriber_hash=subscriber_hash,
+                    )
```
mbeacom: When we start running automatic linting/styling checks, the indentation on these parentheses will be automatically moved up to the previous line or the indent will move inward by one with the trailing `,`

app/marketing/views.py
```diff
@@ -407,15 +412,30 @@ def account_settings(request):
...
+            except Exception as e:
+                print(e)
```
mbeacom: Can we use a logger here?

app/marketing/views.py
```diff
@@ -407,15 +412,30 @@ def account_settings(request):
...
+            # remove email
+            try:
+                client = MailChimp(mc_user=settings.MAILCHIMP_USER, mc_api=settings.MAILCHIMP_API_KEY)
+                result = client.search_members.get(query=es.email)
+                subscriber_hash = result['exact_matches']['members'][0]['id']
+                client.lists.members.delete(
+                    list_id=settings.MAILCHIMP_LIST_ID,
+                    subscriber_hash=subscriber_hash,
+                    )
+            except Exception as e:
```
mbeacom: Which exception(s) are we expecting here? IndexError and KeyError per line 432 and probably a `MailChimp` connection error?

app/marketing/views.py
```diff
@@ -153,6 +155,15 @@ def privacy_settings(request):
     return TemplateResponse(request, 'settings/privacy.html', context)
 
 
+def record_form_submission(request, obj, _type):
```
mbeacom: Can we rename `_type` to `record_type` or `submission_type` to avoid confusion as a private var?

owocki: updated

app/marketing/views.py
owocki: any exception.  the point is that any exception that happens shouldn't cause the page to 500

app/marketing/views.py
owocki: updated

app/marketing/views.py
owocki: ok.  maybe i should download a sublime linter.

app/marketing/views.py
```diff
@@ -50,6 +51,9 @@
 from retail.helpers import get_ip
 
 
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
owocki: just fixed

app/marketing/views.py
SaptakS: @mbeacom do we follow a strict PEP8 or flake8?

SaptakS: @mbeacom as mentioned in [PEP-8](https://www.python.org/dev/peps/pep-0008/#indentation), this might be still correct indentation I think:

> The closing brace/bracket/parenthesis on multiline constructs may either line up under the first non-whitespace character of the last line of list

> or it may be lined up under the first character of the line that starts the multiline construct, as in:

SaptakS:
@mbeacom since we are thinking of migrating from cookielaw, I think I will handle that separately from this PR. So it's good to be merged.

Issue #1282, opened 2018-05-26
As a user, I want Gitcoinbot spam post mortem, so no one accidently spams the network


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a user, I want Gitcoinbot spam post mortem, so no one accidently spams the network

### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
See [Post Mortem](https://medium.com/gitcoin/github-notification-spam-post-mortem-e1ed200936e)

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: 
Feature

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

v1:

* Better Docs
* Make Gitcoinbot more cowardly in dev
* What else?

v2: 
* move all contact information off chain (emails, github usernames, web URLs)
* move all communication stuff out of the dapp
* move notifications.py to its own pypi app
* move sync_geth.py to its own pypi app
* make those installable via pypi, but only if explicit
* what else?

PR #1283, opened 2018-05-26
notification spam prevention: cowardly gitcoin bot

owocki:
##### Description

v1:

* Cowardly Gitcoin bot
* Better Docs
* What else?

See [Post Mortem](https://medium.com/gitcoin/github-notification-spam-post-mortem-e1ed200936e)

##### Checklist

- [x] linter status: 100% pass
- [x] changes don't break existing behavior
- [x] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#step-4-commit)

##### Affected core subsystem(s)

bot

##### Testing

tested it

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/1282

PR #1284, opened 2018-05-26
Add checks for bounty/tip notifications on local/dev

mbeacom:
##### Description
The goal of this PR is to introduce checks to mitigate notifications for `mainnet` bounties when the notification originates from a local or development environment.

##### Checklist

bounty, tip, notifications

##### Testing
Locally

##### Refs

Ref: #1282


app/dashboard/models.py
```diff
@@ -1223,7 +1250,7 @@ def update_slack_integration(self, token, channel, repos):
 
     @staticmethod
     def get_network():
-        return 'mainnet' 
+        return 'mainnet'
```
owocki: should we add rinkeby back in here?

app/dashboard/utils.py
```diff
@@ -186,6 +188,9 @@ def getBountyContract(network):
 
 
 def get_bounty(bounty_enum, network):
+    if (settings.DEBUG or settings.ENV != 'prod') and network == 'mainnet':
```
owocki: should this b e in the PR?

app/dashboard/utils.py
```diff
@@ -253,6 +258,9 @@ def get_bounty(bounty_enum, network):
 
 # processes a bounty returned by get_bounty
 def web3_process_bounty(bounty_data):
+    if not bounty_data or (settings.DEBUG or settings.ENV != 'prod') and bounty_data.get('network') == 'mainnet':
```
owocki: might be worth dropping a comment or print statement here s aying what this is for for a future maintainer?

owocki: also.. `sync_geth` is the main way that people get data into their locals... and this will break that.

app/dashboard/models.py
SaptakS: I think something like [this](https://github.com/gitcoinco/web/pull/1283/files#diff-0d2d86c824ef942c5a3a01c83ddfe758R1226) would be good.

app/dashboard/utils.py
SaptakS: +1 for enabling rinkeby bounties.

app/dashboard/models.py
mbeacom: @SaptakS Yeah, that was originally in `master` for a while.  I'll re-add it here.

app/dashboard/utils.py
mbeacom: Yeah. I intentionally added this to avoid even syncing `mainnet` bounties on local.

app/dashboard/utils.py
mbeacom: This won't stop syncing anything other than `mainnet` bounties on non-prod environments.

PR #1317, opened 2018-05-30
docker-compose was wrong

owocki:
it was loading from local.env instead of `.env`, and therefore the following info was set

```
GITHUB_API_USER=None
GITHUB_API_TOKEN=None
GITHUB_CLIENT_ID=xxxx
GITHUB_CLIENT_SECRET=xxx
GITHUB_APP_NAME=MyGitcoinApp
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,7 @@ services:
       - PYTHONUNBUFFERED=1
       - PYTHONPATH=/code/app
     env_file:
-      - app/app/local.env
+      - app/app/.env
     build: .
     volumes:
       - .:/code


owocki:
cc @mbeacom since he's the docker czar on the team

mbeacom:
Thanks for catching this. I think I originally left it like thus to avoid failures for new contributor environments where they don't follow the instruction to create the env file. This makes the most sense, though!

mbeacom:
Just a note...  The web app should still respect the values from `.env` over the `local.env` variables passed as environment variables to the container via `docker-compose`.  🤔

owocki:
i wasnt experiencing that.. my github creds were def overwritten before this change

mbeacom:
`django-environ` must give precedence to existing env vars over the env file values.  That would explain why this started occurring following: https://github.com/gitcoinco/web/commit/0d2027f7db76dd14bb5563f1345f5641a7f40706

In any case, `.env` should be the default env file. 🙌

PR #1318, opened 2018-05-30
Don't save tip information in local storage except current user's

### LINKED ISSUE #1315: Tip form pre-populates with last sent tip information
### User Story

As a user, I don't want the tip form to be pre-populated with my last sent tip field data

### Why Is this Needed?

*Summary*:  The current behavior can have undesired results since: to email, to github username, and comments are restored from the last sent tip.
If a user doesn't opt to expand the form to: advanced, they might submit the form with unwanted field entries from their last sent tip.

### Description

*Type*: Enhancement

### Current Behavior

screenshot 2018-05-30 12 56 07

### Expected Behavior
Don't pre-populate the form with the last sent tip information.

### Reproduction
Send a tip.

### Solution

*Summary*: Remove the default field values from the .html template.

### Definition of Done

The updated tip form no longer pre-populates previously sent tip information like the recipient email, github username, public and private comments, etc.

### Additional Information

If necessary, feel free to ping me or another team member on slack or in this thread for additional guidance.

arun1595:
##### Description

Doesn't store sent tip info in local storage and hence doesn't pre-populate tip info in send tip form.

Also wraps `localStorage.setItem` calls around `try...catch` as the browser might throw an exception in case there's no space.

##### Checklist

##### Refers/Fixes

Fixes #1315


app/assets/yge/youvegoteth/send.js
```diff
@@ -27,7 +27,7 @@ var unPackAddresses = function() {
       }
     }, 3000);
   }
-  localStorage.setItem('addresses', null);
+  localStorage.setItem('addresses', '');
```
arun1595: This change is required because `localStorage` serializes the value `null` to a string. So, `localStorage.getItem('addresses')` will return a string `'null'` which is truthy in JS. So, the condition in line 22 won't catch this issue.

/cc @mbeacom

app/assets/yge/youvegoteth/send.js
```diff
@@ -54,27 +54,12 @@ window.onload = function() {
   if (localStorage['amount']) {
     $('amount').value = localStorage['amount'];
   }
-  if (localStorage['username']) {
-    $('username').value = localStorage['username'];
-  }
-  if (localStorage['issueURL']) {
```
thelostone-mc: Why was this taken away ? Cause `send.js` right now is used a few more places , not just tip
So I believe if we remove this  -> those might get affected (like the ETHOS flow )

Also these are just used to autopopulate the input fields , I would be a lot more comfortable if this stayed until we get around to clean up 😅 

Your thoughts @arun1595 ?

@mbeacom  / @SaptakS

SaptakS: I agree. Also in try-catch, you seem to have removed some of the `setItems`

arun1595: Oh, I didn't realize this was used in other places of the app. Let me fix it.

arun1595: @thelostone-mc @SaptakS I've extracted the pre-population code to a function and it only gets called if the current page is not a send tip page. This is just a temporary fix until we clean up the dependency.

Let me know what you guys think.

/cc @mbeacom

app/assets/yge/youvegoteth/send.js
```diff
@@ -43,40 +44,54 @@ var updateEstimate = function(e) {
...
-  jQuery('#token').on('change', updateEstimate);
-
-  unPackAddresses();
-
-  var min_send_amt_wei = 6000000;
+function isSendTipPage() {
+  return (/\/tip\/send/).test(window.location.pathname);
+}
 
+function prePopulateFounderFields() {
```
arun1595: I am not happy with the function name I came up with. Feel free to suggest a better alternative.

app/assets/yge/youvegoteth/send.js
```diff
@@ -43,40 +44,54 @@ var updateEstimate = function(e) {
...
   }
   if (localStorage['fromEmail']) {
     $('fromEmail').value = localStorage['fromEmail'];
   }
+  if (localStorage['expires']) {
+    $('expires').selectedIndex = localStorage['expires'];
+  }
+}
+
+function prePopulateGitcoinerFields() {
```
arun1595: Ditto. Feel free to suggest a better name.

app/assets/yge/youvegoteth/send.js
arun1595: lol my bad :see_no_evil: 

Fixed.

app/assets/yge/youvegoteth/send.js
```diff
@@ -43,40 +44,54 @@ var updateEstimate = function(e) {
...
+window.onload = function() {
+  jQuery('#amount').on('keyup blur change', updateEstimate);
+  jQuery('#token').on('change', updateEstimate);
+
+  unPackAddresses();
+
+  var min_send_amt_wei = 6000000;
+
+  prePopulateFunderFields();
+  if (!isSendTipPage) {
```
thelostone-mc: @arun1595 `isSentTipPage()` ?

arun1595:
@thelostone-mc I've fixed the merge conflicts, PTAL :slightly_smiling_face:

app/assets/yge/youvegoteth/send.js
thelostone-mc: wait didn't you add `isSendTipPage` function [here](https://github.com/gitcoinco/web/pull/1318/files#diff-69887365dcfb83895cb00a1c9639f716R47) ? 🤔 
It looks like you wanted it to populate the gitcoinerField when the url isn't `/funding/new`

Also `isSentTipPage` is a function so wouldn't you need to invoke it rather than simply using it ?

arun1595: We don't want to populate gitcoiner fields if the current page is `/tip/send` right?

Derp, didn't understand what you meant in your other comment, fixed.

arun1595:
@mbeacom @thelostone-mc I've resolved the conflict. PTAL :slightly_smiling_face:

PR #1325, opened 2018-05-31
toolbox: avoid $.get when url is undefined

thelostone-mc:
###### Issue 
x1

##### Checklist

- js

### DIFF
--- a/app/assets/v2/js/toolbox.js
+++ b/app/assets/v2/js/toolbox.js
@@ -7,7 +7,8 @@ setInterval(function() {
   if (document.preloads.length) {
     var url = document.preloads.pop();
 
-    $.get(url);
+    if (url)
+      $.get(url);
   }
 }, interval);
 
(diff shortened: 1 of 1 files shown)


thelostone-mc:
Fixes the failed get calls but still do we need to invoke it in a setInterval ? 

```
  $('.cards .img img').each(function(index, element) {
    document.preloads.push($(element).data('hover'));

  });
```

@owocki  Can't we just directly do `$.get` here instead ? 
Haven't changed this cause I wanna check why this was written this way (so that I don't break anything unknowingly )

mbeacom:
@kziemianek completed the original bounty for this.  Any thoughts?

SaptakS:
@thelostone-mc @mbeacom I am removing the `$.get` code for the time. Hope you both are okay with it?

mbeacom:
Sounds good!

mbeacom:
screenshot 2018-05-31 16 39 31
Beautiful.

PR #1332, opened 2018-06-01
shared: refactored + removed dead code

thelostone-mc:
##### Description

- removed filter code as the filter options are no longer
  shown on bounty_detail page
- removed attach_close_button in shared as it's not being
  used anywhere
- removed sidebar_redirect_triggers in shared as the sidebar
  filter are using only in the explorer
- moved custom input type function before document entirely
  loads to ensure fields appear on devices with slow internet
  connection

##### Checklist

- js 

##### Testing

Tested it myself but would like to be tested once more
@SaptakS @mbeacom 

@vs77bb this ensure that the input fields appear along with the other firlds


mbeacom:
Deployed to stage

PR #1343, opened 2018-06-02
WIP: add 'stop work & slash' button

### LINKED ISSUE #1340: Admin: Stop Work + Stop Work & Slash
### User Story

[comment]: # (As an admin/funder, I want to differentiate between 'Stopping Work' on someone and 'Stop Work + Slash' to only block participants who are maliciously spamming the network.)

### Why Is this Needed?
Currently, slashing conditions are in place for all Gitcoiner's where an admin clicks 'Stop Work'. 

*Type*: Feature

### Solution
[comment]: # (Provide a summary of the solution and a task list on what needs to be fixed.)
*Summary*: Both a 'Stop Work'  and a 'Stop Work + Slash' button to differentiate between punishment / no punishment. 

screen shot 2018-06-01 at 3 17 48 pm
screen shot 2018-06-01 at 3 17 48 pm

Punishment -- if a user has been slashed, then they cannot start work without asking admins to remove their slash.  this is partially implemented already => https://github.com/gitcoinco/web/blob/master/app/dashboard/views.py#L191-L195

### Reviewers
@mbeacom @owocki

jvmaia:
##### Description
implement the 'stop work & slash' feature
##### Refers/Fixes
closes #1340


jvmaia:
The code is losing ~0.07% because the if/else statement in dashboard/views.py is extended. I can write it in one line, but i don't know if it would be a good idea.

owocki:
reviewing now

jvmaia:
Not a big game changer, but using a int in url we can reduce the size of the if/else statement inside views.uninstered().

app/assets/v2/js/shared.js
```diff
@@ -220,13 +230,21 @@ var mutate_interest = function(bounty_pk, direction, data) {
 };
 
 
-var uninterested = function(bounty_pk, profileId) {
-  var request_url = '/actions/bounty/' + bounty_pk + '/interest/' + profileId + '/uninterested/';
+var uninterested = function(bounty_pk, profileId, slash = false) {
+  var slashed = '0';
+  var success_message = 'Contributor removed from bounty.';
+
+  if (slash) {
```
SaptakS: I think it's better to send `true` and `false` instead of 0 and 1. And then [here](https://github.com/gitcoinco/web/pull/1343/files#diff-3e886cd5b50df8f0cabc86d33ca734e6R352) you can write something like `if slash == 'true'` ? cc: @mbeacom

mbeacom: I agree with this.  Additionally, we could probably swap this out to simply use a POST param versus the URL defined variable, defaulting to `False` if the `request.POST` param isn't present.

jvmaia: Updating the code to use a POST param, as the @mbeacom said.

jvmaia: Do you think it's better send the POST params only when the `slash` is true?  Makes more sense to me always send it.

app/app/urls.py
```diff
@@ -84,7 +84,7 @@
     path('actions/bounty/<int:bounty_id>/interest/new/', dashboard.views.new_interest, name='express-interest'),
     path('actions/bounty/<int:bounty_id>/interest/remove/', dashboard.views.remove_interest, name='remove-interest'),
     path(
-        'actions/bounty/<int:bounty_id>/interest/<int:profile_id>/uninterested/',
+        'actions/bounty/<int:bounty_id>/interest/<int:profile_id>/uninterested/<int:slash>/',
```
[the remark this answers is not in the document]
jvmaia: Yep, code updated!

jvmaia:
Code updated! =)

owocki:
> @owocki Are we married to "slash" here? Thoughts on alter_rep / ding_rep or something like that? After discussing it with @SaptakS, slash seems fairly misleading, since we use slash in multiple places to represent the presence of / and might prove difficult to maintain or approach by community members that are unfamiliar with the terminology.

> Note: I understand the hunter wrote this to the spec, so I wouldn't expect him to modify this. Just something we might want to consider before we push it live.

I just used slash bc its what Vlad Zamfir uses as a verb when he talks about Proof of Stake.  I'm not married to it; and I don't feel strongly in either direction (as this is an admin only functionality)

PR #1356, opened 2018-06-04
Funder form v3

### LINKED ISSUE #1224: Detailed Funder Flow with multiple issue support
### Description

As a funder, I would like a funding experience where I can start editing funding information of multiple issues simultaneously and save them in draft form so as to work again later. Also, a well detailed, well separated form with proper UX helps a lot.
*Type*: feature.

### Definition of Done
- Designs and rearrangement of the forms and fields are done.
- User has option to add new issue while editing current issue. (#1323)
- User has the ability to submit one issue and leave other issues in draft state. (#1323)
- User can switch between issues and edit. (#1323)
- User can add deadlines
- Bigger description box
- Add project types and permissions (#1098 )
- User can enter bounty in USD to get the corresponding ETH or other token.
- User can select which gas speed from the recommended list or customize.

### Reviewers
@PixelantDesign @owocki @mbeacom @thelostone-mc

### Assets
1 copy 10

SaptakS:
##### Description

Improves the Funder Form
- [x] Rearranging the fields into different sections
- [x] Increase size of description texts
- [x] Increase the overall width of the form
- [x] Allow USD to ETH conversion for better UX
- [x] Provide basic recommended Gas Settings which users can use directly off their plate.
##### Checklist

##### Refers/Fixes

Fixes #1224


SaptakS:
I know this has all the commits from the quickstart screen PR. I will rebase once that gets merged to avoid the redundant commits. But most of the commits were needed for this PR, so built on top of it.

thelostone-mc:
@SaptakS good for review?

app/dashboard/utils.py
```diff
@@ -404,7 +404,7 @@ def record_user_action_on_interest(interest, event_name, last_heard_from_user_da
 
 
 def get_context(ref_object=None, github_username='', user=None, confirm_time_minutes_target=4,
-                confirm_time_slow=10, confirm_time_avg=4, confirm_time_fast=1, active='',
+                confirm_time_slow=90, confirm_time_avg=30, confirm_time_fast=1, active='',
```
thelostone-mc: fancy

thelostone-mc:
When I'm not logged in -> I get this alert text
screen shot 2018-06-18 at 2 08 16 pm

Need to change the font of the Login to `12px` to match the rest of the form
screen shot 2018-06-18 at 2 10 33 pm

thelostone-mc:
@PixelantDesign  / @SaptakS  consider making the background totally white ?

screencapture-0-0-0-0-8000-new-2018-06-18-14_12_58

SaptakS:
@thelostone-mc the `unknown` minute won't be occuring in live/staging. You need to sync your gas_conf_time to get the minutes, every 30 minutes. But anyways I have added space between `unknown` and `minutes`

thelostone-mc:
All the queries have been addressed 👍 
Here's a 🍪 for the quick fix post review  @SaptakS  


@PixelantDesign could you answer the questions I've left for you -> Post that this should be good for merge

owocki:
form shows an error when `get_prices` has not been run [screenshot](http://bits.owocki.com/2X0i1w2b3a38/Screen%20Shot%202018-06-18%20at%2012.35.30%20PM.png)

note: sometimes eth gas station (where we get the gas info from) goes down.. so we will need to build around this scenario and handle it gracefully

owocki:
from slack:

I think the primary issue now is that the average speed/price is the same as the slow

i wonder if we might not want to change the way we’re providing the estimates

so basically right now the network is humming along.  you can do a FAST and CHEAP transaction at 1 gwei
https://ethgasstation.info/predictionTable.php
ethgasstation.info
ETH Gas Station
ETH gasprice recommendations
maybe if the network is under that scneario we just show a little module that says theres no tradeoff between fast/cheap
but if the network slows down THEN we show the module that we’ve already designed

SaptakS:
> sometimes eth gas station (where we get the gas info from) goes down.. so we will need to build around this scenario and handle it gracefully

@owocki What would be a graceful way of handling this? Should I show something like . "No Information Found"? I don't think UI wise that would look good. @PixelantDesign @thelostone-mc what do you suggest?

PixelantDesign:
@thelostone-mc the background is white.
We talked about removing email sign up in yesterday am ftux & avatar builder zoom sync up.

Will think on showing no tradeoff

PixelantDesign:
Sharing so everyone is up to speed, @SaptakS and I talked about the following for the two outstanding scenarios. Let me know if I'm understanding these use cases right.

When all recommendations are fast and there are no tradeoffs. We default to one option and highlight it for the user with accompanying message above the selection.
screen shot 2018-06-19 at 9 22 58 pm

When the gas is unknown, we display a message and recommend the user input a gas value in the customize gas box. 

screen shot 2018-06-19 at 9 39 30 pm
OR - should we default to 5 Gwei as a recommendation and show a message of - We aren't sure of the speed and price, we recommend setting it high at 5 Gwei" instead of directing users down to the custom entry box?? Thoughts?

PR #1357, opened 2018-06-04
HOLD: Remove django mail logger

mbeacom:
##### Description
The goal of this PR is to remove the Django `mail_logger` and replace it with direct CloudWatch logging for persistent logs that we can setup alarms on.

##### Checklist

logging

##### Testing
Locally


app/app/settings.py
```diff
@@ -189,37 +190,63 @@
...
+        aws_access_key_id=AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+        region_name=AWS_DEFAULT_REGION
+    )
+
     LOGGING = {
         'version': 1,
         'disable_existing_loggers': False,
+        'root': {
+            'level': 'ERROR',
```
SaptakS: So we are not logging the warnings? 😅

mbeacom: This is the root/default logging behavior and aligns with the default logger provided by Django as far as I know. 🤔

mbeacom: Watchtower is set to log whatever the level is that's passed to: `AWS_LOG_LEVEL` in this case: `DEBUG` - So, much more verbose than what we're even logging currently. @SaptakS

SaptakS: Okay. Makes sense.

PR #1381, opened 2018-06-05
Clean hashbang from gh urls

mbeacom:
##### Description
The goal of this PR is to introduce extremely basic cleaning of the bounty vc/github URL on bounty save and in a few other places we are performing general checks on the associated URL.

For example, the bounty associated with: https://github.com/balance-io/balance-manager/issues/187 was created with a `#issuecomment` portion included and ultimately breaks our url lookup.

##### Checklist

bounties

##### Testing
Local

### DIFF
--- a/app/assets/v2/js/pages/new_bounty.js
+++ b/app/assets/v2/js/pages/new_bounty.js
@@ -137,7 +137,7 @@ $(document).ready(function() {
         _alert(gettext('You are on an unsupported network.  Please change your network to a supported network.'));
         return;
       }
-      
+
       var data = {};
       var disabled = $(form)
         .find(':input:disabled')
@@ -153,7 +153,7 @@ $(document).ready(function() {
       // setup
       loading_button($('.js-submit'));
       var githubUsername = data.githubUsername;
-      var issueURL = data.issueURL;
+      var issueURL = data.issueURL.replace(/#.*$/, '');
       var notificationEmail = data.notificationEmail;
       var amount = data.amount;
       var tokenAddress = data.deonomination;
@@ -352,7 +352,7 @@ $(document).ready(function() {
 
         // bounty is a web3.js eth.contract address
         // The Ethereum network requires using ether to do stuff on it
-        // issueAndActivateBounty is a method definied in the StandardBounties solidity contract.
+        // issueAndActivateBounty is a method defined in the StandardBounties solidity contract.
 
         var eth_amount = isETH ? amount : 0;
         var _paysTokens = !isETH;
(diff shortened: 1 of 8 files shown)


owocki:
the backend code looks good to me.  but did you test creation of a new rinkeby bounty under this condition perchance? IIRC the funding form takes you to http://localhost:8000/issue/details?url=[URL]  and it might not find the correct bounty URL if the frontend is not updated too..

SaptakS:
I agree with @owocki. It needs to be modified in JS as well else the URL still has the hash bang

app/dashboard/utils.py
```diff
@@ -421,3 +421,25 @@ def get_context(ref_object=None, github_username='', user=None, confirm_time_min
...
+
+    TODO:
+        * Deprecate this in favor of Django forms.
+
+    Returns:
+        str: The cleaned Bounty URL.
+
+    """
+    try:
+        return url.split('#')[0]
```
thelostone-mc: would be an overkill to use `regex` right ? 
The `[0]` always makes me uncomfortable 😅

mbeacom: 😂Understandable. `[0]` will always be present on a `split()` since it returns a list, regardless of whether or not the characters split on are present in the string. @thelostone-mc

mbeacom:
@SaptakS @owocki I understand what you're both saying here.  I'll change it in JS, but we're checking against the stripped down version of whatever the JS is passing.

Regardless of what is passed by the FE to the BE, we're comparing cleaned URLs versus the absolute url provided.

app/assets/v2/js/pages/new_bounty.js
```diff
@@ -352,7 +352,7 @@ $(document).ready(function() {
 
         // bounty is a web3.js eth.contract address
         // The Ethereum network requires using ether to do stuff on it
-        // issueAndActivateBounty is a method definied in the StandardBounties solidity contract.
+        // issueAndActivateBounty is a method defined in the StandardBounties solidity contract.
```
thelostone-mc: 👁

app/dashboard/tests/test_dashboard_utils.py
```diff
@@ -50,3 +50,10 @@ def test_get_ordinal_repr():
         assert get_ordinal_repr(22) == '22nd'
         assert get_ordinal_repr(23) == '23rd'
         assert get_ordinal_repr(24) == '24th'
+
+    @staticmethod
+    def test_clean_bounty_url():
+        """Test the cleaning of a bounty-esque URL of # sections."""
+        assert clean_bounty_url(
```
SaptakS: 👍 for the unittest!!

PR #1412, opened 2018-06-08
landing page: final changes

thelostone-mc:
##### Description
- updated navbar
- swap fund & login button
- custom newsletter

##### Checklist

- landing page

### DIFF
--- a/app/retail/views.py
+++ b/app/retail/views.py
@@ -54,6 +54,9 @@ def index(request):
         'slides': slides,
         'slideDurationInMs': 6000,
         'active': 'home',
+        'hide_newsletter_caption': True,
+        'hide_newsletter_consent': True,
+        'newsletter_headline': _("Get the Latest Gitcoin News! Join Our Newsletter."),
     }
     return TemplateResponse(request, 'index.html', context)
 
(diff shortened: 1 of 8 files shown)


thelostone-mc:
Nav changes

x1


Also @mbecom / @Saptak -> The news letter content doesn't seem to get updated based on the params I'm passing from `views.py`
Am I missing anything ?

mbeacom:
This is on staging.

app/retail/views.py
```diff
@@ -54,6 +54,9 @@ def index(request):
         'slides': slides,
         'slideDurationInMs': 6000,
         'active': 'home',
+        'hide_newsletter_caption': True,
```
mbeacom: What's the utility of this bool? Are we altering this value at any point? 🤔

mbeacom:
screenshot 2018-06-08 15 04 39

app/retail/views.py
SaptakS: @mbeacom I think what @thelostone-mc has done here is he is sending this boolean only to landing-page, which makes it non-hidden in all the other pages. Hence both the caption and consent are shown in every other page except the landing page.

PR #1416, opened 2018-06-09
Remove Get started with Gitcoin link

arun1595:
##### Description

image

Currently, the 'Get Started with Gitcoin' link (shown in the above screenshot) links to the new bounty page which doesn't make sense from the perspective of a developer who's interested in working on bounties.

Not sure what to replace it with as we're already linking to `/explorer` from 'View Available Bounties' link.

Hence, this PR removes the  'Get Started with Gitcoin' link. Let me know if I should retain it and link to some other page.

##### Checklist

### DIFF
--- a/app/retail/templates/index.html
+++ b/app/retail/templates/index.html
@@ -35,8 +35,6 @@
           <span class="text-highlight-gc-purple">{% trans "Want to work on OSS Projects?" %}</span>
           <div class="ml-3 text-highlight-gc-purple">
             <a class="text-highlight-gc-purple" href="{% url "explorer" %}">{% trans "View Available Bounties" %}</a>
-            |
-            <a class="text-highlight-gc-purple" href="{% url "new_bounty" %}">{% trans "Get Started with Gitcoin" %}</a>
           </div>
         </div>
       </div>


thelostone-mc:
This is the funder landing page ! 
The 1st step they'd take would be to create a bounty as opposed to visiting the explorer 


Once we have the FTUX rolled out => we'd be linking it to that

arun1595:
@thelostone-mc 

> This is the funder landing page !

Yeah, it is. But the link in question is followed by "Developers! Want to work on OSS projects?" which makes it obvious we have a small screen space for developers (contributors) on the funder landing page. That's why there's a 'View available bounties' link I assume. The 'Get Started with Gitcoin' link for developers should link to the explorer (IMO) but we do have 'View available bounties' linking to the same which makes it redundant. Am I making sense?

Do let me know if I'm missing something.

thelostone-mc:
Oh derp derp! Didn't read that 
As per @PixelantDesign design, the developer would  end up navigating to the  contributor landing page ( the one you are working on )  


@PixelantDesign until that comes can we remove this or link it to something else?

PR #1444, opened 2018-06-12
no flex on alert messages

owocki:
it seems to me that the alert messages `.alert .message` should not be `display:flex`

see this gif with a before and after for why:

screen recording 2018-06-12 at 07 26 am

### DIFF
--- a/app/assets/v2/css/base.css
+++ b/app/assets/v2/css/base.css
@@ -583,7 +583,6 @@ input[type=text].loading {
 
 .alert .message {
   align-items: center;
-  display: flex;
 }
 
 .alert .message::before {
--- a/app/assets/v2/css/base.css
+++ b/app/assets/v2/css/base.css
@@ -583,6 +583,10 @@ input[type=text].loading {
 
 .alert .message {
   align-items: center;
+  display: flex;
+}
+.alert .message .content{
+  display: inline;
 }
 
 .alert .message::before {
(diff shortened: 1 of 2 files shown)


app/assets/v2/css/base.css
```diff
@@ -583,7 +583,6 @@ input[type=text].loading {
 
 .alert .message {
   align-items: center;
-  display: flex;
 }
```
SaptakS: Instead of this I suggest changing the `_alert` function in shared.js to something like:

```javascript
var _alert = function(msg, _class) {
  if (typeof msg == 'string') {
    msg = {
      'message': msg
    };
  }
  var numAlertsAlready = $('.alert:visible').length;
  var top = numAlertsAlready * 66;

  var html = function() {
    return (
      `<div class="alert ${_class}" style="top: ${top}px">
        <div class="message">
          <div class="content">
            ${alertMessage(msg)}
          </div>
        </div>
        ${closeButton(msg)}
      </div>;`
    );
  };

  $('body').append(html);
};
```

The extra `<div class="content">` ensures that the content remains unharmed while also maintaining the flex property between the icon and the content.

cc: @thelostone-mc @mbeacom

owocki: d5b87be617af5a82e87ecd90e33c0996cdb55c3f

PR #1447, opened 2018-06-13
save recent searches in tabs above search bar -- like http://www.hipmunk.com/ does

owocki:
##### Description

save recent searches in tabs above search bar -- like http://www.hipmunk.com/ does
screen recording 2018-06-12 at 11 18 pm

makes for easy returning to recent searches

##### Checklist

search

##### Testing

tested it 
##### Refers/Fixes

Self, hacking for the night


owocki:
three things after sleeping on this:
1. what do we do about this, relative to the legacy 'save search' which goes to email?
2. when do we want to allow the user to start a new tab up top.  its probably excessive to start one every time a new search is performed.   maybe we (a) only start one if the new search is diff than the previous search (b) or when the user explicitly clicks a "new tab" button?
3. the tabs in the search UI could use some visual love.

owocki:
@PixelantDesign how do you feel about the aesthetics of https://user-images.githubusercontent.com/513929/41331624-26dfc120-6e97-11e8-8bad-58f873751a1c.gif ?

owocki:
> what do we do about this, relative to the legacy 'save search' which goes to email?

i vote we nuke the legacy 'save search'.  only 232 peopole have ever used it.  

> when do we want to allow the user to start a new tab up top. its probably excessive to start one every time a new search is performed. maybe we (a) only start one if the new search is diff than the previous search (b) or when the user explicitly clicks a "new tab" button?

i think i'm going to go with (a) unless anyone stops me

thelostone-mc:
I feel we don't need the search feature all together

Reason : 
- We save the latest search criteria in localStorage & users end up visiting the explorer via the same browser on the same machine
- This is nice -> but clutters up the space ,I for one would not use it 

If the team feels it's needed -> then we can roll with this

@PixelantDesign @mbeacom @SaptakS

owocki:
@thelostone-mc i think that where this extends the existing functionality is that it allows you to easily toggle between several different searches you might be doing.  

so for me, as a funder on gitcoin, i want to search by

1. issues funded by me
2. issues funded by others on gitcoin
3. work that is open (as an admin, this is useful for me to see why / whats not getting picked up)
4. work that is started (is it stalled?)

and being able to easily toggle back and forth would be huge for me

thelostone-mc:
Hmm you make sure valid point ! Alright let's roll with it

thelostone-mc:
@owocki updated the UI a bit

Any suggestions on improvs ? 

screen shot 2018-06-17 at 12 07 10 am

owocki:
i really like the new UI!  

@mbeacom maybe we can deploy this to stage so that @PixelantDesign can play with it and tell us what she thinks?


here is the direction i think we should take this https://github.com/gitcoinco/web/pull/1447#issuecomment-397099293

owocki:
e68bc472ef5e1c9f2e11caddf28eb7500208c650 accomplishes https://github.com/gitcoinco/web/pull/1447#issuecomment-397099293

thelostone-mc:
Closing this out for now cause we've removed bits and pieces and resolving those conflicts will be a nightmare
We'll get a fresh PR out with the relevant changes later

owocki:
are we still interested in this featureset?  personally i think it'd be useful, enough so that i'd be willing to resolve the merge conflicts and make it happen.

owocki:
df122bc3f25b18503eaf10542b8858f5d9975213 brings this up to date with current codebase 

though to your point @thelostone-mc we could maybe refactor this to use the URL bar instead of storing a bunch of crazy info in localstorage

mbeacom:
Ship this Monday unless there are strong objections? @owocki @thelostone-mc ?

owocki:
OMG YES pls .... i haven't tested in a while, but assuming it works, i'm very very excited that this will be available in our platform!

Issue #1341, opened 2018-06-01
As a user, i'd like faucet distributions to scale up and down with the eth network backlog so i can always have enough gas to submit a standard bounties tx


owocki:
# user story

As a user, i'd like faucet distributions to scale up and down with the eth network backlog so i can always have enough gas to submit a standard bounties tx

# what

spurred by feedback from a user http://bits.owocki.com/1m2f3p2q2B2f/Screen%20Shot%202018-06-01%20at%202.30.07%20PM.png

we've given out faucet distributions for 0.001(ish) for a bit.. it doesnt make sense to do the same faucet distro when the recommended gas price is 1 gwei as it does 10 gwei.

# bounty 

* we should send enough gas for the user to submit a 318730 gas tx that clears within 5 minutes
* Put another weay, each faucet should scale up and down according to the response of `recommend_min_gas_price_to_confirm_in_time(5)`
* the FaucetRequest.amount should record the amoutn sent

owocki:
feedback from a user http://bits.owocki.com/1m2f3p2q2B2f/Screen%20Shot%202018-06-01%20at%202.30.07%20PM.png

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.1 ETH (56.61 USD @ $566.12/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/1341/551).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $10,695.77 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.

 1. [@shad-k](https://gitcoin.co/profile/shad-k) 

 has committed to working on this project to be completed 4 weeks from now.


@owocki, __please see the below comments / questions regarding approach for this ticket from the bounty hunter(s):__ 
- [@shad-k](https://gitcoin.co/profile/shad-k): In faucets folder I will add the recommended gas to the faucet amount in the views file. Does the faucet amount need to scale according to a particular formula?

gitcoinbot:
[...]

PR #1466, opened 2018-06-15
Add min gas price to faucet to confirm in 5 minutes

shad-k:
Fixes: https://github.com/gitcoinco/web/issues/1341


owocki:
this doesnt actually update the amount that is sent to the user.. only updates the database object that is created after the fact.  

you'll need to update line 107 @shad-k 

also, `recommend_min_gas_price_to_confirm_in_time()` returns a value in gwei, whereas `faucet_amount` is in ETH.

PR #1476, opened 2018-06-15
Base implementation of django-pipeline

mbeacom:
##### Description
The goal of this PR is to implement `django-pipeline` to provide ES6 support and asset compression/uglify without retooling/reworking the entire frontend.

##### Checklist

###### TODO

- [x] Implement base package setup
- [x] Implement ES6 handler
- [x] Implement in head css via pipeline
- [x] Implement in footer scripts js via pipeline

fe, be, static, assets

##### Testing
- [ ] Local
- [ ] Stage


requirements/test.txt
```diff
@@ -1,5 +1,5 @@
 -r base.txt
-pytest==3.5.0
```
owocki: dont we want to pin the versions here?

Dockerfile
```diff
@@ -3,12 +3,20 @@ ENV PYTHONUNBUFFERED 1
...
     pip install --upgrade pip wheel setuptools && \
     pip3 install dumb-init && \
     rm -rf /var/lib/apt/lists/*
 
+RUN apt-get update && apt-get install curl -y && \
+    curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash - && \
+    apt-get install -y nodejs && \
+    npm install -g yuglify uglify-js && \
+    npm install --save-dev babel-cli babel-preset-env && \
+    apt-get remove curl -y && \
```
thelostone-mc: Benefits of removing this ? (Just saving space ? )

app/app/settings.py
```diff
@@ -238,12 +240,142 @@
...
+)
 STATIC_ROOT = root('static')
 
 STATIC_HOST = env('STATIC_HOST', default='')
 STATIC_URL = STATIC_HOST + env('STATIC_URL', default='/static/')
 
+PIPELINE = {
+    'PIPELINE_ENABLED': True,
+    'PIPELINE_COLLECTOR_ENABLED': True,
+    'COMPILERS': ('pipeline.compilers.es6.ES6Compiler', ),
```
thelostone-mc: Would the compiler run against all js files or only the ones mentioned in the Pipeline ?
Should we export this to a new file (cause this is gonna keep growing in size)

thelostone-mc:
@mbeacom 

-   Need to update this  `'PIPELINE_ENABLED': env('PIPELINE_ENABLED', default='True')` to ensure easier debugging (for the [doc](http://django-pipeline.readthedocs.io/en/latest/usage.html#templatetags))

- How do we get it to rebuild each time a file is changed ?  (Would it be [cache-bursting](http://django-pipeline.readthedocs.io/en/latest/usage.html#cache-busting) ) ?

Dockerfile
mbeacom: It's just to make the image smaller since we only using it during the build stage.

app/app/settings.py
mbeacom: 🙌I was wondering if I should split it off... Yeah, we can move this into it's own file and import it in settings.py 👍

app/app/static_storage.py
```diff
@@ -1,13 +1,18 @@
...
+from os.path import basename
+from secrets import token_hex
 
 from django.conf import settings
 from django.contrib.staticfiles.storage import ManifestFilesMixin
 
-class SilentFileStorage(PipelineMixin, CompressedManifestStaticFilesStorage):
-    """Define the static storage using whitenoise with hashing.
+from pipeline.storage import PipelineMixin
+from storages.backends.s3boto3 import S3Boto3Storage, SpooledTemporaryFile
```
thelostone-mc: you did the exact same thing I did -.-

mbeacom:
Looks like the only thing we need to fix (from what I'm seeing at least):

screenshot 2018-09-25 17 41 45

PR #1477, opened 2018-06-15
Added the gas amount to the right place

shad-k:
Please recommend if there is a better way to convert gwei to ETH.

Fixes: https://github.com/gitcoinco/web/issues/1341

### DIFF
--- a/app/faucet/views.py
+++ b/app/faucet/views.py
@@ -103,7 +103,8 @@ def process_faucet_request(request, pk):
     except FaucetRequest.DoesNotExist:
         raise Http404
 
-    faucet_amount = settings.FAUCET_AMOUNT
+    """Converting gwei to eth by dividing the gwei amount by 1000000000"""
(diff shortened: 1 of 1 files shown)


app/faucet/views.py
```diff
@@ -104,7 +103,8 @@ def process_faucet_request(request, pk):
     except FaucetRequest.DoesNotExist:
         raise Http404
 
-    faucet_amount = settings.FAUCET_AMOUNT
+    """Converting gwei to eth by dividing the gwei amount by 1000000000"""
+    faucet_amount = settings.FAUCET_AMOUNT + (recommend_min_gas_price_to_confirm_in_time(5) / 1000000000)
```
owocki: upon further review / experimentation in my console, i think that this should be 

> faucet_amount = settings.FAUCET_AMOUNT * recommend_min_gas_price_to_confirm_in_time(5)

since the base faucet amount assumes 1 gwei and the total gas the user will need to use will be `gasPrice` * `gasLimit`

shad-k:
Can you check now?

owocki:
thanks!

PR #1482, opened 2018-06-15
Base livestream sessions tracking and redirect

mbeacom:
##### Description
The goal of this PR is to implement livestream session user tracking and a mechanism of gating access to the livestream to gitcoin authenticated users only.

##### Checklist


owocki:
i love this idea!

1. any way we can tie the zoom account to github username?  in the case of trolls, it'd be
2. how do we updathe zoom id on a weekly basis?  it'll be important to get that info to @vs77bb 
3. should we manage the livestream agenda via this system?

mbeacom:
@owocki

1. This pr introduces tracking of attendees by their Gitcoin profile. I'm tracking them on the attendees m2m and creating a useraction.
2. I was thinking we could add a recurring job, slack command, and/or some admin functionality to schedule unique livestreamsession entries or tie into the zoom api to schedule these and retrieve the relevant zoom ID and password.
3. We could add the agenda as the description and maybe create something like a list view to display all historical and upcoming livestreams. I figured we can upload the recorded stream and have those available for past streams.

owocki:
1. ya i just wonder what vivek and i would do if someone is a troll -- i guess we'd have to go to their admin page and slash them
2 and 3 make sense to me

thelostone-mc:
Do we need this level of tracking ? it is a nice thing to have

PR #1494, opened 2018-06-18
WIP - email: added ftux 

thelostone-mc:
##### Description

This PR has two commits -

- added email template which would be sent out to contributor upon completing ftux.  (needs review)
- refactored other email templates + updated css + re-indented (Boy scout rule)

screencapture-0-0-0-0-8000-_administration-email-suggested_bounties-2018-06-18-13_37_47

Tasks left: 

- [x] build template (txt + html)
- [ ] fix query  
- set trigger point (to be sent only for first time users)

##### Checklist

- email

### DIFF
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -221,6 +221,7 @@ urlpatterns = [
     url(r'^_administration/email/new_match$', retail.emails.new_match, name='new_match'),
     url(r'^_administration/email/quarterly_roundup$', retail.emails.quarterly_roundup, name='quarterly_roundup'),
     url(r'^_administration/email/new_work_submission$', retail.emails.new_work_submission, name='new_work_submission'),
+    path('_administration/email/suggested_bounties', retail.emails.suggested_bounties, name='suggested_bounties'),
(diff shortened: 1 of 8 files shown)


thelostone-mc:
@mbeacom / @owocki few questions 

Ideally this mail would be sent out once they finish the ftux and are redirected to the explorer ->
(We show the first three search results which need to be sent out )

### Where do i add the trigger point ? 
Is it when the user completes the last step in onboard, which means we'd trigger it then ?
(which means it falls under the onboard view)

OR
After we redirect it to the user the explorer where we three issues for them and they click `Got it`  ? 
(which means it falls under the dashboard view)

screenshot-2018-6-16 issue explorer

### Would this have to be a API fire ? 

### Should be sent out only when the users has logged in for the very first time 
(can we do that with the existing tables cols ) ?

thelostone-mc:
Closing as this does not seem to be needed at this point!
General CSS updates + html refactor to the email templates have been bundled in  a seperate commit & merged ^_^

PR #1495, opened 2018-06-18
Fix pytest version

### LINKED ISSUE #1492: travis fails with 'The command "pytest -p no:ethereum" exited with 1.'
The command "pytest -p no:ethereum" exited with 1.
example build

https://travis-ci.org/gitcoinco/web/builds/393249455?utm_source=github_status&utm_medium=notification

SaptakS:
##### Description

@owocki @leonprou this should fix the travis error in pytest.
@mbeacom review.

##### Checklist

##### Refers/Fixes

Fixes #1492 
Refs #1491

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ jobs:
         - codecov
       env:
         - PYTHONPATH="${TRAVIS_BUILD_DIR}/app/"
+        - DJANGO_SETTINGS_MODULE="app.settings"
     # Deploy to Docker Hub and initiate a rolling update for web.
     - stage: deploy-stage
       if: branch = master AND NOT type = pull_request
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,5 +1,5 @@
 -r base.txt
-pytest==3.5.0
+pytest>3.5.0
 pytest-cov
 pytest-isort
 pytest-django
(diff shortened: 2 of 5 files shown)


SaptakS:
@mbeacom I am not entirely sure how it was working but I think in latest pytest, they don't allow environment variables to be declared in the `setup.cfg`. So I modified the commands.

.travis.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         - codecov
       env:
         - PYTHONPATH="${TRAVIS_BUILD_DIR}/app/"
+        - DJANGO_SETTINGS_MODULE="app.settings"
```
mbeacom: Is this needed?

requirements/test.txt
```diff
@@ -1,5 +1,5 @@
 -r base.txt
-pytest==3.5.0
+pytest>3.5.0
```
mbeacom: Should we just pin this to the appropriate version or completely drop the version pinning for pytest?

SaptakS: I think dropping version pinning is better. Pinning is what caused a problem here. I would have however done `>=`. But that might not upgrade the package installed because the current 3.5.0 will anyways satisfy.

.travis.yml
SaptakS: Yes. Seems so. The environment variable declaration in setup.cfg under `tool:pytest` seems to not been working most probably in the newer versions. They need the environment variables to be set before the pytest can be run. Without this travis fails.

Issue #1457, opened 2018-06-14
z-index issue on gitcoin.co/ens


owocki:
<img width="1439" alt="screen shot 2018-06-13 at 8 01 01 pm" src="https://user-images.githubusercontent.com/513929/41387519-8b08aa6a-6f44-11e8-8a17-f79cae7de576.png">

premachb:
Hey! Just checking out this project and I have a fix for this. As well the page had a problem where the container would expand vertically because of the rain effect and I've fixed that too. I can include the fix for both in one PR or just make two separate PRs. Thoughts?

thelostone-mc:
@premachb Nah just send it one PR with screenshots and referencing this issue ^_^

PR #1556, opened 2018-06-26
Fix for #1457 (Updated zIndex for Cookie Banner)

premachb:
##### Description

Updated the zIndex for the cookie banner on the ENS page so that the login box doesn't go above it. 

Before: 

before

After:

after

This also includes a fix for the body vertically expanding due to the rain effect on the page. Modified body to use overflow rather than just overflow-x. As the yge and ens both use this main.css module, I made sure this didn't break both which it doesn't seem to.

I can revert this if there is a specific reason for using overflow-x here.

##### Checklist

ENS

##### Testing

Simple CSS change so mostly just by checking both pages that use the css module I modified. Let me know if I missed anything

##### Refers/Fixes

Refs: #1457

### DIFF
--- a/app/assets/yge/css/main.css
+++ b/app/assets/yge/css/main.css
@@ -22,7 +22,7 @@ body {
   line-height: 1;
   font-family: futura-pt, sans-serif !important;
   -webkit-text-size-adjust: none;
-  overflow-x: hidden;
+  overflow: hidden;
   width: 100%;
   height: 100%;
   color: #000000;
--- a/app/assets/cookielaw/css/cookielaw.css
+++ b/app/assets/cookielaw/css/cookielaw.css
@@ -4,7 +4,7 @@
   bottom: 0;
   background-color: #F2F6F9;
   width: 100%;
-  z-index: 20;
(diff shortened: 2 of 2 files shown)


app/assets/yge/css/main.css
```diff
@@ -22,7 +22,7 @@ body {
   line-height: 1;
   font-family: futura-pt, sans-serif !important;
   -webkit-text-size-adjust: none;
-  overflow-x: hidden;
+  overflow: hidden;
```
SaptakS: This will actually change the overflow in all the pages and prevent scrolling all other pages according to me. Can you please check?

premachb: ezgif-4-ccb4d55776

GIF showing scrolling on home page and some other pages with the changes. Let me know if there's any page in particular you're worried about.

SaptakS: @premachb well it happens in the url `/gas` but I guess we need to make design modifications in there to fit it without scroll because the unnecessary scroll because of the background animation also takes place.

Issue #1554, opened 2018-06-26
as a user, i want to be prompted to login when creating my avatar so i dont waste time building an avatar and can't save it


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a user, i want to be prompted to login when creating my avatar so i dont waste time building an avatar and can't save it

screenshot -- http://bits.owocki.com/3g3D2P3c0S2Y/Screen%20Shot%202018-06-26%20at%2012.12.44%20PM.png

### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
because it stinks to create an avatar and not be able to save it

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: 
ux issue

### Current Behavior
[comment]: # (Describe what actually happened.)
not prompted to login before creating avatar

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
prompted to login before creating avatar

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 15.0 DAI (15.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/1554/695).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $28,055.76 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 6 days, 22 hours from now. Please review their questions below:


1. [premachb](https://gitcoin.co/profile/premachb) has started work. 	
 * Q: Interested in working on this one. Getting more familiar with this codebase.

premachb:
@mbeacom Question about the UX flow. Should it just redirect to GitHub to login our should there be some sort of UI before the user gets redirected to GitHub?

owocki:
i think just a redirect is fine

gitcoinbot:
[...]

PR #1587, opened 2018-06-30
Ensure avatar auth

premachb:
##### Description

Ensure user is logged in before going through onboarding flow by redirecting to GitHub login if a user is not authenticated. 

##### Checklist

Dashboard folder more specifically the Onboarding Flow.

##### Testing

Simple change that checks and redirects and I've confirmed this doesn't break if the user is logged in.

##### Refers/Fixes

Refs: #1554


app/dashboard/views.py
```diff
@@ -522,6 +522,11 @@ def send_tip_2(request):
 
 def onboard(request, flow):
     """Handle displaying the first time user experience flow."""
+
+    if not request.user or not request.user.is_authenticated:
```
SaptakS: This is wrong. This will cause anyone who goes to the onboarding flow to be redirected to the github login. That is not expected. We want this only for the standalone avatar page. So this check should be only if someone is trying to access endpoints like: `/onboard/contributor/?steps=avatar` or `/onboard/profile` and so on.

mbeacom:
@premachb Thanks for your contribution!

cc @owocki @SaptakS 
I have adjusted this PR to only redirect to login when the user is hitting `onboard` outside of a flow containing the `github` step.

This will ensure we're redirecting to auth for any user passing `steps` without `github` or entering with the `profile` flow.

Also, I updated the if check by dropping `if not request.user` - `request.user` will always be present, but it might represent an unauthenticated user.  In cases of auth, we want to check `request.user.is_authenticated` and `hasattr(request.user, 'profile')` to verify the user's auth flow has successfully completed and they have been assigned a `profile` object.

Thanks again for the contribution!

SaptakS:
@mbeacom makes sense.

PR #1598, opened 2018-07-02
moves tokens from being hardcoded in the codebase to being stored in the db

owocki:
##### Description

moves tokens from being hardcoded in the codebase to being stored in the db

screen shot 2018-07-02 at 2 46 54 pm
screen shot 2018-07-02 at 2 46 43 pm

##### Checklist

tokenization

##### Testing

tested end to end

##### Refers/Fixes

self, retreat convo


owocki:
this should make submitting new tokens to the system way easier! 

@ecorpnu @mbeacom @thelostone-mc @SaptakS

app/dashboard/tokens.py
```diff
@@ -15,185 +15,15 @@
...
-    {"addr": "0xfa6f7881E52fDF912c4a285D78a3141B089cE859", "name": "AVO", "decimals": 18},
-    {"addr": "0x58b6a8a3302369daec383334672404ee733ab239", "name": "LPT", "decimals": 18},
-    {"addr": "0x09617f6fd6cf8a71278ec86e23bbab29c04353a7", "name": "ULT", "decimals": 18},
-    {"addr": "0x4CEdA7906a5Ed2179785Cd3A40A69ee8bc99C466", "name": "AION", "decimals": 8},
-    {"addr": "0x4162178B78D6985480A308B2190EE5517460406D", "name": "CLN", "decimals": 18},
 
-]
+
+def get_tokens():
+    from economy.models import Token
```
SaptakS: Is this imported here to prevent infinite loop in import?

owocki: yes

owocki:
@SaptakS @mbeacom pls review

app/retail/views.py
```diff
@@ -840,3 +876,12 @@ def youtube(request):
...
     return redirect('https://www.youtube.com/watch?v=cZZMDOrIo2k')
+
+
+def tokens(request):
+    context = {}
+    networks = ['mainnet', 'ropsten', 'rinkeby', 'unknown', 'custom']
+    for network in networks:
+        key = f"{network}_tokens"
+        context[key] = Token.objects.filter(network=network, approved=True)
+    return TemplateResponse(request, 'tokens_js.txt', context, content_type='text/javascript')
```
SaptakS: @owocki Is there no other way, rather than rendering it into a txt file which we then make seem like a javascript? Seems a bit messy to me. 😅 Can't we just make an API request from a js file to set the tokens? This approach makes it hardcode the `<script>` src. Also, @mbeacom @thelostone-mc do you think this might cause problem if we use transpiling? I feel it might give some hard time.
@owocki are you planning to like this SHIP this and then reiterate? I am cool in that case.

thelostone-mc: I agree ! API call + setting it vai JS seems a lot cleaner 
Else transpiling might give us issues later

owocki: > Is there no other way, rather than rendering it into a txt file which we then make seem like a javascript? Seems a bit messy to me.

all javascript files are already text files..

> Can't we just make an API request from a js file to set the tokens?

if we make an API request on every page request that needs tokens, then 
1. it'll slow down the functionality of the page because we'll have to do another request to the API, (which has different, less liberal, caching settings than the cloudfront `static/` dir)
2. add a bunch of callback functionality for the app to only perform certain functions when the tokens (or other dynamic data) are loaded.

perhaps #2 isnt a big deal for this use case (tokens are only really needed to be listed on /new and /tip ) but in a way, this PR def sets a precedent for how other dynamic data which is stored in the DB will be transferred to the frontend.  (i know that in https://github.com/gitcoinco/web/pull/1631 we might want to take that information and move it to the frontend)

> are you planning to like this SHIP this and then reiterate? I am cool in that case

i'll only ship it when you are comfortable.  i'm willing to go in your direction if i'm outvoted.  not a big deal.

owocki: @mbeacom wat do u think

owocki: upon thinking about it further, i agree, API route is probably the better route to go

mbeacom: I don't think the django template is all that bad.  Transpiling shouldn't give us too many issues since we're not handling the django template in that manner.  In any case, I'm inclined to agree that the API route could prove easier to maintain.

owocki: after doing some initial research on migrating this to apis, i'm realizing it's more complicated than i originally thought... in particular, because the scope isn't just contained to the new bounty and new tip pages.  theres a bunch more places we're going to have to either update the API responses or add a bunch of stringed callbacks in the UI in order to get the token details.

```

/Users/kevinowocki/Sites/gitcoin/web/app/assets/v2/js/tokens.js:
    1  
    2  var tokenAddressToDetails = function(addr) {
    3:   return tokenAddressToDetailsByNetwork(addr, document.web3network);
    4  };
    5  
    6: var tokenAddressToDetailsByNetwork = function(addr, network) {
    7    var _tokens = tokens(network);
    8  

/Users/kevinowocki/Sites/gitcoin/web/app/assets/v2/js/pages/bounty_details.js:
  511    // setup
  512    var decimals = 18;
  513:   var related_token_details = tokenAddressToDetailsByNetwork(result['token_address'], result['network']);
  514  
  515    if (related_token_details && related_token_details.decimals) {

/Users/kevinowocki/Sites/gitcoin/web/app/assets/v2/js/pages/dashboard.js:
  437        // setup
  438        var result = results[i];
  439:       var related_token_details = tokenAddressToDetailsByNetwork(result['token_address'], result['network']);
  440        var decimals = 18;
  441  

```


i'm giong to merge this and add a follow up ticket for moving it over to the API

owocki: NVM; i thought someone had approved the PR, but i guess not.

in any case, here is the followup ticket. ( https://github.com/gitcoinco/web/issues/1652 )   let me know if you guys want to discuss the strategy for making the whole app token-aware (considering the new constraint of having to add async API request and callbacks on the above pages), or if we should merge this as-is.

owocki:
[last call](https://gitcoincore.slack.com/archives/CAXFMVDGD/p1531160730000082)

owocki:
merged; can hit the refactor next

PR #1601, opened 2018-07-03
email: reindent + update css

thelostone-mc:
##### Description

- redient  + update css


thelostone-mc:
screenshot_2018-07-03 screenshot

PR #1607, opened 2018-07-03
WIP: new bounty email should only show bounties which match keywords

owocki:
##### Description

new bounty email should only show bounties which match keywords

this is based off  a complaint from @dogopupper that he was getting emails that werent relevant to him

previously we were searching the whole metadata package, which includes everything we know about the ipfs package

##### Checklist

emails, bounties

##### Refers/Fixes

email thread


owocki:
open question: what should the filtering algorithm be?  should we search the whole metadata package (as we were doing before) or should we search keywords OR title or DESC and include if one of those matches?

owocki:
@dogopupper what do you think

app/marketing/management/commands/new_bounties_email.py
```diff
@@ -29,17 +29,16 @@
...
-            idx_status__in=['open'],
-            )
-        for bounty in relevant_bounties.filter(web3_created__gt=(timezone.now() - timezone.timedelta(hours=hours_back))):
-                new_bounties_pks.append(bounty.pk)
-        for bounty in relevant_bounties:
-                all_bounties_pks.append(bounty.pk)
+    relevant_bounties = Bounty.objects.filter(
+        network='mainnet',
+        current_bounty=True,
+        idx_keywords__ioverlap=keywords, #TODO: do we search the title/desc too?
```
mbeacom: @owocki I think we need to shift from `ArrayField` keywords to taggit, honestly.  That would allow all us to simply relate keywords with `Profile`, `Organization`, and `Bounty` objects.  Title and description matching might get weird.  If we establish user expectation that `keywords` will indeed attract the appropriate hunters/funders, it shouldn't be an issue and would make landings, explorer filtering/sorting, and skill/bounty matching much simpler.

owocki:
need to test this before it goes live

app/marketing/management/commands/new_bounties_email.py
owocki: i agree with that direction for the long term; but whether we merge this in the short term depends upon how far out that is in the roadmap.  right now people are getting emails every day with bounties that seemingly don't match with their preferences -- so there is urgency in doing something soon

owocki:
closing in favor of the much simpler 86755127191cba104f097603cfb7095c57043f06

PR #1609, opened 2018-07-03
tree interactivity on landing page - experiment

owocki:
screen recording 2018-07-03 at 04 52 pm

### DIFF
--- a/app/assets/v2/css/landing_page.css
+++ b/app/assets/v2/css/landing_page.css
@@ -298,6 +298,7 @@ body {
 
 #gc-tree {
   text-align: center;
+  cursor: pointer;
 }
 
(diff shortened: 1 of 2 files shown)


SaptakS:
We can maybe change color of each leaf (or make it grow or make pulses) on hover. But to be honest, I feel right now it is placed at a segment, where most people might just scroll by it. Though the effort looks awesome.

owocki:
> But to be honest, I feel right now it is placed at a segment, where most people might just scroll by it. Though the effort looks awesome.

maybe they will see the cursor is clickable and will click on it? 

alternatively, we can do something like this as they scroll by.. i think that scroll interactivity is a nice way to show attention to detail to the user

PR #1631, opened 2018-07-05
programming languages stored in db

owocki:
##### Description

programming languages stored in db, so that anyone can add a new /results page

##### Checklist

db

##### Testing

tested it

##### Refers/Fixes

self


owocki:
yes we can hold off.. but if a `Keyword` model is the way we want to go.. then i'd express the concern that we don't seem to be at a point with the bounties -> profile FKs and profile -> user FKs that inspires a lot of confidence.  for example

1. one of my `owocki*` users doesnte have a profile
`. a handful of the legacy `Bounty` objects don't have `Profile` FKs to them

mbeacom:
I understand your comment, but I am confused as to how the FK user <-> profile relationship relates to my desired approach to this.

1. The `owocki*` profile issue is definitely an odd one that I'm tracking in a separate issue.  I have yet to be able to reproduce this on my end to date.  I combed through the `socialaccounts` repo and issues and haven't seen much to explain the bug either.  I am raising an issue on the relevant repo to request more information about the bug.
When this originally occurred (from a corrective action standpoint), I migrated all relevant data from `bountywill` and your additional user objects over to the correct user objects.

2. The reason some legacy profiles don't have user's is because they have never logged in again since the migration to the Django user framework.  

At least from my own individual outlook on user management, relations, and the querying/performance/association aspects, I'm much more confident with using the relations we've established for surfacing data associations than querying each potential association based on flat values.

I think there is a whole lot that still needs deprecated or migrated from original (and new) logic that's using value querying that makes me uncomfortable.  From a data integrity, simplicity of maintenance, and performance stance... I think creating and optimizing one queryset manager to pull relevant data through a relation is preferable over the alternative.

With that said, if we want to proceed with using all charfields and querying each as necessary, I suppose that's fine too.  I figured we wanted to move towards making full use of the relational database and simplify maintaining like queries and association across the platform though.

owocki:
@mbeacom makes sense as a long term strategy.   

my endgame for this PR is to make it easy to track new programming languages / markets we launch.  i don't want perfect to be the enemy of good, so what do we want to do for short term? 

maybe for now, i can adjust this PR to use the `Keyword` class in `marketing.models` and add a `type` (and type==programming_language for this use case) to that model ?  looking at the data in that table, it looks like its kind of a mess though..

owocki:
let me know what you think about the above action plan ^^

owocki:
this seems to have stalled out.  let me know what you think of the above; if i dont hear back, ill proceed with the proposed action plan

thelostone-mc:
Closing it out as the PR is super outdated + we are good with what we have for now! 
We'll open up a fresh PR if this is super important

PR #1658, opened 2018-07-09
Tip receive flow - display a red banner if the user's metamask is not…

### LINKED ISSUE #1606: in tip flow, user should be notified if they are on the wrong network
per https://github.com/gitcoinco/gitcoinco/issues/46#issuecomment-398208880

if the user is not on the network that the tip was sent to, on the receive page, the user should be notified they need to change their metamask settings

pinkiebell:
… on the same network.

Fixes #1606


owocki:
actually already did this in the YGE refactor so not going to merge this.. will still pay it out!

PR #1668, opened 2018-07-09
Show Just 1 Alert Banner for ERC20 Token

### LINKED ISSUE #1566: Show Just 1 Alert Banner for ERC20 Token
Two alert banners show when i've looked at two ERC 20 tokens, and not activated it.

We should show just 1 alert banner.

screen shot 2018-06-27 at 7 14 17 pm

pinkiebell:
Fixes #1566

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -854,7 +854,18 @@ var promptForAuth = function(event) {
 
     token_contract.allowance.call(from, to, function(error, result) {
       if (error || result.toNumber() == 0) {
-        _alert("You have not yet enabled this token.  To enable this token, go to the <a style='padding-left:5px;' href='/settings/tokens'> Token Settings page and enable it</a>. (this is only needed one time per token)", 'warning');
+        if (!document.alert_enable_token_shown) {
+          _alert(
+            gettext('To enable this token, go to the ') +
+            '<a style="padding-left:5px;" href="/settings/tokens">' +
+            gettext('Token Settings page and enable it.') +
+            '</a> ' +
+            gettext('This is only needed once per token.'),
+            'warning'
+          );
+        }
+        document.alert_enable_token_shown = true;
+
         $('input, textarea, select').prop('disabled', 'disabled');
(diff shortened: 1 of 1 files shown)


app/assets/v2/js/shared.js
```diff
@@ -825,7 +825,11 @@ var promptForAuth = function(event) {
 
     token_contract.allowance.call(from, to, function(error, result) {
       if (error || result.toNumber() == 0) {
-        _alert("You have not yet enabled this token.  To enable this token, go to the <a style='padding-left:5px;' href='/settings/tokens'> Token Settings page and enable it</a>. (this is only needed one time per token)");
+        if (!document.alert_enable_token_shown) {
+          _alert("You have not yet enabled this token.  To enable this token, go to the <a style='padding-left:5px;' href='/settings/tokens'> Token Settings page and enable it</a>. (this is only needed one time per token)");
+        }
+        document.alert_enable_token_shown = true;
```
thelostone-mc: I think it should be a warning alert like [here](https://github.com/gitcoinco/web/blob/639b0de163aa2a7d251efe6d605492264e9c9b6f/app/assets/onepager/js/send.js#L115)

pinkiebell:
Ahh, will rebase ;)

thelostone-mc:
@pinkiebell could you squash the commits into one neat commit  ? 
Also what's the content of `error.toString()` 
Need to make sure we aren't showing info that's not needed for the user

pinkiebell:
@thelostone-mc 
Hopefully a sophisticated error message. But I really don't know, the error object is always null no matter what I pass. However that might change in the future and could provide more info is something unexpected happens.

Let me know if you want me to revert back to just the warning message.

pinkiebell:
@thelostone-mc 
https://github.com/ethereum/web3.js/blob/54e40bb4bbb97ba9590c3ba6f933951c7a04dd16/lib/web3/errors.js

thelostone-mc:
Let's just stick to warning for now and do away with displaying the error message 👍

pinkiebell:
@thelostone-mc 
I pushed a little update, is this good to go?

Issue #1678, opened 2018-07-10
Coop Bounty - 1 Start Work Missing


PixelantDesign:
from Tim Vanmourik

https://github.com/TimVanMourik/GiraffeTools/issues/72

A coop bounty, and two applications. However, the second one (@Thoughtscript) doesn't show up in the list, and the first one twice.

![image 10](https://user-images.githubusercontent.com/238235/42521029-cb4ccc04-8435-11e8-886d-582d212b7f27.png)

owocki:
@pinkiebell is this something we missed in the last PR?

happy to tip out something extra if you can find a fix

pinkiebell:
@owocki I will take a look

pinkiebell:
@owocki can you trigger create_activity_records? Thanks

pinkiebell:
My guess: User's network connection was reset/other error after clicking on start work - modal dialog.
After that we update the ui and display a green success banner instead of an error.

owocki:
the application to start work went through, but not the `Activity` object.. so I dont think it was a network issue

pinkiebell:
@owocki 

Hmm. Then the user should be in 'bounty.interests' ? (looking at create_new_interest_helper)

https://gitcoin.co/actions/api/v0.1/bounties/?github_url=https://github.com/TimVanMourik/GiraffeTools/issues/72&network=mainnet&standard_bounties_id=724

pinkiebell:
Ohh wait. Maybe the second application @TimVanMourik means is a duplicate email from the first applicant?

TimVanMourik:
Hi @pinkiebell, sorry, my bug report wasn’t very elaborate. The timeline of the problem was this:
1. @Weezyana07 applied, I received an email and I approved him
2. @Thoughtscript applied (“@TimVanMourik - Signed up! Thanks!”). I did not receive any email. I checked out the page and saw screenshot above: a duplicate of @Weezyana07, although I could not approve him.
3. You, @pinkiebell,  applied as a test, I did receive the email and the gitcoin message was updated accordingly. I did not check out the bounty page to see if I could confirm you before you removed your application.

pinkiebell:
@TimVanMourik 

Thanks!  'Weezyana07 Applied' & 'Worker Weezyana07, Approved'  are two different events, so that is fine.

That means @Thoughtscript has not applied or he tried and had a network error and got no error message; which means he just has to hit 'Start Work' again ;).

I think the issue is closed then? @owocki

[...]

PR #1681, opened 2018-07-10
Display an error banner on network error on start-/stop work..

pinkiebell:
...instead of showing nothing.

Related #1678

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -266,7 +266,11 @@ var mutate_interest = function(bounty_pk, direction, data) {
     }
     return false;
   }).fail(function(result) {
-    alert(result.responseJSON.error);
+    _alert('Network or API error. Please reload the page and try again.', 'error');
+
+    if (result) {
+      console.log(result.responseJSON.error);
+    }
   });
 };
 
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -266,7 +266,7 @@ var mutate_interest = function(bounty_pk, direction, data) {
     }
     return false;
   }).fail(function(result) {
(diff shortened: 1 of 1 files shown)


app/assets/v2/js/shared.js
```diff
@@ -266,7 +266,11 @@ var mutate_interest = function(bounty_pk, direction, data) {
     }
     return false;
   }).fail(function(result) {
-    alert(result.responseJSON.error);
+    _alert('Network or API error. Please reload the page and try again.', 'error');
```
SaptakS: Please use `gettext()` to ensure translation

owocki:
> Display an error banner on network error on start-/stop work instead of showing nothing

I wasnt around to review this before.. but it seems clear to me from before this PR that there was an `alert()` that displayed the error before.  Why does the title of this ticket allege 'showing nothing' ?

pinkiebell:
@owocki 
We failed silently on network errors before.
In case of API errors, well you are right, we should show the error message to the user.
The reason I did a `console.log` was because that happened only on 'glitches'.
At least I assumed that back then ;)

owocki:
> We failed silently on network errors before.

Maybe we should check and see if the response has any error code in it and if it does, show an alert with the error message in it.  If not, show a generic error msg

pinkiebell:
@owocki #1918 

@PixelantDesign 
The red banner with possible messages:
            'error': _('You must be authenticated via github to use this feature!'),
            'error': _(f'There is already someone working on this bounty.'),
            'error': _(f'You may only work on max of {num_issues} issues at once.'),
            'error': _('Because a staff member has had to remove you from a bounty in the past, you are unable to start'
                       'more work at this time. Please leave a message on slack if you feel this message is in error.'),
            'error': _('You have already started work on this bounty!'),

Issue #1682, opened 2018-07-10
as a user, i want this module on /about to be clickable, so i can click it.


owocki:
page link: https://gitcoin.co/about
heatmap screenshot: 
<img width="1102" alt="screen shot 2018-07-10 at 1 58 22 pm" src="https://user-images.githubusercontent.com/513929/42534316-ac767dac-8449-11e8-9fb6-817dbe1d8d97.png">

we've learned from our heatmap screenshots (see above) that users expect this module on gitcoin.co/about to be clickable.

i would like you, bounty hunter, to take this image, and make it into HTML.  please add hover / click effects to the text areas. and please make sure the section is responsive

due date: 1 week from today.  please only claim bounty if you can turnaround before then

olafghanizadeh:
Is the click supposed to lead somewhere?

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 6 days, 17 hours from now. Please review their questions below:


1. [olafghanizadeh](https://gitcoin.co/profile/olafghanizadeh) has started work. 	
 * Q: Going to try to finish this one today :) Hopefully no merge conflicts (lol).

thelostone-mc:
@owocki It doesn't make sense to make them clickable 🤔

owocki:
> @owocki It doesn't make sense to make them clickable 🤔

i just think it's a nice touch to show that we've built interactivity into the implementation (given we already know people are clicking, why not invest in a little animation to show our attention to detail?).  i agree there is no functional action to be taken upon click.

octavioamu:
I think if people are clicking means they want some information about the "call" they clicked. I think is kind of identification 
![image](https://user-images.githubusercontent.com/1358093/42669419-c3ca3f32-862b-11e8-9941-5630b38cd0b6.png)

If I'm a repo maintainer I will click expecting to understand how gitcoin specifically will give me value, how I will use it, is an opportunity to convince that "persona" 
Ex. 
when "repo maintainer" is clicked could go to https://gitcoin.co/bounty/quickstart so with that quick guide I understand "what and how" of https://gitcoin.co/how/funder

contributor > https://gitcoin.co/how/contributor
job hunter > don't know yet but since what I had read about GitHub is in the roadmap, so just maybe to a capture form or incentive to post tasks so he could finally hire people already working in their organization

gitcoinbot:
[...]

PR #1688, opened 2018-07-11
Clickable bounty hunters #1682

olafghanizadeh:
refs: https://github.com/gitcoinco/web/issues/1682

- [ ] Make bounty hunters clickable
- [x] Test for responsiveness

- [x] Switch assets to SVG and text instead of image
- [x] Recreate HTML with bootstrap


thelostone-mc:
@owocki That seems confusing in my head! Wouldn't users end up waiting longer to see something happen 🤔

owocki:
ok im not married to the clickability idea if you guys dont think its a good idea

olafghanizadeh:
How should we proceed here?

thelostone-mc:
@olafghanizadeh just leave them as non clickable :P

PR #1690, opened 2018-07-11
nav_auth.html: make nav_avatar an <a> tag and link it to /profile/.

pinkiebell:
So that we have a shortcut to the profile page.


thelostone-mc:
Code looks alright!

@PixelantDesign this look okay to you? A direct shortcut does seem nice

owocki:
i just tested this and, at least on my local, clicking the avatr in the top right doesnt take me to my profile.  maybe it has something to do with the `.click()` handler on that element being overriden?

pinkiebell:
@owocki Fixed! If I remember correctly, I wrote the change inplace; in the browser. Maybe that's why I didn't catch the click handler 👍

Issue #1364, opened 2018-06-04
Link to Etherscan from Bounties Funded/Completed List on Profile


PixelantDesign:
### User Story

[comment]: # (As a <user>, I want to <a link to etherscan> so that <i can easily have a reference to bounties i've funded or completed>.)

### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
Currently its difficult to reference bountied work to etherscan transactions

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: 
Feature Enhancement

### Current Behavior
[comment]: # (Describe what actually happened.)
None

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

When I click the arrow icon in the list (right side), I should see the row expand and reveal the etherscan link. See below

The icon should expand and collapse.

<img width="1333" alt="screen shot 2018-06-04 at 12 57 54 pm" src="https://user-images.githubusercontent.com/238235/40931212-3d09b51e-67f8-11e8-84e3-6b97eae9a297.png">

pinkiebell:
I take care of that 👍

owocki:
note that the txhash isn't currently stored in the DB so theres some backend work associated with this.

pinkiebell:
Should that be done with management/sync_listener or should I aim at other solutions?

owocki:
i'm actually not sure what the best way is to connect a specific bounty's updates with a specific txid.  it can certainly be done in `sync_listener`, and via the `sync/web3`url route.. but I'm not sure the best way to string it through via `sync_geth`.  maybe @mbeacom  would have an idea?

pinkiebell:
I could upgrade the sync_listener to use WebsocketProvider and event filters,
and then call web3_process_bounty(bounty, tx_hash)

pinkiebell:
Btw, the above solution makes sync_geth unnecessary.

owocki:
> Btw, the above solution makes sync_geth unnecessary.

sure, assuming that sync_listener catches everything (which we dont have confidence in right now) . sync_listener is about 1 week old

pinkiebell:
@owocki 
Ready for review ;)

owocki:
link me the PR?

thelostone-mc:
@PixelantDesign do we need to display the txn hash on our page? Not sure if folks need to see the txn hash on this page.
[...]

PR #1722, opened 2018-07-12
WIP/HOLD: Switch to Web3 WebsocketProvider & rework sync_listener

pinkiebell:
Related #1364


owocki:
is this WIP or ready for review?

app/dashboard/management/commands/sync_listener.py
```diff
@@ -22,64 +22,107 @@
...
+        bounty = get_bounty(bounty_id, network)
+        did_change, old_bounty, new_bounty = web3_process_bounty(bounty)
+        if event == 'BountyActivated':
+            new_bounty.tx_hash_activated = tx_hash
+            new_bounty.save()
+            return True
+        if event == 'FulfillmentAccepted':
+            new_bounty.tx_hash_accepted = tx_hash
+            new_bounty.save()
+            return True
```
owocki: is it worth adding the other events to this list?

app/dashboard/migrations/0091_auto_20180712_1800.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dashboard', '0090_auto_20180709_1525'),
+    ]
+
+    operations = [
```
owocki: i wonder if this should just be a json field with a list of many tx hashes... @mbeacom what do you think?

requirements/base.txt
```diff
@@ -26,7 +26,7 @@ sendgrid
 slackclient
 eth-tester
 websockets
-web3==4.0.0b10
+web3==4.4.1
```
owocki: might be worth a regression test of functionality elsewhere if were bumping 4 minor versions

app/dashboard/management/commands/sync_listener.py
pinkiebell: @owocki 
Most of the time we have > Issued|Activated|ContributionAdded| > BountyFulfilled|BountyKilled > FulfillmentAccepted.

We can save all events and make those hash properties a list.

owocki: > Most of the time 

sure, but what about IncreasePayout event?  and the newFulfillment event?  those are pretty common too.

owocki: yes that makes sense to me

owocki:
is this still WIP

owocki:
also, how extensively has this been tested?  how long can it be expected to run without dying?

pinkiebell:
I ran it several times for 2 - 11 hours. It works fine, except that infura is somtimes really slow or completely unresponsive, in this case; the socket will timeout and the code picks up from the last block.

pinkiebell:
@owocki Let me know if you want me to decouple the tx_hashes & UI part from this PR.

mbeacom:
Is this still a WIP @pinkiebell ? Can you resolve the merge conflicts?

pinkiebell:
@owocki 
No WIP. Rebased 👍

pinkiebell:
@mbeacom 
s/owocki/mbeacom

pinkiebell:
This actually worked very well for a few weeks, but infura is too unreliable recently (esp. for event logs).
So I pushed it back to WIP/HOLD to improve things a bit/to cope with infura performance 👍

owocki:
cc @mbeacom in case our web3 RPC is still up and running

pinkiebell:
@mbeacom @owocki 
Any thoughts on running an own geth instance?

mbeacom:
@pinkiebell We've been discussing this a bit and are tracking here:  https://github.com/gitcoinco/web/issues/2394

I don't think it'd be a problem. We were running our own for a while, but ultimately turned it off and reverted back to Infura because we no longer saw a need for it.  I still have our geth k8s charts and configuration files, so I could have one up relatively quickly...  It's just a matter of determining the TCO and cost/benefit of doing this.  I think we'd see a major performance boost with respect to sync times.

That said, we'd want to run more than a single node to ensure HA/redundancy/resilience for production use.

PR #1756, opened 2018-07-15
removed the hiding of searchbar on mobile screens.

whb07:
##### Description

This PR solves the issue of rendering properly in a mobile screen the "sort issues by" select input. As it stands in a mobile phone the SORT select bar is hidden making the explorer unusable. As seen by the image below in an Iphone 8 (according to Safari's dev tools):
 
screen shot 2018-07-15 at 2 05 36 pm

##### Checklist

- mobile UI

##### Testing
- Screenshot of new behavior (this time from Firefox's dev tools) below. As you can see the changes on the smallest screen at 320px now renders a nice styling and much more importantly a functional filter to explore issues. There are media queries which handle the width and padding-right to reset them back at 481px ( which is already being used throughout project). 

screen shot 2018-07-15 at 2 10 57 pm

### DIFF
--- a/app/assets/v2/css/base.css
+++ b/app/assets/v2/css/base.css
@@ -1462,11 +1462,7 @@ input.is-invalid {
   .bounty_def {
     display: none;
   }
-
-  .search_bar span {
-    display: none;
-  }
-
+  
   .interior .body {
     width: 100%;
     flex: 0 0 100%;
--- a/app/assets/v2/css/jquery-ui.css
+++ b/app/assets/v2/css/jquery-ui.css
@@ -811,6 +811,12 @@ button.ui-button::-moz-focus-inner {
   width: 14em;
 }
 
+@media (max-width: 481px) {
+  .ui-selectmenu-button.ui-button {
+    width: 10em;
+  }
+}
+
 .ui-selectmenu-icon.ui-icon {
   float: right;
   margin-top: 0;
--- a/app/assets/v2/css/search_bar.css
+++ b/app/assets/v2/css/search_bar.css
@@ -150,3 +150,9 @@
     display: none;
   }
 }
+
+@media (max-width: 481px) {
+  .search_bar div .heading {
+    padding-right: 0px;
+  }
+}


thelostone-mc:
Most of our user base come to gitcoin via laptops / desktops
Do we wanna keep the sort / hide the whole thing as such on mobile 🤔 

The latter seems neater IMHO

@PixelantDesign @SaptakS  @whb07 your thoughts ?

whb07:
Hmmm...the “latter” here being to hide the sorting feature for a list of results in all mobile phones ?

thelostone-mc:
> Hmmm...the “latter” here being to hide the sorting feature for a list of results in all mobile phones ?

Yes! It's not a super needed functionality on mobile IMO which is why I'd favor removing it all together

> As I make this comment from my phone btw

Ah if you notice -> github also does hide a few options on mobile which they feel isn't super relevant

whb07:
Hmmm...While i agree with you that cluttering/offering too many features on a mobile screen is a bad thing. I do not see this tiny little widget to sort the results as being part of the “not useful on mobile list”.

Can you point out a couple major projects/companies/sites that gives users a searchable list and hides the sorting widget? Off the top of my head I am having a hard time thinking of it.

As a potential user I might come in to Gitcoin via mobile and as I’m seeing the site and looking at the explorer, be confused when I can’t interact with what clearly is a standard list. Then I’d leave and possibly never come back.

That story is essentially mine. I never thought I’d see mobile being placed as a second class citizen in 2018. 

I’ll close the PR and keep having the mobile users be wondering why the explorer list is unusable 🦄👍

PixelantDesign:
Hi @whb07 Thank you for the thoughts an input to this. The filters in general can become overwhelming on a small screen which is why we initially hid it when we rolled out this screen on mobile. We are open to your ideas. If there is an elegant way to create an experience where multiple filters can be accessed at the same time, we would love your contribution.

SaptakS:
I completely agree with @thelostone-mc and @PixelantDesign. The search filters and the search bar doesnt only make it a little too overwhelming but also reduces the visibility space of the issues altogether in smaller screens. We want to focus on showing the issues more. However I think a search bar might make sense, but in that case we might want a more elegant way, like maybe a button which on click, a search bar slides down. Something of that line I would say.

whb07:
I mean have you guys used it on mobile yet? If you tilt your screen to landscape(horizontal) the sorting select is rendered and still very much usable. There is very little cluttering (without the changes on CSS via this PR).

This project is in the business of selling issues to developers to stop by and work on them. Allowing better and friendlier interfacing with the issues seems like a critical thing to work on. if a user is unable to sort and explore issues by their choosing (value, newest, etc) they will leave. 

If you're to look at the dev tools for a screen 320px wide which is one of the smallest widths out there, it does not appear to be crowding out anything. The widths and paddings were modified to account for this as well.

thelostone-mc:
> I mean have you guys used it on mobile yet? 

It's safe to say we have which is why I said IMO the search filter isn't super relevant on mobile ^_^
Be default the list would be sorted from the latest created issues + search box should suffice.
 
I also checked the heatmaps -> and I don't see users (apart from a few) using the sort feature on mobile landscape

Would you be open to hiding it on mobile for now and if users tell us they'd want it in (cause we've got a pretty open community) -> we'd get it back in ^_^

whb07:
Maybe users leave the moment the site is unusable? So instead of having repeat numbers via mobile they either quit entirely and never come back or if they do want to stick around just use the site via desktop.

idk, I honestly don’t care. I’m only bringing it up because as a new user myself I can tell you that the design choice is a bad one.

PR #1778, opened 2018-07-18
urls.py + about

thelostone-mc:
##### Description

- updated fonts
- removed weird yellow button 
- replaced tree png with svg
- remove extra animation which can get annoying
- update urls -> repath

##### Checklist


thelostone-mc:
@mbeacom I didn't update all the url -> repath (Updating after testing if repath works)
Should I just go ahead and do it for all / continue it in batches ?

mbeacom:
I'm fine with doing this in batches.  The main reason I haven't transitioned them all over is to avoid unnecessary merge conflicts with other PRs.

As for using one or the other (re_path/path), it might not always be appropriate to use only `path` - I think we should standardize the URLs all together.  We need to decide if we want all trailing slashes or no trailing slashes.  This has implications on search indexing and internal metrics as well.

PR #1793, opened 2018-07-19
Admin actions as a seperate view

owocki:
bounty details was too crowded before

before:

screen shot 2018-07-19 at 10 12 24 am

after:
screen shot 2018-07-19 at 10 11 46 am

new view:

screencapture-localhost-8000-issue-admin-2018-07-19-10_10_51


thelostone-mc:
@owocki I like it but an extra click + new page might get annoying for moderators 

cc @vs77bb @ceresstation

vs77bb:
hmmm yeah i tend to agree with @thelostone-mc ... i do like keeping the details page quiet for funders, but i like having the moderator options right on issue details for speed as a power-user

owocki:
>@owocki I like it but an extra click + new page might get annoying for moderators

we can remove the 'new page' thing. i just couldnt find where to update that at first.

> hmmm yeah i tend to agree with @thelostone-mc ... i do like keeping the details page quiet for funders, but i like having the moderator options right on issue details for speed as a power-user

sure; but they clog the screen when youre using the site for non-moderation tasks

maybe a comprimise would be to put them on a modal on-screen on the bounty details page?

ceresstation:
I personally like having moderator options on the issue itself, a modal would work just as well though.

SaptakS:
I am thinking if a dropdown rather than an entire new page makes more sense. Like on clicking `admin actions` a dropdown opens where we show the different options..

owocki:
> I am thinking if a dropdown rather than an entire new page makes more sense. Like on clicking admin actions a dropdown opens where we show the different options..

interesting, so you would like to use the select2 UI element here, i presume?  

some of these actions require user input (like the 'change bounty status' one), which right now is hacked together with an alert view. 

for this reason, i wonder if a modal might be better so we have more real estate to add user input options.

PixelantDesign:
I agree - having the actions in context of the issue would help.

SaptakS:
@owocki Yeah true. Modal sounds a better option if we are thinking about more inputs from the user. Completely agreed.

owocki:
just pushed up some changes to put all these actions in a modal 

[screenshot](http://bits.owocki.com/f562959ff43f/Screen%20Shot%202018-08-22%20at%204.33.15%20PM.png)

PixelantDesign:
@owocki should I close #2045? Didn't realize this was still open.

owocki:
hmm i didnt realize moderator controls == admin controls.

i think there are 4 levels of control:

0. general public
1. bounty hunter
2. funder
3. admin

maybe we can use one design for admin and one for funder?

PixelantDesign:
The buttons seem the same for the most part? What is the reasoning for distinguishing/having two sets?

I thought moderator was for anyone that we wanted to give us access to, to help us do issue explorer checks, etc.

Admin controls I thought were only for core team.

frankchen07:
We could keep the designs the same and "whiteout" options that aren't available at each step? Tradeoff is it potentially could confuse funders or admins why they may or may not have access to a certain option.

PixelantDesign:
cc @vs77bb

thelostone-mc:
@PixelantDesign we've already got this [here](https://user-images.githubusercontent.com/5358146/44994225-410e4700-afbb-11e8-8536-345617d43cd5.gif)
 
We've grouped the admin actions and pushed it to the top as a stick nav on scroll

@vs77bb @ceresstation @owocki

vs77bb:
It's taking a bit of getting used to, but I like!

PR #1802, opened 2018-07-20
Utilize hosted IPFS node and fallback to Infura

mbeacom:
##### Description

The goal of this PR is to utilize our hosted IPFS nodes over Infura and fallback to Infura if connectivity fails while introducing some logging.

##### TODO

Before merging this, we need to:

- [x] Spin up IPFS nodes on k8s for staging and prod
- [x] Test on staging

##### Checklist

ipfs, bounties


owocki:
i heard from @mbeylin that infura auto-pins everything through their infrastructure..

i wonder if we'll need to do the same or explicityly call `pin()` on each thing we uplaod

docker-compose.yml
```diff
@@ -9,17 +9,17 @@ services:
     volumes:
       - pgdata:/var/lib/postgresql/data
 
-  pgweb:
-    image: sosedoff/pgweb
-    environment:
-        - DATABASE_URL=postgres://postgres:postgres@db:5432?sslmode=disable
```
owocki: did you mean to remove the pgweb image?

app/app/settings.py
```diff
@@ -408,6 +408,15 @@
 COLO_ACCOUNT_ADDRESS = env('COLO_ACCOUNT_ADDRESS', default='')  # TODO
 COLO_ACCOUNT_PRIVATE_KEY = env('COLO_ACCOUNT_PRIVATE_KEY', default='')  # TODO
 
+IPFS_HOST = env('IPFS_HOST', default='ipfs')
```
owocki: shouldnt this default to infura?

docker-compose.yml
mbeacom: Yes.  Unless there is a lot of demand for this read-only web ui, we might as well strike it from the `docker-compose.yml`.

mbeacom:
@owocki Think I should write a migration to pin all `IPFS` hashes in our system?

app/app/settings.py
mbeacom: We can definitely default to infura, if you want.  But I figured we could default to the local docker container of IPFS (all ipfs data will persist just as long as the DB data, so this should suffice and reduce latency) and use our hosted IPFS nodes as the primary on staging/live.

owocki:
> @owocki Think I should write a migration to pin all IPFS hashes in our system?

it can't hurt, though i wonder if it's worth the effort given that infura is supposedly pinning them.

docker-compose.yml
SaptakS: I have always wondered if anyone uses this, because I usually log into the pgsql image and use psql to check db related stuff.

thelostone-mc: I actually used it! :P

PR #1813, opened 2018-07-23
shows how it works on landing page / external nav

owocki:
this PR shows how it works on landing page / external nav.

im not sure why it was hidden, but this information architecture makes more sense to me.  if i'm new to the site (eg on the landing page), i might be curious how it works


mbeacom:
This was originally requested by @PixelantDesign - Why did we want to hide it?

PR #1851, opened 2018-07-30
Fix assemble_leaderboards to catch missing all_ cases

mbeacom:
##### Description

The goal of this PR is to resolve broken logic that is resulting in missed `LeaderboardRank` impressions and fix pytest to allow tests to pass.

##### Checklist

tests, leaderboards, backend

##### Testing

Locally

##### Refers/Fixes

https://github.com/gitcoinco/web/commit/76e330a55cf2164465b2aeae7b39b0c341537898

### DIFF
--- a/app/marketing/management/commands/assemble_leaderboards.py
+++ b/app/marketing/management/commands/assemble_leaderboards.py
@@ -1,20 +1,22 @@
-'''
-    Copyright (C) 2017 Gitcoin Core
+# -*- coding: utf-8 -*-
+"""Define the management command to assemble leaderboard rankings.
 
-    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


app/marketing/management/commands/assemble_leaderboards.py
```diff
@@ -112,11 +119,19 @@ def sum_bounties(b, usernames):
                 breakdown = 'yearly'
                 sum_bounty_helper(b, breakdown, username, val_usd)
 
+        add_element('all_all', username, b._val_usd_db)
```
mbeacom: @owocki Looks like we were missing cases when `idx_status` doesn't equal `done` as we were previously.

owocki: thanks

PR #1903, opened 2018-08-01
Moderators v1

mbeacom:
##### Description

The goal of this PR is to introduce the base moderator functionality for non-staff moderators.

##### Checklist

moderators, bounties

##### Testing

Locally and on staging


app/assets/v2/js/pages/dashboard.js
```diff
@@ -462,7 +463,7 @@ var refreshBounties = function(event) {
       var project_type = ucwords(result['project_type']) + ' &bull; ';
 
       result['p'] = project_type + (result['experience_level'] ? result['experience_level'] + ' &bull; ' : '');
-      
+
```
thelostone-mc: ❤️

app/dashboard/management/commands/create_activity_records.py
```diff
@@ -30,29 +30,29 @@ def set_created(activity, date):
...
-            act = helpers.record_bounty_activity('work_done', bounty.prev_bounty, bounty, fulfillment)
+            act = record_bounty_activity('work_done', bounty.prev_bounty, bounty, fulfillment)
             set_created(act, fulfillment.accepted_on)
             done_recorded = True
     if bounty.status == 'done' and not done_recorded:
-        act = helpers.record_bounty_activity('work_done', bounty.prev_bounty, bounty)
+        act = record_bounty_activity('work_done', bounty.prev_bounty, bounty)
         set_created(act, bounty.fulfillment_accepted_on)
     if bounty.status == 'cancelled':
-        act = helpers.record_bounty_activity('killed_bounty', bounty.prev_bounty, bounty)
```
thelostone-mc: why was this used like this  in the first place ? O.o

app/dashboard/router.py
```diff
@@ -226,6 +227,23 @@ def get_queryset(self):
                 interested__profile__handle__iexact=self.request.query_params.get('interested_github_username')
             )
 
+        # Retrieve all mod bounties.
+        # TODO: Should we restrict this to staff only..? Technically I don't think we're worried about that atm?
```
thelostone-mc: It shouldn't be a problem 👍 
It's only a query after all.

app/dashboard/templates/shared/sidebar_search.html
```diff
@@ -84,6 +84,42 @@
       </div>
     {% endif %}
 
+    {% if profile_id and user.profile.is_moderator or user.profile.is_staff %}
+    <div class="col">
```
thelostone-mc: Could we indent the by 1 ?

mbeacom: I guess we can.  I typically don't indent things inside of the `{% if %}` blocks because the indentation carries over into the html.

app/dashboard/management/commands/create_activity_records.py
mbeacom: 😂  Great question

Issue #1443, opened 2018-06-12
DESIGN - As a user, I want gitcoin.co/is_not_a_shitcoin to exist, so I can understand why Gitcoin did not do a token / token sale.


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
DESIGN - As a user, I want gitcoin.co/is/not/a/shitcoin to exist, so I can understand why Gitcoin did not do a token.

### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
There are so many ICOs and I might think Gitcoin is one of them if I was none the wiser.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: 
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
Doesnt exist

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Exists

I have a document that describes the content of the page.  If you are approved to work on the task I will share it with you.

### Bounty

This bounty is for *design* of said page.  Please feel free to use anything in [this repo](https://github.com/gitcoinco/creative) especially the [brand guide](https://github.com/gitcoinco/creative/tree/master/Brand).  Also, https://gitcoin.co/mission and https://gitcoin.co/results are a good place to draw some inspiration.

If you want to work on this bounty, please do two things
1. Post a link to your portfolio
2. Request access to the above document.

Design will have to be responsive.

I would very much like to have a first draft of this page by 20 days from now, so please do not apply unless you can make that happen.

Thanks, looking forward to working with you.
- Kevin

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [groundRachel](https://gitcoin.co/profile/groundRachel) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/1443/607?mutate_worker_action=approve&worker=groundRachel) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/1443/607?mutate_worker_action=reject&worker=groundRachel))_. 	
 * Q: ewrwerfsdf

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

[...]

PR #1906, opened 2018-08-01
not_a_token page 

octavioamu:
##### Description

Add front end page  /not_a_token  

I make a video because fullsize capture is without the elements (since the appear with the scroll of viewport) 
https://share.vidyard.com/watch/bUE87J9T4ZtC7MrgzcgAhL

##### Checklist

UI
##### Refers/Fixes

Refs: #1443


app/app/urls.py
```diff
@@ -189,6 +189,7 @@
     re_path(r'^about/?', retail.views.about, name='about'),
     re_path(r'^mission/?', retail.views.mission, name='mission'),
     re_path(r'^vision/?', retail.views.vision, name='vision'),
+    re_path(r'^not_a_token/?', retail.views.not_a_token, name='not_a_token'),
```
thelostone-mc: lol

app/retail/templates/shared/nav.html
```diff
@@ -26,7 +26,7 @@
       <img id="logo" src="{% static "v2/images/logo_large.png" %}" data-hover="{% static "v2/images/logo_med_hover.gif" %}" width="150" height="50" />
     {% endif %}
   </a>
-  {% if not request.path|is_in_list:"/,/contributor,/about,/mission,/vision,/results,/slack,/help,/how/funder,/how/contributor,/tools" %}
+  {% if not request.path|is_in_list:"/,/contributor,/about,/mission,/vision,/not_a_token,/results,/slack,/help,/how/funder,/how/contributor,/tools" %}
```
thelostone-mc: you would have to resolve the conflicts here 😅 
added a regex template tag to ensure this shows up only on relevant pages 
rebase and you'll see the changes :P

thelostone-mc:
@octavioamu mind squashing commits + rebasing + resolving conflicts ?

app/app/urls.py
octavioamu: @thelostone-mc first one was "is_not_a_shitcoin" so "not_a_token" is great hahahaha

app/retail/templates/shared/nav.html
octavioamu: hahaha I'm looking at it right now 2 new pages more and we will  going end with a 1mb page with that " if not request.path" hahah I will fix it

octavioamu:
@thelostone-mc 
Great I will change it to path

About the sizes are the ones in the design file, I found it big too but that page also have only a few content. Tell me if we want to make it smaller. 

I can disabled that animations, Im not fan at all of scroll interactions but that lib was already used hahah Personally I prefer without animations 

I will fix the mobile view. 

Yep the ones I have added are compressed but I will check if the others are too

thelostone-mc:
Let's do away with the animation in that case ! 
Hmm could we give it a shot with a smaller font size ? ( It's your call to make )

octavioamu:
@thelostone-mc take a look, I think is better like this: 
localhost_8000_not_a_token

octavioamu:
Also better with the title smaller too
localhost_8000_not_a_token 1

octavioamu:
@thelostone-mc everything done! 

kapture 2018-08-02 at 13 50 56

PR #1925, opened 2018-08-02
sets the bounties limit if not specified, for performance reasons

owocki:
since right now this endpiont hits 800 bounties if not specifiecd


app/dashboard/router.py
```diff
@@ -259,7 +259,10 @@ def get_queryset(self):
         queryset = queryset.distinct()
 
         # offset / limit
-        limit = self.request.query_params.get('limit', None)
+        limit = self.request.query_params.get('limit', 20)
```
mbeacom: This is what I was thinking as well...  But in this case, how are we handling paging through the results, for example in the explorer?

owocki: explicitly set the limit to 500 in the explorer until the pagination PR is merged

mbeacom: The explorer is already equipped to step through each limit up to 500 incrementing the requesting offset?

owocki: no it doesnt make new requests as you scroll down the page.  it loads them all initially into a client side cache and them paints them as you scroll down the page
b2e29bc0bbfd5d86fd708b59b235480f3bda51c8

owocki:
i think we could ship this if 

1. we updated the lazy loading of explorer results to use offset and limit params.
2. need to figure out what to do about the stats bar at the top of the explorer
3. also worth checking the iOS app to see if this would affect it at all.

owocki:
hey @mbeacom i just put up a commit that lazily makes requests on the explorer as you scroll down the page.

i had to temporariliy disable the stats module though, for now.    given enough time it'd be trivial to just create a new api/stats endpoint

take a look and let me know what you think.

app/dashboard/router.py
```diff
@@ -254,10 +254,15 @@ def get_queryset(self):
...
-        limit = self.request.query_params.get('limit', None)
+        limit = int(self.request.query_params.get('limit', 100))
+        max_bounties = 100
+        if limit > max_bounties:
+            limit = max_bounties
         offset = self.request.query_params.get('offset', 0)
         if limit:
-            queryset = queryset[int(offset):int(limit)]
+            start = int(offset)
+            end = start + int(limit)
```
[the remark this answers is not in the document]
thelostone-mc: could we remove this ?

app/assets/v2/js/pages/dashboard.js
```diff
@@ -14,6 +14,8 @@ var sidebar_keys = [
   'misc'
 ];
 
+document.results_limit = 50;
```
thelostone-mc: Does this need to be stored at the document level as opposed to 
`dashboard.results_limit` ?

app/dashboard/router.py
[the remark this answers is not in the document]
owocki: removed

app/assets/v2/js/pages/dashboard.js
owocki: updated

PR #1997, opened 2018-08-14
bounty_details.js: Show 'Submit Work' only for approved &-or started bounties.

### LINKED ISSUE #1951: Submit Work Prior To Approval
**Expected behavior**
Users should not be able to 'Submit Work' on a bounty until they have been approved to work on the bounty.

**Screenshots**
[Example here.](https://github.com/OriginProtocol/origin-dapp/issues/329)

Individuals were not yet 'approved' this task. However, 3 submitted work. This should have not been the case. You should only be able to submit work if you've been chosen to work the bounty. 

**Thoughts**
The downside would be that someone is pending 'approval' before working, but it has the upside of ensuring that random submissions don't come across for a bounty which wasn't anticipating it.

pinkiebell:
Fixes #1951


thelostone-mc:
could you throw in a screenshot ?

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -719,6 +719,23 @@ const is_current_user_interested = function(result) {
...
+  const needs_approval = result['permission_type'] === 'approval';
+  const interested = result.interested || [];
+  let len = interested.length;
+
+  while (len--) {
+    const interest = interested[len];
+    const handle = interest.profile ? interest.profile.handle : '';
+
+    if (handle && handle === document.contxt.github_handle) {
+      return needs_approval ? interest.pending === false : true;
```
SaptakS: I think just checking interest.pending isn't enough. We also need to check the `interest.status` and the `interest.acceptance_date` to  ensure that it is actually accepted and not some other state. cc: @mbeacom

pinkiebell:
@thelostone-mc The 'Submit Work' button is just not shown on these conditions. ;)
I saw a PR from @owocki which greys them out with additional tooltips, but that was for another scenario.

app/assets/v2/js/pages/bounty_details.js
pinkiebell: @SaptakS I will grep through the codebase and take a look at it 👍

pinkiebell: Seems like we rely on this too at
https://github.com/gitcoinco/web/blob/master/app/dashboard/models.py#L557
and we mutate 'pending' in
https://github.com/gitcoinco/web/blob/master/app/dashboard/views.py#L859

pinkiebell: and
https://github.com/gitcoinco/web/blob/master/app/marketing/management/commands/pending_start_work_actions.py#L28

thelostone-mc: @pinkiebell we should update that too!
@SaptakS does have a valid point -> those extra checks ensures we don't display it in the right states.

Any reason you would be against it  ?
PS: could you fix resolve the conflicts too

SaptakS: Sorry for late reply here. I would say the presence of `interest.acceptance_date` in the table would ensure. However, I am okay with resolving this thing later all over the codebase if that is more easier since I think we should get this shipped already. :)

pinkiebell:
@SaptakS That's ok. I also acknowledge the fact that we lookup how we handle the 'pending' property - properly 👍 . 
That may also play a role with the inactivity warnings the gitcoinbot plays out to a bunch of people,
if I think correctly here 😄 .

PR #2027, opened 2018-08-17
deprecates old tip receive flow

owocki:
deprecates old tip receive flow, v1 and v2, which aren't used anymore.


mbeacom:
Testing on staging - I'll merge if everything is 💚

mbeacom:
Everything seems fine on stage.  Merging!

PR #2091, opened 2018-08-24
adds leaderboard by city/country/continent

owocki:
##### Description

adds leaderboard by city/country/continent

useful for seeing where our users are, and for marketing purposes

##### Checklist

##### Refers/Fixes
marketing call today


app/marketing/views.py
```diff
@@ -615,7 +618,7 @@ def leaderboard(request, key=''):
     else:
         amount_max = 0
 
-    is_linked_to_profile = '_tokens' in key or '_keywords' in key
+    is_linked_to_profile = '_tokens' in key or '_keywords' in key or '_cities' in key or '_countries' in key or '_continents' in key
```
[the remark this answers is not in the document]
mbeacom: If we think this might continue to grow, we could also use something like:

```python
profile_keys = ['_tokens', '_keywords', '_cities', '_countries', '_continents']
is_linked_to_profile = any(sub in key for sub in profile_keys)
```

app/marketing/management/commands/assemble_leaderboards.py
```diff
@@ -36,6 +36,50 @@
...
+        locations = locations + locations
+    return locations
+
+
+def tip_to_location(tip):
+    return profile_to_location(tip.username) + profile_to_location(tip.from_username)
+
+
+def tip_to_country(tip):
+    return list(set([ele['country_name'] for ele in tip_to_location(tip) if ele and ele['country_name']]))
```
mbeacom: Maybe we could adjust this to query the profile and fetch locations once and iterate over these checks against the already retrieved data? We could potentially define tip/bounty to_something once and pass the ele key to check for as well. Either way, this should work.

mbeacom:
We'll need to update the leaderboard test case for these changes:

```
_________________ TestAssembleLeaderboards.test_handle_command _________________
self = <test_assemble_leaderboards.TestAssembleLeaderboards testMethod=test_handle_command>
    def test_handle_command(self):
        """Test command assemble leaderboards."""
        Command().handle()
    
>       assert LeaderboardRank.objects.all().count() == 12
E       AssertionError: assert 0 == 12
E        +  where 0 = <bound method QuerySet.count of <QuerySet []>>()
E        +    where <bound method QuerySet.count of <QuerySet []>> = <QuerySet []>.count
E        +      where <QuerySet []> = <bound method BaseManager.all of <django.db.models.manager.Manager object at 0x7fefe9b11080>>()
E        +        where <bound method BaseManager.all of <django.db.models.manager.Manager object at 0x7fefe9b11080>> = <django.db.models.manager.Manager object at 0x7fefe9b11080>.all
E        +          where <django.db.models.manager.Manager object at 0x7fefe9b11080> = LeaderboardRank.objects
app/marketing/tests/management/commands/test_assemble_leaderboards.py:144: AssertionError
__________________ TestAssembleLeaderboards.test_sum_bounties __________________
self = <test_assemble_leaderboards.TestAssembleLeaderboards testMethod=test_sum_bounties>
    def test_sum_bounties(self):
        """Test sum tips of bounties leaderboards."""
        user = 'freddy'
        total = 3
    
        sum_bounties(self.bounty, [user])
        print(assemble_leaderboards.ranks)
    
        rank_types_user = ['all_all', 'weekly_all', 'monthly_all', 'yearly_all']
        for rank_type in rank_types_user:
>           assert assemble_leaderboards.ranks[rank_type][user] == total
E           KeyError: 'freddy'
```

app/marketing/management/commands/assemble_leaderboards.py
owocki: i strongly agree... it seems like the ip to location stuff is pretty unperformant

owocki: @mbeacom how do you see this working?  should we just have a `Location` model that we `fk` to the Profile object when there is a new IP / location introduced, and just treat the DB as the source of truth for this?

owocki:
```
>       assert LeaderboardRank.objects.all().count() == 12
E       AssertionError: assert 0 == 12
```

i wonder why the code is now created no leaderboard rank items at all.. its like its not able to do GeoIP stuff and completely bombs out

app/marketing/management/commands/assemble_leaderboards.py
mbeacom: Hhmm - we could do that.  Will we want to track the location on every interaction?  We could add a location relation onto `UserAction` and/or drop a location fk on profile.

mbeacom:
merged in: https://github.com/gitcoinco/web/pull/2268

PR #2111, opened 2018-08-28
small tip tweaks

owocki:
- stores user agent
- makes searchable by ip and metadata


app/dashboard/tip_views.py
```diff
@@ -290,6 +290,10 @@ def send_tip_3(request):
     to_emails = list(set(to_emails))
     expires_date = timezone.now() + timezone.timedelta(seconds=params['expires_date'])
 
+    # metadata
+    metadata = params['metadata']
+    metadata['user_agent'] = request.META['HTTP_USER_AGENT']
```
mbeacom: 🤔 - `HTTP_USER_AGENT` *should* always be present in the `META`.  This shouldn't be a problem.

PR #2123, opened 2018-08-29
Fixes version of eth-utils and web3 to make compatible with other pac…

SaptakS:
…kages

##### Description
Bumps up version of eth-utils so it stays <2.0.0,>=1.2.0 because eth-abi and many other packages are not compatible with a lower version.

screen shot 2018-08-30 at 1 20 41 am

##### Checklist

python package upgrade

##### Testing

Travis was breaking because of this which it shouldn't anymore. Locally tests pass.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -25,6 +25,7 @@ sendgrid
 slackclient
 eth-tester
 websockets
+eth-abi==1.1.1
 web3==4.0.0b10
 mailchimp3
 django_chartit==0.2.9


owocki:
should we make sure that anything that relies on eth-utils doesnt have any deprecations?

owocki:
looks like its only `from eth_utils import to_checksum_address`

SaptakS:
@owocki mainly web3 and eth-abi uses eth-utils. We are not using eth-utils anywhere in our code. Which is why I updated web3 as well.

owocki:
given that we just traversed four minor versions (4.0.x => 4.4.x) i think it' feel more comfortable if you did a rinkeby bounty and a tip end to end first..

SaptakS:
It does break, because web3 now seems to allow a [MAX_EXTRA_DATA of 32](https://github.com/ethereum/web3.py/issues/898). Not very sure what is the best approach here.

SaptakS:
Okay. I think it is fixed. Instead of upgrading web3 and eth-utils, I fixed the version of eth-abi which was causing a dependency problem on eth-utils based on the last successful build. So for the time being it is fixed. But I do think maybe we should refactor the code too later.

thelostone-mc:
pulled this and checked! major flows seem to be working fine

Any other way to get this tested ?

owocki:
LGTM thanks saptak!

owocki:
actually saptak looks like this just updates `eth-abi` version not actually `eth-utils` version?

mbeacom:
Post-push of this PR:

```shell
rlp 1.0.1 has requirement eth-utils<2,>=1.0.2, but you'll have eth-utils 1.0.0b2 which is incompatible.
hexbytes 0.1.0 has requirement eth-utils<2.0.0,>=1.0.1, but you'll have eth-utils 1.0.0b2 which is incompatible.
eth-testrpc 1.3.5 has requirement rlp<=0.6.0,>=0.4.7, but you'll have rlp 1.0.1 which is incompatible.
eth-tester 0.1.0b28 has requirement eth-utils<2.0.0,>=1.0.1, but you'll have eth-utils 1.0.0b2 which is incompatible.
eth-rlp 0.1.2 has requirement eth-utils<2,>=1.0.1, but you'll have eth-utils 1.0.0b2 which is incompatible.
eth-account 0.2.2 has requirement eth-utils<2,>=1.0.2, but you'll have eth-utils 1.0.0b2 which is incompatible.
eth-abi 1.1.1 has requirement eth-utils<2.0.0,>=1.0.1, but you'll have eth-utils 1.0.0b2 which is incompatible.
```

SaptakS:
@mbeacom this was there before pushing of this PR too. LOL. And tried to make it compatible but as I mentioned above, our code isn't compatible right now with updated web3 and eth-utils.
screen shot 2018-08-31 at 11 04 51 pm

This PR actually does nothing with eth-utils. It just fixes eth-abi. So whatever you see have always been there.

mbeacom:
@SaptakS Yeah, I know that. XD - What did this PR aim to resolve though? Just `eth-abi` ? Should we follow this PR up with a resolution for the other `eth-*` packages..?

SaptakS:
@mbeacom I think we should follow this code with a migration of our code to support web3 4.4.x at least. If we do that, eth-* packages can easily be updated without any trouble. So I would say focussing on that. Without that, we can't update eth-* package because then web3 dependency will break.

Issue #2010, opened 2018-08-15
Disable Advanced Payout submission if required fields are empty/invalid


mbeacom:
**Describe the bug**
Users can submit the `Advanced Payout` form even if fields that should be required are empty or invalid.  For example, recipient address.

**To Reproduce**
Submit the advanced payout form without a recipient address.

**Expected behavior**
Display error opposed to allowing submission.

rahulpat:
@mbeacom there are some edge cases like sending 0 ETH that can be captured, which should prevent form submission. Do you want me to include those as well?

mbeacom:
@rahulpat That would be ideal, yes!

Thanks

PR #2149, opened 2018-09-03
WOC: Prevent advanced payout form submission when recipient address is empty

rahulpat:
- Captured prevention of advanced payout form submission when recipient github username is empty
Fixes: https://github.com/gitcoinco/web/issues/2010

##### Description

This PR aims to fix the bug where advance payout form submission was completing despite having empty recipient address (which is a required field)

##### Checklist

Advanced Payout form UI

##### Testing

Tested this change locally

Screenshots:

Before:

screen shot 2018-09-04 at 2 24 06 am

After:

screen shot 2018-09-04 at 2 25 28 am

##### Fixes

Fixes: #2010

### DIFF
--- a/app/assets/v2/js/pages/bulk_payout.js
+++ b/app/assets/v2/js/pages/bulk_payout.js
@@ -29,7 +29,7 @@ $(document).ready(function($) {
       self.html(self.html().replace(/(<([^>]+)>)/ig, ''));
     }, 10);
   });
-  
+
 
   $(document).on('click', '#close_bounty', function(event) {
     update_registry();
@@ -157,7 +157,7 @@ $(document).ready(function($) {
 
   $('#acceptBounty').click(function(e) {
     e.preventDefault();
-    
+
     if (!$('#terms').is(':checked')) {
       _alert('Please accept the TOS.', 'error');
       return;
@@ -166,7 +166,16 @@ $(document).ready(function($) {
       _alert('You do not have any transactions to payout.  Please add payees to the form.', 'error');
       return;
     }
+    var usernames = $('.username');
+    
+    for (var i = 0; i < usernames.length; i++) {
+      var username = usernames[i].textContent.trim();
 
+      if (username === null || username === '' || username === '@') {
+        _alert('Please provide a valid recipient Github username', 'error');
+        return;
+      }
+    }
     sendTransaction(0);
   });
 


rahulpat:
@mbeacom Please check this and let me know.

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -166,6 +166,10 @@ $(document).ready(function($) {
       _alert('You do not have any transactions to payout.  Please add payees to the form.', 'error');
       return;
     }
+    if ($('.username').val().trim() == '' || $('.username').val() === null) {
```
SaptakS: Firstly, `$('.username')` isn't an input field. It is a `div` inside a table row. So` .val()` should always return empty string. So it should use `.text()` instead of `.val()`.

Secondly, checking for empty string and null isn't enough. You also need to check if the string contains only '@', because that is what it defaults to.

Thirdly, you need to check in all rows and if there is a row with a percentage amount and no username, it should give error. A scenario as shown in the below screenshot.
screen shot 2018-09-05 at 1 02 23 pm

Fourth, please stick to `===`. I see you used `==` in one part of condition and `===` in another. :)

SaptakS:
@rahulpat and as @mbeacom already said, like you mentioned in the issue you can add a PR for 0val ETH as well and we will be happy to review and merge. :)

app/assets/v2/js/pages/bulk_payout.js
rahulpat: @SaptakS thanks for the feedback, will fix.

rahulpat:
@SaptakS updated! Also, appreciate a tip ;-)

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -166,7 +166,12 @@ $(document).ready(function($) {
       _alert('You do not have any transactions to payout.  Please add payees to the form.', 'error');
       return;
     }
-
+    $('.username').each(function() {
+      if ($(this).text().trim() === '' || $(this).text() === null || $(this).text() === '@') {
```
SaptakS: It should be `$(this).text().trim() === '@'`

SaptakS:
@rahulpat any update on this? Also, I noticed that the way you have written, this won't actually stop the function from going to the `sendTransaction(0)` statement in the code. The return that you have written is inside the callback function, so it returns void from the callback function and not from the enclosing function. And hence it will still go to the next line.

rahulpat:
@SaptakS sorry was traveling. Working on the changes now.

Reg: the early return, JS n00b here. Do you recommend i set a flag inside the callback function and then return if the flag is set ?

app/assets/v2/js/pages/bulk_payout.js
thelostone-mc: @rahulpat ^

thelostone-mc:
@rahulpat just use a `for` loop as opposed to `each` and add the return inside it 👍 
Can you wrap this up today ?

rahulpat:
@thelostone-mc updated with a boolean flag. thought it was more elegant than accessing child divs and iterating over them.

thelostone-mc:
Hmm I'm not too excited about the flag 😓 
Defer to @SaptakS

rahulpat:
@thelostone-mc @SaptakS updated to use a for loop instead of each. please check.

thelostone-mc:
Thanks for the contribution @rahulpat ^_^

Issue #1769, opened 2018-07-17
Associate PR to Existing Open Issue


PixelantDesign:
### User Story

As a Contributor I'd like for Gitcoin Bot to be smart and know that a PR I submit is related to an open issue that I've committed to.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: No detection now

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:Feature

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
When a user creates a pr, it would be nice if Gitcoin bot knew or asked if you want to associate the pr with an open issue you're working on.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- [ ] Review the work that has been done on [this ticket](https://github.com/gitcoinco/web/issues/1769)
- [ ] Fix the problem or propose a new tech approach
- [ ] Once a contributor has started and associated a PR to an issue, the Gitcoin Bot should recognize that updates are being made on the PR.
- [ ] Make Gitcoin Bot update the original Github issue somehow and also denoting activity on the PR. Feel free to brainstorm with the core team here.
- [ ] Review with core team members **@pixelandesign @mbeacom @SaptakS**

PixelantDesign:
@darkdarkdragon any thoughts on this?

mbeacom:
We have this coded up and it worked at one point:  https://github.com/gitcoinco/web/pull/458
Something seems to have broken it a while back.  It needs debugged and fixed.

owocki:
right now you can attach a PR when you submit work and that is linked in gitcoinbot's subsequent comments about the submission.  but i think this ticket is in regards to some other part of the flow?

darkdarkdragon:
@owocki after submitting pr discussion usually moves to the pr, but gitcoin bot doesn't pick that up and issues inactivity warnings to the issue thread

darkdarkdragon:
@mbeacom from reading code in pr #458 looks like it doesn't pickup comments in the pr thread

owocki:
I thought that on #458 we were using the timeline API (which should pick up related PRs).

From the PR

> Uses the GitHub timeline API to detect various types of activity on a given issue performed by the interested party.

In any case, I agree this shoudl be fixed.  It'd be helpful to have an example of a bounty (and associated PR) where it has not been working in prod to debug.

[...]

PR #458, opened 2018-02-16
Detect GitHub issue activity other than comments

JakeStoeffler:
##### Description
Implements: https://github.com/gitcoinco/web/issues/455

Uses the [GitHub timeline API](https://developer.github.com/v3/issues/timeline/) to detect various types of activity on a given issue performed by the interested party. Previously, only comments were considered. In this PR, the following events are considered "activity" and will prevent the inactivity warning email from being sent:

- `cross-referenced` - The issue was referenced from another issue
- `commented` - A comment was added to the issue
- `merge` - The issue was merged by the actor.
- `referenced` - The issue was referenced from a commit message.
- `review_requested` - review_requested

Also sets the `per_page` parameter to the max (100) for the GitHub comments and timeline APIs. This prevents comments from being missed (in most cases) if the API was returning multiple pages of results. HOWEVER, the issue of pagination still needs to be addressed in cases where there are more than 100 comments or timeline events. See [info about pagination in GitHub API](https://developer.github.com/v3/guides/traversing-with-pagination/). We might want to traverse pages of results in the future.

##### Checklist

- [x] linter status: 100% pass
- [ ] changes don't break existing behavior
- [x] commit message follows [commit guidelines](https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#step-4-commit)

##### Affected core subsystem(s)

marketing/management/commands

##### Testing

I tested locally by mocking some values into `expiration_start_work.py`. However, this should also be tested in a production-like environment before merging.

##### Refers/Fixes
Refs: https://github.com/gitcoinco/web/issues/455

PR #2176, opened 2018-09-07
expiration_start_work.py: Honor pull requests

pinkiebell:
As promised my lovely @gitcoinbot :*

Related #1769

### DIFF
--- a/app/git/utils.py
+++ b/app/git/utils.py
@@ -475,6 +475,9 @@ def get_interested_actions(github_url, username, email=''):
         elif 'user' in action:
             gh_user = action['user']['login']
 
+        if gh_user and gh_user == username and action['event'] in activity_event_types:
+            actions_by_interested_party.append(action)
+
         if action['event'] == 'cross-referenced':
             pr_num = action.get('source', {}).get('issue', {}).get('number', '')
             pr_repo_owner, pr_repo = action.get('source', {}).get('issue', {}) \
@@ -499,8 +502,6 @@ def get_interested_actions(github_url, username, email=''):
                     if gh_email and gh_email == email:
                         actions_by_interested_party.append(pr_action)
 
-        if gh_user and gh_user == username and action['event'] in activity_event_types:
-            actions_by_interested_party.append(action)
     return actions_by_interested_party
 
 
--- a/app/marketing/management/commands/expiration_start_work.py
+++ b/app/marketing/management/commands/expiration_start_work.py
@@ -99,28 +99,45 @@ class Command(BaseCommand):
                             print(" - no actions")
                         else:
                             # example format: 2018-01-26T17:56:31Z'
-                            action_times = [
-                                datetime.strptime(action['created_at'], '%Y-%m-%dT%H:%M:%SZ') for action in actions
-                                if action.get('created_at')
-                            ]
-                            last_action_by_user = max(action_times).replace(tzinfo=pytz.UTC)
+                            action_times = []
+                            has_pull_request_by_user = False
 
-                            # if user hasn't commented since they expressed interest, handled this condition
-                            # per https://github.com/gitcoinco/web/issues/462#issuecomment-368384384
-                            if last_action_by_user.replace() < interest_day_0:
-                                last_action_by_user = interest_day_0
+                            for action in actions:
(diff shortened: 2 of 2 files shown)


app/git/utils.py
```diff
@@ -475,6 +475,9 @@ def get_interested_actions(github_url, username, email=''):
         elif 'user' in action:
             gh_user = action['user']['login']
 
+        if gh_user and gh_user == username and action['event'] in activity_event_types:
+            actions_by_interested_party.append(action)
+
         if action['event'] == 'cross-referenced':
```
pinkiebell: Actually, we shouldn't fetch further at this point.
It's likely that we fetch things and (prs) from other issues, that gives us false-positives 🤔

owocki:
> As promised my lovely @gitcoinbot :*

lol. ...


reviewing now!

owocki:
two questions:

1. what does this do that https://github.com/gitcoinco/web/pull/458 doesnt?  sorry, having a hard time following the diff's in the code
2. how did you test this?

app/git/utils.py
owocki: wouldnt fetching more things / prs from other issues give us less false-negatives (more PRs, more chances we could find a place where the contributor commented)

app/marketing/management/commands/expiration_start_work.py
```diff
@@ -99,28 +99,45 @@ def handle(self, *args, **options):
...
-                            delta_now_vs_last_action = timezone.now() - snooze_time - last_action_by_user
-                            last_heard_from_user_days = delta_now_vs_last_action.days
-
-                            # decide action params
-                            should_warn_user = last_heard_from_user_days >= num_days_back_to_warn
-                            should_delete_interest = last_heard_from_user_days >= num_days_back_to_delete_interest
-                            should_ignore = last_heard_from_user_days >= num_days_back_to_ignore_bc_mods_got_it
-
-                            print(f"- its been {last_heard_from_user_days} days since we heard from the user")
+                            action_times = []
```
owocki: whyd you remove the `datetime.strptime(action['created_at'], '%Y-%m-%dT%H:%M:%SZ') ` from this list?

owocki: ah nvm, isee it was moved below

pinkiebell: This was the first bug, see variable assignment. 😄

pinkiebell: Ah, sry that was no bug here. I had my mind on the wrong file

app/git/utils.py
pinkiebell: That was the bug I'm talking about first (variable assignment/overwrite) - the gh_user.

With false-negatives I mean if we fetch further we very likely fetch PRs unrelated to the issue in question.
If we just do one loop we should get all the events we need, at least if the user references the issue in the commits/pull request(s). That may also include false-negatives, but that would be rare.

pinkiebell:
> two questions:
> 
> 1. what does this do that #458 doesnt?  sorry, having a hard time following the diff's in the code
We bail out early if we recognize a pull request (we do not check the time in this case).
> 2. how did you test this?
Locally with advanced loggin 👍

owocki:
> We bail out early if we recognize a pull request (we do not check the time in this case).

makes sense

> Locally with advanced loggin 👍

got it; i have not tested this.. did you create a issue / PR on github to test this?  if so mind passing along the link so i can test it too?

pinkiebell:
> > We bail out early if we recognize a pull request (we do not check the time in this case).
> 
> makes sense
> 
> > Locally with advanced loggin 👍
> 
> got it; i have not tested this.. did you create a issue / PR on github to test this? if so mind passing along the link so i can test it too?

I tested this with the mainnet issues, I just commented out the 'if settings.DEBUG: return' things .

pinkiebell:
@owocki 
btw,
```
should_continue_loop = len(actions) == 100
```
we should change that to `!= 0` because I had seen timeline results with multiple pages but returning < 100 per request. I don't know if that is still the case though

mbeacom:
:star: PR built and available in a preview environment **gitcoinco-web-pr-2176** [here](http://web.jx-gitcoinco-web-pr-2176.ci.gitcoin.co)

Issue #2200, opened 2018-09-11
[Code] Inviting Specific Gitcoiners To Work


vs77bb:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a Funder, I'd like the ability to invite specific Gitcoiners to work on my issue after it is funded so that I can foster continuing working relationships with people who I know and trust. 

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: After the 'Fund Issue' page, while the tx is posting to the blockchain, an option should be provided to send the Gitcoin Issue either a) around the Gitcoin community to top profiles which match the bounty description or b) around the interwebs (on Twitter, Slack, elsewhere) to create a viral loop bringing more dev's into Gitcoin. 

Here's a mock-up from #2038. 

![image](https://user-images.githubusercontent.com/23297747/45382370-9820a700-b5cd-11e8-9f47-0cacd76f56b1.png)

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature 

### Current Behavior
[comment]: # (Describe what actually happened.)
There is no in-app prompting for funders to share / invite people to work on their bounty. 

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
In-app ways for funders to share with a) Gitcoiner's they've worked with in the past, b) Gitcoiner's who they haven't worked with in the past, but have a good reputation / good fit for the issue and c) social sharing outside of Gitcoin to the funder's community. 

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- [ ] Update the bottom portion of the 'Transaction Processing' page into the image provided above (for .sketch files, see #2038). 
- [ ] Re-use existing type down from Tips to search users.
- [ ] This type down should also display wether you've worked with the contributor prior and on how many bounties.
- [ ] The user should be able to clear selected user out of the input field.
- [ ] Please implement and consider failure cases and messaging.
- [ ] Please ensure that the page is responsive.
- [ ] The bounty, once submitted, still should show up in the Issue Explorer as 'Open' status (even though some people have been invited).
- [ ] The invited attendees should receive an e-mail notification / Github notification (if possible) letting them know that they've been invited to work on the task. 

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)
- Please confirm your technical approach with @mbeacom or @thelostone-mc on technical feasibility/ available data prior to beginning.
[...]

PR #2201, opened 2018-09-11
WOC: Reserve Bounties

### LINKED ISSUE #2037: BUILD - Reserve Bounties
### User Story

As a funder, I want to fund and send a bounty directly to a certain group of individuals (or perhaps, just one individual).

### Why Is this Needed
This is needed so that select group of individuals can work on a project.

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: Right now, I sometimes get unnecessary noise on bounties that I would rather send directly to a smaller group of people that I know from prior work on the platform. 

### Description

*Type*: Feature

### Current Behavior
Post a bounty on the main issue explorer, make it approval-only, and approve the worker of my choosing. If I don't get to approve within 24 hours, perhaps deal with auto-approvals of people I didn't want to get involved. 

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

**Funder Form**
- [ ] On the funder form under Permissions Field add the following: Reserved for ___________________.
- [ ] User should be able to type in a username in the input field

**Issue Details**
- [ ] The Issue Details should display that this issue is Reserved for X Contributor with the contributor's icon, linking to their profile.
- [ ] Only the contributor(s) that have been selected to work this issue can click start work.
- [ ] Group information and line the information up like this (Will Tip Extra)
screen shot 2018-08-27 at 7 43 16 am

**Email**
- [ ] User(s) should be notified that an issue has been reserved for them to work on.

### Reviewed by 
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

@owocki @PixelantDesign @thelostone-mc Thoughts?

mul1sh:
##### Description

This PR allows funders to reserve bounties for particular users for 72 hours.

**Funder Form**
[x] On the funder form under Permissions Field add the following: Reserved for _.
[x] User should be able to type in a username in the input field

**Issue Details**
[x] The Issue Details should display that this issue is Reserved for X Contributor with the contributor's icon, linking to their profile.
[x] Only the contributor(s) that have been selected to work this issue can click start work.
[x] Group information and line the information up like this (Will Tip Extra)

**Email**
[x] User(s) should be notified that an issue has been reserved for them to work on.

##### Checklist

web

##### Testing

Yes I have tested this PR, and it will not break anything when merged into master

##### Refers/Fixes

Fixes #2037


app/assets/v2/css/bounty.css
```diff
@@ -393,6 +420,11 @@ a.btn {
   transform: none;
 }
 
+#bounty_reserved_for img {
+  margin-left: 3px;
+}
+
+
```
thelostone-mc: could we remove the extra line ?

thelostone-mc:
@mbeacom / @SaptakS could we have this up on staging once ?

app/assets/v2/css/bounty.css
mul1sh: done 😄

mul1sh:
btw @vs77bb  i added some logic alerting the user reserved for the issue about its existence, via email,  here https://github.com/gitcoinco/web/pull/2201/files#diff-4805da8d907cc7a606621667d13e1744R447.

No sure if the email is send though because I don't have a sendgrid api key to test with

vs77bb:
>  spend some time today trying to move the reserved issue into 'Work Started' and realized it can't be done easily without having to do major refactors to the python backend

@mul1sh Thanks for the detailed explanation. @mbeacom / @thelostone-mc any thoughts on how we could easily default reserved bounties to 'Work Started' with the person it's reserved for? This seems important, because otherwise the issue still shows 'Open' when it's really not. 

Another option is having it 'Open' but hidden from the main explorer. Then, the person who it is reserved for is the only one who can access it via the e-mail they receive. 

Open to ideas here, and @mul1sh we're happy to tip extra for time spent getting this right. 

> btw @vs77bb i added some logic alerting the user reserved for the issue about its existence, via email, here https://github.com/gitcoinco/web/pull/2201/files#diff-4805da8d907cc7a606621667d13e1744R447.

That sounds good -- @mbeacom @SaptakS any way we can test the e-mail as a part of the PR? That's probably the last step to make sure folks are notified that they have been reserved the bounty so they can actually 'Start Work'. Happy to help here if it's on staging // @mul1sh we can probably get you a Sendgrid API key if needed.

app/dashboard/templates/bounty/new.html
```diff
@@ -22,6 +22,7 @@
   {% include 'shared/head.html' %}
   {% include 'shared/cards.html' %}
   <link rel="stylesheet" href={% static "v2/css/submit_bounty.css" %}>
+  <link rel="stylesheet" href="{% static "v2/css/jquery.select2.min.css" %}">
```
mbeacom: This template already `includes` shared/head.html which already has `v2/css/jquery.select2.min.css` in it.  This can be removed.

app/dashboard/templates/bounty/new.html
```diff
@@ -170,5 +171,7 @@ <h5 class="font-subheader pb-2">{% trans "Gas Settings" %}</h4>
 <script src="{% static "v2/js/shared.js" %}"></script>
 <script src="{% static "v2/js/pages/shared_bounty_mutation_estimate_gas.js" %}"></script>
 <script src="{% static "v2/js/pages/new_bounty.js" %}"></script>
+<script src="{% static "v2/js/jquery.select2.min.js" %}"></script>
```
mbeacom: This template already `includes` shared/footer_scripts.html which already has `v2/js/jquery.select2.min.js` in it.  This can be removed.

app/dashboard/models.py
```diff
@@ -232,6 +232,7 @@ class Bounty(SuperModel):
     bounty_owner_profile = models.ForeignKey(
         'dashboard.Profile', null=True, on_delete=models.SET_NULL, related_name='bounties_funded', blank=True
     )
+    bounty_reserved_for = JSONField(default=dict, blank=True)
```
mbeacom: I'm skeptical that this should be `JSONField`.  Would it be more appropriate to use `reserved_for = models.ForeignKey('dashboard.Profile', null=True, related_name='reserved_bounties', on_delete=models.SET_NULL)` ?

app/dashboard/helpers.py
```diff
@@ -419,6 +420,7 @@ def create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id):
                 'bounty_owner_address': bounty_issuer.get('address', ''),
                 'bounty_owner_email': bounty_issuer.get('email', ''),
                 'bounty_owner_name': bounty_issuer.get('name', ''),
+                'bounty_reserved_for': metadata.get('reservedFor', ''),
```
mbeacom: You could move `bounty_reserved_for = metadata.get('reservedFor', '')` outside of this `if not` and remove from below and simply grab it once above this if check, so it's already fetched and assigned to be used below.

mbeacom:
I'll deploy this to staging once feedback is turned around and we've adjusted the `models.py` changes as requested or discuss strong reservations with using an FK to profile over `JSONField`.

app/dashboard/models.py
mul1sh: @mbeacom  I wanted to use a foreign key relation at first but then realized I'd also need to store the datetime when the issue was reserved for the user for comparison purposes in the UI frontend.

Also when the issue details are modified and the reserved for field is modified I also wanted to capture this as well again for display purposes in the UI i.e. check whether 24 hours are past before allowing others to [start work]

mul1sh: so thats why i chose the JSONField 🙂

mul1sh: Also finally storing the Profile as the foreign relation, would mean i do another request in the frontend to get the user details  when I want to manipulate them. So I found this to be unnecessary, so thats why I also stored everything for reserved_for in the JSONField and then just get it all at once in the frontend with a single api call.

thelostone-mc:
@mul1sh heads up ! mind squashing the commits before final review?

mul1sh:
@thelostone-mc sure, squashing them now

app/dashboard/models.py
mbeacom: @mul1sh Thanks for the followup!

I understand what you're saying here, my only hang up is if you're storing it as an FK, we'd be able to fetch the reserved data directly in one query, likewise if you're fetching bounty details, you could query once to fetch the profile AND bounty data (just as long as you're using the appropriate `select_related`/`prefetch_related`.  The only time you should have to fetch the reserved user profile independently would be to add it to the bounty initially.

Also, do we care about the datetime when the bounty reserved for data was established? I assume we allow them to reserve for a user profile during bounty creation, so if it's a reserved bounty, we could check against the DT on `Bounty.created_on` and if the delta exceeds 24 hours, drop the reservation and make it open.

Feel free to elaborate if I'm missing something.

vs77bb:
@PixelantDesign That makes sense... I'm now partial to 'Open' but hidden for 72 hours so it doesn't confuse people on 'Open'. 

We can clarify that reserved means it will be held for that specific bounty hunter for 72 hours, at which point it will be put into an 'Open' status if the hunter doesn't accept the work. 

This also is extensible to #2200 where we could make bounties unlisted for _x_ hours ('invite only') and then make them 'Open' if no one bites.

app/dashboard/models.py
mul1sh: @mbeacom thanks for clarifying, makes sense now why I should have the foreign key, I actually hadn't thought of using `select_related/prefetch_related` but you are very right with a foreign key this will work.

But still for the bounty I had thought of using `Bounty.created_on` to check if the reserved for time period exceeds 24 hours, but what about when the user updates the issue details and re-assigns the reserved issue to another user? Thats also the reason I chose the JSON object so as to capture this DT. So how do you think I should handle this?

mul1sh:
@vs77bb thanks for clarifying, so just to confirm, a reserved issue should be hidden from the issue explorer for 72 hours and then after that if the reserved user hasn't started work, it should be visible for other people to attempt, correct?

app/dashboard/models.py
mbeacom: Maybe we could use `Activity` since we can already obtain that with a query to `Bounty` via `Bounty.activities`.  We'd have to add a type to `Activity.ACTIVITY_TYPES` and create the relevant object following invited/declined/removed based on the user/funder actions.  I'm not entirely married to this idea, though.

mul1sh: Ok thanks for the clarification 🙂 , let me implement this in the next few hours then.

mul1sh: @mbeacom one last thing before i push another update, how I get the avatar to display for the reserved_for user? Right now i'm displaying a placeholder but I want to display the users actual avatar in the next commit. Thanks

mul1sh:
@vs77bb @PixelantDesign  Kindly expound on the 72 hour reserved issue `hidden` logic, I'm about to push another commit and I want to make sure I don't miss anything.

Also should the `contest` and `co-operative` projects be reserved for users or should I hide the field when a funder selects either of these fields under project type? 

Thanks

vs77bb:
> so just to confirm, a reserved issue should be hidden from the issue explorer for 72 hours and then after that if the reserved user hasn't started work, it should be visible for other people to attempt, correct?

I think this would be great, yes. Sorry for the late reply. 

>Also should the contest and co-operative projects be reserved for users or should I hide the field when a funder selects either of these fields under project type?

Hiding the field works, you shouldn't be able to reserve `contest` or `cooperative` bounties.

mul1sh:
@mbeacom I've implemented the changes requested, so now I use a foreign key for the reserved user profile instead of the `JSONField`. I've also modified the email body string to include the `url` of the created reserved bounty.

@thelostone-mc @vs77bb @PixelantDesign I've also modified the logic so that a reserved issue will be hidden in the issue explorer, for at least 72 hours after its created, and after that any other bounty hunter can apply to start working on the issue if approved and if the reserved user has not started working on the issue.

Also the `reserved_for` field only appears in the issue details, if the `project_type` is `traditional` otherwise it will disappear.  Finally here is another video showing the changes i've made https://vimeo.com/290557655 . Thanks

mul1sh:
@mbeacom @vs77bb any more changes I should make to this PR? thanks 🙂

mul1sh:
Hellooo, the silence is worrying 🙁

app/dashboard/helpers.py
```diff
@@ -465,6 +475,17 @@ def create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id):
             logging.error(f'{e} encountered during new bounty creation for: {url}')
             new_bounty = None
 
+        # notify a user that a bounty has been reserved for them once the bounty has been created
+        if reserved_for_user and new_bounty is not None:
+            from_email = _('founders@gitcoin.co')
```
mbeacom: We shouldn't wrap the email in gettext. `_()`

app/dashboard/helpers.py
```diff
@@ -465,6 +475,17 @@ def create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id):
...
+            email_subject = _('Reserved Issue')
```
mbeacom: This whole block should be moved to the `emails.py` file in it's own email method.

app/dashboard/helpers.py
```diff
@@ -465,6 +475,17 @@ def create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id):
...
+            body = _(f'<p>Hi {reserved_for_user.handle},<br><br>An issue has been assigned to you on gitcoin. '
+                     f'The link to the issue is {new_bounty.url}. '
+                     f'Please start working on the issue, in the next 72 hours,'
```
mbeacom: `f` is only required on lines that have variables wrapped in `{}`

app/dashboard/templates/bounty/change.html
```diff
@@ -21,6 +21,7 @@
   {% include 'shared/head.html' %}
   {% include 'shared/cards.html' %}
   <link rel="stylesheet" href={% static "v2/css/submit_bounty.css" %}>
+  <link rel="stylesheet" href="{% static "v2/css/jquery.select2.min.css" %}">
```
mbeacom: These css and js files are already loaded in `head.html` and `footer_scripts.html`

app/dashboard/templates/bounty/details.html
```diff
@@ -58,50 +58,70 @@ <h4 id="title" class="font-title"></h4>
               <div class="col-12">
                 <div id="status"></div>
               </div>
-              <div class="col-12 bounty-info">
-                <div id="timer">
-                  <div>
-                    <span class="bounty-info-heading timeleft">{% trans "Time left" %} </span>
-                    <span class="bounty-info-text" id="expires_date"></span>
+              <div id="bounty-info-row" class="row">
+                  <div class="col-12 col-sm-6 col-xl-3 bounty-info">
```
mbeacom: This indentation needs fixed. 2-space indentation throughout *.html files please :D

app/dashboard/views.py
```diff
@@ -1492,7 +1492,8 @@ def change_bounty(request, bounty_id):
         else:
             raise Http404
 
-    keys = ['experience_level', 'project_length', 'bounty_type', 'permission_type', 'project_type']
+    keys = ['experience_level', 'project_length', 'bounty_type', 'permission_type', 'project_type',
```
mbeacom: 😭

app/dashboard/views.py
```diff
@@ -1518,7 +1519,21 @@ def change_bounty(request, bounty_id):
         for key in keys:
             value = params.get(key, '')
             old_value = getattr(bounty, key)
+
+            if key == 'bounty_reserved_for_user' and value is not None:
```
mbeacom: Value will never be `None` here since it's defaulting to an empty string on L1520

app/dashboard/views.py
```diff
@@ -1518,7 +1519,21 @@ def change_bounty(request, bounty_id):
...
+                if old_value is not None:
+                    old_value = old_value.id
+                else:
+                    old_value = None
```
mbeacom: If `old_value` is `None` as validated by the `if` above, there's no need to reassign this value.

app/dashboard/views.py
```diff
@@ -1518,7 +1519,21 @@ def change_bounty(request, bounty_id):
...
+
+            if key == 'bounty_reserved_for_user' and value is None:
```
mbeacom: Same as above.  This will never be `None`, but rather `''`
In python, `None != ''`.

If you had:


```python
val = ''
val2 = None

if not val:
    # This scope will be reached.
    print('in scope')

# Same here
if not val2:
    # In scope
    print('in scope')

# Not here though
if val is None:
    # Not in scope
    print('You never see me')
```

Also, this can be done with an if/elif vs the double if blocks.

app/dashboard/views.py
```diff
@@ -1518,7 +1519,21 @@ def change_bounty(request, bounty_id):
...
+                value = None
+                old_value = 'old_val'
+
             if value != old_value:
+                if key == 'bounty_reserved_for_user' and value is not None:
```
mbeacom: Same as the above

app/dashboard/views.py
```diff
@@ -1518,7 +1519,21 @@ def change_bounty(request, bounty_id):
...
+
+            if key == 'bounty_reserved_for_user' and value is not None:
+                if old_value is not None:
+                    old_value = old_value.id
+                else:
+                    old_value = None
+
+            if key == 'bounty_reserved_for_user' and value is None:
+                value = None
+                old_value = 'old_val'
```
mbeacom: What's this about?

app/dashboard/views.py
```diff
@@ -1546,7 +1561,13 @@ def change_bounty(request, bounty_id):
 
     result = {}
     for key in keys:
-        result[key] = getattr(bounty, key)
+        if key == 'bounty_reserved_for_user':
+            bounty_res = getattr(bounty, key)
+            result[key] = '{}'
```
mbeacom: Why a string of `{}` ?

app/dashboard/views.py
```diff
@@ -1546,7 +1561,13 @@ def change_bounty(request, bounty_id):
...
+            if bounty_res is not None:
+                result[key] = '{"id":' + str(bounty_res.id) + ',"text":"' + bounty_res.handle + '"}'
```
mbeacom: Why is this a string versus a dict? 🤔

mul1sh:
@mbeacom anything else I should do here?

Issue #2217, opened 2018-09-13
Underscores in URLs results in italicized text


john-light:
**Describe the bug**

When a URL in the task description contains two underscores, the text between the underscores is italicized and the underscores disappear. See screenshot for example.

**To Reproduce**

Put a URL with two underscores in it in a task description

 e.g. `https://monthly.aragon.one/guides/guide_for_submitting_news/`

**Expected behavior**

I expect the URL to have all of the proper characters in it.

**Screenshots**

![underscore_gitcoin](https://user-images.githubusercontent.com/9424721/45521109-68aa9f80-b771-11e8-93cf-87ff05166d39.png)

**Desktop (please complete the following information):**

- OS: Linux Mint 18.3
- Browser: Firefox
- Browser Version: 62.0

mbeacom:
<img width="841" alt="screenshot 2018-09-14 14 26 46" src="https://user-images.githubusercontent.com/7315957/45568304-46288d00-b82a-11e8-91e4-b6b3cdad7ef1.png">
Resolved.  Thanks for reporting this issue!

PR #2218, opened 2018-09-13
Set up showdown to parse links

arun1595:
Fixes https://github.com/gitcoinco/web/issues/2217


arun1595:
@mbeacom FYI, I don't have enough time at the moment to spin up the docker container and test it. 

But I did test this on Showdown live editor.

## With `simplifiedAutoLink: false`
image

## With `simplifiedAutoLink: true`
image

PR #2251, opened 2018-09-20
Fix box position bug in select2

octavioamu:
##### Description

Fix wrong box position when select 2 render first time 

image

##### Checklist

### DIFF
--- a/app/assets/v2/js/user-search.js
+++ b/app/assets/v2/js/user-search.js
@@ -29,6 +29,14 @@ $(function() {
     templateSelection: formatUserSelection
   });
 
+  // fix for wrong position on select open
+  var select2Instance = $('.username-search').data('select2');
+
+  select2Instance.on('results:message', function(params) {
+    this.dropdown._resizeDropdown();
+    this.dropdown._positionDropdown();
+  });
+
   function formatUser(user) {
 
     if (user.loading) {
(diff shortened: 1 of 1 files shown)


octavioamu:
Take that @stickler-ci you couldn't find nothing heee???

octavioamu:
not sure why codecov keeps complaining

mbeacom:
@octavioamu You're fine.  One thing I'm noticing is previous changes being re-added in this PR: https://github.com/gitcoinco/web/pull/2251/files#diff-3e886cd5b50df8f0cabc86d33ca734e6R1015 Would you mind fixing the rebase here?  Other than that, lgtm.  cc @SaptakS

app/dashboard/views.py
```diff
@@ -1012,7 +1012,7 @@ class ProfileHiddenException(Exception):
     pass
 
 
-def profile_helper(handle, suppress_profile_hidden_exception=False):
+def profile_helper(handle, suppress_profile_hidden_exception=False, current_user=None):
```
mbeacom: Seems previous changes are being re-added in this PR.  Mind fixing the rebase?

octavioamu:
@SaptakS @mbeacom I had returning back because in the middle time Mark solves the lint problem in receive.js so both branches master and mine where conflicting so returned to the inicial fix since the lint one is not needed anymore.

app/assets/v2/js/user-search.js
```diff
@@ -29,6 +29,14 @@ $(function() {
     templateSelection: formatUserSelection
   });
 
+  // fix for wrong position on select open
+  var select2Instance = $('.username-search').data('select2');
```
SaptakS: @octavioamu correct me if I am wrong. But won't this return the value in the `data-select2` attribute instead of returning the select2 instance? Shouldn't we be doing something like `$('.username-search').find('data-select2')` ?

octavioamu: Because the plugin need the data to recalculate the size of the box.

octavioamu: In others words is the way to access the plugin functions and data not just the dom element

SaptakS: Okay. 👍

PR #2256, opened 2018-09-20
 Resurface Issue To the top when Expiration has been Extended

### LINKED ISSUE #2184: Build - Resurface Issue To the top when Expiration has been Extended
## Description 

As a user I want to see issues at the top If no one has started work and the expiration has been extended.

## Definiton of Done

- [ ] Build in logic to make an issue refresh and display at the top of the issue explorer if the issue time frame has been extended.

Please review with @pixelantdesign @vs77bb

mul1sh:
##### Description

This PR will push issues to the top, If no one has started work and the expiration date has been extended.

##### Checklist

web

##### Testing

I have tested this change thoroughly, it won't break anything 🙂 

##### Refers/Fixes

Fixes #2184

### DIFF
--- a/app/assets/v2/js/pages/dashboard.js
+++ b/app/assets/v2/js/pages/dashboard.js
@@ -70,7 +70,7 @@ var buildURI = function() {
   let uri = '';
   let _filters = filters.slice();
 
-  _filters.push('keywords', 'order_by');
+  _filters.push('keywords', 'order_by', 'order_by_date');
   _filters.forEach((filter) => {
     if (localStorage[filter] &&
       localStorage[filter] != 'any') {
@@ -86,6 +86,7 @@ var buildURI = function() {
  */
 var save_sidebar_latest = function() {
   localStorage['order_by'] = $('#sort_option').val();
+  localStorage['order_by_date'] = '-expires_date';
 
   filters.forEach((filter) => {
     localStorage[filter] = '';
@@ -289,9 +290,13 @@ var get_search_URI = function(offset) {
   }
 
   var order_by = localStorage['order_by'];
+  var order_by_date = localStorage['order_by_date'];
 
   if (order_by) {
     uri += '&order_by=' + order_by;
+  }
+  if (order_by_date) {
+    uri += '&order_by_date=' + order_by_date;
   }
   uri += '&offset=' + offset;
   uri += '&limit=' + results_limit;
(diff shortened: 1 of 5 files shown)


PixelantDesign:
@thelostone-mc @SaptakS @mbeacom is this good to pay out?

mul1sh:
@PixelantDesign here is a video showing the PR in action https://vimeo.com/290886103

app/assets/v2/js/pages/dashboard.js
```diff
@@ -393,6 +393,36 @@ var reset_offset = function() {
   document.offset = 0;
 };
 
+// function that resurfaces an issue to the top when expiration has been extended
+var resurfaceExtendedExpirationIssue = function(issues) {
```
mbeacom: 🤔 Any reason we're not updating the backend expiration date and allowing the backend to resurface the issue in the appropriate order?

mbeacom:
@PixelantDesign Not until we have merged the PR.

PixelantDesign:
@mul1sh thanks for the video!

So if I as a user came into the issue explorer. I would see all of the issues that have expiration extended at the top, and they would be blue until they faded out, correct?

app/assets/v2/js/pages/dashboard.js
mbeacom: @mul1sh  This will probably break when more results are loaded in from the paginated results coming from the backend, since order is dictated by `order_by` param being passed to the API and if it doesn't, it'll have to run through all results as the user scrolls down and attempts to render more results.  If we want to display any kind of visual indicator that the expiration was recently updated, this is something we could also determine in the router and pass as a key/value in the response.

We probably don't want to manipulate the `order_by` results of the queried data on the frontend.  Since the expiration is being updated, why not just adjust the filtering on the explorer from `-created` to `-modified_on` ?  Plus this might hinder performance, since we're passing all results coming from the backend into this function.

mul1sh:
@mbeacom I totally agree with you, having the filtering in the frontend might create more problems than needed, so I've moved everything to the backend.

also the date i'm using for the ordering is `expires_date` since for this issue to be solved, I have to move an issue to the top where the expiration date has been extended and this is reflected in the `expires date` not `modified_on` from what i've seen.

app/assets/v2/js/pages/dashboard.js
mul1sh: https://github.com/gitcoinco/web/pull/2256#issuecomment-423342065

mul1sh:
@mbeacom anymore changes to be made here?

PR #2263, opened 2018-09-21
Add VSCode remote debugging handling and documentation

mbeacom:
##### Description

The goal of this PR is to add vscode remote debugging support (optional) and necessary setup documentation.

##### Checklist

##### Testing

Locally

### DIFF
--- a/app/manage.py
+++ b/app/manage.py
@@ -1,21 +1,28 @@
 #!/usr/bin/env python3
-'''
-    Copyright (C) 2017 Gitcoin Core 
+"""Define the Django main entry.
 
-    This program is free software: you can redistribute it and/or modify
-    it under the terms of the GNU Affero General Public License as published
-    by the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
+Copyright (C) 2018 Gitcoin Core
 
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-    GNU Affero General Public License for more details.
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
 
-    You should have received a copy of the GNU Affero General Public License
-    along with this program. If not, see <http://www.gnu.org/licenses/>.
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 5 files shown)


timborden:
@mbeacom & @SaptakS - just a heads up...looks like `ptvsd` needs to be added to https://github.com/gitcoinco/web/blob/master/requirements/base.txt as well. Not a prod issue, as it only errors when `VSCODE_DEBUGGER_ENABLED` is _"on"_ ....and even then, it appears to fail gracefully in the development environment....but VS Code debugging won't work without it.

mbeacom:
@timborden Please check out the documentation at: https://docs.gitcoin.co/mk_contributors/#vscode-remote-debugger

[This section](https://github.com/gitcoinco/web/pull/2263/files#diff-9b74db5a5a6c7358724d34a227099303R27) in the additions installs `ptvsd` on initial launch of the container if `VSCODE_DEBUGGER_ENABLED=on`.  I'm fine with adding the dep to `base.txt`, but I was trying to avoid changes for users that don't use vscode (or the debugger).

If you execute: `docker-compose down; docker-compose up -d`, it *should* automatically install `ptsvd` on initial spin up IFF `VSCODE_DEBUGGER_ENABLED =on` is already present in the `.env` when you start the compose stack.

timborden:
Ahh....I spotted the `VSCODE_DEBUGGER_ENABLED=on` after running `make fresh` and didn't realise it was used in `docker-command.bash` as well.

I like the approach and agree it's better to keep the package out of the requirements for those that don't use VS Code.

Thanks for clarifying! (...nice work, by the way)

timborden:
I needed to change my `"remoteRoot": "/code/app"` in `launch.json` to `"remoteRoot": "/code"`...not sure if that's specific to my setup.

mbeacom:
@timborden 🤔 That's odd.  Which version of vscode and OS are you running on?

Do you happen to know what the error read?  The provided `launch.json` snippet works on my end, but I want to catch this case before other users start adopting it.

Also, a random note...  I was doing to include a `make` entry to automatically add the provided launch configuration to the project level `launch.json`, but Microsoft uses `//` comments in the automatically generated flavor of the config file that yields errors without stripping them out first.  I didn't want to take the liberty of modifying user config files to that degree.  I'm open to discussion about whether or not this process **should** be abstracted further, though.

timborden:
I'm on OSX 10.13.6 and VS Code 1.27.2...here's the warning:
```
pydev debugger: warning: trying to add breakpoint to file that does not exist: /code/app/app/marketing/views.py (will have no effect)
```

Interesting idea. It's probably enough to note the instructions in the Markdown file. There's a chance that someone could already have a `launch.json` file setup (unlikely, but possible) so the code would need to add/update that file....which (to me) feels like it could be a rabbit hole.

mbeacom:
@timborden It appears this was due to a modification in one of **my** testing configs.  I have updated the documentation accordingly.  Thanks @timborden

PR #2276, opened 2018-09-25
WIP: Wallet base implementation

mbeacom:
##### Description

The goal of this PR is to replace `preferred_payout_address` while not modifying the legacy logic of how we handle the payout address to avoid FE modifications for the time being.

*Note: This is a work in progress and will be revisited following the shipment of a few other features...*

##### Checklist

profile, payouts

##### Testing

Locally


app/dashboard/views.py
```diff
@@ -1568,15 +1569,15 @@ def get_users(request):
...
-            if user.avatar_id:
-                profile_json['avatar_url'] = user.avatar_url
-            profile_json['preferred_payout_address'] = user.preferred_payout_address
+            profile_json['id'] = profile.id
+            profile_json['text'] = profile.handle
+            profile_json['email'] = profile.email
+            profile_json['avatar_id'] = profile.avatar_id
+            if profile.avatar_id:
+                profile_json['avatar_url'] = profile.avatar_url
+            profile_json['preferred_payout_address'] = profile.get_default_wallet().address
```
jasonrhaas: Would it be cleaner to have a property like `profile.default_wallet`?

app/dashboard/models.py
```diff
@@ -1543,6 +1550,62 @@ def tips(self):
...
+    def wallet_exists(self, address):
+        """Check the provided address against existing associated wallets.
+
+        Returns:
+            bool: Whether or not the address is already an associated wallet.
+
+        """
+        return self.get_wallet_by_address(address).exists()
+
+    def get_default_wallet(self):
```
jasonrhaas: Related to the other comment below.  Could make a property:

```
@property
def default_wallet_address(self):
    try:
        wallet = self.objects.wallets.get(is_default=True)
    except Wallet.DoesNotExist:
        return None
    else:
        return wallet.address
```

app/economy/models.py
```diff
@@ -118,6 +118,30 @@ def reverse_conversion_rate(sender, instance, **kwargs):
...
+        is_default (bool): Whether or not this Wallet is the preferred payout address.
+        profile (dashboard.Profile): The owner of this wallet.
+        wallet_users (dashboard.Profiles): All Profiles that have used this wallet.
+
+    """
+
+    address = models.CharField(max_length=255, db_index=True, unique=True)
+    is_default = models.BooleanField(default=False)
+    profile = models.ForeignKey('dashboard.Profile', on_delete=models.CASCADE, related_name='wallets')
+    wallet_users = models.ManyToManyField('dashboard.Profile', blank=True)
```
jasonrhaas: Is the `wallet_users` field so that one user can have multiple wallets?

PR #2321, opened 2018-10-02
fixed avatar builder color handling

### LINKED ISSUE #2150: bug - avatar is wrong skin color
[screenshot](http://bits.owocki.com/a3b86f8690f7/Screen%20Shot%202018-09-03%20at%203.37.17%20PM.png)
https://gitcoin.co/profile/owocki

Repro steps
1. make an avatar. save it
1. go to avatar builder
1. wrong skin color is shown

expected behaviour:
1. skin color is the same one as it is in my profile photo

evgeniuz:
##### Description

Fixes parsing colour from path in avatar builder. The issue was that on `localhost` splitting with `'.'` was working correctly, but in production environment `path` includes actual domain with dots and splitting returns incorrect portion. Fixed this by taking portion not from beginning, but from end (where there's no domain name).

##### Checklist

Does not affect core subsystems.

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/2150.

### DIFF
--- a/app/assets/v2/js/avatar_builder.js
+++ b/app/assets/v2/js/avatar_builder.js
@@ -82,10 +82,13 @@ function getIdFromPath(path, option) {
 function getColorFromPath(path, option) {
   let color = '';
 
+  // filename without path and extension
+  const basename = path.split('.').slice(-2)[0].split('/').slice(-1)[0];
+
   if (option === 'FacialHair' || option === 'Accessories') {
-    color = path.split('.')[0].split('/').slice(-1)[0].split('-')[2];
+    color = basename.split('-')[2];
   } else {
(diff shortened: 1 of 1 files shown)


owocki:
yayyy -- thanks for the quick fix @evgeniuz !  @mbeacom i just tested this on my local and it seems to work

evgeniuz:
@owocki, note that issue was caused by dots in production domain, so if your local domain does not have dots in it (e. g. `localhost`) issue wasn't appearing there anyway.

mbeacom:
Testing on staging.

mbeacom:
Tested on staging.  LGTM!

@evgeniuz Thanks for the contribution!  Good to pay out @owocki !

PR #2296, opened 2018-09-28
payout: added autocomplete

thelostone-mc:
- Ported `truncate-hash.js`
- Adds autocomplete feature to bulk payout
- Shows address if available
- introduced flag to ensure the address shows up only in payout and not tip

x

PR #2337, opened 2018-10-03
take users word for it if no gitcoin profile

owocki:
fixes https://gitcoincore.slack.com/archives/CAXQ7PT60/p1538571121000100

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -1578,6 +1578,15 @@ def get_users(request):
                 profile_json['avatar_url'] = user.avatar_url
             profile_json['preferred_payout_address'] = user.preferred_payout_address
             results.append(profile_json)
+        if len(profiles) == 0:
+            profile_json = {}
+            profile_json['id'] = -1
+            profile_json['text'] = q
+            profile_json['email'] = None
+            profile_json['avatar_id'] = None
+            profile_json['preferred_payout_address'] = None
+            results.append(profile_json)
+
         data = json.dumps(results)
     else:
         raise Http404


app/dashboard/views.py
```diff
@@ -1578,6 +1578,15 @@ def get_users(request):
                 profile_json['avatar_url'] = user.avatar_url
             profile_json['preferred_payout_address'] = user.preferred_payout_address
             results.append(profile_json)
+        if len(profiles) == 0:
```
mbeacom: Can we adjust this to `if not profiles:` ?

owocki:
hey @mbeacom what do you think of ff8155bd6f65dead7d08696d6020e3af4f865c31 which searches github too?

mbeacom:
@owocki lgtm - The only thing I'm concerned with is potentially hitting the rate limit for search (30 rph).

owocki:
@mbeacom is it worth either (1) rate limiting the users on github lookups or (2) using the logged in users GITHUB key instead of ours?

for the latter, 

1. is there any readymade examples where that is done i can copy?  what field off user/profile do i want
2. we could possibly just provide a message that says 'login to search github' if a user is not logged in / has no github oauth.

app/dashboard/views.py
```diff
@@ -1578,6 +1579,28 @@ def get_users(request):
...
+                profile_json = {}
+                profile_json['id'] = -1
+                profile_json['text'] = result['login']
+                profile_json['email'] = None
+                profile_json['avatar_id'] = None
+                profile_json['avatar_url'] = result['avatar_url']
+                profile_json['preferred_payout_address'] = None
+                results.append(profile_json)
+        # just take users word for it
+        if not profiles:
```
mbeacom: should we check `if not profiles and not results` here to avoid the dupe if truthy? @owocki 
I'm about to commit a pygithub based search method to use our current token flow.  You'd just pass `request.access_token` / `access_token` from the context to: `search_users(query, token=access_token)` from `git.utils`

owocki: lgtm

app/assets/v2/js/user-search.js
```diff
@@ -1,7 +1,7 @@
 $(function() {
   $('.username-search').select2({
     ajax: {
-      url: '/api/v0.1/users_search/',
+      url: '/api/v0.1/users_search/?token=' + currentProfile.access_token,
```
owocki:
```
Uncaught ReferenceError: currentProfile is not defined
```

owocki: this is while im logged in

mbeacom:
@owocki All set.

mbeacom:
search_gh

app/assets/v2/js/user-search.js
```diff
@@ -1,7 +1,7 @@
 $(function() {
   $('.username-search').select2({
     ajax: {
-      url: '/api/v0.1/users_search/',
+      url: '/api/v0.1/users_search/?token=' + currentProfile.githubToken,
```
thelostone-mc: Should we put in a check there to ensure the token is available ?

thelostone-mc: fixed it here https://github.com/gitcoinco/web/pull/2296

mbeacom: Yeah 😛

Issue #2212, opened 2018-09-13
as a data analysist, i want the utm* params of a login UserAction stored in the DB


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a data analysist, i want the utm* params of a login UserAction stored in the DB

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

`utm_*` params are the way we know what marketing campaign engaged a user.  for example, if a user comes through a campaign like this http://gitcoin.co/?utm_medum=xxx&utm_source=yyy&utm_campaign=zzzz 

we would like to be able to attribute actions that they take on the platform to the 

utm_medium
utm_campaign
utm_source

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)

utm_medium
utm_campaign
utm_source 

not tracked in the db

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

anytime a user visits the site with a  

utm_medium
utm_campaign
utm_source

we store that info into a cookie, and that cookie is transmitted to the server with actions they take.

anytime a 

`UserAction.objects.create()` of `action==Login, Logout` happens during that session, I would like for these three parameters to be tracked on the User Action.

This will allow us to track utm parameters in our metabase dashboard

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 75.0 DAI (75.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/2212/1239).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/2212/1239).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $22,569.92 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


[...]

PR #2348, opened 2018-10-04
WOC: Implement utm* params of a login UserAction stored in the DB feature #2212

unclebean:
##### Description

This PR is for saving utm* params to DB for data analysis, relevant story is [2212](https://github.com/gitcoinco/web/issues/2212). 
Whenever user access gitcoin with "utm_medum, utm_source, utm_campaign" will store them in cookie, after user login or user logout will store to dashboard_useraction table.

##### Checklist

UserAction data model.

##### Testing

Test with below 3 scenarios.

**Scenario:** Record utm* in Cookie
**Give:** utm_medum, utm_source, utm_campaign in URL
**When:** User open gitcoin
**Then:** Should record utm* in Cookie

**Scenario:** Record utm* in DB
**Give:** utm* in Cookie
**When:** User click "LOG IN" button
**Then:** Should record utm* in DB

**Scenario:** Record utm* in DB
**Give:** utm* in Cookie
**When:** User click "Log Out" button
**Then:** Should record utm* in DB


app/dashboard/utils.py
```diff
@@ -125,6 +135,30 @@ def create_user_action(user, action_type, request=None, metadata=None):
...
+
+    """
+    utm_source = request.COOKIES.get('utm_source')
+    utm_medium = request.COOKIES.get('utm_medium')
+    utm_campaign = request.COOKIES.get('utm_campaign')
+
+    try:
+        utm_campaign = json.loads(urllib.parse.unquote(utm_campaign))
+    except Exception:
+        utm_campaign = None
```
mbeacom: Maybe we can plop all of these values into a dict here and return that versus all three unique vars? 🤔

unclebean: i just think they are same to use a dict or three vars. otherwise we want to store these values in one column in DB but it’s not good for analysis.

Issue #2362, opened 2018-10-05
Update Tips with Updated Auto Complete


PixelantDesign:
### User Story

Take what @thelostone-mc did on #2240 and implement it into the tips flow.

### Why Is this Needed

We want funders to understand where their bounty funds are going.

Display associated wallet address:
<img width="242" alt="screen shot 2018-10-05 at 3 16 03 pm" src="https://user-images.githubusercontent.com/238235/46555540-a94d9280-c8b1-11e8-896a-baf941da18ce.png">

Display secure messaging on users with no wallet address:
<img width="312" alt="screen shot 2018-10-05 at 3 16 58 pm" src="https://user-images.githubusercontent.com/238235/46555573-bd918f80-c8b1-11e8-903a-44b764a7d05a.png">

### Definition of Done

- [ ] Follow contributor guidelines outlined in:  https://docs.gitcoin.co/mk_contributors/ and https://docs.gitcoin.co/mk_styleguide/
- [ ] Implement updated type down with relevant address pre-fill from #2240 
- [ ] Test it locally
- [ ] Review with core team members @thelostone-mc @mbeacom @PixelantDesign 
- [ ] Doesn't **break** existing functionality
- [ ] Travis, Codecov, and Stickler CI are passing
- [ ] PR is merged

https://gitcoin.co/tip

thelostone-mc:
Ah I ended up throwing it after talking with @owocki  😓  Did not know we had this out for hacktoberfest 

https://github.com/gitcoinco/web/pull/2372

owocki:
i'm so so so happy with this.. thanks for the design / coding to both of you

PR #2372, opened 2018-10-07
tip : autocomplete shows payout address

thelostone-mc:
##### Description

x

@octavioamu I removed the default function call which happens within `user-search.js` and whichever page needs -> needs to individually invoke the function. 

If you want only the name invoke function as `userSearch('element-identifier', false);`
If you want the set payout address also -> `userSearch('element-identifier', true);`

Fixes: https://github.com/gitcoinco/web/issues/2362


octavioamu:
Late but lgtm :)

PR #2365, opened 2018-10-05
Fixing advanced payout - Default to Bountied Amount for Payment, etc.

Anish-Agnihotri:
##### Description

This PR resolves #2294 by introducing: 

1. Alerts upon advanced payment > 100%.
2. Automatic table altering upon X click. 
3. Other misc. payout preview cleaning (removing mentions of bounty stake, etc.).

[See it in action!](https://streamable.com/dirw9)

##### Checklist

- [x] Remove Return Bounty Stake Checkbox
- [x] Default all payments to pay from the Initial Funded Bounty Amount
- [x] Add an alert (bg color #E66700, #FFF for copy) when user exceeds 100% or Bountied value in ETH.
- [x] Add wallet information and overage amount to the alert.
- [x] Display the overage amount on the payout preview in #E66700
- [x] Remove refunded Eth
- [x] Remove mention of bounty stake
- [x] If multiple users have been added to the pay list, and a user is removed, the payout preview should reflect the removal.
- [x] This issue is tested

##### Affected core subsystem(s)

UI is the only core system affected, since all changes are aesthetic. 

##### Testing

Tested across two environments across two platforms (W10, Mojave). Feel free to test yourself as well.

##### Refers/Fixes

Fixes #2294.

PR #2386, opened 2018-10-09
Adding overage alert to advanced payout.

### LINKED ISSUE #2294: Advanced Pay Fix - Default to Bountied Amount for Payment, Add Guidance on Overage
### User Story

As a funder I'd like the advanced pay feature to be more intuitive and provide guidance on where funds are coming from.

### Description

Update the build so that:
- All funds come from the bountied amount.
- When overage is detected, an alert message notifies the user of the difference and where the funds will be paid from. This is not meant to be an error, but more of a warning.
- Clean up the Payout preview (remove confusing terms and unecessary information).

### Definition of Done

- [ ] Remove Return Bounty Stake Checkbox
- [ ] Default all payments to pay from the Initial Funded Bounty Amount

**Alert**
- [ ] Add an alert (bg color #E66700, #FFF for copy) when user exceeds 100% or Bountied value in ETH.
- [ ] Add wallet information and overage amount to the alert.

**Payout Preview**
- [ ] Display the overage amount on the payout preview in #E66700
- [ ] Remove refunded Eth
- [ ] Remove mention of bounty stake

**Other**
- [ ] If multiple users have been added to the pay list, and a user is removed, the payout preview should reflect the removal.

### Related Issues

This ticket is part of a series of fixes for Advanced Pay
- **[Issue 2215](https://github.com/gitcoinco/web/issues/2215)** Display Tooltip **Where is My Money Going** Bountied to XXX
- Default to Bountied Amount for Payment
- **[Issue 2240](https://github.com/gitcoinco/web/issues/2240)** Add Auto Complete & Make it clearer who I'm paying (verified/unverified)

### Design 
screen shot 2018-09-28 at 9 32 36 am

Anish-Agnihotri:
##### Description

This PR resolves #2294 by introducing: 

1. Alerts upon advanced payment > 100%.
2. Other misc. payout preview cleaning (removing refunded amount in payout preview).

[See it in action! Before new changes.](https://streamable.com/dirw9)

A lot of the changes that were made in the [last](https://github.com/gitcoinco/web/pull/2365) PR were addressed by @thelostone-mc in his PR introducing user profiles. Thus, code that I had previously written for that purpose (removing users, bounty stake, etc.) has been removed from this revised PR.

##### Checklist

- [x] Add an alert (bg color #E66700, #FFF for copy) when user exceeds 100% or Bountied value in ETH.
- [x] Add wallet information and overage amount to the alert.
- [x] Display the overage amount on the payout preview in #E66700
- [x] Remove refunded Eth
- [x] This issue is tested

UI is the only core system affected, since all changes are aesthetic. 

##### Testing

Tested across two environments across two platforms (W10, Mojave). Feel free to test yourself as well.

Currently, this PR is not functional since the Payout Preview paid and net values are broken (cc @thelostone-mc). Once that is functioning again this PR will be fully functional. 

##### Refers/Fixes

Fixes #2294.

### DIFF
--- a/app/assets/v2/js/pages/bulk_payout.js
+++ b/app/assets/v2/js/pages/bulk_payout.js
@@ -209,9 +209,23 @@ var update_registry = function() {
   var denomination = $('#token_name').text();
   var original_amount = $('#original_amount').val();
   var net = round(original_amount - tc, 2);
+  var over = round((original_amount - get_total_cost()) * -1, 4);
+  var addr = web3.eth.coinbase.substring(38);
 
   $('#total_cost').html(tc + ' ' + denomination);
   $('#total_net').html(net + ' ' + denomination);
+  $('#total_overage').html(over + ' ' + denomination);
+  $('#address_ending').html(addr + ' ');
+  $('#preview_ending').html(addr + ' ');
+  $('#preview_overage').html(over + ' ' + denomination);
+
+  if (over > 0) {
+    $('.overageAlert').css('display', 'inline-block');
+    $('.overagePreview').css('display', 'inline-block');
+  } else {
+    $('.overageAlert').css('display', 'none');
+    $('.overagePreview').css('display', 'none');
+  }
 
   let transactions = [];
 
--- a/app/dashboard/templates/bulk_payout_bounty.html
+++ b/app/dashboard/templates/bulk_payout_bounty.html
@@ -99,6 +99,14 @@
                   <input type="hidden" id="original_amount" value="{{ bounty.value_true }}">
                   <span id="token_name">{{ bounty.token_name }}</span>
                 </div>
+                <div class="w-100 overageAlert">
+                  <div>
+                    <img src="{% static "v2/images/warning.svg" %}">
+                  </div>
+                  <div>
+                    <p>{% trans "You have exceeded the bounty's funded amount. The difference of " %}<span id="total_overage"></span> {% trans "will be paid from your metamask wallet ending in ..." %}<span id="address_ending"></span></p>
+                  </div>
+                </div>
                 <div class="w-100 my-3">
                   <label for=bountyFulfillment>{% trans "Payout" %}</label>
                   <label id="tooltip">{% trans "Where is my Eth going? " %}<i class='fa fa-info-circle'></i></label>
@@ -108,7 +116,7 @@
                     </div>
                     <div class="mt-2" id="trans_preview">
                       <h5>{% trans 'Payout Preview' %}</h5>
-                      <p class="mb-0">{% trans 'Refunded' %}: <span id="total_returned">{{ bounty.value_true }} {{ bounty.token_name }}</span></p>
+                      <p class="mb-0 overagePreview"><strong><span id="preview_overage"></span>{% trans ' to be paid from wallet ending in ...' %}<span id="preview_ending"></span></strong></p>
                       <p class="mb-0">{% trans 'Paid' %}: -<span id="total_cost"></span></p>
                       <p class="mb-0">{% trans 'Net' %}: <span id="total_net"></span></p>
                       <p class="mt-1 mb-0">{% trans 'Transactions:' %}</p>
@@ -215,7 +223,36 @@
       .entry.active{
         background-color: #bbb;
       }
-
+      .overageAlert {
+        background-color: #E66700;
+        color: #FFF;
+        border-radius: 5px;
+        margin-top: 15px;
+        margin-bottom: 5px;
+        display: none;
+        padding-left: 15px;
+      }
+      .overageAlert > div:first-child {
+        width: 25px;
+        float: left;
+        padding-right: 10px;
+      }
+      .overageAlert > div:last-child {
+        width: calc(100% - 40px);
+        float: left;
+        margin-left: 15px;
+        padding-right: 15px;
+        padding-top: 6.5px;
+      }
+      .overageAlert > div:first-child > img {
+        width: 17px;
+        margin-top: 10px;
+      }
+      .overagePreview {
+        color: #E66700;
+        padding-bottom: 5px;
+        display: none;
+      }
     </style>
     {% include 'shared/bottom_notification.html' %}
     {% include 'shared/analytics.html' %}


Anish-Agnihotri:
CC @pinkiebell, @thelostone-mc, @mbeacom, @SaptakS, @PixelantDesign for re-review. 

Thanks :).

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -162,7 +162,7 @@ $(document).ready(function($) {
       <tr>
         <td class="pl-0 pb-0">
           <div class="pl-0">
-            <select id="username" onchange="update_registry()" class="username-search custom-select" style="width: 100%; margin-left: auto; margin-right: auto;"></select>
+            <select onchange="update_registry()" class="username-search custom-select" style="width: 100%; margin-left: auto; margin-right: auto;"></select>
```
thelostone-mc: thanks man ^_^

Anish-Agnihotri:
Just pushed out fix. Codecov error is not due to this edit. 

Let me know if it's good to go :).

mbeacom:
:star: PR built and available in a preview environment **gitcoinco-web-pr-2386** [here](http://web.jx-gitcoinco-web-pr-2386.ci.gitcoin.co)

app/dashboard/templates/bulk_payout_bounty.html
```diff
@@ -215,7 +223,36 @@ <h5>{% trans 'Payout Preview' %}</h5>
...
+      }
+      .overageAlert > div:first-child > img {
+        width: 17px;
+        margin-top: 10px;
+      }
+      .overagePreview {
+        color: #E66700;
+        padding-bottom: 5px;
+        display: none;
+      }
```
thelostone-mc: Looks like this is piling up @Anish-Agnihotri 
Could we push all the css into a new file  ? ^_^

app/dashboard/templates/bulk_payout_bounty.html
```diff
@@ -215,7 +223,36 @@ <h5>{% trans 'Payout Preview' %}</h5>
       .entry.active{
         background-color: #bbb;
       }
-
+      .overageAlert {
```
SaptakS: Hi @Anish-Agnihotri. Can you please move these styles and the other existing styles to a separate`.css` file? Else it's gonna add a lot of tech-debt later. Thanks in advance. :)

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -209,9 +209,23 @@ var update_registry = function() {
   var denomination = $('#token_name').text();
   var original_amount = $('#original_amount').val();
   var net = round(original_amount - tc, 2);
+  var over = round((original_amount - get_total_cost()) * -1, 4);
+  var addr = web3.eth.coinbase.substring(38);
 
   $('#total_cost').html(tc + ' ' + denomination);
   $('#total_net').html(net + ' ' + denomination);
+  $('#total_overage').html(over + ' ' + denomination);
```
SaptakS: This section I feel is unnecessary in case `over < 0` right? Since it will anyways be hidden. Can't we put this inside the condition check to optimize the code?

app/dashboard/templates/bulk_payout_bounty.html
Anish-Agnihotri: For sure, makes sense!

app/dashboard/templates/bulk_payout_bounty.html
Anish-Agnihotri: For sure, @thelostone-mc suggested the same :).

app/assets/v2/js/pages/bulk_payout.js
Anish-Agnihotri: That would be correct and it makes absolute sense to do so. I've restructured it in the new commit!

Anish-Agnihotri:
@thelostone-mc @SaptakS 

Feel free to re-review when you have time! Thanks :).

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -185,12 +185,12 @@ $(document).ready(function($) {
 });
 
 var get_total_cost = function() {
-  var num_rows = $('#payout_table tbody').find('tr').length;
+  var num_rows = $('#payout_table').find('tr').length;
```
thelostone-mc: @SaptakS @Anish-Agnihotri so would it make sense to keep `tbody` and set `i = 0` ?  Or meh not worth it  ?

`$('#payout_table')` -> first `tr` would be `thead`'s row 
`$('#payout_table tbody') ` -> first `tr` would be the row we need

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -185,12 +185,12 @@ $(document).ready(function($) {
...
   var total = 0;
   var i = 1;
```
thelostone-mc: do we need this ?

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -185,12 +185,12 @@ $(document).ready(function($) {
...
 
   for (i = 1; i < num_rows; i += 1) {
```
thelostone-mc: change to `for( let i = 0` ?

- Init `i` to 0 cause of [this](https://github.com/gitcoinco/web/pull/2386/files#r224689252) 
- declaring variable here as opposed to [L190](https://github.com/gitcoinco/web/pull/2386/files#r224689394)

Anish-Agnihotri:
@thelostone-mc 

Although I've restructured the commits to include `tbody`, I'm not really happy with where it went. `nth-child()` is a function that doesn't follow the usual `[0, 1, 2, 3]` scheme but rather `[1, 2, 3, 4]`. 

Because of that, simply changing the i variable to 0 would not work, since `nth-child(0)` would not work. But, if I made set `i = 1`, although `nth-child(1)` would work, it would break the function since `i < num_rows` would no longer be valid. 

The only way to fix this issue was to append `+1` to the `num_rows` variable, which would then allow you to use `i = 1` and receive `nth-child(1)`, `nth-child(2)`, and so on (since in this case `num_rows` would always be one more than `i`). 

Of course, this might actually be a more complicated solution than simply not using `tbody` in the first place.

I hope @SaptakS or you have some better idea at how to combat this, since this doesn't seem to be the easiest way to do this. Possibly my JS is just all over the place?

thelostone-mc:
Ah 😓 meh ! I'm good with what we have if that's the case

> nth-child() is a function that doesn't follow the usual [0, 1, 2, 3] scheme but rather [1, 2, 3, 4].

did now know about that 😅

SaptakS:
@Anish-Agnihotri can't you just do `i <= num_rows`? That way you don't have to hard code the +1 to the number of rows.
Secondly, I don't think nth-child is even needed. You can get all the rows by `var rows = $('#payout_table tbody').find('tr')`. Then you can just loop like `rows.forEach(function(row) { // code here });` . That way you can completely avoid using nth-child. Using nth-child is over engineering I feel.

Anish-Agnihotri:
Lol, I was completely overthinking myself. Thanks @SaptakS. I've pushed out the new (hopefully ready to merge, now) commit.

I realized that there was a lot of over-engineering and removed the need for `num_rows`, `i`, and the `nth-child` altogether. Now we are just going through a simple function that runs on `each` instance of `#payout_table tbody tr`.

Do you think this is a better idea?

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -185,21 +185,20 @@ $(document).ready(function($) {
...
+    var username = $(this).find('.username-search').text();
     var is_error = !$.isNumeric(amount) || amount <= 0 || username == '' || username == '@';
 
     if (!is_error) {
       total += amount;
     }
-  }
+  });
+
   return total;
```
SaptakS: @Anish-Agnihotri okay. Sorry but `.each()` function won't be very helpful here because callback functions will be executed asynchronously, and hence when you return total outside, no guarantee that it will give the total.

Something like this will be good.
```js
var rows = $('#payout_table tbody tr');
for (i = 0; i < rows.length; i += `) {
    var $rows = $(rows[i]);
    var amount = parseFloat($row.find('.amount').text());
    // and all the other code
}
return total;
```

Anish-Agnihotri: Ah, yes, async execution. Thanks for the reviews, always a good learning experience. Pushing it out according to your recommendation now.

app/assets/v2/css/bulk_payout_bounty.css
```diff
@@ -0,0 +1,122 @@
...
+  background-color: #E66700;
+  color: #FFF;
+  border-radius: 5px;
+  margin-top: 15px;
+  margin-bottom: 5px;
+  display: none;
+  padding-left: 15px;
+}
+
+.overageAlert>div:first-child {
```
pinkiebell: Cosmetic / Space to breath 😄 
`.overageAlert > div:first-child`

app/assets/v2/css/bulk_payout_bounty.css
```diff
@@ -0,0 +1,122 @@
...
+  width: 25px;
+  float: left;
+  padding-right: 10px;
+}
+
+.overageAlert>div:last-child {
```
pinkiebell: Same ^

app/assets/v2/css/bulk_payout_bounty.css
```diff
@@ -0,0 +1,122 @@
...
+  width: calc(100% - 40px);
+  float: left;
+  margin-left: 15px;
+  padding-right: 15px;
+  padding-top: 6.5px;
+}
+
+.overageAlert>div:first-child>img {
```
pinkiebell: Same ^

app/assets/v2/js/pages/bulk_payout.js
pinkiebell: Add `w-100 ml-auto mr-auto` to the class attribute and we can get rid of the `style=...` attribute. :)

app/assets/v2/js/pages/bulk_payout.js
```diff
@@ -185,14 +185,13 @@ $(document).ready(function($) {
...
-  var num_rows = $('#payout_table tbody').find('tr').length;
+  var rows = $('#payout_table tbody tr');
   var total = 0;
-  var i = 1;
 
-  for (i = 1; i < num_rows; i += 1) {
-    var $row = $('#payout_table tobdy').find('tr:nth-child(' + i + ')');
-    var amount = parseFloat($row.find('.amount').text());
-    var username = $row.find('.username-search').text();
+  for (i = 0; i < rows.length; i += 1) {
```
pinkiebell: Maybe we want to use let / ES5/6 style ? cc @thelostone-mc @SaptakS 
Like:
```
for (let i = 1; i < num_rows; i++) {
  let row = rows[i]; // XXX: Do we need $(rows[i])?
```

Anish-Agnihotri: I think Gitcoin is moving towards ES6 styling if I remember correctly, so I'd be glad to switch these over.

Anish-Agnihotri:
Thanks @pinkiebell for the review 🙌 ! Just pushed out new commit.

Anish

app/assets/v2/js/pages/bulk_payout.js
SaptakS: It's okay for now. We will anyways do a complete migration from old-school to ES6 soon.

Anish-Agnihotri:
🎉 Perfect! We should be good for merge then :)

app/assets/v2/js/pages/bulk_payout.js
arun1595: @SaptakS @Anish-Agnihotri Hey guys, chiming in to clarify something.

> Sorry but .each() function won't be very helpful here because callback functions will be executed asynchronously,

jQuery `each` function is _synchronous_. You can verify this by adding a `console.log(total)` after the `each` function call. You can also refer to [this](https://stackoverflow.com/questions/7371942/is-jquery-each-function-synchronous) SO thread for more information.

PR #2387, opened 2018-10-09
concept - quickfund an issue

owocki:
##### Description

quickfunds an issue.. only shows advance fund options IFF user requests them http://bits.owocki.com/c00d3edc1db2/Screen%20Recording%202018-10-08%20at%2008.32%20PM.mov

##### Checklist

funding 
##### Testing

tested it 
##### Refers/Fixes

self

### DIFF
--- a/app/assets/v2/css/submit_bounty.css
+++ b/app/assets/v2/css/submit_bounty.css
@@ -133,3 +133,7 @@ textarea:read-only {
   display: inline;
   margin-right: .5rem;
 }
+
+.quickfund .hide_if_quickfund{
+  display: none;
+}
--- a/app/assets/v2/css/submit_bounty.css
+++ b/app/assets/v2/css/submit_bounty.css
@@ -135,5 +135,5 @@ textarea:read-only {
 }
 
 .quickfund .hide_if_quickfund{
-  display: none;
+  display: none !important;
 }
--- a/app/assets/v2/js/pages/new_bounty.js
+++ b/app/assets/v2/js/pages/new_bounty.js
@@ -52,7 +52,6 @@ $('#sync-issue').on('click', function(event) {
 });
 
 
-
 $('#issueURL').focusout(function() {
   setInterval(function() {
     $('#last-synced span').html(timeDifference(new Date(), new_bounty.last_sync));
@@ -89,14 +88,14 @@ $(document).ready(function() {
 
 
   $('#switch_quickfund').click(function(event) {
-    if($("#primary_form").hasClass('quickfund')){
-      $("#primary_form").removeClass('quickfund');
-      $(".text-center.title").text(gettext("Fund Issue"));
-      $(this).text(gettext("Switch to QuickFund"));
+    if ($('#primary_form').hasClass('quickfund')) {
+      $('#primary_form').removeClass('quickfund');
+      $('.text-center.title').text(gettext('Fund Issue'));
+      $(this).text(gettext('Switch to QuickFund'));
     } else {
-      $("#primary_form").addClass('quickfund');
-      $(".text-center.title").text(gettext("QuickFund Issue"));
-      $(this).text(gettext("Switch to Advanced Fund"));
+      $('#primary_form').addClass('quickfund');
+      $('.text-center.title').text(gettext('QuickFund Issue'));
+      $(this).text(gettext('Switch to Advanced Fund'));
     }
     event.preventDefault();
   });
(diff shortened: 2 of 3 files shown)


PixelantDesign:
Looking good!

Could we default to the current funding form and add a link to Quick Fund under the title? 

For the link to switch to quicksand, could we use the words _Normal Funding Form_ instead of _Advanced Funding Form_?

I still think most of our use case will be funders will still need to review project types and think about the details prior to funding an issue.

PixelantDesign:
For the Quickfund view - could we list somewhere what all the defaults are, even though users don't need to make selections?

owocki:
> Could we default to the current funding form and add a link to Quick Fund under the title?

I was kind of hoping we would default to the fast / easy funder form and only go to 'advanced' if user required it.  kind of like how you go to google.com and just get the easy small UX, and then can click 'advanced search' 

> For the link to switch to quicksand, could we use the words Normal Funding Form instead of Advanced Funding Form?

yes

> I still think most of our use case will be funders will still need to review project types and think about the details prior to funding an issue.

i'll be curious to see the breakdown of both.  i dont have a good intuition on what people would use the most, but i do greatly like the idea of making it faster/easier to fund issues, so i hope quickfund would get a bunch of use

> For the Quickfund view - could we list somewhere (in small plain text) what all the defaults are (work scheme, tags, time commitment), even though users don't need to make selections?

i wonder if a middle ground here would be being able to view each of these small defaults and edit them inline on the quickfund form... hmmm

thelostone-mc:
I like it.Only thing is as a new funder personally I may not click on advanced  funding
would it make sense to use the empty space on the right side to display what extra can be done if they switch to advance/make it more prominent? 
Not sure if that's the right thing to do

cc my design man @willsputra

PixelantDesign:
>  i dont have a good intuition on what people would use the most, but i do greatly like the idea of making it faster/easier to fund issues, so i hope quickfund would get a bunch of use.

Who are we optimizing for?

PixelantDesign:
@frankchen07 what do you think of the above? Is there anything we can look at in terms of what users are selecting the most as a project type? Perhaps we default to that?

frankchen07:
This is interesting. Generally, I'm more of a minimalist, so the simpler the better. I do have some thoughts. 

If I were a funder, I would want a say to the issue details that I'm putting out, so I like the idea of showing defaults on the quick fund form. This allows user control. However, the purpose of a quick fund is to reduce friction, so perhaps the important default items pertinent to the issue should pop up only after the issue and pricing fields are finished (this way they don't see a ton of things to input before they even start).

One thing we should exercise caution with in regards to quick funding is the email address. I think it's prudent to ensure that the user has that choice to be more protective of their information regardless if it's quick or advanced. 

So I see quick fund looking like this:

1. User comes to the quick fund page.
2. User inputs the Github issue link and syncs it.
3. User inputs price.
4. Project type, permissions, issue details, contact information, and contact information sharing option pop up with recommended defaults (respectively: traditional project type, permissionless permission, intermediate skill level, hours time commitment, feature bounty type, email pre-filled, willing to share information == True).
5. Ensure that the user has the option to visit all fields that are modifiable post-submitting to the blockchain. 

screen shot 2018-10-19 at 14 46 18

PixelantDesign:
Thanks @frankchen07.

I agree with Franks recommendations on reordering fields and suggested defaults displaying below the price section based on the data.

We do get keywords/tags from the repo as well. Will tags still come in as part of the github issue info or populate after ? @willsputra is doing some thinking on how we can encourage more accurate tag display on the issue explorer and usage since language is a big determining factor on a contributor taking on an open issue.

Frank should we also reorder components on the full form for consistency?

cc @mbeacom

frankchen07:
I'd say the ability to edit tags that are pulled in from the Github issue should pop up only after the issue and pricing has been inputted. Those two actions (especially price) are indecision points. 

Everything that pops up after seems like fact-filling (types, difficulties, tags, etc).

It makes sense for the advanced form to have a similar layout to the quick fund layout. I'm undecided and unsure about the pop-up feature for the advanced form though. 

Also open to other suggestions above. I took a look at HotJar to see if we had a high-level overview of the funder form, but I do not think we're tracking that right now.

PixelantDesign:
Could we hold off on this for right now? This is very related to the tagging issue that is in cue. I also think it’s worth updating based on other recommendations above.

mbeacom:
So, closing this for now?

PixelantDesign:
I'd like to leave this open if it's alright @mbeacom

thelostone-mc:
Closing this out now as it's been built out

PR #2418, opened 2018-10-11
Create Grant, Subscription & Approve functionality

captnseagraves:
##### Description

Adding in web3 functionality (without tokensubminer)

##### Checklist

- grants


app/grants/templates/grants/fund.html
```diff
@@ -138,4 +135,7 @@ <h2 class="fund__header">
 <script src="{% static "v2/js/shared.js" %}"></script>
 <script src="{% static "v2/js/ipfs-api.js" %}"></script>
 <script src="{% static "v2/js/ipfs.js" %}"></script>
+
```
mbeacom: Can we drop these new lines?

app/grants/templates/grants/index.html
```diff
@@ -47,9 +47,7 @@ <h1 class="grants-hero__title mb-2">{% trans 'Dev Grants' %}</h1>
           support for some of the best OSS projects in the space.
           {% endblocktrans %}
         </p>
-        <a class="button button--primary font-weight-bold" href="{% url 'grants:new' %}">
-          {% trans 'Create a Grant' %}
```
mbeacom: Do we no longer want the `Create a Grant` button here?

app/grants/views.py
```diff
@@ -62,6 +62,9 @@ def grant_details(request, grant_id):
     except Grant.DoesNotExist:
         raise Http404
 
+    print("this is the web3 instance", w3)
```
mbeacom: Can you omit the new prints throughout this file?

app/grants/views.py
```diff
@@ -117,19 +120,23 @@ def grant_fund(request, grant_id):
...
-        # subscription.subscriptionHash = request.POST.get('input_name')
-        # subscription.contributorSignature = request.POST.get('description')
-        # Address will come from web3 instance
-        # subscription.contributorAddress = request.POST.get('reference_url')
+
+        print("it fired")
+        print(request.POST)
+
+
+        subscription.subscription_hash = request.POST.get('subscription_hash')
```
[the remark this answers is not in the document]
mbeacom: ^

mbeacom: I know they shouldn't be getting to a point where they are POSTing without these parameters, but can you default to empty string and/or check for `if not var:` for these?  I'm thinking it might make the most sense for us to migrate this view to use django forms like the one introduced in `grants/forms.py` so we don't have to validate all of these entries in the view itself.

app/grants/views.py
```diff
@@ -117,19 +120,23 @@ def grant_fund(request, grant_id):
...
-        # subscription.tokenAddress = request.POST.get('denomination')
+        subscription.token_address = request.POST.get('token_address')
         subscription.gas_price = request.POST.get('gas_price')
-        # network will come from web3 instance
-        # subscription.network = request.POST.get('amount_goal')
+        subscription.network = request.POST.get('network')
         subscription.contributor_profile = profile
         subscription.grant = grant
         subscription.save()
+        return redirect(f'/grants/{grant.pk}')
```
mbeacom: Can you use `reverse()` with the `redirect()` here, so if the route changes, the view doesn't need updated?

app/grants/views.py
```diff
@@ -148,9 +155,14 @@ def subscription_cancel(request, subscription_id):
     subscription = Subscription.objects.select_related('grant').get(pk=subscription_id)
     grant = getattr(subscription, 'grant', None)
 
+    print(subscription)
+    print(grant)
```
mbeacom: Can we remove these prints?

app/grants/views.py
```diff
@@ -148,9 +155,14 @@ def subscription_cancel(request, subscription_id):
...
+
     if request.method == 'POST':
         subscription.status = False
         subscription.save()
+        return redirect(f'/grants/{grant.pk}')
```
mbeacom: `reverse()` with `redirect()`

app/grants/templates/grants/fund.html
```diff
@@ -35,62 +35,57 @@ <h2 class="fund__header">
...
-        </div>
-      </div>
-
-      <div class="col-4">
-        <div class="grant__progress">
-          <div class="progress">
-            <div class="progress-bar" style="width:50%"></div>
-            <div class="progress-bar progress-bar--fund" role="progressbar" style="width:10%">
-              <div class="contribution">
-                <span>10 ETH</span>
```
mbeacom: Do we want to omit this?

app/grants/templates/grants/detail.html
```diff
@@ -30,8 +30,9 @@ <h1 class="grant__title">{{ grant.title }}</h1>
             </div>
           </div>
         <div class="col-6">
-          <button class="button button--primary button--full">Fund this Project</button>
-
+          <a href="{% url 'grants:fund' grant.id %}">
+            <button class="button button--primary button--full">Fund this Project</button>
```
mbeacom: Can you wrap this copy in `{% trans "copy" %}`

app/grants/models.py
```diff
@@ -69,7 +69,7 @@ class Subscription(SuperModel):
     amount_per_period = models.DecimalField(default=1, decimal_places=4, max_digits=50)
     token_address = models.CharField(max_length=255, default='0x0')
     gas_price = models.DecimalField(default=1, decimal_places=4, max_digits=50)
-    network = models.CharField(max_length=8, default='mainnet')
+    network = models.CharField(max_length=16, default='mainnet')
```
mbeacom: No troll - which network will be 16 characters?

app/grants/migrations/0003_auto_20181010_2148.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.2 on 2018-10-10 21:48
```
mbeacom: These migrations need to be deleted and recreated as a single migration at some point

app/assets/v2/js/grants/fund_grant.js
```diff
@@ -1,124 +1,149 @@
...
+                  let postData = {
+                    subscriptionContract: data.contract_address,
+                    parts: parts,
+                    subscriptionHash: subscriptionHash,
+                    signature: signature
+                  };
+
+                  console.log('postData', postData);
+
+                  fetch('http://localhost:10003/saveSubscription', {
```
mbeacom: This will need updated to take a uri fed in from the django app

app/grants/templates/grants/index.html
captnseagraves: We do. Not sure why it got deleted... Will add back in.

app/grants/models.py
captnseagraves: testrpc has 13. I gave an extra 3 for flex. Can reduce to 13. But was getting errors during dev.

captnseagraves: Although docs say it defaults to current unix time on start. We could specify a network id when using ganache.

app/grants/templates/grants/fund.html
captnseagraves: I dont think we do. Not sure exactly where it should be replaced.

app/grants/views.py
captnseagraves: Makes sense

app/grants/models.py
mbeacom: I thought we used `local` or `testrpc` for `network` on other objects on the platform right now.  I don't think we need the full network name for the local testrpc instance here do we?

captnseagraves: Sure, we can have it as `local` or `testrpc`. That works.

app/grants/templates/grants/fund.html
captnseagraves: Added back in

app/grants/models.py
captnseagraves: Reverted to 8 characters

app/assets/v2/js/grants/fund_grant.js
captnseagraves: Is there anything to execute on this right now?

captnseagraves:
@mark Travis-ci is failing on `npm install` and doesnt have any obvious errors in the stack trace. Any ideas on how to resolve this?

mbeacom:
@captnseagraves Can you resolve the conflicts here and I'll investigate the `npm install` issue on travis?

captnseagraves:
@mbeacom Conflicts resolved

captnseagraves:
@mbeacom just added `period_seconds` to the models and circleci seems to break on a sql query because of it. Do I need to run migrations to fix this error? 

Also, travis-ci is still breaking on `npm i` with no obvious error

package-lock.json
```diff
@@ -3461,11 +3909,44 @@
...
+          "resolved": "http://registry.npmjs.org/combined-stream/-/combined-stream-1.0.6.tgz",
+          "integrity": "sha1-cj599ugBrFYTETp+RFqbactjKBg=",
+          "requires": {
+            "delayed-stream": "~1.0.0"
+          }
+        }
+      }
+    },
+=======
+>>>>>>> grants
```
mbeacom: Bad rebase

PR #2457, opened 2018-10-13
bounty template layout adjustments

pinkiebell:
### Description
Adjusts the layout for bounty rows and refactor the template a bit.


thelostone-mc:
@pinkiebell could you throw in a before and after screenshot ?

pinkiebell:
I rebased & pushed another version of this.
Layout is close to the original except for the removed background color and the added flex-wrap.
Before that change, there was a possibility that the funding tags got clipped' off.

Before
localhost_8000_explorer_network mainnet order_by null

After
localhost_8000_explorer_network mainnet order_by null 1
localhost_8000_explorer_network mainnet order_by null 2

PR #2511, opened 2018-10-22
Cancel view clean up

captnseagraves:
Cancel view clean up


app/grants/views.py
```diff
@@ -35,6 +36,11 @@
 from marketing.models import Keyword
 from web3 import HTTPProvider, Web3
 
+now = datetime.datetime.now()
```
mbeacom: Can we move this into the method itself? @captnseagraves

captnseagraves: sure

captnseagraves:
Comments addressed here: https://github.com/gitcoinco/web/pull/2513

PR #2517, opened 2018-10-23
grants/new: add any token option

thelostone-mc:
##### Description

screen shot 2018-10-23 at 10 38 06 pm

@captnseagraves is this what you were hoping for ?


captnseagraves:
@thelostone-mc Yes! We need to find a way to take ETH off as well though

captnseagraves:
LGTM

mbeacom:
@captnseagraves @thelostone-mc As discussed previously, we will want to probably update the `tokens` BE method with an optional arg/kwarg to handle entry for grants.  If grants, set choice of `any` to `0x0...`.  This would eliminate the need for this change.

Issue #2483, opened 2018-10-17
Stop Work button even if request is not sent


santteegt:
**Describe the bug**
Stop work button appears event if the request to work is not sent

**To Reproduce**
A user with all three working issue slots assigned tries to send a request to start working on a new bounty. An error of "max of 3 issues at once" appear but the Stop Work button is shown.

**Expected behavior**
[comment]: # (A clear and concise description of what you expected to happen.)
Start work should be in place

**Screenshots**
![out](https://user-images.githubusercontent.com/926341/47056128-86a85d00-d180-11e8-82f2-62ec4ecb736c.gif)

**Desktop (please complete the following information):**

- OS: MacOs High Sierra 10.13.6
- Browser Google Chrome
- Browser Version 69.0.3497.100

PR #2565, opened 2018-10-30
fix 'stop work' button when the request is not sent

jvmaia:
##### Checklist

ui

##### Testing

tested the starting/stoping work.
##### Refers/Fixes
Fixes: #2483


octavioamu:
Thanks for helping
Things have changed a little in master now is just changing the buttons on success as promise also because we now have "cancel reason".
I think we can close this PR for now. 
CC @thelostone-mc

thelostone-mc:
Thanks @jvmaia ! Appreciate this but will be closing this one out as this use-case is now avoided 😅

Issue #2586, opened 2018-10-31
bug - kudos tx ran out of gas


owocki:
https://etherscan.io/tx/0x063b85c9e070e57ede8cfdd8f92cd8841c59d0ef0d6fecd8d58c64227ddc9228
kudos to username: TimDaub
original report: https://twitter.com/TimDaub/status/1057353311655993344

success criteria:
1. identify and fix the root cause of this issue.
2. communicate with user to let them know its fixed and to try again (you might have to send them more gas)

octavioamu:
Not sure what happens but also the data of his kudo is messy
https://gitcoin.co/profile/TimDaub
![image](https://user-images.githubusercontent.com/1358093/47791580-c763b280-dcf8-11e8-942c-0143dc77c97b.png)

kuhnchris:
I think that the gas calc is messed up to begin with, not just kudos.
my default gas was 2 on kudos, but I barely get anything verified with 2 anymore. 2.5wei seems to be the minimum right now. The calculator may simply be screwed.

octavioamu:
#2570

jasonrhaas:
Do we know if this is a Kudos Direct Send or Kudos Indirect Send?  Ie, if the Kudos needed to be "received".  The link on etherscan.io says the transaction is from https://etherscan.io/address/0xa5a3835689008041296d2cd3f94c307e05f82630

Looking at the transactions, it looks like it is probably the proxy account, which means Indirect Send.  Hard to be sure  without looking at the database directly though.

jasonrhaas:
OK so I added a 10% ETH buffer for the proxy account to add some extra gas money for when the Kudos actually gets minted on the receiving end.

Before buffer:
<img width="359" alt="screen shot 2018-10-31 at 4 47 48 pm" src="https://user-images.githubusercontent.com/7018237/47820993-2b07d300-dd2d-11e8-8e78-d574089427b0.png">


After adding a 10% buffer:
<img width="358" alt="screen shot 2018-10-31 at 4 50 35 pm" src="https://user-images.githubusercontent.com/7018237/47820999-34913b00-dd2d-11e8-83eb-297116b3013d.png">

This means that the upfront cost for Kudos Indirect Send will be a bit higher.  We also need to make sure that in the *kudos_receive.js* that we are returning any leftover money to the original Kudos sender.

owocki:
https://github.com/gitcoinco/web/pull/2656

owocki:
[...]

PR #2656, opened 2018-10-31
fix for tx's running out of gas

owocki:
gas limit fix.  this is a fix for 

https://github.com/gitcoinco/web/issues/2586

it needs to be tested before deployment.  i can't do that currently bc i have bad internet..

### DIFF
--- a/app/assets/v2/js/pages/kudos_receive.js
+++ b/app/assets/v2/js/pages/kudos_receive.js
@@ -155,7 +155,7 @@ $(document).ready(function() {
  
         kudos_contract.clone.estimateGas(forwarding_address, tokenId, numClones, {from: holding_address, value: kudosPriceInWei}, function(error, gasLimit) {
           console.log(gasLimit);
-          var buffer = new web3.BigNumber(0);
+          var buffer = new web3.BigNumber(100000);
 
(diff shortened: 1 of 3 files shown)


mbeacom:
Seeing an [out of gas error](https://rinkeby.etherscan.io/tx/0x29f590258542caefdb70a4930d001ba8860c9ca6e25c3f4bf4b2a5d7d1ddfb23) when using this on rinkeby

jasonrhaas:
> Seeing an [out of gas error](https://rinkeby.etherscan.io/tx/0x29f590258542caefdb70a4930d001ba8860c9ca6e25c3f4bf4b2a5d7d1ddfb23) when using this on rinkeby

@mbeacom this was a Kudos Indirect Send right?

owocki:
@mbeacom https://github.com/gitcoinco/web/pull/2656/commits/f93b8345f4daa4fb768716b8c6f8273c0074afce should fix that

jasonrhaas:
Doesn't this line https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/pages/kudos_send.js#L508

already account for the Kudos gasLimit?  It seems like we are sending the gas money twice.  I suppose it's better to have extra gas in there anyway.  We just need to make sure that we return the extra gas money leftover.

PR #2657, opened 2018-10-31
Adjust lazy_load_kudos to catch nonexistent profile

### LINKED ISSUE #2655: Don't assume a profile will exist without catching with try/except
https://sentry.gitcoin.co/gitcoin/gitcoin-prod-backend/issues/1719/

```
DoesNotExist: Profile matching query does not exist.
(3 additional frame(s) were not displayed)
...
  File "dashboard/views.py", line 1211, in lazy_load_kudos
    profile = Profile.objects.get(handle=handle)
  File "cacheops/query.py", line 354, in get
    return qs._no_monkey.get(qs, *args, **kwargs)

DoesNotExist: Profile matching query does not exist.
```

mbeacom:
##### Description

The goal of this PR is to resolve the exception encountered in:  #2655 when attempting to query for a profile that doesn't exist in `lazy_load_kudos`

##### Checklist

##### Testing

Local

##### Refers/Fixes

Fix #2655

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -1202,22 +1202,25 @@ def profile(request, handle):
 @csrf_exempt
 def lazy_load_kudos(request):
     page = request.POST.get('page', 1)
-    address = request.POST.get('address')
     context = {}
     datarequest = request.POST.get('request')
     order_by = request.GET.get('order_by', '-modified_on')
     limit = int(request.GET.get('limit', 8))
     handle = request.POST.get('handle')
-    profile = Profile.objects.get(handle=handle)
-    if datarequest == 'mykudos':
-        key = 'kudos'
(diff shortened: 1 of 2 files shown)


octavioamu:
LGTM , tested in my local

app/dashboard/views.py
```diff
@@ -1202,22 +1202,25 @@ def profile(request, handle):
 @csrf_exempt
 def lazy_load_kudos(request):
     page = request.POST.get('page', 1)
-    address = request.POST.get('address')
     context = {}
     datarequest = request.POST.get('request')
     order_by = request.GET.get('order_by', '-modified_on')
```
octavioamu: looks like the order was removed from the front end so if we are not going to have it there is not need to have it in the request

mbeacom: I'll remove it in a follow up. Thanks for the heads up!

Issue #2663, opened 2018-11-01
Track status on KudosTransfer objects


mbeacom:
`KudosTransfer` should carry a `status` field or something similar acting as an indicator of `Token` status onchain.

Prompted by:

```
jhaas [14 hours ago]
Part of the problem is that we are just associating the Kudos objects with the profile, disregarding anything on the blockchain.  Will lead to confusion.  Also the problem with using KudosTransfer is that its not a real kudos.  Could have a scenario where the transaction fails, so the kudos would show up in the profile, but not exist on chain.

One possibility is showing the KudosTransfers as “pending” Kudos, so they can see the kudos right away.  But then when the kudos actually mines + syncs, replace that with the real kudos from the Token table.
```

owocki:
https://github.com/gitcoinco/web/pull/2673

PR #2673, opened 2018-11-02
lookup and sync Kudos/Tips txids success statuses

owocki:
##### Description

lookup and sync Kudos/Tips txids success statuses

also:

* Threads through tx status success to a bunch of diff places
* Updates analytics

##### Checklist

Kudos/Tips

##### Testing

I ran the `update_tx_status` management command.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/2663

##### TODOs

* If a Tip/Kudos drops/fails, We should notify the sender
* Perhaps we only trigger 'you have a new X' emails after the tx clears?  
* what else?


owocki:
@mbeacom particularly interested in your thoughts on this:
```
What do we want to do with this information? Perhaps we only trigger 'you have a new X' emails after the tx clears?
If a Tip/Kudos drops/fails, We should notify the sender
There are probably points in our analytics where we should only include tx's that succeeeded.
what else?
```

owocki:
@mbeacom any thoughts on the todos listed in the PR desc?

owocki:
@mbeacom ^^

owocki:
@frankchen07 when this goes live, youll ahve to check not only for a non falsy `txid`, but also that tx_status == 'success'

owocki:
@mbeacom can i get an 👀

Issue #2538, opened 2018-10-25
As a contributor, I'd like to see the correct time estimate on a Gitcoin Issue.


frankchen07:
### User Story

As a contributor, I'd like to see the correct time estimate on a Gitcoin Issue.

### Why Is this Needed

317 years is not an option that a funder can select in the "expires in" section of the funder form.

### Description
Example: https://gitcoin.co/issue/MetaMask/metamask-extension/3127/1155

On the issue details page:
![screen shot 2018-10-24 at 14 15 58](https://user-images.githubusercontent.com/7516920/47535815-05348700-d872-11e8-9bbf-46fdc9ceb72b.png)

On the funder page:
![screen shot 2018-10-25 at 16 27 03](https://user-images.githubusercontent.com/7516920/47536015-f13d5500-d872-11e8-9539-8e6a27b8981b.png)

### Current Behavior
"Time left" estimates should match what the funder put up in the funding form.

### Expected Behavior
Never should show up as "never" or something like an infinity sign.

### Definition of Done
1. The "Expires in" value selected by a the funder for an issue should match the "time left" countdown on the issue details page.

octavioamu:
Great @Anish-Agnihotri Im fixing this one in the bounty detail and explorer
Bounty detail:
![image](https://user-images.githubusercontent.com/1358093/47947541-2e51b900-defd-11e8-897e-e0abd30a25cd.png)

Explorer
![image](https://user-images.githubusercontent.com/1358093/47947557-85578e00-defd-11e8-8bd8-1fa851c25b1e.png)

Anish-Agnihotri:
Looks good! Thanks @octavioamu!

frankchen07:
Looks great! Thanks for taking a look at this @Anish-Agnihotri and @octavioamu!

PR #2676, opened 2018-11-03
Fix never expires

octavioamu:
##### Description

This fixes the estimate expiration when user choose "Never"

In bounty detail:
image

In explorer:
image

##### Checklist

##### Refers/Fixes

Fixes: #2538

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -1012,6 +1012,7 @@ function renderBountyRowsFromResults(results, renderForExplorer) {
     const dateNow = new Date();
     const dateExpires = new Date(result['expires_date']);
     const isExpired = dateExpires < dateNow && !result['is_open'];
+    const isInfinite = dateExpires - new Date().setFullYear(new Date().getFullYear() + 1) > 1;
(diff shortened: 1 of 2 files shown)


app/assets/v2/js/shared.js
```diff
@@ -1047,7 +1048,13 @@ function renderBountyRowsFromResults(results, renderForExplorer) {
       const timeLeft = timeDifference(dateNow, dateExpires);
       const expiredExpires = dateNow < dateExpires ? 'Expires' : 'Expired';
 
-      result['p'] += ('Opened ' + openedWhen + ' ago, ' + expiredExpires + ' ' + timeLeft);
+      if (isInfinite) {
+        const expiredExpires = 'Never expires';
```
thelostone-mc: `expiredExpires` seems confusing as it's already declared above.
Would it make sense to change it to

```
if (isInfinite) {
  const expiredExpires = 'Never expires';
 ....
} else {
  const expiredExpires = dateNow < dateExpires ? 'Expires' : 'Expired';
  ...
}
```

octavioamu: Totally agree, I had made it in that way not sure why I moved to outside the statement.

PR #2689, opened 2018-11-04
PR@2473: Add new background pattern category "wallpaper"

### LINKED ISSUE #2473: as a user, i would like support for patterned backgrounds in the avatar builder
as a user, i would like support for patterned backgrounds in the avatar builder

like this one:

success criteria:
- refactor the current background scheme to support having a `.svg` which contains a pattern as the background

kuhnchris:
##### Description
Adds new background pattern category

##### Checklist

dashboard

##### Testing
done

##### Refers/Fixes
Fixes #2473


owocki:
`make fix` will handle many of the linting issues!

also need to resolve the merge conflicts

kuhnchris:
Yeah I'll take care of that ASAP, thanks for pointing it out!

kuhnchris:
image
image
image

owocki:
looking good!

kuhnchris:
Thanks @thelostone-mc for fixing that!

owocki:
@kuhnchris i wonder what i'm doing wrong with the new assets i added ... 

https://github.com/gitcoinco/web/issues/2746

owocki:
@kuhnchris here is a SVG i commited : https://github.com/gitcoinco/web/blob/master/app/assets/v2/images/avatar/Wallpaper/gears.svg

there are others in the same directory as it 

what are the specs of the viewport i should set?

kuhnchris:
Hi there, I never figured that out aswell, but it seems to vary wildly between template and display in avatar builder, hence my request if we could make a unified system that uses the same components instead of squashing them. 
For repeating background we actually need to define a pattern as seen in the anchors.svg, it's complicated and has to be done by hand. Sadly I know that sucks.

What focus do you want @owocki : rewriting the avatar generation or fixing those SVGs? I think that the problems only continue to grow with the current system...

owocki:
per convo on slack, @kuhnchris is gonna submit a short term fix now and we'll break the larger scope into another ticket

Issue #2272, opened 2018-09-25
Redefine Work Started


PixelantDesign:
### User Story

As a funder I'd like a simple and clear distinction between contributors that express interest and  contributors that have _actually_ started work.

As a contributor I'd like to show my interest in working on a project but not have it be counted as started until I actually begin the work.

### Why Is this Needed

- Improve issues in a dangling state
- Make the different states clear to funders and contributors

### Description

Break apart expressed interest and work started.

### Current Behavior
Both are bucketed into one state.

### Solution
[comment]: # (Provide a summary of the solution and a task list on what needs to be fixed.)
*Summary*:

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

mbeacom:
`work started` is `expressed interested`, just renamed copy on the frontend.  The only action we have offchain is "work started" (express interest).
Are you thinking: `work started` would mean they have an open PR or an associated commit?  Oddly enough, this is exactly how `work started` began xD - That's why the BE all references `interest` vs `start work/work started`.

What do we want to consider `work started` vs hitting a CTA for "Start work" ?

thelostone-mc:
@PixelantDesign seems to be you are asking for a `follow` issue button. Is that it ?

PixelantDesign:
@thelostone-mc

So currently *work started* consists of *contributors who want to work on an issue* and *contributors who have actually started work*.

The thinking is that if we broke out *Express Interest* from actual *work started* then it would be *clearer* where work has actually started because interested contributors will know what they can apply to work on.

For issues that don’t need approval. There would not be an express interest phase. A contributor would go straight to start work.

Thoughts?

Renaming it on the front-end could partially work @mbeacom.

frankchen07:
I actually didn't know that "work started" simply meant "express interest" (according to @mbeacom ), so that's good to know. It also makes sense for the admin controls since the `interested` field is populated when work has been started. 

I think work started should be exactly what the meaning is == they've started work. This means they've opened a PR, or have code changes committed, etc. 

[...]

PR #2691, opened 2018-11-05
bounty: rename text to express interest for permissioned bounty

thelostone-mc:
Button states `Express Interest` and `Stop` for permission bounty 
Button states `Start Work` and `Stop Work` for permissionless bounty

untitled

Fixes: https://github.com/gitcoinco/web/issues/2272

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -866,10 +866,17 @@ var do_actions = function(result) {
 
   if (show_start_stop_work) {
     const enabled = true;
+    let text;
+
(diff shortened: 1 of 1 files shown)


thelostone-mc:
> Could we also update this language everywhere else that is displaying Start work?

@PixelantDesign this would be again only for permissioned bounties right ?

owocki:
i haven't looked closely but i bet dollars to donuts that there are other places its called 'start work' in the app (activities feed, emails, gitcoin bot, etc)

thelostone-mc:
@owocki we could address it all in one PR or follow it up with another PR like what @octavioamu stated

PR #2725, opened 2018-11-07
Semaphore changes

mbeacom:
##### Description

The goal of this PR is to ensure we're not generating blocking gunicorn threads when a semaphore is present on a bounty if the bounty is being processed.
Additionally, this PR introduces: `dashboard.utils.release_bounty_lock` to manually release a semaphore in the event of unforeseen failure.

##### Checklist

redis, bounties

##### Testing

Locally


app/dashboard/utils.py
```diff
@@ -561,3 +559,13 @@ def test(addrstr):
     # return priv key, pub key, address
 
     return priv.to_string().hex(), pub.hex(), checksum_encode(address)
+
+
+def get_bounty_semaphore_ns(standard_bounty_id):
+    return f'{SEMAPHORE_BOUNTY_NS}_{standard_bounty_id}_{SEMAPHORE_BOUNTY_SALT}'
+
+
+def release_bounty_lock(standard_bounty_id):
```
owocki: is this being called anywhere?

app/dashboard/helpers.py
```diff
@@ -517,11 +519,18 @@ def process_bounty_details(bounty_details):
...
         return (did_change, latest_old_bounty, latest_old_bounty)
 
-    new_bounty = create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id)
+    semaphore_key = get_bounty_semaphore_ns(bounty_data['id'])
+    semaphore = get_semaphore(semaphore_key)
 
-    if new_bounty:
-        return (did_change, latest_old_bounty, new_bounty)
-    return (did_change, latest_old_bounty, latest_old_bounty)
+    try:
```
owocki: nice refactor!

owocki:
left a few comments

app/dashboard/utils.py
mbeacom: No - merely introduced it as a mechanism for us to manually lift a lock in the event we needed to.  @owocki

owocki: makes sense

owocki:
i thought about this a bit more tonight.  we run the risk of any thread (gunicorn or not) dying and the bounty being sync'd being blocked forever.    what we really need is to use `redis:setex` instead of redis:set` so that one can assume that after ~ 60 seconds, that the sync thread has died

owocki:
@mbeacom 17ffa689bb4f04becbdc8af51de98749fa2e160f should accomplish the above.. lmk if you concur

upstream code => https://github.com/bluele/redis-semaphore/blob/master/redis_semaphore/__init__.py#L18

PR #2727, opened 2018-11-07
removes IOS app from site

owocki:
i just got an email from a user who was upset that someone applied to work on their bounty and was automatically approved
after looking into it, they applied with the iOS app..  which does not support pending workers
i think this is the nail in the coffin for the iOS app for me… which has never been used really, and is more of a headache to maintain than is worth it i think.

what this does:
- removes iOS banner
- removes iOS page
- disables ios endpoints

still TODO:
- email all users of the iOS app and let them know its not supported anymore


owocki:
http://bits.owocki.com/b1b892306bf8/Screen%20Shot%202018-11-07%20at%201.33.57%20PM.png

PR #2734, opened 2018-11-08
fixes the logic for avatar save checks

owocki:
previous_* was being set every time, when really it should only updated if the field its dervied from is updated

per https://gitcoincore.slack.com/archives/CBZEHJCN7/p1541628897002900


app/avatar/models.py
```diff
@@ -82,14 +82,14 @@ def save(self, *args, force_insert=False, force_update=False, **kwargs):
         """Override the save to perform change comparison against PNG and SVG fields."""
         if (self.svg != self.__previous_svg) or (self.svg and not self.custom_avatar_png):
```
owocki: alternatively we could just remove the if statement and update every single time

PR #2756, opened 2018-11-10
search history saved in database

owocki:
##### Description

search history saved in database

why:
- people won't explicitly tell us their code preferences.  by saving their searches, we can more easily understand the way people are trying to splice the work on the platform, and to use it to target bounties / grants to them in the future.

##### Checklist

API

##### Testing

tested it

##### Refers/Fixes

self


app/dashboard/router.py
```diff
@@ -264,6 +265,14 @@ def get_queryset(self):
             end = start + int(limit)
             queryset = queryset[start:end]
 
+
+        # save search history 
+        SearchHistory.objects.create(
+            user=self.request.user if not self.request.user.is_anonymous else None,
```
owocki: for some reason, i get `AnonymousUser` everytime I call `self.request.user` in the router.py context.. it seems to not be Anonymous in regular views

mbeacom: @owocki Fixed in:  https://github.com/gitcoinco/web/pull/2756/commits/04bf2fdcd49244dc1f173396d9277830a402bbf3

mbeacom: Ref:  [If no class authenticates, request.user will be set to an instance of django.contrib.auth.models.AnonymousUser, and request.auth will be set to None.](https://www.django-rest-framework.org/api-guide/authentication/#how-authentication-is-determined )

PR #2757, opened 2018-11-10
Remove full name entry + hide in form.

### LINKED ISSUE #2748: Remove the "full name" entry and the option to hide it on the Gitcoin funder form.
### User Story

As a funder in the open source community, true anonymity is not really possible. When funding an issue, I want as few decision points as possible from start to successfully funding an issue. 

On an Issue Details page, "Anonymous" is shown under the Funder section, but if the funder interacts on the Github issue, they eventually have their information exposed in the all activity section.

screen shot 2018-11-09 at 10 54 04

Anonymity only works if it's completely carried throughout the "All Activity" section and throughout Github (which isn't possible). It doesn't make sense to be in open source, funding issues, and remain anonymous - that's part of the "reputation" of open source.

### Why Is this Needed

This issue was brought up by a community member: https://github.com/gitcoinco/web/issues/2579

In order to tidy up the Issue Details all activity section, using Github usernames by default is more consistent, and follows the fact that true anonymity (in this case) is difficult, especially on Github and in open source. 

What logically follows is to remove the decision point to enter funder full names and just use their Github handle to indicate who is funding the bounty. The decision to enter and hide their email will remain.

This has the added benefit of not having difficulties following up with funders as necessary.

### Current Behavior

On the funder form, one can enter their full name and decide whether or not they want to remain anonymous. 

### Expected Behavior

A funder's Github handle will be shown by default.

### Definition of Done

1. Remove on the funder form:

screen shot 2018-11-09 at 10 43 35

Anish-Agnihotri:
##### Description

This is a simple commit that simply defaults the `details.html` page to take username as default value rather than full name, and removes the option to add a full name altogether from `new.html`. It's a simple commit which does this through: 

1. Removing the option to add a name. 
2. Defaulting occurrences of `bounty_owner_name` to `result.bounty_owner_github_username` in `bounty_details.js`. Since the option to anonymize full name has been removed altogether, I've also taken out the code that was committed previously for #2273. 

##### Checklist

- [X] commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)

User Interface. 

##### Testing

Tested cross-platform as usual and a few times with various issues on Rinkeby to ensure function. 

##### Refers/Fixes

Fixes #2748, removes need for #2273. 

CC: @frankchen07


thelostone-mc:
@Anish-Agnihotri added it into this PR ^_^

Issue #2579, opened 2018-10-30
Missing post-synchronisation for issue founders


kuhnchris:
**Describe the bug**
Apparently some funders of issues aren't properly connected to their profiles.

**To Reproduce**
Not sure, can't reproduce on purpose

**Expected behavior**
Have the issue funder's name displayed

**Screenshots**
![image](https://user-images.githubusercontent.com/761911/47752968-52f82780-dc96-11e8-9e4f-383c9a46b5b5.png)

**Desktop (please complete the following information):**

- OS: Mac OS X
- Browser Chrome
- Browser Version Latest Stable

kuhnchris:
For reference: The affected bounty is this one:
https://gitcoin.co/issue/gitcoinco/web/2508/1650

PixelantDesign:
Thanks for the report @kuhnchris!

frankchen07:
Ah, beat me to it. 😄 good catch!

thelostone-mc:
So what probably happened is that the options below were checked by default causing the name to be rendered `Anonymous`

<img width="805" alt="screen shot 2018-11-09 at 12 11 16 am" src="https://user-images.githubusercontent.com/5358146/48219977-0ad2b680-e3b4-11e8-9c41-0330e109d71e.png">

For Activities we track user action which doesn't taken into account if the funder has selected hide options.

We could either 
- remove those options 
  (personally I don't think we would need it cause it openness == transparency )
- update the Activity to keep track of that ( which might be a bit more tricker )

Here is the API response for that bounty :
https://gitcoin.co/actions/api/v0.1/bounties/?github_url=https://github.com/gitcoinco/web/issues/2508

@PixelantDesign @frankchen07

kuhnchris:
![image](https://user-images.githubusercontent.com/761911/48227107-baeaf400-e3a1-11e8-9ed7-3e5dd0a3faf0.png)
That seems redundant at best. if any of those values is not "Anonymous" then it wouldn't make sense to hide 'any' of em...

thelostone-mc:
So we always display the github username in the activities stream but on the footer we display the name of the funder which in this case @PixelantDesign has chosed to hide which is why that is anonymous.

The most sensible thing at this point in my head is
- Show the github username as opposed to the name of the funder in the funder info.
- Or do a check to see if funder doesn't want his name to be shown -> instead show github username

frankchen07:
I like your thought that the default should be showing the Github username - let's implement that. 

[...]

PR #2759, opened 2018-11-10
details : remove full name 

thelostone-mc:
- removed funder name from details page
- ensures email label + field is shown only if available
- cleans up extra code

Fixes: https://github.com/gitcoinco/web/issues/2579

screen shot 2018-11-11 at 12 43 38 am

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -26,8 +26,8 @@ var gitcoin_ize = function(key, val) {
 var email_ize = function(key, val) {
 
   if (val == 'Anonymous' || val == '') {
-    $('#bounty_owner_email').hide();
-    $('#bounty_owner_email_label').hide();
+    $('#bounty_owner_email').remove();
+    $('#bounty_owner_email_label').remove();
   }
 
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/pages/bounty_details.js
```diff
@@ -20,10 +20,16 @@ var gitcoin_ize = function(key, val) {
...
     return [ null, null ];
   }
-  return [ key, '<a href="https://gitcoin.co/profile/' + val + '" target="_blank" rel="noopener noreferrer">@' + val.replace('@', '') + '</a>' ];
+  return [ key, '<a href="https://gitcoin.co/profile/' + val + '" target="_blank" rel="noopener noreferrer">' + val.replace('@', '') + '</a>' ];
 };
 
 var email_ize = function(key, val) {
+
+  if (val == 'Anonymous' || val == '') {
+    $('#bounty_owner_email').hide();
```
octavioamu: Maybe is better to just .remove() the elements since we don't want to show it.

thelostone-mc: derp derp ! why didn't I think of that 😂

PR #2776, opened 2018-11-12
Ks cancel fixes

captnseagraves:
Ks cancel fixes


captnseagraves:
@mbeacom Just trying to fixes a few things in the cancel_subscription functionality, but have a bunch of other changes that come through from my migration issues last week. Do you have a suggested course of action to resolve the issues?

app/assets/v2/fonts/Muli/muli-v12-latin-600.svg
```diff
@@ -1,309 +0,0 @@
-<?xml version="1.0" standalone="no"?>
```
mbeacom: Are these file deletions right for this branch?

captnseagraves: No I dont think so. not sure where they came through. should I just copy them from another branch and reimplement?

captnseagraves: @mbeacom file deletions undone, and named as in `Master`, my only question is on the migration files. Are those ok to add?

PR #2791, opened 2018-11-13
Add ability to send a Kudos using only Eth address

owocki:
what the title says ^^

screen recording 2018-10-15 at 04 50 pm


app/dashboard/models.py
```diff
@@ -1061,7 +1061,7 @@ class SendCryptoAsset(SuperModel):
     emails = JSONField(blank=True)
     url = models.CharField(max_length=255, default='', blank=True)
     tokenName = models.CharField(max_length=255, default='ETH')
-    tokenAddress = models.CharField(max_length=255)
+    tokenAddress = models.CharField(max_length=255, null=True)
```
mbeacom: We should use `blank` in favor of `null` for text/char specific field types.

[Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL](https://docs.djangoproject.com/en/2.1/ref/models/fields/#null)

owocki: 2a4655a14f3b5d66bcf1c5e0e2a2a29f9

owocki:
tested this.. its working well!

Issue #2764, opened 2018-11-11
Canceling "started" bounties shows wrong error message


kuhnchris:
**Describe the bug**
When a started bounty needs to be cancelled, the wrong error message appears.
This issue is in here:
https://github.com/gitcoinco/web/blob/a52adc187cee91df3ffd6a80d743d13de760ce8e/app/assets/v2/js/pages/kill_bounty.js#L68
Basically we need a split between "is_open" and "is_started".

**To Reproduce**
Fund issue, have someone start it, try to cancel it

**Expected behavior**
Error message: "You cannot cancel a bounty that has a worker working on it".

**Screenshots**
![image](https://user-images.githubusercontent.com/761911/48318230-2f6d9f00-e5fd-11e8-9631-72f46c6e322c.png)
![image](https://user-images.githubusercontent.com/761911/48318231-34325300-e5fd-11e8-8a7f-084956c466b9.png)
![image](https://user-images.githubusercontent.com/761911/48318226-267ccd80-e5fd-11e8-8e53-e085cdd46fbe.png)

reported by RichatMakerDAO on Slack.

PixelantDesign:
Thanks for creating this ticket @kuhnchris !

owocki:
ahh shoot, yeah [this looks wrong](http://bits.owocki.com/2c9c9de6105f/Screen%20Shot%202018-11-12%20at%2010.44.43%20AM.png)

kuhnchris:
We even anaylzed the structure we received - it's definitly because is_open is not true anymore. Result does not contain a "is_started", we would need to add that to the python and then additionally check against that in that block, so the proper error is shown.

owocki:
i posted about this on the gitcoin slack

pasting details now

```
# on the triage call:

this issue has already been paid

its admin override status is ‘started’ thats why its showing as not paid in the frontend.

# to the team :

team, there have been an increasing number of support requests from people whose bounties have an override status on them

for example, this one above has been paid out already and its override status is ‘started’.  the funder has for months been really confused about why they can’t pay it out.

for a fix it twice, we need to either
1) be more responsible with our usage of override statuses and remember individually to check back on them and remove them when the override is not needed anymore
2) or, we need to make it so that overrides expire after a few days / weeks
3) or we need to remove them altogther
```

kuhnchris:
oh, interesting! didn't hear about that one, my bad for making assumptions.

frankchen07:
[...]

PR #2811, opened 2018-11-13
[WOC] PR@2764: Update corner case canceling started issue

kuhnchris:
##### Description
Additional check for open == false but other things being true to show additional info message.

##### Checklist
- [ ] test

UI

##### Testing
Cannot test due to broken pyvips :-(

##### Refers/Fixes
Fixes: #2764


mbeacom:
@kuhnchris I was hoping to include the restriction on the BE as well, but if you're opposed to adding it right now, I guess we can merge it and I'll try to tackle it at some point, time permitting.

kuhnchris:
Your decision, that PR at least fixes the front-end side (for now).

thelostone-mc:
@kuhnchris would be super neat if we could throw it into the BE if you are up for it ? ^_^

mbeacom:
Anything you need from us here? @kuhnchris

kuhnchris:
ah pardon me, didn't see the reply. want me to implement that check on backend side aswell (just that check) or want me to create a build-ticket to check this in general?

thelostone-mc:
@kuhnchris would be ideal :P or do you just wanna resolve the conflicts and roll with this ?

kuhnchris:
I do not mind either way. If the official statement is "fix be aswell" then I'll fix it, we just do not want to let this fall through the roof.

kuhnchris:
@thelostone-mc you still know if this is still out there and causing issues? else I'd suggest we throw this away! Thanks!

app/assets/v2/js/pages/kill_bounty.js
```diff
@@ -80,13 +80,18 @@ window.onload = function() {
...
           let errormsg = undefined;
 
-          if (bountyAmount == 0 || is_open == false) {
-            errormsg =
-                gettext('No active funded issue found at this address.  Are you sure this is an active funded issue?');
-          }
-          if (fromAddress != web3.eth.coinbase) {
-            errormsg =
-                gettext('Only the address that submitted this funded issue may kill the bounty.');
+          if (bountyAmount > 0 && !is_open) {
```
octavioamu: parseInt(0.09, 10)>0 will be false 
0.09 > 0 will be true (what we want)

PR #2849, opened 2018-11-15
Ks emails update

captnseagraves:
Ks emails update

Added grant_cancellation and subscription_terminated emails


captnseagraves:
CI Failures have to do with Leaderboard issues

PR #2872, opened 2018-11-17
Fix rare `is_metamask_approved is not defined` race condition.

### LINKED ISSUE #2813: ReferenceError: is_metamask_approved is not defined
**Describe the bug**
Once in a while ```Reference error is_metamask_approved is not defined in console.log```. It does not show any network in the gitcoin page also.

**To Reproduce**
Run into some bad luck

screencast from wednesday 14 november 2018 08_54_19 ist

**Desktop:**

- OS: Ubuntu 18.04
- Browser Chrome

pinkiebell:
Sometimes, the window's `load` handler get executed in another order
which breaks variable lookup.
Fix that by explicit declaring `is_metamask_approved` and `is_metamask_unlocked`.

Fixes #2813

### DIFF
--- a/app/assets/v2/js/metamask-approval.js
+++ b/app/assets/v2/js/metamask-approval.js
@@ -1,3 +1,7 @@
+
+var is_metamask_approved = is_metamask_approved;
+var is_metamask_unlocked = is_metamask_unlocked;
+
 async function metamaskApproval() {
   if (window.ethereum) {
     window.web3 = new Web3(ethereum);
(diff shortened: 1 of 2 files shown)


thelostone-mc:
@pinkiebell not sure if this is right ! 
Would it make sense to just declare the variable once and let it have global scope cause it's used throughout the app ?

pinkiebell:
@thelostone-mc Sure, I just added that for redundancy. I will drop the declarations from `shared.js`.

thelostone-mc:
Also wouldn't `var is_metamask_approved = is_metamask_approved;` still throw an undefined error ? 
Would it make more sense to just leave it as `var is_metamask_approved` ?

pinkiebell:
@thelostone-mc 
Nah, thats perfectly valid. I intentionally wrote it that way to signal that it might be used elsewhere,
so we don't overwrite the variables if that happens.
Just paranoia 😆

thelostone-mc:
Haha this makes me a lot more comfortable :P

Issue #2869, opened 2018-11-17
speed up collectstatic on gitcoin's codebase


owocki:
```
(gitcoin-37) ubuntu@HOST:~/gitcoin/coin/app$ time python3 manage.py collectstatic --noinput -i other
DEBUG 2018-11-17 19:03:10,568 base 11888 140056514545408 Configuring Raven for host: https://sentry.gitcoin.co
2 static files copied.

real	3m10.866s
user	0m40.784s
sys	0m2.356s
```

Collect static takes too long on our codebase in production.  I'm looking for someone who can help make it an order of magnitude, or two, faster.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.5 ETH (87.89 USD @ $175.78/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/2869/1781).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/2869/1781).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $53,728.34 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [the-rileyj](https://gitcoin.co/profile/the-rileyj) has started work.**

I'll see if there's anything I can do with the existing code to speed things up (I would assume this is most likely not going to be the case, however it provides a good starting point), then move onto concurrent solutions to see if that would help.

Are you ok with adding dependencies which are not from the standard library?

Are you ok with having concurrent (asynchronous, multi-threaded, multi-processed) solutions? If so, which would you prefer, if any.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/2869/1781).

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

[...]

PR #2879, opened 2018-11-18
Sped up static collect, minor changes to RUNNING_LOCALLY.md in docs

the-rileyj:
## Main Change:

Adding the variable which indicates the number of threads desired for collectfast to use in ```~/web/app/app/settings.py```
* ```COLLECTFAST_THREADS = 20```

Results on a Dual Core laptop

### Synopsis:

I don't have very powerful hardware, allocating 20 threads when I truly only have 4 isn't prudent, however I would imagine that the machine collect static is typically ran on has many more threads; so this is a decent starting place

## Other Changes:

Rationale for the Following Changes:
    Jumping in with no background, there were a few gotchas when it came to gathering all of the dependencies locally (I understand that docker is used in production); I would imagine that this would prove to be a barrier to entry for any future contributors. Which is why I made a few additions to make the local installation more robust.
    Note: I was doing all of this on Ubuntu, so if there are any discrepancies in the additions I made to some of the shell commands in the ```~/web/docs/RUNNING_LOCALLY.md``` file that may be a reason why (using apt instead of brew as a package manager, etc.)

Changed brew install packages
* changed ```python``` to ```python3``` for the sake of clarity
* added ```libvips libvips-dev libvips-tools``` which are needed for the python pyvips python library to function

Added line to ```~/web/docs/RUNNING_LOCALLY.md``` to assure as many packages as possible are installed for running locally in the case an error occurred with the requirements file

```shell

awk '!/^(-r)/' requirements/base.txt requirements/test.txt | xargs -n 1 pip3 install

```

Made minor changes to the portions in ```~/web/docs/RUNNING_LOCALLY.md``` which concerned python but were not explicit in the version
* ex. ```pip``` -> ```pip3```, ```virtualenv ...``` -> ```python3 -m virtualenv```

Made minor changes to ```~/web/docs/RUNNING_LOCALLY.md``` to indicate the context for which commands are executed

From

```shell

python3 -m virtualenv ...

```

To

```shell

~/web$ python3 -m virtualenv ...

```

##### Checklist

- [ x ] changes don't break existing behavior
- [ x ] commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)

Changes to ~/web/doc/RUNNING_LOCALLY.md
1 line addition to ~/web/app/app/settings.py

##### Testing

No changes to core code base.

##### Refers/Fixes

https://gitcoin.co/issue/gitcoinco/web/2869/1781

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -431,6 +431,7 @@ DJANGO_REDIS_IGNORE_EXCEPTIONS = env.bool('REDIS_IGNORE_EXCEPTIONS', default=Tru
 DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = env.bool('REDIS_LOG_IGNORED_EXCEPTIONS', default=True)
 COLLECTFAST_CACHE = env('COLLECTFAST_CACHE', default='collectfast')
 COLLECTFAST_DEBUG = env.bool('COLLECTFAST_DEBUG', default=False)
+COLLECTFAST_THREADS = 20
 REDIS_URL = env('REDIS_URL', default='rediscache://redis:6379/0?client_class=django_redis.client.DefaultClient')
 SEMAPHORE_REDIS_URL = env('SEMAPHORE_REDIS_URL', default=REDIS_URL)
 
--- a/docs/RUNNING_LOCALLY.md
+++ b/docs/RUNNING_LOCALLY.md
@@ -3,10 +3,12 @@
 *Note: This setup method is not recommended. To ensure a consistent environment, please check out the [Docker Setup Guide](https://docs.gitcoin.co/mk_setup/).
 
 ```shell
-brew install libmaxminddb automake pkg-config libtool libffi gmp python openssl
-git clone https://github.com/gitcoinco/web.git
-cd web/app
-cp app/local.env app/.env
+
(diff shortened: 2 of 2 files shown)


mbeacom:
@the-rileyj Thanks for the contribution and welcome to the community!

I'd love to enable multithreading for `collectstatic`/`collectfast` and I intended on doing so when I first implemented `collectfast`.  The problem is, we are using the `HashedFilesMixin` to enable us to cache static files for extremely long periods of time and simply upload new assets if they've changed while updating the salt.

There is some incompatibility with collectfast multithreading and the hashing mixin:  https://github.com/antonagestam/collectfast/issues/107

Once this is resolved, I'd be more than happy to add `COLLECTFAST_THREADS = env.int('COLLECTFAST_THREADS', 20)` to start with.

the-rileyj:
Understood, I'll try to do more.

mbeacom:
@the-rileyj Feel free to PR the documentation updates

the-rileyj:
Will do! Expect to see a new PR soon

owocki:
> The problem is, we are using the HashedFilesMixin to enable us to cache static files for extremely long periods of time and simply upload new assets if they've changed while updating the salt.

i wonder why collectstatic takes 3 minutes to run if it's only uploading new assets that have changed  ( see profile on  https://github.com/gitcoinco/web/issues/2869 )

Issue #2862, opened 2018-11-16
KS Grants QA - 11.16.18


captnseagraves:
**Index**
- [x] Implement index revamp design
- [x] Need a link to ‘Your Grants’ in nav or in user dropdown
- [ ] Need paginator instead of infinite scroll
- [x] Change ETH in Card progress bar to be DAI
- [x] Short images shorten card length. We should enforce a certain height to maintain consistency. 

**New**
- [x] It would be nice to spice up design
- [x] If goal value in DAI is 0 all views break because progress bar divides by 0, which is NaN. Need validation to ensure amount goal is greater than 0. 
- [x] Rename “title” input tag so that google doesn’t try to guess with your name
![screen shot 2018-11-16 at 2 34 49 pm](https://user-images.githubusercontent.com/22862292/48651853-feed8280-e9b9-11e8-97ab-70eb01870b66.png)
- [x] "create Grants button should be same width as form and same color as other action buttons. See "fund Grant" button for example
- [x] When the ‘waiting_room’ screen is up, the ALPHA warning banner still sitting at the bottom of the page
- [x] remove ETH from token drop down and include "any"

**What is Grants**
- [x] Introduce yourself or your team bullet appears twice
- [x] “Contribute with any ERC-20 token’ bullet says that you can pay with ETH, but you can’t

**Details/Admin**
- [x] Title input cannot handle a long title. Cuts title off after ~15-20 characters. 
- [x] Same for url, but a little longer since the link column is wider
- [x] Need to change ETH in progreess bar to DAI
- [ ] Should users be able to edit the their monthly goal?
- [x] Need to wire up updates
- [x] POST on updates fire the cancel_grant functionality on frontend. Will likely have the same issue when grant edit info is saved. I have a solution, but need to implement. 
- [x] ‘Waiting_room’ scene need to be implemented for cancel_grant functionality
- [x] Metamask check need to be implemented for cancel_grant funcitonality
- [x] Need to smooth out design when cancel_grant fired and “This grant has ended. You can find all active grants here.” is displayed.
- [x] After a user cancels their subscription, grants/deatils should no loner show them as a supporter

**Milestones**
- [x] Would be nice to spice up design
- [ ] Should we suggest people create milestones in the Grant Creation Process?
- [x] Date entered should be the deadline completion data, not the data the milestone is started. 
- [ ] The complete button should mark the milestone as completed and the actual date the milestone is marked completed with an associated github commit hash. 

**Your Grants**
[...]

PR #2889, opened 2018-11-19
Qa fixes

captnseagraves:
Qa fixes relevant to https://github.com/gitcoinco/web/issues/2862#event-1972438367. 

All formatting implemented by pre-commit hooks.


app/grants/models.py
```diff
@@ -179,25 +234,43 @@ class Contribution(SuperModel):
...
     token_amount = models.DecimalField(
-        default=1, decimal_places=4, max_digits=50, help_text=_('The promised contribution amount per period.'),
+        default=1,
+        decimal_places=4,
+        max_digits=50,
+        help_text=_('The promised contribution amount per period.'),
     )
     period_seconds = models.DecimalField(
-        default=0, decimal_places=0, max_digits=50, help_text=_('The number of seconds thats constitues a period.'),
+        default=0,
```
mbeacom: This formatting looks like it's from `yapf`, but you're removing `yapf` from the pre-commit hook.  Do we want to leave this enabled?

mbeacom:
Other than the yapf comment, lgtm.

app/grants/models.py
captnseagraves: @mark Unfortunately, no, currently either isort or yapf need to be disabled, or we need to sort through their configurations and discern where they are conflicting and changing each other back and forth. It was preventing anything from being committed. On Friday I had removed isort and committed. I believe this commit is where the yapf formatting came from. After discussing the conflicting configurations with you briefly it sounded like isort might be the preferred linter, so I added isort back and removed yapf. Let me know if yapf is actually preferred.

PR #2902, opened 2018-11-21
More Qa fixes

captnseagraves:
More Qa fixes for https://github.com/gitcoinco/web/issues/2862#event-1972438367

remove ETH and add Any Token to token drop downs. Also, ensure that user cant select token if one required by grant creator. Core systems shouldnt be affected, but will test in morning.


app/grants/views.py
```diff
@@ -291,6 +291,7 @@ def grant_fund(request, grant_id):
         'active': 'fund_grant',
         'title': _('Fund Grant'),
         'subscription': {},
+        'grant_has_no_token': True if grant.token_address == '0x0000000000000000000000000000000000000000' else False,
```
mbeacom: @captnseagraves Somewhat similar to what we discussed about the `not` reserved word, you can do something like:

`'grant_has_no_token': grant.token == '0x0000000000000000000000000000000000000000'`

PR #2910, opened 2018-11-21
qa changes

thelostone-mc:
screen shot 2018-11-21 at 1 46 47 pm

##### Description

- updated default image 
- updated grants/new qa

Note: the default background looks ugly on contribute grant and cancel contribution. 
those will be updated once we get new mocks (cc @willsputra )

##### Checklist


app/retail/templates/emails/grants/thank_you_for_supporting.html
```diff
@@ -114,7 +114,7 @@ <h2 style="text-transform: none;">{% trans "The world of open source is a better
 <hr>
 <div class="grantInfo">
   <a href="{% url 'grants:details' grant.id grant.slug %}">
-    <img src="{% if grant.logo and grant.logo.url %}{{ grant.logo.url }}{% else %}{% static 'v2/images/bitmap.png' %}{% endif %}" width="50" height="50" />
+    <img src="{% if grant.logo and grant.logo.url %}{{ grant.logo.url }}{% else %}{% with grant_logo='v2/images/grants/logos/' id=grant.id|modulo:3|add:1 %} {% static grant_logo|addstr:id|add:'.png' %} {% endwith %} {% endif %}" width="50" height="50" />
```
mbeacom: One downside to this approach over using `static()` in the template tag is that we no longer serve the hashed asset.  I'm not too concerned with this, since it's doubtful these will change, but just something to consider!

app/grants/templatetags/grants_extra.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    return str(value) + str(arg)
+
+
+@register.filter
+def modulo(num, val):
+    """Modulo of number
+
+        Args:
+            num : Number
```
mbeacom: nit: for our pep-257 docstrings, we could use:

```python
"""Get the modulo of the provided number and value.

    Args:
        num (int): Something describing the number.
        val (int): Something describing the value.

    Usage:
        {{ num|modulo(val) }}

    Returns:
        int: The modulo of number and value.

"""
```

PR #2868, opened 2018-11-17
Adding Success Rate Estimation In Pricing Section

nanspro:
##### Description
Estimating probability of success in terms of hourly rate!! 

##### Checklist

- [x] linter status: 100% pass
- [x] changes don't break existing behavior
- [x] commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)

##### Affected core subsystem(s)
UI

##### Testing
locally
*video attached

##### Refers/Fixes
Fixes #2158

PR #2912, opened 2018-11-21
hide estimator when hours is 0 / not defined

thelostone-mc:
##### Description

untitled


thelostone-mc:
@PixelantDesign @frankchen07 

need to sync on what to do for tips cause of this 
https://github.com/gitcoinco/web/pull/2868#issuecomment-440492243

frankchen07:
For now let's hide it. Hourly rates on tips doesn't make sense to me, and it has the potential to be confusing.

Thanks to everyone here for taking fast action on this.

PR #2921, opened 2018-11-22
grant/fund redesign

thelostone-mc:
##### Description

untitled

~~@captnseagraves / @mbeacom  could you help test this out before merge ?~~
~~I'm having issues at me end getting token and I'm not able to test out the contribute flow~~

~~There aren't any changes in the JS code but still as a precaution~~

Update: got it working :P 

screen shot 2018-11-26 at 2 13 02 pm

### DIFF
--- a/app/assets/v2/css/base.css
+++ b/app/assets/v2/css/base.css
@@ -124,6 +124,10 @@ span {
   border-radius: 4px !important;
 }
 
+.font-weight-semibold {
+  font-weight: 600;
+}
+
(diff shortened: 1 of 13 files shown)


willsputra:
@thelostone-mc lgtm design-wise. 😁

thelostone-mc:
I realized that the connect metamask check wasn't added and we didn't do the metamask check in the contribute fund section. 
So added that in another individual commit

untitled

Issue #2836, opened 2018-11-14
As a contributor, I want to be able to search issues by organization.


frankchen07:
### User Story

As a contributor, I want to be able to search issues by organization.

### Why Is this Needed

Let's say I'm stoked about Gitcoin and want to exclusively look for Gitcoin issues. Adding "Gitcoin" to the search bar returns more issues by organization beyond Gitcoin.

### Current Behavior

The current search function seems to only filter by tags, not necessarily by organization.

without "Gitcoin":
![screen shot 2018-11-14 at 14 44 39](https://user-images.githubusercontent.com/7516920/48517695-f5c6af00-e81b-11e8-835a-ce03942193ed.png)

with "Gitcoin":
![screen shot 2018-11-14 at 14 44 53](https://user-images.githubusercontent.com/7516920/48517700-fb23f980-e81b-11e8-87e6-cc616ffe7296.png)

### Expected Behavior

The search function should be able to account for organizational input. 

### Definition of Done

- [x] searching by organization only returns that organization's issues

PR #2924, opened 2018-11-23
Filter bounties by organization 

octavioamu:
##### Description

Add a text field to filter bounties by organization name.

##### Checklist

##### Refers/Fixes

Ref: #2836

### DIFF
--- a/app/dashboard/router.py
+++ b/app/dashboard/router.py
@@ -194,11 +194,21 @@ class BountyViewSet(viewsets.ModelViewSet):
             urls = self.request.query_params.get('github_url').split(',')
             queryset = queryset.filter(github_url__in=urls)
 
-        # filter by urls
+        # filter by orgs
         if 'org' in param_keys:
-            org = self.request.query_params.get('org')
-            url = f"https://github.com/{org}"
-            queryset = queryset.filter(github_url__icontains=url)
+            for orgs in param_keys:
+                val = self.request.query_params.get('org', '')
+                values = val.strip().split(',')
+                values = [value for value in values if value and val.strip()]
+                if values:
+                    _queryset = queryset.none()
+                    for value in values:
+                        args = {}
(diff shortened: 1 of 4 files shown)


octavioamu:
@mbeacom I added you because I have change a little the backend side to support multiple orgs search, for sure that can be improved, feel free to guide me or even to edit if needed.

app/dashboard/router.py
```diff
@@ -194,11 +194,21 @@ def get_queryset(self):
             urls = self.request.query_params.get('github_url').split(',')
             queryset = queryset.filter(github_url__in=urls)
 
-        # filter by urls
+        # filter by orgs
         if 'org' in param_keys:
-            org = self.request.query_params.get('org')
-            url = f"https://github.com/{org}"
-            queryset = queryset.filter(github_url__icontains=url)
+            for orgs in param_keys:
```
mbeacom: This will iterate through every parameter they send while making the request.
We don't seem to be doing anything with `orgs` here.  Can we eliminate that part to something like:

```python
# filter by orgs
val = self.request.query_params.get('org', '')
values = val.strip().split(',')
values = [value for value in values if value and val.strip()]
if values:
    _queryset = queryset.none()
    for value in values:
        org = value.strip()
        _queryset = _queryset | queryset.filter(github_url__icontains=f'https://github.com/{org}')
    queryset = _queryset
```

octavioamu: perfect, I will make the changes here. Thanks @mbeacom

thelostone-mc:
@octavioamu should we use bootstrap tooltip? 🤔

octavioamu:
> @octavioamu should we use bootstrap tooltip? 🤔

I think we can discuss about it because is a big work to change all to use bootstrap tooltip in this page.

thelostone-mc:
maybe in product sync :P

Issue #2779, opened 2018-11-13
DESIGN - As a user, I want to be able to upload Kudos instead of add them to the YAML file / codebase, so Kudos are easier to deploy.  As  a Gitcoin founder, I want there to be an easy way for new artists to contribute to the platform to enable more liquidity and therefore revenue.


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a user, I want to be able to upload Kudos instead of add them to the YAML file / codebase, so Kudos are easier to deploy.  

As  a Gitcoin founder, I want there to be an easy way for new artists to contribute to the platform to enable more liquidity and therefore revenue.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

Its a cumbersome 8 step process to add a new Kudos to the system right now.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Current Behavior
[comment]: # (Describe what actually happened.)

You must 

1. Commit a kudos to the git repo [example]
2. Push it to master; push it live.
3. Run the command `python manage.py mint_all_kudos mainnet ~/gitcoin/coin/app/kudos/kudos.yaml --gitcoin_account --filter=GREP_EXPRESSION --live` to mint the kudos to the blockchain.
4. Run `bash scripts/run_management_command_if_not_already_running.bash sync_kudos mainnet opensea --start 1` or wait for it to run automatically.

To get a new kudos on the system.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

1. User goes to 
https://gitcoin.co/mint/kudos
2. User uploads all of this information to the system:

```
- name: Kauri Pioneer
  description: Cheers to you, one of the first Kauri contributors!
  priceFinney: 3
  numClonesAllowed: 100
  tags: 
    - web3
    - kauri
    - ethereum
    - documentation
    - knowledge
  image: kauri_1.svg
  to_address: 0xF8aE578d5d4e570De6c31F26D42eF369C320aE0b
```
3. Uploads the SVG artwork.
4. Form sends an email to founders@gitcoin.co asking them to approve the Kudos.
5. Upon approval, It is minted to the blockchain and added to the DB after the form is submitted.
6. Upon failure, an error message is shown that specifies why the failure.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
See above specification.

### Additional Information
[...]

PR #2936, opened 2018-11-24
[WOC] new kudos upload / mint process

owocki:
##### Description
new kudos upload / mint process

##### Checklist

kudos 

##### Testing

i still need to test it. in particular:
- how do we get our gitcoin IPFS host to serve the items as an image (via the `/ipfs/` route instead of the `cat` route, and in both cases, via content type text/svg) ?
- hook up the kudos minting process to the admin web interface

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/2779


app/kudos/models.py
```diff
@@ -465,3 +466,51 @@ class BulkTransferRedemption(SuperModel):
     def __str__(self):
         """Return the string representation of a model."""
         return f"coupon: {self.coupon} redeemed_by: {self.redeemed_by}"
+
+
+
+class TokenRequest(SuperModel):
+    """Define the Token model."""
```
mbeacom: `TokenRequest`

app/kudos/views.py
```diff
@@ -690,3 +691,45 @@ def receive_bulk(request, secret):
...
+
+@csrf_exempt
+def newkudos(request):
+    context = {
+        'active': 'newkudos',
+        'msg': None,
+    }
+
+    if request.POST:
+        required_fields = ['name', 'description', 'priceFinney', 'artist', 'platform', 'numClonesAllowed', 'tags', 'to_address', 'artwork_url']
```
mbeacom: We could move this to a django form to handle this type of validation.

app/kudos/views.py
```diff
@@ -690,3 +691,45 @@ def receive_bulk(request, secret):
...
+                platform=request.POST['platform'],
+                numClonesAllowed=request.POST['numClonesAllowed'],
+                tags=",".split(request.POST['tags']),
+                to_address=request.POST['to_address'],
+                artwork_url=request.POST['artwork_url'],
+                network='mainnet',
+                approved=False,
+                metadata={
+                    'ip': get_ip(request),
+                    'email': request.POST['email'],
```
mbeacom: `email` isn't validated as a required field in https://github.com/gitcoinco/web/pull/2936/files#diff-9b7a42d8513c01971adb785752bd3f67R704 - We'll probably want to use `.get('key', 'default_value')` for all of the params that aren't required here.

owocki:
> how do we get our gitcoin IPFS host to serve the items as an image (via the /ipfs/ route instead of the cat route, and in both cases, via content type text/svg) ?

@mbeacom i could use your help with this after grants ships.  right now i can get the IPFS response to show up, but i can't get any browser to show it as an actual SVG.  i think it has to do with the content type of the reponse

app/kudos/models.py
owocki: updated

app/kudos/views.py
owocki: 6865ed40568af4b3397f5a91b1c4d80870d95760

owocki:
https://github.com/gitcoinco/web/pull/2936#issuecomment-442616633

owocki:
@mbeacom if you get a sec, i'd love to jam on this

> how do we get our gitcoin IPFS host to serve the items as an image (via the /ipfs/ route instead of the cat route, and in both cases, via content type text/svg) ?

this is the big blocker IMHO

mbeacom:
I assume we can jam on this towards the end of the week or next week.

PR #2951, opened 2018-11-26
Ks contract versioning

captnseagraves:
Implemented contract versioning


owocki:
screen shot 2018-11-26 at 2 16 34 pm

tried leaving this comment then my computer froze bc the PR is so big

app/grants/migrations/0005_grant_slug.py
```diff
@@ -4,13 +4,13 @@
...
 
-def update_grants(apps, schema_editor):
-    """Handle the data migration for grant.slug."""
-    Grant = apps.get_model('grants', 'Grant')
-
-    grants = Grant.objects.all()
-    for grant in grants:
-        grant.save()
+# def update_grants(apps, schema_editor):
+#     """Handle the data migration for grant.slug."""
```
owocki: did you mean to commit this?

app/grants/migrations/0005_grant_slug.py
```diff
@@ -24,6 +24,7 @@ class Migration(migrations.Migration):
             model_name='grant',
             name='slug',
             field=django_extensions.db.fields.AutoSlugField(blank=True, editable=False, populate_from='title'),
-        ),
-        migrations.RunPython(update_grants),
+        )
+        # ,
+        # migrations.RunPython(update_grants),
```
owocki: this one too

app/grants/migrations/0007_auto_20181126_1652.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.1.2 on 2018-11-26 16:52
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
owocki: @SaptakS @thelostone-mc do these migrations look ok to you?

app/grants/migrations/0005_grant_slug.py
captnseagraves: Yeah, probably dont need to comment this function

app/grants/migrations/0005_grant_slug.py
captnseagraves: But this is the point at which I was receiving the error last week and you said to just comment out the offending line so we could move forward

owocki: @SaptakS did you run into this issue when you rebased?

captnseagraves: I ran into there issues again this morning when I pulled down the lastest `grants`

captnseagraves: Since I already have eveything migrated there are no errors while running `make migrations` or `make migrate`. I can uncomment the line.

app/grants/models.py
```diff
@@ -112,6 +112,12 @@ class Meta:
         default='0x0',
         help_text=_('The contract address of the Grant.'),
     )
+    contract_version = models.DecimalField(
```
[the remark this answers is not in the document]
owocki: > Can we switch this to a charfield? If we're going by semantic versioning, decimalfield won't make sense: 0.1.0 for example. If we're using integers like: 0 or 1, we should use positiveintegerfield here.


this makes sense to me

> Should we include a process for handling the ABI based on contract version, so it outputs the appropriate static URL or a given contract version..?

as does this

PR #2955, opened 2018-11-27
Implemented details updates

captnseagraves:
Added Update model in grants and implemented POST functionality


app/grants/models.py
```diff
@@ -164,6 +164,27 @@ def percentage_done(self):
         return ((self.amount_received / self.amount_goal) * 100)
 
 
+class Update(SuperModel):
+    """Define the structure of a Grant."""
```
mbeacom: Update docstrings

app/grants/models.py
```diff
@@ -164,6 +164,27 @@ def percentage_done(self):
...
+        default='',
+        max_length=255,
+        help_text=_('The title of the Grant.')
+    )
+    description = models.TextField(
+        default='',
+        blank=True,
+        help_text=_('The description of the Grant.')
+    )
+    grant = models.ForeignKey(
```
mbeacom: If we want this to show up in activity feeds, we might want to move to `Activity.comment` and just add it there.

app/grants/views.py
```diff
@@ -80,61 +80,36 @@ def grant_details(request, grant_id, grant_slug):
...
-        'title': 'Awesome Update',
-        'date': '08.02.2018',
-        'description': 'Some awesome update about this project.',
-    }, {
-        'title': 'Stellar Update',
-        'date': '08.02.2018',
-        'description': 'Another stellar update about this project.',
-    }]
+        print(request.POST)
+        if 'contract_address' in request.POST:
```
mbeacom: This shouldn't be the latest revision to grants..? I know I've fixed the ability for *anyone* to no longer cancel a grant...

captnseagraves: Yes, only an admin is shown the "Cancel This Grant" button. I have added here a series of if statements, based on the names of inputs in the different forms, to handle the different form submissions on the grants/details page. Currently, there is a form for the "Cancel This Grant" button, the updates form, and the edit details form.

mbeacom: @captnseagraves I'm saying in the actual backend here, *anyone* can simply `POST` and delete it.

mbeacom:
@captnseagraves Please cherry pick your changes here and move them over to a branch from `grantz`

app/grants/views.py
captnseagraves: @mbeacom Gotcha, makes sense. Unfortunately, I have not seen your implementation for that. I pulled the latest grants and added to the `if request.method == 'POST':` statement for each form.  Would the solution look something like an additional `if` statement, `if profile == grant.admin_profile:`

captnseagraves: hmmm... No that wouldn't work. Can you point me in the right direction?

PR #2959, opened 2018-11-27
QA + waiting state loader

thelostone-mc:
##### Description

This PR

- Fixes the metamask check on cancel
- enables alert + network pill across all grant pages which need metamask interaction
- adds waiting state loader to fund a grant, cancel a grant contribution
- cleanup a bit of code + DRY + re-indent (boy scout rule)
- fixed card height
- added stub grant page for non staff users

untitled

untitled

---

For @captnseagraves 
- [ ]  On cancel contribution page -> I get this error (so the redirection to bounty details fails)
screen shot 2018-11-27 at 6 07 23 pm

- [x] On cancel grant is [this right place ](https://github.com/thelostone-mc/web/blob/grants/app/assets/v2/js/grants/detail.js#L65-L73)to add the waiting state loader ?

For @SaptakS 

- [ ] @PixelantDesign stated that when we cancel the grant -> we need a modal window confirming the delete operation. cc @willsputra 
- [ ] On the waiting state loader the etherscan transaction link doesn't get mapped. Instead it points to the grant link
screen shot 2018-11-27 at 6 56 15 pm


app/assets/v2/js/grants/cancel_subscription.js
```diff
@@ -67,16 +67,17 @@ window.onload = function() {
...
                   });
               });
             })
             .on('confirmation', function(confirmationNumber, receipt) {
               console.log('receipt', receipt);
+              window.location = $('#grant-link').val();
             })
             .on('error', function(err) {
               console.log('err', err);
+              // TODO: Redirect to same page and throw error ?
```
owocki: maybe at least throw an error and log something to sentry?

app/assets/v2/js/grants/detail.js
```diff
@@ -60,15 +60,21 @@ $(document).ready(function() {
       let deployedSubscription = new web3.eth.Contract(compiledSubscription.abi, data.contract_address);
 
       web3.eth.getAccounts(function(err, accounts) {
-
-
-        deployedSubscription.methods.endContract().send({from: accounts[0], gasPrice: 4000000000})
+        deployedSubscription.methods.endContract()
+          .send({from: accounts[0], gasPrice: 4000000000})
+          // .on('transactionHash', function() {
+          //   // Waiting State screen
```
owocki: is this intended to be commented out?  if so can we remove it?

app/assets/v2/js/grants/fund.js
```diff
@@ -202,3 +187,12 @@ const updateSummary = (element) => {
     $('#summary-frequency').html($('input#period').val());
   });
 };
+
+const enableWaitState = () => {
```
owocki: coulkd maybe go into a shared file so it can be re-used?

app/assets/v2/js/grants/new.js
```diff
@@ -71,64 +71,39 @@ $(document).ready(function() {
...
-
-              $.each($(form).serializeArray(), function() {
-                data[this.name] = this.value;
-              });
-              console.log(data);
-              form.submit();
+          }).send({
+            from: accounts[0],
+            gas: 2500000,
+            gasPrice: 4000000000
```
owocki: why the hardcoded gas price?  does this pattern exist all across grants?  there is lots of good gas price info in the backend and in the gas price module.

app/grants/templates/grants/detail.html
```diff
@@ -30,6 +31,22 @@
       {% include 'grants/nav.html' %}
     </div>
 
+    <div class="row no-gutter grant_waiting" style="display: none;">
```
owocki: can we DRY this module?

app/grants/templates/grants/fund.html
```diff
@@ -48,6 +49,21 @@
       </div>
     </div>
 
+    <div class="row no-gutter grant_waiting" style="display: none;">
```
owocki: DRY

app/grants/templates/grants/cancel.html
```diff
@@ -48,6 +49,21 @@
       </div>
     </div>
 
+    <div class="row no-gutter grant_waiting" style="display: none;">
```
owocki: DRY

app/assets/v2/js/grants/new.js
thelostone-mc: I just re-indented that ! 😅 
@captnseagraves would you be able to answer this ?

app/assets/v2/js/grants/fund.js
thelostone-mc: coolio

app/assets/v2/js/grants/detail.js
thelostone-mc: I'm not sure if that is the right way to add it , which is why I tagged @captnseagraves to calrify on that ! 
This is essentially when to load waiting screen -> once user clicks on cancel grant 
would you have context on this @owocki  ?

app/assets/v2/js/grants/new.js
captnseagraves: @owocki @thelostone-mc Hardcoded gasPrice because there is a frequent error on Rinkeby when the gasPrice is not set and defaults to 1 gwei. Implementing the gasPrice functionality already within Bounties has not been a priority up to now, so has yet to be done. The above only prefills the value in metamask and still allows the user to adjust as they see fit.

app/assets/v2/js/grants/detail.js
captnseagraves: @owocki @thelostone-mc Yes, this is how we should implement the waiting screen. We can uncomment this code.

app/assets/v2/js/grants/new.js
```diff
@@ -71,64 +71,39 @@ $(document).ready(function() {
...
             });
+            console.log(data);
+            form.submit();
+          });
         });
       });
     }
   });
 
-  $('#new-milestone').on('click', function(event) {
```
captnseagraves: Why has this been deleted? Was it unused?

thelostone-mc: yeah we don't allow adding `milestones` in new grant right ? 
This was added way back if I'm not mistaken

app/assets/v2/js/grants/cancel_subscription.js
thelostone-mc: like an alert?

app/assets/v2/js/grants/new.js
captnseagraves: Ah, yes, you are right. We'll likely eventually add back in milestones to grants/new, but we can delete it for now.

app/assets/v2/js/grants/detail.js
owocki: 👍

thelostone-mc:
@captnseagraves the token I used had 4 decimals

PR #2991, opened 2018-11-28
primary email was wrong type

owocki:
fixes the primary emails being the wrong type 

see 
https://gitcoin.co/_administrationdashboard/tip/
https://gitcoin.co/_administrationkudos/kudostransfer/457/change/

### DIFF
--- a/app/dashboard/tip_views.py
+++ b/app/dashboard/tip_views.py
@@ -283,7 +283,7 @@ def send_tip_3(request):
             primary_email = to_emails['events'][0]
         else:
             print("TODO: no email found.  in the future, we should handle this case better because it's GOING to end up as a support request")
-    if type(primary_email) is list and len(primary_email):
+    if primary_email and isinstance(primary_email, list):
         primary_email = primary_email[0]
 
     # If no primary email in session, try the POST data. If none, fetch from GH.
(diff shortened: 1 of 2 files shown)


app/dashboard/tip_views.py
```diff
@@ -283,7 +283,8 @@ def send_tip_3(request):
             primary_email = to_emails['events'][0]
         else:
             print("TODO: no email found.  in the future, we should handle this case better because it's GOING to end up as a support request")
-
+    if type(primary_email) is list and len(primary_email):
```
owocki: it might be creating a method named `get_primary_email()` that accepts `params`.. but geez, `kudos/views` and `tip_views.py` just have so much to consolidate it hurts to start

mbeacom: We can omit `len()` and `type() is` in favor of `if primary_email and isinstance(primary_email, list):` - checks for val followed by the type assertion.

app/kudos/views.py
```diff
@@ -324,6 +324,8 @@ def send_3(request):
             primary_email = to_emails['events'][0]
         else:
             print("TODO: no email found.  in the future, we should handle this case better because it's GOING to end up as a support request")
+    if type(primary_email) is list and len(primary_email):
```
mbeacom: Can also be written as: `if primary_email and isinstance(primary_email, list):`

app/dashboard/tip_views.py
owocki: dbec78ef4edab0ce4e996426e44d847cc6c2d813

app/kudos/views.py
owocki: dbec78ef4edab0ce4e996426e44d847cc6c2d813

PR #3012, opened 2018-11-29
Ks grantz details edit update

captnseagraves:
Ks grantz details edit update


app/assets/v2/js/grants/detail.js
```diff
@@ -49,36 +71,34 @@ $(document).ready(function() {
...
+      deployedSubscription.methods.endContract()
+        .send({from: accounts[0], gasPrice: 4000000000})
+        .on('transactionHash', function() {
+          enableWaitState('#grants-details');
+        })
+        .on('confirmation', function(confirmationNumber, receipt) {
+          $.ajax({
+            type: 'post',
+            url: '',
+            data: { 'contract_address': contract_address},
```
octavioamu: not sure you have access to `contract_address` from this scope

app/assets/v2/js/grants/detail.js
```diff
@@ -49,36 +71,34 @@ $(document).ready(function() {
...
+            success: function(json) {
+              console.log('cancel grant POST successful');
```
octavioamu: Maybe a page reload here so the changes will be reflected.

app/assets/v2/js/grants/detail.js
captnseagraves: It is available. The ajax call inherits from ln76. Just tested again and functionality works.

app/assets/v2/js/grants/detail.js
captnseagraves: yes, the redirects in views.py dont seem to be firing. Aditya had added some but i removed them thinking we didnt need them. I will add back in here and ln55.

octavioamu: Yes because is an ajax sent not normal form so you will need something like `location.reload();` in the success callback.
Also to get accepted by stickler you need to remove all the console.log()

captnseagraves: any suggestions for the failure case besides a console.log?

SaptakS: You can probably make an alert rather than a console log. Use the `_alert()` function maybe.

octavioamu:
LGTM

Issue #2949, opened 2018-11-26
Community QA - Grants -- Week of 11/26


PixelantDesign:
# Alisa QA 11/26/2019

**Must Have**
- [x] Homepage Summary - Correct spelling of dependancy ---> dependency
- [x] Homepage Summary - Once EIP1337 Link is clicked - it is not legible. Please use the light blue Gitcoin link color.

- [x] Cancel Grant - Need an interstitial modal to explain what is about to happen if they click cancel a grant. Ex: By clicking cancel, you will delete all of the information associated, etc, etc. Include any important information the user should know about their data and funds. 
This should display prior to the metamask dialog box popping up.

**Should Have**
- [x] Homepage - Summary is a little wordy. Maybe there is a better way to explain what grants is...
Will users understand what dependency tree is? 
Suggestion: Show your appreciation for maintainers by elevating their contributions or giving them core team responsibility.  

- [x] View Grants - the columns between the 3 sections are uneven as you view the page from top to bottom.

- [x] View Grants - Description needs header similar to Team

**Won't Have**
- [ ] What is grants tab seems like it should be part of the homepage or an FTUX screen to create grant, but noting this so we can fix later. 

# Alisa QA 11/30/2019

**Should Have**

- [x] Alpha Banner - Banner is difficult to read and copy is alarming.
Suggestions: 
Invert the background and font color for legibility.
Remove the word Warning and the scary copy.
Use Copy: Gitcoin Grants is in ALPHA release. Please use caution if funding at high values.

**Nice to Have**
- [ ] Create Grant Form - Project URL - Provide hint text examples on whether this could be a normal url or a GitHub url.

- [ ] Create Grant Form - Team Members - Provide hint text guidance whether this has to be a Github user name or a regular person’s name or an email address.

- [ ] IF we default the monthly sought amount to Dai, should we default to Dai for Required Token Subscription?

- [x] Fund Grant - Change Number of Periods to Number of Recurring Payments. 

- [ ] ~~If I'm the creator of a grant, I should also be able to add funding to my own grant.~~

- [ ] Getting Started with Grants - Take Me Back link and paradigm is in a weird location, recommend removing.

# Kevin QA 2018/11/28

## LIKES
* new index page is hott
* 'my grants' page is really nice design.
* in general, the CSS seems really well baked.

## MINOR STUFF
M.png)
[...]

PR #3029, opened 2018-11-30
qa fixes

thelostone-mc:
- wired up active state so you can see both active and completed grants
- show alpha banner only when needed

Refs: https://github.com/gitcoinco/web/issues/2949

untitled

### DIFF
--- a/app/grants/templates/grants/card.html
+++ b/app/grants/templates/grants/card.html
@@ -51,7 +51,7 @@
           </div>
         </div>
         <div class="grant-item__state mt-3">
-          {% if grant %} <!-- TODO: wire up grant state -->
+          {% if grant.active %}
             <div class="grant-item__state-active">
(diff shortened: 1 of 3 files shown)


app/grants/templates/grants/card.html
```diff
@@ -51,7 +51,7 @@ <h2 class="grant-item__title font-subheader"><a href="{% url 'grants:details' gr
           </div>
         </div>
         <div class="grant-item__state mt-3">
-          {% if grant %} <!-- TODO: wire up grant state -->
+          {% if grant.active %}
```
mbeacom: This works, but if a grant is cancelled, it's going to show as `Completed` for now xD

captnseagraves: True. Though eventually we do need fidelity on this.

PR #3035, opened 2018-11-30
work around for contract deploy working

captnseagraves:
work around for contract deploy working

Deploy working even if getting error back. For some reason the web3js function hangs if an external function is called from within .on() so callFunctionWhenTransactionMined contained within .on()


thelostone-mc:
@captnseagraves I guess the `avatar/utils.py` got re-indent by the linter.
Could we undo that ?

captnseagraves:
@thelostone-mc just pushed the fix

PR #3052, opened 2018-12-03
Ks successful contribution def

captnseagraves:
Ks successful contribution def

@owocki is this the sort of thing you were expecting?


app/grants/models.py
```diff
@@ -314,34 +325,12 @@ class ContributionQuerySet(models.QuerySet):
 class Contribution(SuperModel):
     """Define the structure of a subscription agreement."""
 
-    tx_id = models.CharField(max_length=255, default='0x0', help_text=_('The transaction ID of the Contribution.'))
-    from_address = models.CharField(
+    tx_id = models.CharField(
         max_length=255,
         default='0x0',
-        help_text=_('The wallet address tokens are sent from.'),
-    )
```
owocki: you just need to `./manage.py makemigrations grants` on this so it creates a migration

app/grants/models.py
```diff
@@ -304,6 +304,17 @@ def __str__(self):
...
+    def successful_contribution(self, kwargs):
+        """Create a contribution object."""
+        contribution_kwargs = {
+            'tx_id': kwargs.tx_id,
+            'gas_price': kwargs.gas_price,
+            'nonce': kwargs.nonce,
+            'subscription': self
+        }
+        contribution = Contribution.objects.create(**contribution_kwargs)
+        # TODO: Email user that there payment has gone through
```
owocki: can you generate this email in this PR?

owocki: are there any other `TODO`s here?

captnseagraves: Sure, I can create the template and wire everything up. Will probably need to polish after the fact.

app/grants/models.py
captnseagraves: will do

app/grants/models.py
captnseagraves: I dont think there are any other todo's here

PR #3085, opened 2018-12-05
grants/new: updated look & feel

thelostone-mc:
Making the form similar to the `bounty/new`

Ignore the all blue thing -> gif recorder doesn't pick the color diff 😅 

untitled


owocki:
code looks fine to me.. left some UX comments on slack

Issue #3069, opened 2018-12-04
grants - failed tx's because approve() amount is wrong


owocki:
see dteails on https://gitcoincore.slack.com/archives/CBDTKB59A/p1543950428135100

we need to account for the cost of the subminer in the `approve()` txs

PR #3093, opened 2018-12-05
Ks miner fixes

captnseagraves:
##### Description

This pull request fixes bugs in subminer and has transactions consistently submitting successfully as well as saving contribution data. 

##### Checklist

- [ ] commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)


app/grants/management/commands/subminer.py
```diff
@@ -65,19 +61,20 @@ def process_subscription(subscription, live):
                 if live:
                     txid = subscription.do_execute_subscription_via_web3()
                     print(f"   -- *waiting for mine* (txid {txid}) ")
-                    while not has_mined(txid, subscription):
+                    while not has_tx_mined(txid, subscription.grant.network):
                         time.sleep(10)
-                    status, __ = get_tx_status(txid, subscription.grant.network, timezone.now())
+                        print("   -- *waiting 10 seconds*")
+                    status, timestamp = get_tx_status(txid, subscription.grant.network, timezone.now())
```
mbeacom: We're still not using `timestamp` anywhere here.

Can you revert this back to:  `status, __ = get_tx_status(txid, subscription.grant.network, timezone.now())`

app/grants/models.py
```diff
@@ -385,13 +385,6 @@ def _do_helper_via_web3(self, fn, minutes_to_confirm_within=5):
         signed_txn = web3.eth.account.signTransaction(tx, private_key=settings.GRANTS_PRIVATE_KEY)
         return web3.eth.sendRawTransaction(signed_txn.rawTransaction).hex()
 
-    def do_cancel_subscription_via_web3(self, minutes_to_confirm_within=5):
```
owocki: why was this removed?

app/grants/models.py
```diff
@@ -402,9 +395,11 @@ def do_execute_subscription_via_web3(self, minutes_to_confirm_within=5):
     def helper_tx_dict(self, minutes_to_confirm_within=5):
         """returns a dict like this: {'to': '0xd3cda913deb6f67967b99d67acdfa1712c293601', 'from': web3.eth.coinbase, 'value': 12345}"""
         from dashboard.utils import get_nonce
+        from dashboard.utils import get_web3
+        web3 = get_web3(self.grant.network)
         return {
             'from': settings.GRANTS_OWNER_ACCOUNT,
-            'nonce': get_nonce(self.grant.network, settings.GRANTS_OWNER_ACCOUNT),
+            'nonce': web3.eth.getTransactionCount(settings.GRANTS_OWNER_ACCOUNT),
```
owocki: why was this changed?  this will cause any pending transactions to pile on top of each other at the same nonce.

app/grants/models.py
```diff
@@ -474,22 +469,24 @@ def get_hash_from_web3(self):
...
-
-    def successful_contribution(self, kwargs):
+    def successful_contribution(self, tx_id):
         """Create a contribution object."""
         from marketing.mails import successful_contribution
         self.last_contribution_date = timezone.now()
-        self.next_contribution_date = timezone.now() + timezone.timedelta(seconds=self.real_period_seconds)
+        self.next_contribution_date = timezone.now() + timezone.timedelta(seconds=int(self.real_period_seconds))
         self.save()
+        # TODO: add gasPrice and nonce
```
owocki: whats wrong with how this was before?  how do you want to resolve this TODO thats different than how this was beofre?

app/grants/models.py
```diff
@@ -474,22 +469,24 @@ def get_hash_from_web3(self):
...
-            'tx_id': kwargs.tx_id,
-            'gas_price': kwargs.gas_price,
-            'nonce': kwargs.nonce,
+            'tx_id': tx_id,
             'subscription': self
         }
         contribution = Contribution.objects.create(**contribution_kwargs)
         grant = self.grant
-        grant.amount_received = (grant.amount_received + convert_amount(self.amount_per_period, self.token_symbol, "USDT", timezone.now()))
+        grant.amount_received = (
```
owocki: what if the token_symbol can't be converted to USDT?  seems to me we should handle this case..

owocki: i wonder if its worth making `amount_received` a JSON blob that has a keyvalue pair of all currencies being added to the grant

i.e.
```
{ 'WETH': 0.01, 'DAI': 12}
```

app/avatar/utils.py
```diff
@@ -71,109 +71,124 @@ def get_avatar_context():
...
-            'options': (
-                'anchors', 'circuit', 'jigsaw', 'lines', 'gears', 'clouds', 'signal', 'polka_dots', 'polka_dots_black',
-                'squares', 'shapes', 'sunburst', 'sunburst_pastel', 'rainbow',
-            ),
-            'paid_options': {
-                'sunburst_pastel': 0.01,
-                'rainbow': 0.01,
-            },
-        }],
+        },
```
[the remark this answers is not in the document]
captnseagraves: Sure. Just wondering though, why do we maintain them? `Make fix` changes them every time.

app/retail/emails.py
```diff
@@ -121,7 +121,7 @@ def render_successful_contribution_email(grant, subscription):
     params = {'grant': grant, 'subscription': subscription}
     response_html = premailer_transform(render_to_string("emails/grants/successful_contribution.html", params))
     response_txt = render_to_string("emails/grants/successful_contribution.txt", params)
-    subject = "Your subscription on Gitcoin Grants has been cancelled by the Grant Creator"
+    subject = "Your contribution of {{ subscription.amount_per_period }} {{ subscription.token_symbol }} to {{ grant.title }} was successful!"
```
owocki: is this meant to be an fstring?

app/grants/models.py
owocki: `get_nonce()` handles this scenario successfully

app/dashboard/utils.py
```diff
@@ -717,6 +718,7 @@ def get_nonce(network, address):
     try:
         nonce_from_db = JSONStore.objects.get(key=key, view=view).data[0]
         nonce_from_db += 1 # increment by 1 bc we need to be 1 higher than last txid
+        print('nonce_from_db', nonce_from_db)
```
owocki: extraneous print

app/grants/models.py
captnseagraves: The miner should never handle a `cancelSubscription` call. This is a direct transaction on chain performed by the user.

app/avatar/utils.py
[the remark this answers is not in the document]
owocki: +1

app/grants/models.py
owocki: these changes are in the model, not in the miner..

app/grants/models.py
captnseagraves: Any attempted transactions that didnt reach the chain increased the nonce in the DB which caused all transactions for the account to fail. I can change it back since the subminer is reaching the chain now and DB should be able to stay in sync.

owocki: >  I can change it back since the subminer is reaching the chain now and DB should be able to stay in sync.

makes sense to me

app/grants/models.py
owocki: > I do wonder why we save these though. Where do we intend to use them?

 +1 for removing them... i can't recall why we'd need them either

app/grants/models.py
owocki: ok but we can't merge this PR into the stable branch if theres a `ConversionNotFound` exception that's not handled

owocki: i'm not sure what the scope of this PR is so i can't say.

it doesnt seem like a blocking thing to me though if the scope of this is to cleanup the subminer

app/retail/emails.py
owocki: then the `""` should probably be preceded with an `f` like so `f""`

app/grants/models.py
captnseagraves: The rest of Grants calls the contract from javascript. `cancelSubscription` is handled in `grants/cancel_subsription.js` The miner is the only thing part of Grants that calls to the contract from the model. The miner will never call `CancelSubscription`. So it is unused code.

app/grants/models.py
captnseagraves: Gotcha. This PR is for `grantz-cleanup`. Not trying to move it `master`.  Wanting to get the functional miner where other can start using it though.

app/grants/models.py
owocki: while i agree that its not being called anywhere, i think its useful to have in the codebase in case we need it for administrative reasons.  checkout the precedent set at https://github.com/gitcoinco/web/pull/2725/files#diff-0552410df3354f05f06d86d522aab24fR622

captnseagraves: Makes sense

captnseagraves:
@owocki @mbeacom  Besides the token issues, which I hope to address today, is there anything else this PR needs to be merged into `grantz-cleanup`?

owocki:
> Besides the token issues

what are the token issues?  is there a github link?

app/grants/models.py
owocki: its only a couple line fix to wrap this code in this PR in a try/catch, but if you want to create a github issue to track this and do it later, that works for me.

owocki:
just left a comment or two. 

is the 'token issues' this ? https://github.com/gitcoinco/web/issues/3069

captnseagraves:
@owocki Token issues are the one you mention, https://github.com/gitcoinco/web/issues/3106,  https://github.com/gitcoinco/web/issues/3102 and  https://github.com/gitcoinco/web/issues/3103

PR #3116, opened 2018-12-06
Saving web3 tx_id's

captnseagraves:
##### Description

This PR saves all tx_id's for all web3 actions. This data can then be used to show the user proof of on chain action in waiting screen, transaction data, and emails.


thelostone-mc:
Closing this out in favor in favor of  #3117

PR #2967, opened 2018-11-27
[WIP] Notifications app

octavioamu:
##### Checklist

- [ ] linter status: 100% pass
- [ ] changes don't break existing behavior
- [ ] commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)

##### Refers/Fixes

Refs: #2404

Issue #2404, opened 2018-10-10
Build Front-End of Notifications for Gitcoin


PixelantDesign:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: Currently we email community members on things they need to act upon but there isn't a central place on Gitcoin that users can go to to view all notifications and take action.

### Description
This task is to build the front-end of the notification system.

### Definition of Done

- [x] Review the attached mockups and [Gitcoin's contribution guidelines](https://docs.gitcoin.co/mk_contributors/) and [front-end guidelines](https://docs.gitcoin.co/mk_styleguide/).
- [x] Add a way for users to access the notification dropdown (authenticated).
- [x] Users should not see the notification bell if anauthenticated.
- [x]  A notification count should be displayed at the top of dropdown
- [x] User should be able to view top notifications in the dropdown.
- [x] From the menu users should be able to and click **view al**l to see it all on a page.
- [ ] Dropdown should not show items that have been read.
- [x] The dropdown should show a green dot to denote unread
- [x] See all view should show all notifications.
- [ ] This should work in responsive views.
- [x] notifications can be selected or unselected.  there will be 'select all' or 'deselect all' buttons on the top, and there will also be 'mark selected as read', 'delete selected' and 'mark selected all as unread' buttons'
- [x] notifications will be sorted chronologically and paginated
- [x] notifications can be marked as read, marked as unread, or deleted
- [x] when a user clicks on the notification will be marked as read and takes user to the appropriate issue.
- [ ] PR is merged

Notifications will be generated each time a user needs to take action upon something (approve worker, disapprove worker, accept worker, receive tip, etc)

Here are the potential notifications we are hoping for:

**For Funders**

- You have workers to approve on this issue.
- Your bounty hunters haven't responded on this issue in a few days.
- Remove them if you haven't heard from them?
- There is a submission against your bounty.
- Your bounty hunter has stopped work
- Your bounty is expiring soonYour bounty has crowdfunding added to it.

**For Hunters**

- You haven't responded to this issue in x days.
- This issue has been remarketed and has your skill sets. Are you interested?
- You have been approved for a bounty
- You have been declined for a bounty
- You have been removed from a bounty due to no responseYou have received a tip.
- Your submission has been approved
- Your submission has been declined.
[...]

PR #3119, opened 2018-12-07
Changes to notifs backend

usmanmuhd:
- Returns only notifs for the particular user
- Made changes to models
- Added notifications to admin
- Delete a notification
- Mark a notif as read
- Mark a notif as unread
- Added helper function to generate new notif

Refs: #2404 and #2967


octavioamu:
Looking good @usmanmuhd !!

PR #3127, opened 2018-12-07
sets num kudos limit higher

owocki:
sets num kudos limit higher, so you can see 100 ppl who own those kudos instead of just 20

https://gitcoin.co/kudos/430/resilience

inspired by this kudos, which has 100 sends.

i hope this PR opens a discussion about performance and how many http requests were making... i'm open to disucssion about how to reduce server load here and still show the images


mbeacom:
Thanks!

Following our brief discussion about avatars and the load throughout kudos views due to the heavy reliance on our web server to middleman assets: https://github.com/gitcoinco/web/issues/3129

owocki:
awesome!

PR #3136, opened 2018-12-08
grant/new: add image

thelostone-mc:
##### Description

untitled

Ignore the weird off color that's cause of my svg recorded. need to find a new one


owocki:
This icon makes the page so much nicer.. Thanks!

Issue #3067, opened 2018-12-04
grants - migrate from 4 gwei to dynamic gas prices 


owocki:
grants - migrate from 4 gwei to dynamic gas prices (ideally, driven by the same logic as bounties, displayed below)

<img width="791" alt="screen shot 2018-12-04 at 11 36 29 am" src="https://user-images.githubusercontent.com/513929/49464697-dbe41080-f7b8-11e8-8807-947bc1327490.png">

thelostone-mc:
Todo - add dynamic gas widget on cancel grant popup

PR #3154, opened 2018-12-09
grant/new: add gas-setting

thelostone-mc:
This PR adds dynamic gas pricing to

- grants/new
- grants/fund
- grant/fund/cancel

Refs: https://github.com/gitcoinco/web/issues/3067

untitled

Questions ( @owocki / @captnseagraves  ):
- where else does this need be added ? 
  (If yes for subscription -> we have 2 contract deployment, so we use the same gas price for both ?)
- should we be converting `gas` in [grantz](https://github.com/thelostone-mc/web/blob/gas_setting/app/assets/v2/js/grants/new.js#L82) to hex like in [bounties](https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/pages/new_bounty.js#L424-L426)

### DIFF
--- a/app/assets/v2/css/base.css
+++ b/app/assets/v2/css/base.css
@@ -798,6 +798,11 @@ input[type=text].loading {
   display: inline-block;
 }
 
+#metamask_label img {
+  position: relative;
+  top: -2px;
+}
+
 #metamask_context .red_warning {
   background-color: #fbe0d6;
   color: #fb9470;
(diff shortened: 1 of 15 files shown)


thelostone-mc:
@SaptakS Two things 

- Could you throw the dynamic filter into grant/cancel 
-  Also for fund a grant and cancel a grant the `$` amount doesn't get populated even though the gas gets updated. I remember you doing something to fix this up in bounties 

Could you take care of that ? 

screen shot 2018-12-10 at 5 21 10 pm

thelostone-mc:
@owocki ah thanks for that ! that got it running
I've kept the amount as `318730`. Let me know if that needs be changed

Issue #3114, opened 2018-12-06
EPIC Signal looking for a Job (Contributor)


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a user, i want a 'im looking for a job' field in onboarding, so i can signal i want a job

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

gitcoin is a great place to hire, we just lack the data of whos on the market

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
new feature

### Current Behavior
[comment]: # (Describe what actually happened.)
Does not exist

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
User should be able to let Gitcoin know that they're looking for a job.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- [ ] In the on boarding flow determine where it makes sense to display a checkbox to capture that user is looking for a job.
- [ ] Display in profile
- [ ] In Settings display the same checkbox to indicate user is looking for a job.
- [ ] In every email that is sent, the system should remind me that i AM/AM NOT looking for a job, as of YYYY/MM/DD, so i remember to update it if it's changed.
- [ ] @frankchen07 should built a data dashboard of who's available for hire, ranked by whose done the most bounties.

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

kuhnchris:
![image](https://user-images.githubusercontent.com/761911/49698259-8d23e580-fbc1-11e8-8174-bc284e55e7ed.png)
added the "settings" side of things.
Just a problem with FTUX, I cannot make this as the "last" step because the only "writing back" is done during a step that's called multiple times, and sadly, that's not always the proper value...
Either I separate this and do that save back in it's own sub-path, or I reorder so that the looking for work step is coming earlier (after MetaMask for example)

kuhnchris:
Also, can someone (@mbeacom  ?) please explain to me why the "logic" for the Settings views is in app/marketing(/views.py), while the templates themselves live at app/retail ?

frankchen07:
Just a thought - here in Silicon Valley (or maybe everywhere), most people are "always on the market." So, even if I'm not actively looking for a job, I'm willing to click "I'm looking for a job," just because having more opportunities than not is better. They've got nothing to lose, right?

[...]

PR #3162, opened 2018-12-10
WIP - #3114 - Looking for work

kuhnchris:
##### Description
Implementation "looking for work" issue #3114


kuhnchris:
Let's throw this away, since we already have something like this in the hover/user directory, OK? @PixelantDesign

Issue #3013, opened 2018-11-29
BUILD - As a Gitcoiner, I'd like to have a weekly recap on how my bounties are doing


PixelantDesign:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a funder, I'd like to have a weekly recap on how my bounties are doing.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: If a funder has multiple bounties being worked on, it's hard to see the status of his/her bounties at a glance.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- [ ] Build an email for weekly recap emails
- [ ] This email should have a title
- [ ] The user name should be dynamic
- [ ] Please display all of the information in the mockup
- [ ] View Bounty links to the appropriate bounty
- [ ] Email must be responsive
- [ ] Review with core team members

### Reference

Please see #2544 

![48913556-b2bbaa00-eeb3-11e8-8333-ce2957339d08](https://user-images.githubusercontent.com/238235/49252924-54cb1d00-f3f3-11e8-9209-7fe52a09e366.jpg)

@willsputra could you provide the sketch file?

willsputra:
@PixelantDesign edited your original post to fix the link under 'References'.

Sketch file here!
https://drive.google.com/open?id=1aKejF0wEmBXk6ezjT6DpJBOB0j1sx9Wc

PixelantDesign:
@kuhnchris as we chatted earlier lets focus on bounties first and then if we have some time at the end add in kudos sent/received below that.

kuhnchris:
Hi there @owocki @mbeacom helped me fixed everything up, so I already managed to get enough bounties in:

![image](https://user-images.githubusercontent.com/761911/49834479-c8c6d700-fd9c-11e8-9a90-99019f18e646.png)

So those are already live data, (rinkeby), Links work, titles work, the / new_bounty is for me, since I am currently categorizing them. 

One more thing: the picture on the left, do you want it to be the default profile picture, or your own (e.a. the bounty's image) or the activity person's image? If so then we also should put in (for example for apply work or drop work or whatever) the name of the person doing the change, don't you think so too @PixelantDesign ?

PixelantDesign:
Wow looking great @kuhnchris!

[...]

PR #3217, opened 2018-12-14
[WIP] Weekly Founder Recap Mail 

kuhnchris:
##### Description
See discussion over at #3013


SaptakS:
Please run `make fix isort` to fix travis issues and push

PixelantDesign:
Hey @kuhnchris please see comments from @SaptakS 

Aautomation sending: you will have to first create a management command for the same and then run the management command to the cron tasks.

Please fix linting. You'll need to run the command `make fix isort` to fix linting and for travis to pass.

SaptakS:
We can always test it in staging without pushing into production to check how it looks. :) So I think it is better to get it in with the complete management command. Thanks.

kuhnchris:
Sure, no problemo. Will attempt to implement it today then.

kuhnchris:
One question remaining @PixelantDesign @SaptakS do we want to send "empty" bounty mails aswells, as in, "nothing changed, how about you put up a bounty" incentive, or do we just want to skip those people?

Thanks,
Chris

SaptakS:
I think skip. But @PixelantDesign  would be a right person to take a call on it.

PixelantDesign:
great suggestions @kuhnchris! I think we can open new tickets for those use cases.

app/retail/emails.py
```diff
@@ -818,10 +906,13 @@ def render_new_bounty_roundup(to_email):
...
     return response_html, response_txt, subject
 
+# DJANGO REQUESTS
 
 
-
-# DJANGO REQUESTS
+@staff_member_required
+def weekly_recap(request):
+    response_html, _ = render_weekly_recap("mark.beacom@consensys.net")
```
thelostone-mc: @kuhnchris are we leaving this hardcoded here ?

thelostone-mc:
@kuhnchris `weekly_founder_recap.txt` -> It's an empty file
Could we populate that with just data ?

Also hardcoding @mbeacom. Shouldn't we remove that? 

Could you address the stickler issues please ?

app/retail/emails.py
kuhnchris: This is just for /_administration/, we can put in someone else if you want, we just need someone who had regulary bounty activities so it outputs "something".

kuhnchris:
Ah the .txt one is for the non-html one, right? sure I can add a bit of content there.
Regarding mbeacom, it's just for administration/admin, we can add someone else if you have any suggestion.
addressing stickler, sure.

thelostone-mc:
@kuhnchris Ah it's good ^_^
will merge after CI goes green

kuhnchris:
ci won't go green. ever. until we fix it.

PR #3225, opened 2018-12-15
Fund with decimal, remove nonce, make recurring funding work

captnseagraves:
This PR: 
- [x] adds the functionality to fund with a decimal
- [x] adds gas for all future payment periods to the approve amount
- [x] removes the nonce from the contract and web3 functionality to allow recurring transactions


app/grants/models.py
```diff
@@ -370,8 +370,8 @@ def get_is_ready_to_be_processed_from_db(self):
         if not self.subscription_contribution.exists():
             return True
         last_contribution = self.subscription_contribution.order_by('created_on').last()
-        period = last_contribution.created_on + timedelta(0, round(self.real_period_seconds))
```
captnseagraves: Somone must have changes this underneath me. I assume both work. is there a preferred implementation

app/assets/v2/js/grants/fund.js
```diff
@@ -60,10 +60,10 @@ $(document).ready(function() {
 
         $('#gas_price').val(realGasPrice);
 
-        let realApproval = Number(((data.approve * 10 ** decimals) + realGasPrice) * data.amount_per_period);
-
         let realTokenAmount = Number(data.amount_per_period * 10 ** decimals);
 
+        let realApproval = Number((realTokenAmount + realGasPrice) * data.approve);
```
owocki: shouldnt this be (amount_in_tokens_per_period + gas_price_per_period) * num_periods ?

app/grants/models.py
mbeacom: @captnseagraves I did, as mentioned on slack.

```
period = last_contribution.created_on + timedelta(0, round(self.real_period_seconds))
return period > timezone.now()
```

This avoids the need to force `period` to `float` and each datetime to timestamp.  We can perform the same checks against the datetime object itself without the need for this.
Also, in the changes proposed here, you've flipped the operator.  This will now return `True` when last contribution + period is less than now.

app/grants/models.py
```diff
@@ -513,7 +513,7 @@ def successful_contribution(self, tx_id):
         """Create a contribution object."""
         from marketing.mails import successful_contribution
         self.last_contribution_date = timezone.now()
-        self.next_contribution_date = timezone.now() + timezone.timedelta(seconds=int(self.real_period_seconds))
+        self.next_contribution_date = timezone.now() + timezone.timedelta(seconds=float(self.real_period_seconds))
```
mbeacom: Quick question...  why are we dealing in floats/ints on decimalfields? Why are we even storing these as decimals if we're not using them as decimals in most places?

app/assets/v2/js/grants/fund.js
captnseagraves: @owocki that is what it is. realApproval = (converted_token_amount + converted_gas_price_per_period) * num_periods. I can change a variable name or two to make it more little more readable.

app/grants/models.py
captnseagraves: @mbeacom cool. if we dont need to force the type changes that if great. However, as I look at this I realize we can just use `next_contribution_date`, which i can calculate as `self.next_contribution_date = timezone.now() + timedelta(0, round(self.real_period_seconds))` in the `successful_contribution` function. We do want to flip the operator though. We want to know when the `next_contributoin_date` is less than now. This tells us that the subscription is now ready to be processed.

app/grants/models.py
captnseagraves: I don't have a good answer. This as likely due to my ignorance of django/python when creating these fields. What type should be used?

captnseagraves: Is this a blocker for this PR going out today?

captnseagraves:
@owocki @mbeacom is there an easy way to format grants/abi.py file without doing ti by hand? Stickler is not happy and `make fix` doesn't fix it.

PR #3262, opened 2018-12-19
opens up grants perms

owocki:
makes all pages visibile to all users except 'new grant', which is still limited to users we approve

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -55,12 +55,6 @@ def get_keywords():
 
 def grants(request):
     """Handle grants explorer."""
-    if not request.user.has_perm('grants.view_grant'):
-        params = {
-            'active': 'dashboard',
-            'title': _('Grants Explorer')
-        }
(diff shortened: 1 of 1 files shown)


mbeacom:
Uh. https://github.com/gitcoinco/web/tree/grants-launch

app/grants/views.py
```diff
@@ -375,7 +366,7 @@ def subscription_cancel(request, grant_id, grant_slug, subscription_id):
         return TemplateResponse(request, 'grants/shared/error.html', params)
 
     if request.method == 'POST' and (
-        profile == subscription.contributor_profile or request.user.has_perm('grants.change_subscription')
+        profile == subscription.contributor_profile
```
mbeacom: This means the user has *full* permissions to modify subscriptions.  Do we want to remove this?

thelostone-mc: Yeah could we test run it with permissions and then release this as a subsequent PR.
Would that be cool @owocki  ?

Issue #3278, opened 2018-12-22
CODE - As a funder I'd like to understand all of the offerings that Gitcoin has to help me develop, grow, and maintain my open source project.


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a funder I'd like to understand all of the offerings that Gitcoin has to help me develop, grow, and maintain my open source project.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
To understand Gitcoin's product suite.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
no `/products` page

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
`/products` page

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

Code this page on `/products` => 
![50039270-a0cdb280-ffec-11e8-8c91-5ad20d94cf71](https://user-images.githubusercontent.com/513929/50376457-268fc600-05ca-11e9-8495-7735799c848a.jpg)

Details
* Please make sure it looks good in mobile/tablet/desktop views.
* Please make sure to attach our standard footer.  The designer forgot to put the standard footer on there.

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)
DESIGN artifacts http://bits.owocki.com/e47a076a788d/product_final.sketch
http://bits.owocki.com/acddc53eeadd/alternative.svg
Design ticket https://github.com/gitcoinco/Marketing/issues/8

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 2 days, 23 hours from now.
Please review their action plans below:


**1) [mukulverm4](https://gitcoin.co/profile/mukulverm4) has started work.**

I will use the given design artifacts to develop the /products page.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/3278/2041).

kuhnchris:
Oh, guess someone was quicker than me, hahaha, no probs.
@mukulverm4 keep us posted, if you got a PR i'll be glad to review it ;-)

mukulverm4:
Okay, sure @kuhnchris. :)

[...]

Issue #2867, opened 2018-11-17
Community discussion -- What should live on https://gitcoin.co/


owocki:
what does the community think of changing our landing page at https://gitcoin.co/ to be an index of products, like https://occipital.com/ has ?

reason: it’s misleading to put the bounty product front and center when the gitcoin brand is really a suite of products.

we  could namespace everything for bounties under a `/bounties` path.

i’d even go so far as to say that `https://github.com/gitcoinco/web/issues/2778` could become the new `gitcoin.co` landing page, and we could move the facebook marketing targets to a new path namespaced under `/bounties`

owocki:
@pinkiebell  @kuhnchris @Anish-Agnihotri @nanspro would love ur feedback on this

kuhnchris:
It depends. You want to separate the "project" from a specific product? Let's make it even better and give them their dedicated subdomain. So gitcoin.co and www.gitcoin.co redirect to the landing page you proposed (and the old links still forward to the new bounties.gitcoin.co subdomain, so we do not break existing links), and the other "products" get their own subdomain. 
That'd currently be:
grants.gitcoin.co
bounties.gitcoin.co
kudos.gitcoin.co

AFAIK.
That would have multiple effect as in: gitcoin.co links still work and are valid, the user can be redirected after a period of "transition time" to the main page on gitcoin,co after we assume that most people switched over and the other "old" links would be dead anyways, we have a natural split between operation and project(s), and we do have the possibility to deploy parts of project(s) separatly and can update and enable/disable them on demand if there is some bug or issue or problem. (right now everything is one giant mass)

pinkiebell:
Hah, I actually waited for something like that.
So yes, a product overview as the landing page sounds good 👍

Anish-Agnihotri:
That certainly sounds like something I could see happening!

From a user perspective, the simplification of product-suite on one landing page, along with sub-domains if possible, as suggested by @kuhnchris, would be fantastic, and make the platform far more concise. 

From a developers perspective, I already only focus on the Issue Explorer to begin with + the black bar at the top to switch to and from Kudos, so I doubt it would disrupt my (or others) workflow as much. 

All in all it would certainly enhance the Gitcoin image/brand promise upon landing.

Might be worth looking at your heatmapping tools though (to see what the home-page is being most used for), so that you can plan ahead to divert that traffic - although I'm confident you and the team have already discussed that :).

Sounds good to me!

owocki:
[...]

PR #3303, opened 2018-12-26
Gitcoin is a suite of OSS incentivization tools.  This PR makes the new product page front and center

owocki:
Given that gitcoin is no longer just a bounty platform, and to clarify the brand navigation, i believe it would behoove us to make the landing page at gitcoin.co/ representative of our suite of products and how they fit together.

This PR makes the new product page ( https://github.com/gitcoinco/web/issues/3278#issue-393685127 ) front and center, (it will load on https://gitcoin.co/)  For targeted campaigns, we can and should still still send users to `/funder` or `/contributor`.

If we achieve consensus that this should be merged, I think we should also
1. Put all of the current brochureware pages (products, mission, about, etc -- pictured [here](https://github.com/gitcoinco/web/issues/2867#issuecomment-449581486)) into a mini nav as specified in [this design](https://github.com/gitcoinco/web/issues/2867#issuecomment-449581486)
1. update the nav on all of the brochureware pages to be a brochureware-specific nav (just like we have a kudos nav, grants nav, etc)
1. namespace all of the brochureware pages to `/about/mission`, `/about/not_a_token`, etc
1. namespace all of the bounty-specific stuff to `/bounties/explorer`, `/bounties/spacemeshos/app/21/1556`, etc
1. whatever updates we make, make sure we make the same updates on codefund.

Doing the above would make it clear to the user that Gitcoin is a parent brand, and that Kudos, Grants, CodeFund, Bounties, are all sub-brands.


app/retail/templates/shared/top_nav.html
```diff
@@ -18,7 +18,7 @@
 {% load i18n static matches %}
 <div class="top-nav g-font-muli {{class}}">
   <a class="link d-md-block d-none" href="{% url 'products' %}">Grow Open Source</a>
-  <a class="{% if request.path == '/' %}selected{% endif %}" href="/">
+  <a class="{% if request.path == '/' %}selected{% endif %}" href="{% url 'funder' %}">
```
mbeacom: Should we keep these consistent? 🤔

owocki: good catch.  will update IFF @PixelantDesign is okay with the direction i'm taking this.

vs77bb:
I think we need to add this to the Blue Sprint board. Thoughts, @PixelantDesign @frankchen07?

frankchen07:
I think this might belong on Green Squad - I was poking around to see if there was any type of hypothesis testing. There might be some reusable stuff that we can re-purpose for the "that was easy" value proposition.

thelostone-mc:
Closing this out as we've got the name homepage coming in and that would address this

PR #3324, opened 2018-12-27
grants: disable cancel action if metamask address doesn't match

### LINKED ISSUE #3247: Add Address Warning - Grants
As a user, I want to know when my metamask is set to a different address than the one I used to create a grant or subscription so that I dont try and submit a transaction that will fail. 

This error keeps happening in testing and we need to check what wallet a user is using, create a warning to let the user know, as well as disable any necessary buttons so that a user can't process any transaction until they change to the correct wallet address. 

To get this done we should add a check in web3 enabled views that metamask is using the same address the user created the grant or subscription with. If the address does not match we should provide an alert(). This will prevent users from getting mysterious error messages with no apparent cause.

thelostone-mc:
##### Description

untitled

This closes #3247

### DIFF
--- a/app/assets/v2/js/grants/cancel_subscription.js
+++ b/app/assets/v2/js/grants/cancel_subscription.js
@@ -2,6 +2,7 @@
 $(document).ready(() => {
   $('#period').select2();
 
+  setInterval (notifyOwnerAddressMismatch, 1000);
 
   $('#js-cancelSubscription').validate({
     submitHandler: function(form) {
@@ -12,7 +13,6 @@ $(document).ready(() => {
       });
 
       // need to delete subscription from miner so it isn't checked every 15 seconds.
-      //
 
       let deployedSubscription = new web3.eth.Contract(compiledSubscription.abi, data.contract_address);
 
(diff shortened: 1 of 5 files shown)


app/assets/v2/js/grants/cancel_subscription.js
```diff
@@ -78,3 +78,22 @@ $(document).ready(() => {
...
+const notifyOwnerAddressMismatch = () => {
+  if (!web3 || !web3.eth)
+    return;
+  web3.eth.getAccounts((error, accounts) => {
+    if (document.contxt.github_handle == $('#subscriber-handle').val() &&
+        accounts[0] != $('#subscriber-address').val()) {
+      if ($('#cancel-subscription').attr('disabled') != 'disabled') {
+        $('#cancel-subscription').attr('disabled', 'disabled');
+        _alert({
+          message: 'Looks like you\'ve funded this grant with ' + $('#token_address').val() + '. Switch to that to take action on your subscription.'
```
owocki: seems like this would be more appropriate as a tooltip on the cancel button instead of a giant red banner on the page when i go to it.

app/assets/v2/js/grants/detail.js
```diff
@@ -190,3 +190,22 @@ const copyDuplicateDetails = () => {
     });
   });
 };
+
+const notifyOwnerAddressMismatch = () => {
```
owocki: can we DRY this with the other `notifyOwnerAddressMismatch`?

thelostone-mc: Done 👍  thanks for catching that

Issue #3348, opened 2018-12-29
Bounty API 500 response


kziemianek:
**Describe the bug**
Week or two ago there was no problem with accessing single bounty through endpoint `https://gitcoin.co/api/v0.1/bounties/<bounty_pk>/`, today I noticed that api returns http code 500 for every request.

`{"active": "error", "code": 500, "title": "Error 500"}`

**To Reproduce**
curl -v https://gitcoin.co/api/v0.1/bounties/9/

**Expected behavior**
response 200 with bounty json

JeremiGendron:
I will check the logs

kuhnchris:
Guess that's an issue with the router string, I'll check for you in a bit.

Issue #623, opened 2018-03-14
Refactor the Gitcoin API 


owocki:
##### Description
router.py

We need to refactor quite a bit of the API using more functionality in `django-filter. DjangoFilterBackend` and the `OrderingFilter`

##### Possible Implementation
See above ^^

gitcoinbot:
__This issue now has a funding of 0.05 ETH (32.9 USD @ $658.0/ETH) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/issue/gitcoinco/web/623).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/issue/gitcoinco/web/623)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $3593.01 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 0.05 ETH (25.98 USD @ $519.59/ETH) funding by__: 
 1. [@zoek1](https://gitcoin.co/profile/zoek1) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

 On the above list? Please leave a comment to let the funder (@owocki) and the other parties involved what you're working, with respect to this issue and your plans to resolve it.  If you don't leave a comment, the funder may expire your submission at their discretion. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/issue/gitcoinco/web/623)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $2777.46 more Funded OSS Work Available at: https://gitcoin.co/explorer

zoek1:
@owocki the description is some confused to me, could you detail ir more?

owocki:
@zoek1 - sure; the basic idea behind this issue is to take all of the functionality that is in dashboard/router.py and move it to a more standard direction for django-request-framework handling ( mentioned `django-filter. DjangoFilterBackend` and the `OrderingFilter` classes as starting points)... 

@mbeacom anything to add here?

zoek1:
[...]

PR #3364, opened 2018-12-31
fix for Bounty API 500 response

owocki:
fixes https://github.com/gitcoinco/web/issues/3348 by making a new queryset


app/dashboard/router.py
```diff
@@ -269,6 +270,18 @@ def get_queryset(self):
...
+        pks = queryset.values_list('pk', flat=True)
+        queryset = Bounty.objects.filter(pk__in=pks)
+
+        # order
+        # this is applied twice.. if you change this block, 
+        # also change the block 20 lines up
+        order_by = self.request.query_params.get('order_by')
+        if order_by and order_by != 'null':
+            queryset = queryset.order_by(order_by)
+
```
SaptakS: is it worth making this into a DRY function so can be reused rather than writing twice (or maybe more times in future?) ? @owocki @mbeacom

mbeacom: @SaptakS I don't disagree.  We should DRY this up, but we should probably address: #623 sooner rather than later.

If we're documenting to update both, should we consider moving these off to another method and/or dropping one of them outright?

app/dashboard/router.py
```diff
@@ -269,6 +270,18 @@ def get_queryset(self):
             end = start + int(limit)
             queryset = queryset[start:end]
 
+        # workaround for https://github.com/gitcoinco/web/issues/3348
+        pks = queryset.values_list('pk', flat=True)
+        queryset = Bounty.objects.filter(pk__in=pks)
```
mbeacom: I don't think we should need to perform a whole new query here 🤔  Looks like the shelf life of our logic has been exhausted. If we weren't slicing the QS on [line 271](https://github.com/gitcoinco/web/pull/3364/files#diff-a0ceb478149ae8595c669666edec2decR271), this wouldn't be an issue.  Thoughts on eliminating this in favor of standard DRF pagination/sorting/filtering?

owocki: whats the time estimate on that?  i haven't seen much action on the refactor ticket, don't even know if it's current

app/dashboard/router.py
kziemianek: Seems like https://stackoverflow.com/a/38390480/2582437 is exactly what is needed here but i'm not aware of performance.

SaptakS: I would go for moving this off to another method.

app/dashboard/router.py
SaptakS: I am not sure of the time estimate. If it is a priority I can also work on the DRF. Even if not now, DRF pagination does sound the way to go. Even if not that, how about storing the non-sliced QS in another variable `queryset_full` and then later just do `queryset = queryset_full`. At least that will reduce the number of queries to the DB which is usually a very costly operation. cc @mbeacom @owocki

PR #3486, opened 2019-01-10
active bounties list in admin and change "too much work" calculation

owocki:
shows active bounties list in admin


app/dashboard/models.py
```diff
@@ -2078,6 +2078,11 @@ def get_profile_referral_code(self):
     def job_status_verbose(self):
         return dict(Profile.JOB_SEARCH_STATUS)[self.job_search_status]
 
+    @property
+    def active_bounties(self):
+        active_bounties = Bounty.objects.current().filter(idx_status__in=['open', 'started'])
```
octavioamu: @owocki should this be changed to count only the ones I got approved to work? I know that will introduce a problem if someone apply to work in many things at the same time, but for other side a contributor will be blocked to work if doesn't cancel the "start work" or get rejected by the funder.

Issue #3489, opened 2019-01-10
grants -- brand image is stretched on /fund page


owocki:
![screencapture-gitcoin-co-grants-24-prysm-by-prysmatic-labs-fund-2019-01-10-08_18_48](https://user-images.githubusercontent.com/513929/50977860-6272c880-14b0-11e9-84f7-8e9da6968ffe.png)

we probably dont want to screw up our funders logos :)

thelostone-mc:
Hmm I can't seem to reproduce it :P 

<img width="1440" alt="screen shot 2019-01-10 at 10 04 10 pm" src="https://user-images.githubusercontent.com/5358146/50982709-ca162b80-1523-11e9-9078-ad7ce0686cf6.png">

owocki:
1800x1200 is my reso.. 

try going to the `/fund` page to repro it @thelostone-mc !

owocki:
not page, not grant detail page

PR #3495, opened 2019-01-10
fix: #3489

thelostone-mc:
fixes: #3489

untitled


app/assets/v2/css/grants/fund.css
```diff
@@ -3,6 +3,16 @@
...
 
+.banner-img {
+  max-height: 22rem;
+  overflow-y: hidden;
+}
+
+.banner-img img {
+  height: auto;
+  width: 100%;
+}
```
octavioamu: you can use `object-fit: cover;` to be sure is always is going to cover the space.

owocki:
screenshot looks great

PR #3501, opened 2019-01-10
Ks browser inbetween

captnseagraves:
This PR saves grants/new data immediately upon transaction submission and should prevent the loss of most data, except for contract_address. Solution for contract address to come in following PR. 

I will shortly be implementing a similar solution for grants/fund. We can either wait for that, to be delievered in the next hour, or I can submit an additional PR.


app/assets/v2/js/grants/fund.js
```diff
@@ -70,6 +70,8 @@ $(document).ready(function() {
 
           $('#contributor_address').val(accounts[0]);
 
+          let url;
```
thelostone-mc: @captnseagraves Doesn't look like this url is being assigned a value anywhere ! Do we need it?

app/assets/v2/js/grants/fund.js
```diff
@@ -104,19 +133,38 @@ $(document).ready(function() {
...
           }).on('confirmation', function(confirmationNumber, receipt) {
-            $('#real_period_seconds').val(realPeriodSeconds);
 
             waitforData(function() {
-              $.each($(form).serializeArray(), function() {
-                data[this.name] = this.value;
-              });
               document.suppress_loading_leave_code = true;
-              form.submit();
+              window.location = url;
```
thelostone-mc: This would result in `/undefined` added at the end of the url cause it's not assigned a value right ?

app/assets/v2/js/grants/fund.js
thelostone-mc: init as empty string?

app/assets/v2/js/grants/fund.js
```diff
@@ -104,19 +133,38 @@ $(document).ready(function() {
...
+                      'signature': signature,
+                      'csrfmiddlewaretoken': $("#js-fundGrant input[name='csrfmiddlewaretoken']").val(),
+                      'sub_new_approve_tx_id': $('#sub_new_approve_tx_id').val()
+                    };
+
+                    $.ajax({
+                      type: 'post',
+                      url: '',
+                      data: data,
+                      success: function(json) {
```
thelostone-mc: arrow functions ? :P

app/assets/v2/js/grants/fund.js
```diff
@@ -104,19 +133,38 @@ $(document).ready(function() {
...
+                        console.log('successfully saved subscriptionHash and signature');
+                        url = json.url;
+                        $('#wait').val(false);
+                      },
+                      error: function() {
```
thelostone-mc: arrow functions ? :P

app/assets/v2/js/grants/fund.js
```diff
@@ -104,19 +133,38 @@ $(document).ready(function() {
...
+                        _alert({ message: gettext('Your subscription failed to save. Please try again.') }, 'error');
+                      }
+                    });
+                  }
                 });
               });
             });
           }).on('confirmation', function(confirmationNumber, receipt) {
```
thelostone-mc: arrow functions ? :P

app/assets/v2/js/grants/fund.js
```diff
@@ -104,19 +133,38 @@ $(document).ready(function() {
...
-            $('#real_period_seconds').val(realPeriodSeconds);
 
             waitforData(function() {
```
thelostone-mc: arrow functions ? :P

app/assets/v2/js/grants/fund.js
```diff
@@ -150,7 +198,7 @@ $(document).ready(function() {
 });
 
 var waitforData = function(callback) {
-  if ($('#signature').val() != '') {
+  if ($('#wait').val() == false) {
```
thelostone-mc: could we use `if (!$('#wait').val()) {` OR `if ($('#wait').val() === false) {`

app/assets/v2/js/grants/new.js
```diff
@@ -51,7 +51,10 @@ $(document).ready(function() {
         data[this.name] = this.value;
       });
 
+
       $('#token_symbol').val($('#js-token option:selected').text());
+      $('#token_address').val($('#js-token option:selected').val());
+      console.log($('#token_address').val());
```
thelostone-mc: need to remove this before merge :P

app/assets/v2/js/grants/new.js
```diff
@@ -92,19 +95,61 @@ $(document).ready(function() {
...
+                    },
+                    error: function() {
+                      _alert({ message: gettext('Your grant failed to save. Please try again.') }, 'error');
+                    }
                   });
-                  document.suppress_loading_leave_code = true;
-                  form.submit();
+
                 } else {
                   setTimeout(function() {
```
thelostone-mc: setTimeout(() =>  {

app/assets/v2/js/grants/new.js
```diff
@@ -92,19 +95,61 @@ $(document).ready(function() {
...
+
+                  $.ajax({
+                    type: 'post',
+                    url: '',
+                    data: data,
+                    success: function(json) {
+                      document.suppress_loading_leave_code = true;
+                      window.location = json.url;
+                    },
+                    error: function() {
```
thelostone-mc: `error:() => {`

app/assets/v2/js/grants/new.js
```diff
@@ -92,19 +95,61 @@ $(document).ready(function() {
...
+                    'contract_address': result.contractAddress,
+                    'csrfmiddlewaretoken': $("#create-grant input[name='csrfmiddlewaretoken']").val(),
+                    'transaction_hash': $('#transaction_hash').val()
+                  };
+
+                  $.ajax({
+                    type: 'post',
+                    url: '',
+                    data: data,
+                    success: function(json) {
```
thelostone-mc: `success: json => {`

app/assets/v2/js/grants/new.js
```diff
@@ -92,19 +95,61 @@ $(document).ready(function() {
...
+            };
+
+            $.ajax({
+              type: 'post',
+              url: '',
+              data: data,
+              success: function(json) {
+                console.log('successfully saved grant');
+              },
+              error: function() {
```
thelostone-mc: arrow

app/assets/v2/js/grants/new.js
```diff
@@ -92,19 +95,61 @@ $(document).ready(function() {
...
+              'network': $('#network').val(),
+              'team_members': $('#input-team_members').val(),
+              'csrfmiddlewaretoken': $("#create-grant input[name='csrfmiddlewaretoken']").val()
+            };
+
+            $.ajax({
+              type: 'post',
+              url: '',
+              data: data,
+              success: function(json) {
```
thelostone-mc: arrow function

app/assets/v2/js/grants/fund.js
```diff
@@ -150,7 +198,7 @@ $(document).ready(function() {
 });
 
 var waitforData = function(callback) {
```
thelostone-mc: can we update this function to a `const`  as opposed to `var` to avoid hoisting

app/assets/v2/js/grants/fund.js
owocki: +1

app/assets/v2/js/grants/fund.js
```diff
@@ -83,6 +85,33 @@ $(document).ready(function() {
             $('#sub_new_approve_tx_id').val(transactionHash);
             const linkURL = etherscan_tx_url(transactionHash);
 
+            let data = {
```
owocki: if `$form` is your form, you can just use `$form.data()` and avoid this big dict creation i think!

app/assets/v2/js/grants/fund.js
```diff
@@ -83,6 +85,33 @@ $(document).ready(function() {
...
+              'network': $('#network').val(),
+              'csrfmiddlewaretoken': $("#js-fundGrant input[name='csrfmiddlewaretoken']").val()
+            };
+
+            $.ajax({
+              type: 'post',
+              url: '',
+              data: data,
+              success: function(json) {
+                console.log('successfully saved subscription');
```
owocki: shouldnt we do more than console log here?

owocki:
@captnseagraves whats the state machine for each grant now? 

create grant: (created, has no contract address) => (gets contract address) 

fund grant: (created, has no ??) => (fully created) 

do we need to do anything to the subminer or the views to help them handle new steps in the state machine?

app/assets/v2/js/grants/fund.js
captnseagraves: I will go with solution number one

app/assets/v2/js/grants/fund.js
captnseagraves: The only thing is that I dont want all data in the form to be passed through. That way I can do `if` statements in the view based on the content of the `POST`. If there is a better workaround for that happy to implement it!

app/assets/v2/js/grants/fund.js
captnseagraves: I agree, it would be better not just have a log, but we also dont want to take any action. We just want to next ajax to fire. Do you have any suggestions for what to add here? @owocki?

captnseagraves:
> @captnseagraves whats the state machine for each grant now?
> 
> create grant: (created, has no contract address) => (gets contract address)
> 
> fund grant: (created, has no ??) => (fully created)
> 
> do we need to do anything to the subminer or the views to help them handle new steps in the state machine?

@owocki this PR is an intermeadiate fix. Right now the flow is:

Create Grant: (created and data saved with out contract address immeadiately, once web3 gets confirmation second ajax sent and contract address saved) 
- If the user leaves the page we would need to manually get the contract address and save it. 
- the full fix will be to have a listener watch the transaction and save the contract address and/or flip a confirm_on_chain boolean. 

Fund Grant: (Sub data saved immeaditely w/ approval tx id, then once user signs subscription secind ajax saves sub_hash and contributor_signature) 
- if user leaves the page before signing the sub, which likely wouldnt happen, theyd have to redo the subscription, but if they leave after signing as they wait for the approval transactoin to go through we're all good. 
- the full fix will be to have a listener watch the approval transaction and flip a confirm_on_chain boolean. 

Does that answer your question? Do you have any suggestions?

app/assets/v2/js/grants/fund.js
captnseagraves: Opted for solution two as I also needed to change the value to a string. My original solution worked because the loose equals allowed me to check a string vs. a boolean. JavaScript 🤷‍♂️

owocki:
@captnseagraves can you rebase on master or merge master in?  travis build is failing based upon 

```
    from eth_utils.toolz import curry
ModuleNotFoundError: No module named 'eth_utils.toolz'
The command "pytest -p no:ethereum" exited with 1.
```

which has been fixed in upstream master

owocki:
> Does that answer your question? Do you have any suggestions?

👍 sounds good

PR #3577, opened 2019-01-21
svg to png refactor

owocki:
Over at https://github.com/gitcoinco/web/pull/3249#issuecomment-456205093 we are having many a problem with pyvips and librsvg2-dev / imagemagick.

The purpose of this PR is to create a workaround strategy for converting an SVG to a PNG.    this is encapsulated in `avatar.utils.svg_to_png`.  which will first try to create a png using pyvips, and if it fails, will do so with inkscape.  futher conversions could be added in the future if this doesnt work around.

to play with this

First
```
apk install inkscape
```

then use either 
1. the kudos display funcitonality ( http://localhost:8000/dynamic/kudos/146/web3_advocate ) 
2. make a new avatar.
3. run the `./manage.py debug_test` management command 

to test


app/avatar/utils.py
```diff
@@ -479,6 +479,65 @@ def get_temp_image_file(url):
     return temp_io
 
 
+def svg_to_png(svg_content, width=100, height=100, scale=1):
+    print('creating svg with pyvips')
+    png = svg_to_png_pyvips(svg_content, scale=scale)
+    if not png:
+        print("failed; using inkscape")
+        return svg_to_png_inkscape(svg_content, height=height, width=width)
+    return None
```
kziemianek: I think you should return `png` here if vips conversion succeds.

thelostone-mc:
Closing this out as we've already got this covered in https://github.com/gitcoinco/web/pull/3577

PR #3599, opened 2019-01-23
web3 wallet callout proof of concept

owocki:
#  Why

Sometimes the metamask popup does not show up ( https://github.com/MetaMask/metamask-extension/issues/3759#issuecomment-456908970 )..  In this scenario, the user may click a button expecting a web3 action then think "WTF" when nothing happens.  After a little thinking and hunting around, they will see a little icon in their top right.  
screen shot 2019-01-23 at 11 34 19 am
  But only after looking and even then its small

# Solution

I propose that any web3 action on our site create a callout in the top right asking the user to check their web3 wallet.

Here is a demo (make sure you keep your eye on the top right).

screen recording 2019-01-23 at 11 32 am

# Testing

I have tested this on the tip flow, but still need to integrate it with all web3 actions on our site.  At that point, I can test it site wide.

# TODOS

* [ ] Achieve consensus on this being something we actually want to do.
* [ ] Make sure the design is 👍 
* [ ] Make sure the design on mobile is 👍 
* [ ] Integrate it everywher.
* [ ] Merge and ship.

### DIFF
--- a/app/assets/onepager/js/send.js
+++ b/app/assets/onepager/js/send.js
@@ -281,6 +281,7 @@ function sendTip(email, github_url, from_name, username, amountInEth, comments_p
         };
 
         if (isSendingETH) {
+          prompt_for_web3_wallet_action();
           web3.eth.sendTransaction({
             to: destinationAccount,
             value: amountInWei,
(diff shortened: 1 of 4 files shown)


willsputra:
@owocki 
web3


Sketch file: https://github.com/gitcoinco/creative/pull/46/commits/3d61f7d4ed27feff9a21608285019886dbffa2f9

thelostone-mc:
Closing this out as it's catered to in the fee PRs. Thanks @owocki  ^_^

Issue #3289, opened 2018-12-24
DESIGN - as a user, i want a pricing sheet, so i can see how to take full advantage of gitcoin


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a user, i want a pricing sheet, so i can see how to take full advantage of gitcoin
### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
explains all our revenue options in one place

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:

### Current Behavior
[comment]: # (Describe what actually happened.)
no price sheet

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
remove existing price sheet from gitcoin.co

add this one, which will be put at gitcoin.co/pricing:

![group 3](https://user-images.githubusercontent.com/513929/50389018-5aa9da80-06e1-11e9-8e6f-fa62d04888e0.png)

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

* Page at `/pricing`
* Page Works Responsive
* Put a menu item that points at `/pricing` in the top right nav.
* If I select a plan, then I am taken to a gitcoin grant payment page to pay for that plan. (all info about my pricing will beg prefilled)
* If I select the plan, then all of the promised benefits will be there.

### Additional Info
http://bits.owocki.com/9e5bd92b1e54/Pricing.sketch

owocki:
cc @vs77bb @PixelantDesign

kuhnchris:
*raises hand* please provide the assets in a non-sketch format, thanks

PixelantDesign:
Cool! I'll take up this one @willsputra

owocki:
when this is ready for coding, will do.

thanks @PixelantDesign !

owocki:
another thing we could add to premium featureset only
- funder dashboard
- notifications
- weekly roundups
- instant new issue notifications
- ENS subdomains
- browser extension

@vs77bb @PixelantDesign thoughts on this?  basically anything that's not a basic feature could / should be premium only

kuhnchris:
[...]

PR #3604, opened 2019-01-23
WIP -- Subscription Model Scaffolding

owocki:
This PR creates scaffolding for the subscription model.

See also
* [Original Ticket](https://github.com/gitcoinco/web/issues/3289)
* [Architecture Doc](https://docs.google.com/document/d/1KlRa3CNwOlnZYWnknm4SvJcyfJ6InCiBmCNLeWzXujw/edit#)

TODOS 
 * [x] DB Architecutre
 * [x] DB Entries for new subscription plan
 * [ ] Implement Internal APIS
 * [ ] Hook it up to the frontend
 * [ ] ....
 * [ ] ....
 * [ ] ....
 * [ ] Lots of testing
 * [ ] Ship It


app/revenue/models.py
```diff
@@ -19,13 +19,131 @@
...
+    sku = models.ForeignKey(
+        'revenue.SKU',
+        related_name='alacartegoodpurchases',
+        on_delete=models.CASCADE,
+        help_text=_('The feature that was purchased'),
+        null=False,
+    )
+
+
+class SKU(SuperModel):
```
thelostone-mc: what does SKU mean ? @owocki

app/revenue/models.py
```diff
@@ -19,13 +19,131 @@
...
+from economy.models import SuperModel
 
 
-class DigitalGoodPurchase(SendCryptoAsset):
-    """Model that represents a DigitalGoodPurchase.
+class ALaCartePurchase(SendCryptoAsset):
+    """Model that represents a ALaCartePurchase.
 
     """
     purchase = JSONField(default=dict, blank=True)
```
thelostone-mc: JSONField ? what would go in there ?

app/revenue/models.py
```diff
@@ -19,13 +19,131 @@
...
+class Plan(SuperModel):
+    """Model that represents a Subscription Plan
+
+    """
+    slug = models.CharField(max_length=255, help_text=_('The Slug'))
+    name = models.CharField(max_length=255, help_text=_('The SKU Name'))
+    cost_per_period = models.DecimalField(
+        decimal_places=2,
+        max_digits=50,
+        help_text=_('Cost of this plan in USD'),
```
thelostone-mc: Should we make it to DAI? It doesn't matter but yeah

app/revenue/migrations/0003_auto_20190123_2218.py
```diff
@@ -0,0 +1,102 @@
...
+        migrations.CreateModel(
+            name='Plan',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_on', models.DateTimeField(db_index=True, default=economy.models.get_time)),
+                ('modified_on', models.DateTimeField(default=economy.models.get_time)),
+                ('slug', models.CharField(help_text='The Slug', max_length=255)),
+                ('name', models.CharField(help_text='The SKU Name', max_length=255)),
+                ('cost_per_period', models.DecimalField(decimal_places=2, help_text='Cost of this plan in USD', max_digits=50)),
+                ('period_length_seconds', models.PositiveIntegerField()),
```
danlipert: Looking at the spec and planning docs for this feature, I'm wondering if this will lead to confusion among users when the plan needs to be recharged every 30 days instead of on the same day each month. Looks like a 30-day month is used to calculate the period. A slightly more complicated path forward could use a `DurationField` to store a `timedelta` of 1 month and use that for plan expiry calculations. A simple path forward using the current code would be to change the pricing language from per month to per 30 days.

app/revenue/utils.py
```diff
@@ -17,3 +17,9 @@
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 """
+import random
+import string
+
+
+def sku_generator(size=6, chars=string.ascii_uppercase + string.digits):
```
danlipert: If the SKU represents individually purchased items, especially the ad impressions, we should worry about collisions here with only 6 digits (will likely occur in under 100,000 instances)

app/revenue/models.py
danlipert: Typically a SKU would represent a type of product (a specific car model, like a Red Honda Civic 2019 with sunroof, for example). I think in this case its representing a specific type of A La Carte item purchased (a codefund impression, or kudos, or job board posting).

app/revenue/utils.py
danlipert: Reading the code over more I think the SKU refers to a category of item, with the `PlanItem` being the individually purchased item. If so, is there a reason to randomly generate these or should we just write human-readable SKUs?

danlipert:
This scaffolding looks good - I just had one concern regarding the 30 day duration of plans and a question about the random SKU generation. 👍

PR #3605, opened 2019-01-23
runs linter on the codebase

owocki:
what the title says

am i the only one who keeps getting conflicts on these files because they change every time?


thelostone-mc:
thanks for your contribution @owocki  :) :P

PR #3683, opened 2019-02-05
Fixes links

decanus:
The links on the grants page don't work. I assume this fixes it.


octavioamu:
Anyway I believe that information can came from the server and not from the metamask network

PR #3689, opened 2019-02-06
Rename LISCENSE.txt to LICENSE.txt

abitrolly:
[skip ci]


abitrolly:
Never thought it could be referenced, and it is.

thelostone-mc:
@abitrolly mind updating the references ?

abitrolly:
Just discovered that `grep` is case sensitive by default. Updated.

Issue #3593, opened 2019-01-22
BUILD - As a Gitcoin user, I want a pricing sheet, so I can see how to take full advantage of Gitcoin.


frankchen07:
### User Story

As a Gitcoin user, I want a pricing sheet, so I can see how to take full advantage of Gitcoin.

### Why Is this Needed

This page explains all our revenue options in one place. 

### Description

We can build the initial design right now as a skeleton start, and modify it in the future with incoming funder feedback. This way, we iterate proactively, and not wait until we have perfect information to start the pricing page.

### Current Behavior

No pricing sheet.

### Expected Behavior

1. Remove existing pricing sheet from gitcoin.co.
2. Have the new pricing sheet on a development branch (NO PROD LAUNCH) at gitcoin.co/pricing:

Invision link: https://projects.invisionapp.com/share/42PZCI9CUXB#/340972484_User_Test_-_Home

![pricing_am](https://user-images.githubusercontent.com/238235/50674652-481f8300-0fb6-11e9-865f-32bbedb87c3e.png)

3. Design discussion will continue on https://github.com/gitcoinco/web/issues/3289

### Definition of Done

* Main pricing page on a development branch of `/pricing`.
* Page works and is responsive.
* Selection of a plan leads to a grant payment page for plan payment with information about plan pricing pre-filled. (scope reverted back to full pricing flow, before backend changes 2019-02-05)

PixelantDesign:
Should we build just the one page or the entire experience?

thelostone-mc:
@frankchen07 + @PixelantDesign moving forward could we ensure the sketch file is attached wherever applicable

frankchen07:
> moving forward could we ensure the sketch file is attached wherever applicable 

yes definitely - thanks for the callout!

> Also, I'm moving forward under the assumption that this task has only thing page to be built and nothing else

yes, the task is to build the pages based on the sketches, but no deployment yet. Pricing is part of a larger epic, but having the foundation for the page laid out will reduce future cycles and holdups.

PixelantDesign:
I'll be updating this design

owocki:
> I'll be updating this design based on feedback received.

excited for the next rev on this :) do you have any idea of what youre doing for the new pricing page?  my biggest asks for the next rev of the page:

1. it should be simpler to understand
2. it should be compelling (if i'm in the target user flow, it should be a no brainer, i should want to "pull out my credit card" )

 happy to jam.

frankchen07:
[...]

PR #3703, opened 2019-02-08
pricing + subscription 

thelostone-mc:
#### Description

- added pricing page with nav + redirects to subscription
- build out FE for subscription page with nav. (No BE wiring)
- both views can be accessed by staff only
- code cleanup + create slider component
- fixed cookie UI

https://aditya.stage.gitcoin.co/pricing/

Refs: https://github.com/gitcoinco/web/issues/3593

@danlipert if we were to move to the no of users based as opposed to monthly packs
It might need a few tweaks but nothing major ^_^


danlipert:
@thelostone-mc Whats the login for the staging site? When I try to visit the pricing page it redirects me to the _admin login

app/assets/v2/css/lib/slider.css
```diff
@@ -0,0 +1,46 @@
+.g-switch input { 
+  display:none;
```
danlipert: Why is there different formatting/style in this CSS file? Also, should we get a CSS style checker/linter going?

app/assets/v2/js/animate-number.js
```diff
@@ -0,0 +1,3 @@
+(function(d){var r=function(b){return b.split("").reverse().join("")},m={numberStep:function(b,a){var e=Math.floor(b);d(a.elem).text(e)}},g=function(b){var a=b.elem;a.nodeType&&a.parentNode&&(a=a._animateNumberSetter,a||(a=m.numberStep),a(b.now,b))};d.Tween&&d.Tween.propHooks?d.Tween.propHooks.number={set:g}:d.fx.step.number=g;d.animateNumber={numberStepFactories:{append:function(b){return function(a,e){var f=Math.floor(a);d(e.elem).prop("number",a).text(f+b)}},separator:function(b,a,e){b=b||" ";
```
danlipert: eslint really hates this file 🤣 I think we'll need to mark it as a pass from the linter somehow

app/retail/views.py
```diff
@@ -200,6 +200,127 @@ def index(request):
     return TemplateResponse(request, 'home/index.html', context)
 
 
+@staff_member_required
+def pricing(request):
+
+    plans= [
```
danlipert: Nice configurability! Should we move this into an import to keep this file a little leaner? Also I'm wondering if we should keep this loaded in memory instead of loading it up for each request

docs/STYLEGUIDE.md
```diff
@@ -23,7 +25,18 @@ _usage_
 </div>
 ```
 
-The form styling used in Gitcoin can be found in
+The Multiselect styling used in Gitcoin can be found in
 [select.css](https://github.com/gitcoinco/web/blob/master/app/assets/v2/css/forms/select.css)
 
-_Note: All pages within gitcoin are expected to reuse these classes as applicable as opposed to declaring the `font` within the templates `css` file._
+### Slider
```
danlipert: 💟

app/retail/views.py
SaptakS: @danlipert I do agree with maybe importing. We can actually make a constants.py file and would preferably need cleaning a lot of views. But I didn't get what you mean by keeping this loaded in memory.

app/assets/v2/js/animate-number.js
SaptakS: This is a minified file so no wonder eslint hates it. @thelostone-mc I think just rename it to animate-number.min.js and it will be taken care of.

app/assets/v2/css/lib/slider.css
thelostone-mc: You mean the spacing right? I'll fix that up and yeah we should

app/retail/views.py
thelostone-mc: A single constants file ? where would this reside ? 
I'd be happy to throw that in now or a follow up PR i have coming in a few days

SaptakS: I think a follow up PR makes sense while cleaning every such occurence over different views.

danlipert: @SaptakS @thelostone-mc I think a constants file is a good idea in general, but what I meant by my comment was a bit different: Specifically I'm talking about taking the final `context` variable and moving it outside of the `def pricing(request):` handler definition.

I like to declare psuedo "constants" in all capital letters below the imports. Since this one is many lines of code, it makes sense to import it and keep this file tidy. By doing this, the handler gets a bit speedier in that it doesn't have to re-create the `context` for every request, it already has it available in memory to read and return. We get the same functionality but better performance at the cost of a bit of memory usage.

PR #3706, opened 2019-02-08
preserve manually featured bounties

danlipert:
##### Description

Preserves bounties that have been manually featured when the bounty is updated


app/dashboard/helpers.py
```diff
@@ -485,6 +485,11 @@ def create_new_bounty(old_bounties, bounty_payload, bounty_details, bounty_id):
                 new_bounty.canceled_on = canceled_on
                 new_bounty.save()
 
+            # preserve featured status for bounties where it was set manually
```
owocki: i wonder if we can do this in a more elegant way up on line 404, by updating `bounty_kwargs`

thelostone-mc: As opposed to doing save at every , would it make sense to do it all in a single `new_bounty.save()` ?

Issue #3570, opened 2019-01-20
Some images not uploading for Gitcoin Grants


ceresstation:
**Describe the bug**

In some cases image uploads are failing on Gitcoin Grants. For example, POA tried to upload an image for this grant https://gitcoin.co/grants/35/blockscout-blockchain-explorer-for-inspecting-and but the default image still displays.

**To Reproduce**

Create a grant and upload any jpg image, in the case above it was a ~450 x 350 image as shown below.

https://gyazo.com/9bf4c0260434b7a93d8e0280b6ba224d

**Expected behavior**

Images upload and are displayed on each grant unless the max file size is exceeded (and are ideally adjusted to fit into the space available).

thelostone-mc:
Unable to add Grant logos during creaation :

```
usecase : `grant/new`
issue: when you add a grant image and submit -> the logo doesn't get saved
why : cause we haven't added it to the ajax request 

@captnseagraves since we are sending a file -> we'll  have to convert it to form data and send it via ajax but I had some issues with it when I gave it a shot

Input field in html : https://github.com/gitcoinco/web/blob/master/app/grants/templates/grants/new.html#L112

Where is it to be added while submitting the form
https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/grants/new.js#L85-L99

the server logic : https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L203
```

PR #3710, opened 2019-02-08
Ks grant image fix

captnseagraves:
This PR addresses https://github.com/gitcoinco/web/issues/3570

Grants/new should now be saving images properly.


app/assets/v2/js/grants/new.js
```diff
@@ -82,26 +82,31 @@ const init = () => {
...
-              'token_symbol': $('#token_symbol').val(),
-              'amount_goal': $('#amount_goal').val(),
-              'contract_version': $('#contract_version').val(),
-              'transaction_hash': $('#transaction_hash').val(),
-              'network': $('#network').val(),
-              'team_members': $('#input-team_members').val(),
-              'csrfmiddlewaretoken': $("#create-grant input[name='csrfmiddlewaretoken']").val()
-            };
+            let file = $('#img-project')[0].files[0];
+            let formData = new FormData();
```
danlipert: Since we're using `FormData()` here, why not skip all the append calls below and use `FormData(document.getElementById('create-grant'))`?

thelostone-mc: If we are doing that , we'd have to update the html ids to match like `input_team_members` now become `team_members`

captnseagraves: @danlipert The main reason is that the first POST call will then contain all input id's. Currently, we use the id's being sent through as a way to differentiate between the first (saving most data) and second (saving contract_address and redirecting page) POST calls. This way we can perform different actions after saving data in grants/views.py. If there is a better way to achieve this functionality, I am all for best practices. This was just the solution I came to. Let me know what you think!

PR #3731, opened 2019-02-12
fix: updated django 2.1.5 -> 2.1.6

thelostone-mc:
Fix security vulnerability by upgrading to 2.1.6
No breaking changes ! Locally app boots up while after rebuilding the app

https://docs.djangoproject.com/en/2.1/releases/2.1.6/

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ bs4
 ccxt
 cryptocompare==0.6
 cryptography==2.3
-django==2.1.5
+django==2.1.6
 django-cors-headers==2.4.0
 django-filter==2.0.0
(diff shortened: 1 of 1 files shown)


SaptakS:
Honestly, @owocki I think this could have been one of the reasons for the `Segmentation Fault (Core Dumped)` because seems like this issue was actually causing memory exhaustion, and we did use file uploading, which might have caused it. Just a hypothesis.

owocki:
fwiw i did try upgrading django in order to solve the seg fault. if its the case that this might have been the root cause... then maybe if orgot to restart gunicorn or something after it, when i tested

PR #3758, opened 2019-02-15
hide why-gitcoin for logged in users

thelostone-mc:
As requested by @PixelantDesign


owocki:
@thelostone-mc @PixelantDesign i dont understand the rationale for this. the mission/vision/results pages are all still important reminders of why the user should care about using us, even if they've already converted.

octavioamu:
I agree in general having this unexpected changes in nav could be frustrating to users trying to find something they saw in the menu. I think a better architecture of nav will help more

thelostone-mc:
I tend to agree with @octavioamu + @owocki .
Defer to @PixelantDesign.

PR #3805, opened 2019-02-21
Restore fixed version of grants js

danlipert:


owocki:
what PR broke this functionality?  Should we reopen whatever ticket was associated with that functionality so that we can fix the root cause, but also still manage that ticket correctly?

octavioamu:
Not idea what was the history in grants or why was broken, anyway is working fine now

PR #3858, opened 2019-02-27
integrate codefund into bounty landing pages

thelostone-mc:
##### Description

- Codefund ads are wired into the funder + contributor landing page.
- We've also got ourselves registered over at codefund as a publisher 
- A few ad campaigns don't load. 
The codefund guys are looking into it but ideally that would not change how this works 

I've thrown in some custom CSS with `!important` so as to override the styling provided by codefund cause it breaks the UI

untitled


danlipert:
Nice and clean integration! Is there a open issue for this PR? I'm wondering if putting this on the explorer makes sense... also it's a bit difficult to see the ad here, or tell that its an ad. I see the codefund examples often use a square display area, I'm wondering if using that layout might work better, or may solve the issues with some campaigns not working.

thelostone-mc:
> I'm wondering if putting this on the explorer makes sense... also it's a bit difficult to see the ad here, or tell that its an ad. 

It could be made a lil more distinct. 
Defer to @PixelantDesign  as this layout + design was picked out . by her

> I'm wondering if using that layout might work better

That would fit well into the explorer where we have tons of space on the left side 

With the new homepage coming in -> not sure if integrating codefund page would be as effective as it would be right now

thelostone-mc:
@danlipert  synced up with @PixelantDesign  and it looks like we'll be rolling with this for now and update it later !

PR #3861, opened 2019-02-27
fix create metamask without metamask data

octavioamu:
##### Checklist

- [ ] Read and conforms to the [contributor guidelines](https://docs.gitcoin.co/mk_contributors/)
- [ ] Read and conforms to the [style guidelines](https://docs.gitcoin.co/mk_styleguide/)
- [ ] Linter status: 100% pass
- [ ] Added relevant tests
- [ ] Tests pass and code coverage has not decreased
- [ ] Changes don't break existing behavior
- [ ] Commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)
- [ ] [Code documentation](https://docs.gitcoin.co/mk_contributors/#docstrings)
- [ ] Completed: Code review by core team and any requested changes

###### Contributor

- [ ] Read and followed the [Contributor Guidelines](https://docs.gitcoin.co/mk_contributors/)
- [ ] Tested all changes **locally**
- [ ] Verified existing functionality
- [ ] Ran `make test` and everything passed!

###### Reviewer

- [ ] Affirm contributor guidelines have been followed and requested changes made
- [ ] CI tests and linting pass
- [ ] No conflicts (migrations, files, etc)
- [ ] Regression tested against staging or local deployment

###### Funder

- [ ] Validated requested changes were made to specification
- [ ] Bounty payout released to the contributor

PR #3863, opened 2019-02-27
grants/new hackfix

owocki:
block user from continuing if they haven't initd after a second on the page
replaces https://github.com/gitcoinco/web/pull/3861

### DIFF
--- a/app/assets/v2/js/grants/new.js
+++ b/app/assets/v2/js/grants/new.js
@@ -5,8 +5,21 @@ $(document).ready(function() {
     web3.eth.net.isListening((error, connectionStatus) => {
       if (connectionStatus)
         init();
+        document.init = true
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/grants/new.js
```diff
@@ -5,8 +5,21 @@ $(document).ready(function() {
     web3.eth.net.isListening((error, connectionStatus) => {
       if (connectionStatus)
         init();
+        document.init = true
     });
   }
+  setTimeout(function(){
+      if(!document.init){
+        $('#connect_metamask_error').css('display', 'block');
```
owocki: for some reason i can't access `show_error_banner()` here... maybe a js variable scoping issue? because its defined in document.ready

PR #3905, opened 2019-03-06
adds profile images to header

owocki:
i propose that we make the following change to the gitcoin profile.  see below (left is before, right is after)

business reason: identity is important, and we invested a ton of time/resources into the avatar builder.  we should showcase the beautiful vector avatar identities our system can create front and center

also i propose that you can edit your avatar inline if its your profile 
the hover invites u to update your avatar

screen recording 2019-03-08 at 12 03 pm


owocki:
what it looks like without an avatar

screen shot 2019-03-06 at 11 24 53 am

owocki:
TODO: clean up resonsive design 

screen shot 2019-03-06 at 11 25 22 am

owocki:
fixed the responsive design issues

screen shot 2019-03-06 at 12 01 05 pm

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
 .no-profile-textcolor {
   color: #666666;
 }
+
+#float_avatar_image{
+  z-index: 9999999;
```
octavioamu: can we put just 9 or something lower so we don't if modal or popovers comes in some point don't get behind?

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+  margin-top: -140px;
+}  
+
+#float_avatar_image img{
```
octavioamu: can we change this to class? is easy to overwrite and maintain/reuse

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+  height: 186px;
+  width: 186px;
+  border: 3px solid #FFFFFF;
+  box-shadow: 0 -2px 4px 0 rgba(0,0,0,0.50);
+}  
+
+#float_avatar_image.is_my_profile:hover img{
```
octavioamu: if u use .float_avatar_image in the other place here you just need .is_my_profile:hover without need to overwrite.

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+
+.is_my_profile{
+  cursor: pointer;
+}
+#avatar_edit{
+  text-align: center;
+  background-color: white;
+  border-radius: 50%;
+  opacity: 0.4;
+  filter: alpha(opacity=40);
```
octavioamu: this can be removed

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+  height: 186px;
+  width: 186px;
+}
+#avatar_edit_text {
+  position: absolute;
+  top: -25px;
+  left: 55px;
+  color: black;
+  font-weight: bold;
+  z-index: 999999;
```
octavioamu: Can be lower index to don't overlap modals, popovers, tooltips. Always try to use the index you just need instead of maximum index.

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+  font-size: 20px;
+}
+#avatar_edit:hover {
+  opacity: 0.7;
+  filter: alpha(opacity=70);
```
octavioamu: this can be removed

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+}
+#avatar_edit:hover #avatar_edit_text {
+  color: #aaaaaa;
+}
+
+@media (max-width: 1500px) {
```
octavioamu: remove this?

app/assets/v2/css/profile.css
```diff
@@ -306,3 +310,84 @@
...
+    #float_avatar_image{
+      margin-top: -270px;
+    }  
+}
+
+@media (max-width: 790px) {
+    #float_avatar_image{
+      margin-top: -260px;
+    }  
+    #avatar_edit {
```
octavioamu: using bootstrap display clases will be better I think. like d-md-none will hide the element when md media query is displayed.
Also I think now is ok but using media queries is always better to go mobile first, so instead of max-width 790px going min-width 790px and the default value of margin-top in float_avatar_image with the one needed to the smallest display

app/dashboard/templates/profiles/profile.html
```diff
@@ -26,12 +26,30 @@
       {% endif %}
     </div>
     <div class="container-fluid bg-light pb-5">
-      {% if not hidden %}
+      </div>
         <div class="container profile-card">
           <div class="row">
             <div class="col-12 col-lg-6 profile-header__main-infos">
+              {% if not hidden %}
+              <div id="float_avatar_image" class="{% if is_my_profile %}is_my_profile{% endif %}">
```
octavioamu: Avoid id for styling proposes as much as you can, in that way the code is more modulable, reusable and maintainable. 
also the stylesheet get more comprensible without having to use hierarchy classes  to be able to style when reused or hover states.

app/dashboard/templates/profiles/profile.html
```diff
@@ -26,12 +26,30 @@
...
+                    <div id="avatar_edit">
+                      <div id="avatar_edit_text">
+                        Edit Avatar
+                      </div>
+                    </div>
+                  </a>
+                {% endif %}
+              </div>
+
               <div class="clearfix">
```
octavioamu: is this needed? I think can be removed

app/assets/v2/css/profile.css
owocki: fixed

app/dashboard/templates/profiles/profile.html
owocki: no idea, not part of my code :/

owocki:
i also updated the PR desc for easy reviewing!

app/dashboard/templates/profiles/profile.html
octavioamu: since now is an empty div can be removed without problem

octavioamu:
can we add to L18  into the div the class `pb-5` to avoid  this:
image

owocki:
@octavioamu good catch!  mind pulling the branch and making that update? im not sure exactly what you mean here.

octavioamu:
@owocki done!

owocki:
thanks octavio!  looks really good :)

PR #3944, opened 2019-03-10
Pinned Kudos on Gitcoin Profiles

owocki:
##### Description
Pinned Kudos on Gitcoin Profiles

##### Refers/Fixes

Profile, Kudos

##### Testing and Sign-off

I tested this, and coded it fairly defensively.

##### View

Screen Recording 2019-03-09 at 05 20 PM

##### Edit
Screen Recording 2019-03-09 at 05 20 PM_1


app/assets/v2/js/pages/profile.js
```diff
@@ -8,6 +8,22 @@ $(document).ready(function() {
...
+  $('#kudos-section').on('click', '.pin_to_profile', e => {
+    e.preventDefault();
+    var pk = $(e.target).data('pk');
+    var position = parseInt(prompt('Which position do you want to pin this kudos? (1-5)', '1'));
+    var is_invalid = isNaN(position) || position < 1 || position > 5;
+
+    if (is_invalid) {
+      _alert('Herp Derp!  Invalid input');
+    } else {
+      var target_url = window.location.href.split('?')[0] + '?action=mutate_pinned_kudos&pk=' + pk + '&position=' + position;
```
rafalkowalski: You can use template string here: 
```
const base_url =  window.location.href.split('?')[0];
var target_url = ${base_url}?action=mutate_pinned_kudos&pk=${pk}'&position=${position}
```

app/dashboard/views.py
```diff
@@ -1315,6 +1315,40 @@ def profile_filter_activities(activities, activity_name):
...
+
+    """
+    if not pk.isdigit():
+        return
+    if not position.isdigit():
+        return
+    position = int(position)
+    if position < 1 or position > 5:
+        return
+    kudos_transfer = KudosTransfer.objects.filter(pk=pk)
```
danlipert: can just use .get() here and then refactor some of the lower code to match

app/dashboard/views.py
```diff
@@ -1315,6 +1315,40 @@ def profile_filter_activities(activities, activity_name):
...
+    Returns:
+        nothing
+
+    """
+    if not pk.isdigit():
+        return
+    if not position.isdigit():
+        return
+    position = int(position)
+    if position < 1 or position > 5:
```
danlipert: What do you think of using a `settings.py` entry for the max number of pins? Its mentioned a few times throughout the code and would be good to avoid having a magic number

app/marketing/management/commands/bootstrap_pinned_kudos.py
```diff
@@ -0,0 +1,48 @@
...
+        for username in set(usernames):
+            print(username)
+            profiles = Profile.objects.filter(handle__iexact=username.replace('@', ''))
+            if profiles.exists():
+                print('-')
+                profile = profiles.first()
+                kudos = profile.get_my_kudos.order_by('-id')
+                counter = 0
+                for kudo in kudos:
+                    counter += 1
```
danlipert: Is this command meant to be used in production to set up an initial state for everyone's pinned kudos or is it just for testing?

app/marketing/management/commands/make_all_avatars.py
```diff
@@ -0,0 +1,33 @@
+'''
```
danlipert: I think this file is in here by mistake?

owocki:
from @PixelantDesign 
```
The idea is to show at a glance what the top Kudos the user has right? Should users be able to select which are shown or should we default to last 5?

I like the general idea. Maybe evening out space could help or moving them below the avatar, lined up. If the goal is to give a preview of Kudos the person has.
```

let me know if you want me to just do my best on this @PixelantDesign or if you or @willsputra can give an updated design for this, and i can code thatin

octavioamu:
Is this ready to review?
@PixelantDesign and @willsputra what do you think about this?

danlipert:
Is this PR still valid or are we reworking the design?

thelostone-mc:
^ @owocki / @PixelantDesign  could we get an update on this ^_^

owocki:
i dont think its a priority, so lets close it for now.

Issue #3910, opened 2019-03-06
DESIGN - Design a "nice bot" svg that is situated in the top left corner asking the user to check the Metamask popup when double transactions are necessary.


frankchen07:
### User Story

DESIGN - Design a "nice bot" svg that is situated in the top left corner asking the user to check the metamask popup when double transactions are necessary.

### Why Is this Needed

Context: https://gist.github.com/thelostone-mc/dd89b2824f1c5243fd256a902bb2b29b#gistcomment-2855545

For certain Metamask transactions (Grants, and also the contract % fee), two transactions will be necessary for the user to proceed. Although this UX tradeoff is undesired, it is currently better than heavily modifying the contract (high LOE).

The purpose of the nice bot svg is to prompt the user to look out for that second transaction.

### Current Behavior

No way to notify the user that there's a second Metamask transaction that often doesn't pop up (remains in the top right hand corner in the browser).

### Expected Behavior

Nice bot SVG that directs the user to accept the second transaction.

### Definition of Done

- [ ] Nice bot SVG design that can be added to the appropriate pages while the user is waiting to click on Metamask.

### Data Requirements
- [ ] Google analytics tracking for dropoff once the % fee is released.

### Additional Information
Previous work: https://github.com/gitcoinco/web/pull/3599
@PixelantDesign / @willsputra can you attach the design collateral to this ticket? Depending on Alisa's sign off, this design could already be done!

owocki:
does what @willsputra did on https://gitcoincore.slack.com/archives/CCZFZDT9V/p1551894024002400 count?

frankchen07:
@owocki - yes it does look like what we're looking for.

Ideally, I'd like this up this sprint, but that's contingent on hooking everything up. Generally looks like good progress though.

PixelantDesign:
Can we try something a little more explicit?

<img width="615" alt="Screen Shot 2019-03-13 at 4 10 13 PM" src="https://user-images.githubusercontent.com/238235/54311351-ca29ba80-45aa-11e9-84e5-adab086be27f.png">
[Web3 CalloutAM.zip](https://github.com/gitcoinco/web/files/2963471/Web3.CalloutAM.zip)

frankchen07:
@PixelantDesign I like the wording better on the second (lower) design. It's more succinct. The sentence on the first one (upper) is not a complete sentence, so it didn't make sense to me.

I'd add two things, a comma after "network" and Ethereum should be capitalized.

[...]

Issue #3909, opened 2019-03-06
BUIDL - As a Gitcoin operator, I want to create a dispute resolution faucet to facilitate funder fee refunds.


frankchen07:
### User Story

BUIDL - As a Gitcoin operator, I want to create a dispute resolution faucet to facilitate funder fee refunds.

### Why Is this Needed

Context: https://gist.github.com/thelostone-mc/dd89b2824f1c5243fd256a902bb2b29b#gistcomment-2855545

This is the place where funders can get the fee refunded. for posting a bounty (once the % fee is implemented)

### Description 

1. The funder fee is charged upfront immediately after posting a bounty.
2. If the funder decides to cancel, they will be redirected to the dispute resolution "faucet" in order to get their funds back (for now, reviewed by the Gitcoin team).

We can repurpose gitcoin.co/faucet designs.

### Current Behavior

No faucet or dispute resolution page.

### Expected Behavior

A faucet / dispute resolution page.

### Definition of Done
- [ ] if a funder cancels, the flow should direct them to the refund / dispute resolution faucet.
- [x] sufficient details shoudl be provided to the Gitcoin team (amount to be refunded, bounty issue, cancellation reason, free text reasoning)

### Data Requirements
- [x] ensure that successful and failed refunds are tracked in the db

thelostone-mc:
A few clarifications :
- Can requests be submitted only on cancellation ?
- Should we force users to redirect on cancel bounty / leave a button on cancelled bounty page making it optional? 
- Should the button also be on the bounty details page ?


^ If so could we get mocksups for it ?

frankchen07:
> Can requests be submitted only on cancellation? 

I think requests can be submitted freely as well, and there are two situations that I can see where that's appropriate. This means that the refund page should be accessible on the issue details page as well as on the redirect after cancellation.

1. After cancelling, the funder accidentally closes the fee refund page. They can still access it through the issue details page. 

2. The funder would like a refund on part of the fee (e.g featured bounties fee but not the platform fee).

> Should we force users to redirect on cancel bounty / leave a button on cancelled bounty page making it optional?

I would say it's only fair that we notify them after they cancel a bounty that there is the option to have the fee refunded. To me that's just platform honesty. I am a yes on redirecting a user to the request bounty fee refund page after cancellation.

> Should the button also be on the bounty details page?

Yes, in my opinion, if the two scenarios described above make sense. Are there any other scenarios that make sense in this situation?

I would add some changes to the copy below:

[...]

Issue #3891, opened 2019-03-05
BUIDL - As a Gitcoin operator, I want to properly communicate charging a % fee on a funder for funding a bounty. 


frankchen07:
### User Story

BUIDL - As a Gitcoin operator, I want to properly communicate charging a % fee on a funder for funding a bounty. 

### Why Is this Needed

As we roll out revenue tests and charging funders a percentage fee for posting bounties, the messaging and story arc behind this change is very important. This ticket is the build out of that.

### Description

@thelostone-mc should have the sketch files for the build.

As discussed (2019-03-14):

1. We're keeping the featured bounty fee and the platform fee separate because if we combine those to keep the Metamask popups to a minimum the denomination of different currencies will make that difficult, if not impossible. We'll keep them separate and add messaging in two ways: 
    - messaging under the summary
    - utilizing the blue alert shade at the top
    - Metamask confirmation bot (https://github.com/gitcoinco/web/issues/3910)

This allows ETH or ERC-20 tokens for the fee, ETH only for the featured bounty fee, and ETH or ERC-20 tokens for the bounty.

2. Eventually, we'll want to add currency support for featured bounties which may help the Metamask UX issue, but that's not in scope this round, as if we did end up allowing currency switching, it would add alot of work for this sprint. 

@danlipert & @thelostone-mc - ^ please correct me if I'm wrong or if there are any modifications, or clarifications you guys need. 

### Current Behavior

No messaging or build-out on the funder page.

### Expected Behavior

Fee-description and logic on the funder page.

Design ticket here: https://github.com/gitcoinco/web/issues/3842

### Definition of Done

- [x] messaging on fee is clear
- [x] front-end funder form messaging and elements added
- [x] front-end logic and math for adding the fee is implemented

### Data Requirements
- [ ] bounty amount and any associated fees are stored in the DB by bounty

frankchen07:
Adding Aditya's work preview here - this is when the bounty is funded in ETH:

![Untitled](https://user-images.githubusercontent.com/7516920/54375001-74eab900-463d-11e9-9972-d39e8ed7d95d.gif)

Successful creation: 

<img width="1164" alt="Screen Shot 2019-03-09 at 11 58 09 PM" src="https://user-images.githubusercontent.com/7516920/54375052-8cc23d00-463d-11e9-94c7-ddc895eb8253.png">

When the user rejects the bounty fee:

[...]

PR #3961, opened 2019-03-13
bounties: fee integration

thelostone-mc:
##### Description

- Bounty Fee Integration
- Refund Fee Flow
- Metamask notification
- Ensure db isn't updated when user selects featured bounty and then rejects bounty

Check out [gist for info](https://gist.github.com/thelostone-mc/dd89b2824f1c5243fd256a902bb2b29b)

closes https://github.com/gitcoinco/web/issues/3891
closes https://github.com/gitcoinco/web/issues/3910
closes https://github.com/gitcoinco/web/issues/3909

Pending:  @frankchen07 

- email wiring (designs needed)
- enabling fee for increase bounty fee (design needed)


frankchen07:
@danlipert @thelostone-mc 

First off, great work. Thanks for the amazing teamwork on this, and working through the decisions on contracts and stuff. +1 for Dan putting it up on staging 5 minutes after I asked for it too. Just finished looking at this on staging, and the functioning looks smooth. Some comments below from Aditya's asks.

1. We'll hold off on the email designs. Because this is a lightweight test, I would say we can remove the "request fee refund" option on the issue details page. If they decide to cancel, they can contact us via email / feedback for refunds and we can deal with it ad-hoc. Since we have the code for it already, we can always reactivate it when necessary. Our goal here is to gauge reactions and see what happens.

Screen Shot 2019-03-20 at 10 47 15

2. For the fee charge when increasing the bounty amount, I actually don't see an option to increase the bounty amount (example below). That being said, this is also an edge case that we don't need to worry about as of right now for the purposes of this revenue test. My hypothesis is that the majority of folks posting on Gitcoin aren't trying to game the system. 

Screen Shot 2019-03-20 at 11 07 53

3. I know we store featured bounty revenue in our DB, is the % fee also stored in our DB, and does it have a corresponding bounty_id that we can link to? This is mainly for data reporting purposes, as well as having the proper information to process refunds if necessary from our admin.

thelostone-mc:
@frankchen07  

> actually don't see an option to increase the bounty amount (example below).

The increase bounty is the contribute button & I've seen it used by a few funders.
Click on contribute -> and then increase funding.
Every bounty has this option available to the bounty funder.
I've got the code nearly done but I can revert it if we still feel it's not needed 

>  is the % fee also stored in our DB, and does it have a corresponding bounty_id that we can link to?

we store the % fee directly in the bounty table

Screen Shot 2019-03-21 at 11 39 51 AM

app/assets/v2/css/submit_bounty.css
```diff
@@ -146,6 +146,25 @@ textarea:read-only {
...
+  position: relative;
+  opacity: 0.12;
+  top: 2.2rem;
+}
+
+.fee-circle p {
+  color: #3E00FF;
+}
+
+#total-section {
```
octavioamu: can we avoid id for styles? maybe `.border-top` can be useful here

app/dashboard/templates/bounty/process_refund_request.html
```diff
@@ -0,0 +1,164 @@
...
+{% block breadcrumbs %}{% endblock %}
+
+{% block content %}
+<style>
+
+  body {
+    font-family: 'Muli';
+  }
+
+  #content {
```
octavioamu: `.m-auto`

app/dashboard/templates/bounty/process_refund_request.html
```diff
@@ -0,0 +1,164 @@
...
+    font-size: 0.95rem;
+    margin-top: 1rem;
+    margin-bottom: 1rem;
+  }
+
+  p {
+    font-weight: 600;
+  }
+
+  .section {
```
octavioamu: `.mb-5`

app/dashboard/templates/bounty/process_refund_request.html
```diff
@@ -0,0 +1,164 @@
...
+    margin-bottom: 2rem;
+  }
+
+  textarea {
+    width: 100%;
+    height: 150px;
+    margin-bottom: 1rem;
+  }
+
+  .text-center {
```
octavioamu: already exist in bootstrap

octavioamu: wait is this page not using bootstrap?

app/assets/v2/css/submit_bounty.css
thelostone-mc: The styling is specific to that element. I could make it into a class but since this isn't used anywhere, would it give us any benefit? 🤔  
border-top doesn't give the color we need so I'd still have to add `border-color`.
Do we still wanna do that?

app/dashboard/templates/bounty/process_refund_request.html
thelostone-mc: This is on the admin view! 
Bootstrap styles aren't imported so `m-auto` would not work

app/dashboard/templates/bounty/process_refund_request.html
thelostone-mc: same as [here](https://github.com/gitcoinco/web/pull/3961/files#r269418938)

app/dashboard/templates/bounty/process_refund_request.html
thelostone-mc: same as [here](https://github.com/gitcoinco/web/pull/3961/files#r269418938)
Yup it is not using boostrap as this page is viewed on the django admin. Adding boostrap just for this would make it an overkill 😅

Issue #3881, opened 2019-03-05
Build Front-End Contributor Landing Page Updates


PixelantDesign:
### User Story

As a marketer, I'd like to have the ability to customize certain sections fo the contributor landing page so that it is more robust and has enough info to engage/entice that audience.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: Contributor landing pages are limited now.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Definition of Done

- [ ] Update contributor landing page template so that it follows this new format
- [ ] Add the necessary tags to improve SEO/ organic growth reference comments from Marc on [this ticket](https://github.com/gitcoinco/web/issues/3749)
- [ ] Make sure these changes are implemented on other landing pages (languages)
- [ ] Review with the marketing team

### Reference
[Dean's Google Doc](https://docs.google.com/document/d/1UvG4uJIUrTRkeDXVvaTkq_bVO3ZoccaQIH-19bxrZCg/edit?usp=sharing)

![53750733-8f0df400-3ee5-11e9-94ae-cc0c616468ac](https://user-images.githubusercontent.com/238235/53812048-a15d5180-3f28-11e9-94a7-af0afa280290.jpg)

@willsputra please attache sketch file here.

PixelantDesign:
3

PR #3972, opened 2019-03-15
rebirth: contributor page

thelostone-mc:
- purge old code
- update homepage as per design
- mobile friendly
- added subtle animation

screencapture-localhost-8000-bounties-contributor-2019-03-15-18_10_35

closes  https://github.com/gitcoinco/web/issues/3881


owocki:
this is purdy!

app/retail/templates/bounties/contributor/activity.html
```diff
@@ -16,31 +16,24 @@
 {% endcomment %}
 {% load i18n static  %}
 <link rel="stylesheet" href={% static "v2/css/activity_stream.css" %}>
+<style>
+  .contributor-page .box {
```
rafalkowalski: Could we store styles in css file?

thelostone-mc: I did wanna do that initially but throwing the landing page stylesheet forces me to use `!important` as rules in `activity_stream.css` overwrites them and throwing it onto `activity_stream.css` causes it to be imported in places where it's not needed 😓

PR #4026, opened 2019-03-21
fix node security vulnerabilities

thelostone-mc:
#### Description 

Screen Shot 2019-03-21 at 7 12 03 PM

Ended up running `npm i` -> `npm audit` -> I was asked to update styleint + eslint -> updating those introduced a few errors being thrown in js which are fixed

I'm assuming these packages internally use the packages who's versions have vulnerabilities


app/assets/v2/js/metamask-approval.js
```diff
@@ -17,7 +17,7 @@ async function metamaskApproval() {
         var did_request_and_user_respond = (now_time - start_time) > 1.0;
 
         if (did_request_and_user_respond) {
-          document.location.href = document.location.href;
+          document.location.reload();
```
octavioamu: too bad they added this rule in jslint, why if we want to retry post form data?

PR #4129, opened 2019-04-05
removed duplicate matplotlib

thelostone-mc:
This is because we have matplotlob twice in our requirements. https://github.com/gitcoinco/web/blob/master/requirements/base.txt#L12


requirements/base.txt
```diff
@@ -49,7 +49,6 @@ ipdb==0.11
 django-autotranslate==1.1.0
 svgutils==0.3.0
 Werkzeug[watchdog]==0.14.1
-matplotlib
```
SaptakS: You are our hero ❤️

PR #4221, opened 2019-04-17
hackathon: update static img path

thelostone-mc:
##### Description 

update `/v2/images/..` -> `v2/images/..`


danlipert:
quick fix

PR #4243, opened 2019-04-22
fix sidebar hackathon

octavioamu:

### DIFF
--- a/app/assets/v2/js/pages/dashboard.js
+++ b/app/assets/v2/js/pages/dashboard.js
@@ -20,7 +20,7 @@ local_storage_keys.push('org');
 results_limit = 15;
 
 if (document.hackathon) {
-  results_limit = 50;
+  results_limit = 80;
 }
 
 var localStorage;


app/assets/v2/js/pages/dashboard.js
```diff
@@ -20,7 +20,7 @@ local_storage_keys.push('org');
 results_limit = 15;
 
 if (document.hackathon) {
-  results_limit = 50;
```
owocki: this is going to cause a lot of performance issues as the API is rather slow (i think its like 0.1s per result returned). 

i think a better solution to painting the sidebar is to load the list of orgs into some server side cache (JSONStore anyone?) and then keeping the current infinite scroll scheme with low results_limit

owocki: in leiu of completely rewriting... maybe we can rleease this into prod and checkout the results & iterate from there tho..

octavioamu: totally agree, the initial discussion was there wasn't more than 50 bounties So we decide the go with the lazy (faster) way but yes now looks like we need a endpoint just for orgs, the good thing is we can reuse that enpoint for the explorer filter by org and autocomplete it.

PR #4276, opened 2019-04-25
Get organizations filter by event

octavioamu:


octavioamu:
@owocki I had started this query to fetch just organizations names and make the hackathon filter, but first I want to know if that query performs well. What do you think?

thelostone-mc:
@octavioamu Yo do we need this ?

PR #4278, opened 2019-04-25
get orgs on the backend + improve dashboard performance

owocki:
works around the dashboard / api performance issue by just loading the orgs straight from the DB upon pageload.

tested on chrome / on my local

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -2506,9 +2506,12 @@ def hackathon(request, hackathon=''):
 
     title = evt.name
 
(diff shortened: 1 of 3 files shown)


app/dashboard/views.py
```diff
@@ -2506,9 +2506,12 @@ def hackathon(request, hackathon=''):
 
     title = evt.name
 
+    orgs = set([bounty.org_name for bounty in Bounty.objects.filter(event=evt).current()])
```
octavioamu: do we need to filter by network too?

PR #4285, opened 2019-04-25
bug: handle where fee is 0 for bounty creation + increase funding

thelostone-mc:
##### Description

This PR ensures that bounty creation and increase funding both work as expected when fee is set to 0.
The fee section is hidden when the fee is set to 0, that way the user doesn't see the banner making it less confusing. [The total amount, however, states we charge no fees ](https://user-images.githubusercontent.com/5358146/56760760-7caf8880-67b9-11e9-8d35-45f8cf475f28.png) to make the user know that we are not charging them anything extra. The same thing in increase funding looks like [THIS](https://user-images.githubusercontent.com/5358146/56760860-b2ed0800-67b9-11e9-91f5-a6fd67744d52.png)

Getting a lil more detailed

- updated code to hide fee section when fee is 0 & update content
  as needed.
- avoid extra metamask popup for fee when fee is 0.
- code refactor + css prettify

##### Bounty creation 

- [x] No fee + Featured Bounty
- [x] No fee + Non-Featured Bounty
- [x] Fee + Featured Bounty
- [x] No fee + Non-Featured Bounty

##### Bounty creation 

- [x] No fee + increase funding
- [x] With fee + increase funding

Yes I tested this like a crazy person.

img: see [here](https://user-images.githubusercontent.com/5358146/56760985-f6e00d00-67b9-11e9-8684-594ddf3772cf.png)

cc @frankchen07


app/assets/v2/js/pages/increase_bounty.js
```diff
@@ -36,7 +39,7 @@ $(document).ready(function() {
   var denomSelect = $('select[name=denomination]');
 
   localStorage['tokenAddress'] = denomSelect.data('tokenAddress');
-  localStorage['amount'] = 0;
+  localStorage['amount'] = 0.01;
```
danlipert: What does this change do?

app/assets/v2/js/pages/increase_bounty.js
```diff
@@ -211,27 +214,29 @@ $(document).ready(function() {
...
+            to: to_address,
+            from: web3.eth.coinbase,
+            value: web3.toWei(fee, 'ether'),
+            gasPrice: gas_price
+          }, function(error, txnId) {
+            indicateMetamaskPopup(true);
+            if (error) {
+              _alert({ message: gettext('Unable to pay bounty fee. Please try again.') }, 'error');
+            } else {
+              // TODO: Save txnId + feeamount to bounty;
```
danlipert: does this TODO still need to be added?

app/assets/v2/js/pages/new_bounty.js
```diff
@@ -593,46 +596,48 @@ $(document).ready(function() {
...
+        }
+      };
+
+      const deductBountyAmount = function(fee, txnId) {
+        ipfsBounty.payload.issuer.address = account;
+        ipfsBounty.payload.fee_tx_id = txnId;
+        ipfsBounty.payload.fee_amount = fee;
+        ipfs.addJson(ipfsBounty, newIpfsCallback);
+        if (typeof ga != 'undefined') {
+          ga('send', 'event', 'new_bounty', 'new_bounty_fee_paid');
```
danlipert: do we want to still track here? maybe a different event to indicate a zero fee?

app/dashboard/templates/bounty/increase.html
```diff
@@ -85,31 +85,34 @@ <h3>Increase Funding</h3>
...
+                  <div id="fee-section" class="mt-4">
+                    <h5 class="font-subheader">{% trans "No Suprises" %}</h5>
+                    <div class="row">
+                      <div class="col-12 col-sm-9 mt-3 font-body">
+                        <p>
+                          {% blocktrans %}
+                          Simply pay the bounty amount ( plus a standard
+                          <span class="fee-percentage font-weight-bold">10</span><span class="font-weight-bold">%</span> Gitcoin platform fee ).
+                          This covers our costs for finding quality contributors to join our platform so that you get the best work.
+                          If your business requires additional assitance, please contact us <a href="mailto:founders@gitcoin.co">founders@gitcoin.co</a>
```
danlipert: typo here `assitance` -> `assistance`

app/dashboard/templates/bounty/new.html
```diff
@@ -217,28 +217,33 @@ <h5 class="font-subheader pb-2">{% trans "Gas Settings" %}</h5>
...
+                      {% trans "No Suprises" %}
+                    </h5>
+                    <div class="row">
+                      <div class="col-12 col-sm-9 mt-3 font-body mb-sm-3">
+                        <p>
+                          {% blocktrans %}
+                          Simply pay the bounty amount ( plus a standard
+                          <span class="fee-percentage font-weight-bold">10</span><span class="font-weight-bold">%</span> Gitcoin platform fee ).
+                          This covers our costs for finding quality contributors to join our platform so that you get the best work.
+                          If your business requires additional assitance, please contact us <a href="mailto:founders@gitcoin.co">founders@gitcoin.co</a>
```
danlipert: same typo here - I guess it was there before 😭

app/assets/v2/js/pages/increase_bounty.js
thelostone-mc: This ensures that the min amount is set to 0.01 as opposed to 0. 
I originally updated on the html tag but the value here over-writes that, so had to udpate this too

In Increase funding amount shouldn't be set to 0 during submit.
If the amount is defaulted 0 -> the fee banner doesn't make sense cause it would say we charge 10% fee which is 0 :P 

we set default amount to > 0 so that we don't show 10% fee which equates to 0.

app/assets/v2/js/pages/increase_bounty.js
thelostone-mc: This would be a separate task! 
This would involve updating the bounty model / creating a new model to keep track of the fees.

I had brought this up when we originally built out fee%.
As of now we have no means to store and track it. 
Here I've simple reorganized the code

need to evaluate if it would sense to keep fee in it's model and map bounties and other produxts to it. cc @owocki

app/assets/v2/js/pages/new_bounty.js
thelostone-mc: Added a diff event 🙌  nice catch

thelostone-mc:
@danlipert fixed up the typo + updated the event for no fee scenario 
left the comments for the other two ^_^

Issue #4245, opened 2019-04-22
Ethereal Hackathon Support


PixelantDesign:
Refer to this doc for more context: https://docs.google.com/document/d/15xJSjaxDy0zzfD5y53XdmBNJ5R7JpViAo54AZiP2prs/edit

##  *P0**

- [x] **Problem**
When you open a bounty, then click “Return to Explorer”, it doesn’t take you back to the Hackathon Explorer (takes you to the general issue explorer instead).
**Solution**
Return to Explorer change to Go to back to hackathon explorer

- [x] **Problem**
Microsoft logo is round in flyout
**Solution** Remove round mask
<img width="459" alt="Screen Shot 2019-04-22 at 1 12 04 PM" src="https://user-images.githubusercontent.com/238235/56516366-44355380-6508-11e9-9613-3b851f81093e.png">

- [x] **Problem**
Pay back 10% fee!
Need to pay back all the sponsors, need a list and then we can pay back. Name, 10% fee amount, wallet address.
No need to do for microsoft.

- [ ] **Problem**
Any way to override Github org logo so that Quorum and Kauri can get their logos on there?
ENS funded issue by Kauri pulls ENS Logo in sidebar, can we override?
Can we have options: pull from org logo or upload an image - Admin.
Quorum - Same issue

- [x] **Problem**
Mythx is under Consenys Repo and consensys is not showing in left rail.
**Solution**: PR 4243

- [x] **Problem**
Delete Fluence logo from hackathon
Delete their logos from landing page or hackathon list page
Find out why did they drop out?

## **P1** 

- [ ] **Problem**
Prize title changes not syncing with title changes on Github.
When you click update issue details on the gitcoin Edit bounty details, it might update from github ticket.

- [ ] **Problem**
AV to accept submissions and confirm best way to pay out successful submissions (Advanced payout vs “tip” vs direct payment)
MSFT can’t use DAI/ETH, they will pay with FIAT later
MSFT under understanding that they could pay us later (30 days)
Some work has been submitted/ judged need to accept within 48 hrs
When we get payment, what is the easist way to payout a bounty? 
Use tips/normal pay?
Want to show that the bounty has completed and submission no longer.

- [ ] **Problem**
DAI amounts in prize explorer revert to true amount in smart contract once work is applied for/started (any kind of user activity) despite being changed in admin
Once user activity/submitted work, the Dai amount went back to the initial Dai amount.

- [ ] **Problem**
[...]

PR #4290, opened 2019-04-26
hide bounties value when lower than $1

danlipert:
##### Description

This PR hides the token amount and USDT value when the USDT value is less than $1. This is to hide values for bounties where a placeholder amount of DAI is used to create the bounty, usually on behalf of another organization as part of a managed bounty. This request came from the hackathon support issues.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4245

##### Testing

Tested locally - see video:
https://embed.vidyard.com/share/YoT3rDwFwQociVtoUD829L?


app/assets/v2/js/pages/bounty_details.js
```diff
@@ -905,6 +905,10 @@ var build_detail_page = function(result) {
   });
 
   $('#bounty_details #issue_description code').parent().addClass('code-snippet');
+  if (result['value_in_usdt_now'] != 'undefined' && result['value_in_usdt_now'] <= 1) {
+    $('#value_in_usdt_wrapper').hide();
+    $('.token').hide();
+  }
```
thelostone-mc: Do we need this ? 
we could do these checks within results.html itself using template `if`

thelostone-mc: @danlipert ^ in case you didn't see it

octavioamu: @danlipert this can be removed the changes in the template are doing this check already

octavioamu:
@danlipert why this one was closed? not good?

danlipert:
@octavioamu I thought you were saying this was already done at the template level in another PR :)

octavioamu:
> @octavioamu I thought you were saying this was already done at the template level in another PR :)

no no no, I mean in my review what I was trying to say is your are using the conditional if in the jsrender template if > 1  so that will be processed there so you don't need to do it also in the js file

app/assets/v2/js/pages/bounty_details.js
thelostone-mc: Ideally, this should be done only in live ! we might need to revisit it

danlipert:
Yeah I think this one is dead @thelostone-mc

PR #3983, opened 2019-03-18
[WIP] add web3js and support for tips

octavioamu:
##### Checklist

- [ ] Read and conforms to the [contributor guidelines](https://docs.gitcoin.co/mk_contributors/)
- [ ] Read and conforms to the [style guidelines](https://docs.gitcoin.co/mk_styleguide/)
- [ ] Linter status: 100% pass
- [ ] Added relevant tests
- [ ] Tests pass and code coverage has not decreased
- [ ] Changes don't break existing behavior
- [ ] Commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)
- [ ] [Code documentation](https://docs.gitcoin.co/mk_contributors/#docstrings)
- [ ] Completed: Code review by core team and any requested changes

###### Contributor

- [ ] Read and followed the [Contributor Guidelines](https://docs.gitcoin.co/mk_contributors/)
- [ ] Tested all changes **locally**
- [ ] Verified existing functionality
- [ ] Ran `make test` and everything passed!

###### Reviewer

- [ ] Affirm contributor guidelines have been followed and requested changes made
- [ ] CI tests and linting pass
- [ ] No conflicts (migrations, files, etc)
- [ ] Regression tested against staging or local deployment

###### Funder

- [ ] Validated requested changes were made to specification
- [ ] Bounty payout released to the contributor

PR #4293, opened 2019-04-26
injects web3 0.20.6 if no web3 is present so that opera will work

owocki:
##### Description

injects web3 0.20.6 if no web3 is present so that opera will work

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4209

##### Testing

still need to test this thuroughly on chrome and opera still

### DIFF
--- a/app/assets/v2/js/web3.js
+++ b/app/assets/v2/js/web3.js
@@ -1,0 +1,16362 @@
+/* eslint-disable no-loop-func */
+/* eslint-disable no-console */
+/* eslint-disable nonblock-statement-body-position */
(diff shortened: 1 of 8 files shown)


owocki:
hey all, opening this because yorke from microsoft is big on opera, and wants to be able to use gitcoin across it.

i think that as the world moves from just using metamask to using many web3 browsers... we will have to maintain support for each of them.

am i on the right track?  if so, will continue testing + push this through to non-draft status

octavioamu:
Take a look at #3983 maybe there is something useful there.

owocki:
just commented there

thelostone-mc:
closing this out as we this is now fixed and gitcoin works on opera

PR #3859, opened 2019-02-27
Grant contributions in profile (#3316)

e18r:
##### Description
In order to be able to show the grant contribution/subscription status of a user in her profile, we need to create a few new tabs. These tabs will show the activity related to contributing to, subscribing to, and unsubscribing from grants. However, this activity is not recorded at the moment, so the first thing to do is to record it. 

1. Created a new activity type for one-time grant contributions
2. Created a new foreign key for the object Activity, so that it can be associated with Contribution objects
3. Recorded the three types of grant activity: one-off contributions, recurring subscriptions and cancellations
4. Modified the activity query in the dashboard view so that it also fetches subscription activities
5. Created a new template that formats grant activities

##### Checklist

- [X] Read and conforms to the [contributor guidelines](https://docs.gitcoin.co/mk_contributors/)
- [x] Read and conforms to the [style guidelines](https://docs.gitcoin.co/mk_styleguide/)
- [x] Linter status: 100% pass
- [ ] Added relevant tests
- [ ] Tests pass
- [x] Code coverage has not decreased
- [x] Changes don't break existing behavior
- [x] Commit message follows [commit guidelines](https://docs.gitcoin.co/mk_contributors/#step-4-commit)
- [x] [Code documentation](https://docs.gitcoin.co/mk_contributors/#docstrings)
- [ ] Completed: Code review by core team and any requested changes

##### Affected core subsystem(s)

dashboard, grants

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/3316

##### Testing and Sign-off

I created a few Grants with a custom token on Rinkeby and with another user I contributed to them, subscribed and cancelled subscription. Working as expected.

###### Contributor

- [x] Read and followed the [Contributor Guidelines](https://docs.gitcoin.co/mk_contributors/)
- [x] Tested all changes **locally**
- [x] Verified existing functionality
- [x] Ran `make pytest` and code coverage didn't decrease
###### Reviewer

- [ ] Affirm contributor guidelines have been followed and requested changes made
- [ ] CI tests and linting pass
- [ ] No conflicts (migrations, files, etc)
- [ ] Regression tested against staging or local deployment

###### Funder

- [ ] Validated requested changes were made to specification
- [ ] Bounty payout released to the contributor

PR #4317, opened 2019-04-30
[#4267] Activity data for old grants

### LINKED ISSUE #4267: Display grant history on profiles before #3859
##### Description 

Grant Activity now shows up on the user's profile for newly created grants + subscriptions.
This was introduced in https://github.com/gitcoinco/web/pull/3859

However the activity before this PR went live also needs to be captured and displayed on the profile (which does not do at the moment)

cc @e18r , @gitcoinco/engineer

e18r:
Fixes #4267 

A data migration is introduced, which generates grant-related activity data
that happened before grant-related activity was recorded.

The date before which grant-related activity will be generated was set to `2019-04-24 11:53:00 UTC`, which is when PR #3859 was merged.


e18r:
@owocki @thelostone-mc @danlipert please review my work, thanks

owocki:
migration LGTM .  thanks for doing this @e18r !

e18r:
Hi @danlipert, please don't for get to check this out. Also, if you need anything, don't hesitate. Thank you so much!

e18r:
@danlipert I created a way to check if the activity exists before adding it. Check it out in my latest commit. For unique activity types (create grant and delete grant) it checks that there aren't other activities with that type for the current grant. For other activity types, it also checks that they were made by the same person, and at roughly the same time. The deployment date check is still there as an extra safety layer.

Regarding the unit test, have you guys established a way of unit testing data migrations in Gitcoin? I did some research, and found [this SO question](https://stackoverflow.com/questions/34858074/writing-automated-tests-for-data-migrations) that suggests a way of doing it. However, maybe the simplest way to test it would be to download a subset of the production database and run the migration locally. What do you think would be the best approach? I'm open to doing both.

danlipert:
@e18r Nice! Appreciate the extra care taken here :) Code is looking nice, if you can add some unit tests I'll add a tip to the bounty. 

The strategy that I would take to write the tests is to first create some model objects in the database for grants and subscriptions manually that demonstrate the previous state of the app (i.e. no activity data). Then create some grants and subscriptions through the app itself - either call the underlying helper functions or create mock request objects and call the views individually. This will generate activity data, representing the current state of the app. Then do some asserts showing that there are no duplicates, and that each subscription, cancellation, etc. has the appropriate activity data. You could even have some tests that show what edge cases the helper functions are solving (i.e. `close_enough`).

e18r:
@danlipert I've been working on the unit test. The hard part is getting the migration to run for the test, which is done with something called MigrationExecutor (see link in my previous post), but it has given me plenty of errors.

In order to solve them, I had to modify Gitcoin's `settings.py` to add `'migrations'` in the `INSTALLED_APPS`, and also to include the `MIGRATION_MODULES` dict. Now I'm getting an unexplained KeyError when I call the `migrate` function from the executor.

I'll keep you posted on my progress. Any help appreciated.

e18r:
@danlipert After a lot of work, I located the source of the issue as coming from pytest. When I use `manage.py test`, the KeyError doesn't show up.

However, another, more troubling error appears:

    django.db.migrations.exceptions.IrreversibleError: Operation <RunPython <function generate_activities at 0x7fbf174eb268>> in dashboard.0032_data_for_old_grants is not reversible

In order to go back to a database state in which the migration wasn't applied, django needs to revert the migration, as explained [here](https://stackoverflow.com/questions/37177041/how-to-unittest-a-django-database-migration). However, it doesn't really make sense to write a reversal method by which I delete all the grant-related activities from a certain date in the past, does it?

Please consider employing a more ad-hoc test with production data because I reached a hard limit here.

e18r:
@danlipert @thelostone-mc @owocki What are your thoughts on this?

danlipert:
Hey @e18r - I'm not sure I understand - why is Django trying to run a reverse migration? Maybe instead of running the migration directly, you can have the migration call a helper function, and then in your test just call the helper function instead of the migration itself?

e18r:
@danlipert I have a few questions:

* Do you want me to have the helper function outside the migration file, perhaps in `utils.py`? This could have potentially disastrous effects if someone modifies that helper function in the future.
* Do you want me to have the helper function inside the migration file? In that case: how can I import that helper function from the test file? `from dashboard.migrations.0032_data_for_old_grants import helper` doesn't seem to work.

danlipert:
@e16r ah yes, I think the underscores may cause python's importing to fail :/ I'm thinking you can create a file inside the `migrations` directory like `migrations/utils.py`. Then put your function in there and you should be able to `from dashboard.migrations import utils` then `utils.myfunction()`

e18r:
@danlipert when I create the file `migrations/utils.py`, I get the following error:

    django.db.migrations.exceptions.BadMigrationError: Migration utils in app dashboard has no Migration class

Django thinks I created a new migration and expects a Migration class inside `utils.py`.

danlipert:
> @danlipert when I create the file `migrations/utils.py`, I get the following error:
> 
> ```
> django.db.migrations.exceptions.BadMigrationError: Migration utils in app dashboard has no Migration class
> ```
> 
> Django thinks I created a new migration and expects a Migration class inside `utils.py`.

@e18r Ah, my bad - you'll have to put the helper file elsewhere then

e18r:
@danlipert This PR is ready for review and merge:
* a data migration was introduced
* two failsafe mechanisms prevent the insertion of duplicate data
* a comprehensive unit test suite prevents any bugs

e18r:
@danlipert @thelostone-mc @octavioamu I solved migration dependencies, squashed and rebased. The squash dismissed your reviews, I think, so you probably need to approve them again.

Issue #4321, opened 2019-04-30
Marketing - Update Links on Contributor Landing Pages


PixelantDesign:
Add the links to contributor pages in this paragraph "There are open issues in our Issues Explorer for all Python, JavaScript, Rust, Go, Solidity, Node.js, and many more developers. " 

https://gitcoin.co/bounties/contributor/

kuhnchris:
You mean links to the filtered search results?

mbeharry:
Replace this paragraph;
"There are open issues in our Issues Explorer for all Python, JavaScript, Rust, Go, Solidity, Node.js, and many more developers."

With this one;
There are open issues & bug bounties in our <a href="https://gitcoin.co/explorer">Issue Explorer</a> for <a href="https://gitcoin.co/contributor/python">Python</a>, <a href="https://gitcoin.co/contributor/javascript">JavaScript</a>, <a href="https://gitcoin.co/contributor/rust">Rust</a>, <a href="https://gitcoin.co/contributor/solidity">Solidity</a>, <a href="https://gitcoin.co/contributor/design">Design</a>, <a href="https://gitcoin.co/contributor/html">html</a>, <a href="https://gitcoin.co/contributor/ruby">Ruby</a>, <a href="https://gitcoin.co/contributor/css">CSS</a> and much more.

kuhnchris:
[Easy enough.](https://github.com/gitcoinco/web/pull/4341)

kuhnchris:
👍  thanks for the pull!

octavioamu:
Thanks for the work @kuhnchris

octavioamu:
Your pr got merged is that why it got it closed here, not that we are ungrateful

kuhnchris:
I know, i'm doing this long enough now with you guys. ;-)

PR #4341, opened 2019-05-01
#4321 Update Links for Marketing

kuhnchris:
##### Description
See #4321

##### Refers/Fixes
Fixes: #4321

##### Testing
-


app/retail/templates/bounties/contributor/story.html
```diff
@@ -38,8 +38,16 @@
     </div>
     <div class="col-12 col-md-5 offset-md-2 my-auto mt-4 order-md-2 g-fadein" data-fade-duration=1000 data-fade-direction="right">
       <p>
-        There are open issues in our Issues Explorer for all Python, JavaScript, Rust, Go, Solidity,
-        Node.js, and many more developers.
+        There are open issues & bug bounties in our <a href="/explorer">Issue Explorer</a> for
```
SaptakS: Please use `{% url 'explorer' %}` instead of relative URLs here. And  similarly for all the other links. Thanks.

app/retail/templates/bounties/contributor/story.html
```diff
@@ -38,8 +38,16 @@
...
-        Node.js, and many more developers.
+        There are open issues & bug bounties in our <a href="/explorer">Issue Explorer</a> for 
+        <a href="/contributor/python">Python</a>, 
+        <a href="/contributor/javascript">JavaScript</a>, 
+        <a href="/contributor/rust">Rust</a>, 
+        <a href="/contributor/solidity">Solidity</a>,
+        <a href="/contributor/design">Design</a>,
+        <a href="/contributor/html">HTML</a>, 
+        <a href="/contributor/ruby">Ruby</a>,
+        <a href="/contributor/css">CSS</a>,
```
thelostone-mc: Could we pass this from the backend as an array and simply loop through it here ?

kuhnchris:
@SaptakS Thanks for checking, please check again, better now?

app/retail/templates/bounties/contributor/story.html
kuhnchris: In theory yes, I also thought of that, but we'd need to create a new either table or json in python.
What route do you prefer? Table for editing live, or json/hardcoded json to have a deploy cycle?

thelostone-mc: Harcoded Json for now would work 🙌

SaptakS: @kuhnchris You can just send it as part of the template params as an array of strings.

kuhnchris: I know, just wanting to make sure that's reaaaaaally what you want. But sure, i'll add it.

kuhnchris: This whole bounties/contributor stuff is a whole lot of 💩 .
we cannot use 'url' with that, since it's not a registered route but somehow dynamically generated.

```
django.urls.exceptions.NoReverseMatch: Reverse for 'bounties/contributor' not found. 'bounties/contributor' is not a valid view function or pattern name.
```

```
     {% for subject in contributor_list %}
          <a href="{% url 'bounties/contributor' %}"{=subject.link=}>{=subject.text=}</a>,  <- neither
        {% endfor %} 
        <a href="{% url 'contributor/python' %}">Python</a>,  <- nor this
        <a href="{% url 'contributor' %}/python">Python</a>,  <- nor this
```

Please advice.

SaptakS: It's not whole lot of 💩 , you are just writing it wrong. The code is supposed to be:

```
{% for subject in contributor_list %}
     <a href="{% url 'contributor_bounties' tech_stack=subject.link %}">{{ subject.text }}</a>,  <- neither
 {% endfor %} 
```
This is assuming that `subject.link` is `/python` and `subject.text` is `Python`

I am not sure where you got the above syntax from. I would really advice you to go through this documentation to learn django templating: https://docs.djangoproject.com/en/2.2/topics/templates/

kuhnchris: Thanks for the heads up, still thinking that contributor + tech_stack is 💩 (especially cause I need to comment views.py L567-568 just to make it work on a local dev environment, else those pages 404) , but I couldn't care less. 
grafik

Also, thanks for relinking me to the template documentation, I tried to get the url '' part working so I tried working with various types of {= {% ... to ensure it wasn't me fucking this up.

Committed, please check, worked over here.
grafik

Just letting you know for some reason you dropped "Go" but it's on the infographics on the left, so I might reconsider dropping it. @PixelantDesign

thelostone-mc: ^ @PixelantDesign let us know if that needs to be updated

Issue #4343, opened 2019-05-01
BUG - After Completion Ratings - Rinkeby Bounties are counted in.


kuhnchris:
**Describe the bug**
I didn't fund more than 1 bounty on mainnet, dispite my profile showing 13.
![grafik](https://user-images.githubusercontent.com/761911/57020768-a2f67d80-6c2a-11e9-86d6-ff18a4e15728.png)
After checking, this shows some rinkeby bounties.
![grafik](https://user-images.githubusercontent.com/761911/57020795-c3bed300-6c2a-11e9-821f-d7e7d6c94ac7.png)
That's... probabbly not correct.

**To Reproduce**
Create bounties on testnet. Check Profile on Mainnet

**Expected behavior**
Only selected network in configuration is taken into account (gitcoin.co -> mainnet, local dev -> rinkeby)

kuhnchris:
Checking with #4342 since it seems to be on the same corner.

PR #4344, opened 2019-05-01
[WIP] 4343 - fixing wrong network bounties

kuhnchris:
##### Description
Fixes the issue of selecting wrong network bounties

##### Refers/Fixes
Fixes: #4343

##### Testing
Currently in progress...

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -1886,23 +1886,29 @@ def profile(request, handle):
     context['avg_rating'] = profile.get_average_star_rating
 
     context['unrated_funded_bounties'] = Bounty.objects.prefetch_related('fulfillments', 'interested', 'interested__profile', 'feedbacks') \
-        .filter(
-            bounty_owner_github_username__iexact=profile.handle,
-        ).exclude(
+        .filter(bounty_owner_github_username__iexact=profile.handle) \
+        .filter(network=settings.KUDOS_NETWORK) \
+        .exclude(
             feedbacks__feedbackType='approver',
             feedbacks__sender_profile=profile,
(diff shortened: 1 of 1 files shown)


kuhnchris:
@owocki do we have any "easy" way to find out which network we are on? Right now I use the KUDOS_NETWORK variable since that one should be close enough, and SHOULD always be the same to the "mainnet being used".
Else we need some variable like "PROFILE_NETWORK" to fix running issues from this problem. On live we do have rinkeby and normal bounties. That screws up our totals.
Please advise.

Also, maybe we should start providing some data via dumpdata and loaddata in django?
https://docs.djangoproject.com/en/2.2/howto/initial-data/

I know there's an open ticket, but I don't think there's any progress on it yet?

Thanks

kuhnchris:
Did you manage to test it @thelostone-mc ? I cannot test it for the life of myself, I cannot get that rating stuff working on my local instance...

SaptakS:
@kuhnchris you can use the settings.DEBUG to check if it is a production environment. If it is a prod environment then the network is mainnet, else it is rinkeby.

kuhnchris:
settings.DEBUG also isn't set when there's some trouble on prod and you guys are debugging it @SaptakS ?
If we can ensure it's always displaying properly on prod then I'm all for it 👍

kuhnchris:
Since I assume the answer to that question is "no" I'll simply go for
```
settings.DEBUG == True => Rinkeby 
Else => Mainnet
```

Update coming soon

kuhnchris:
so we can throw this one away @thelostone-mc ?

Issue #932, opened 2018-04-21
ConnectionError: HTTPSConnectionPool(host='ipfs.infura.io', port=5001): Max retries exceeded with url: /api/v0/cat/QmXxkBASF92QvFFwChwZAECSRsjoMstMHJ2A9bwXuchXbA (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7f6791


mbeacom:
View details in Rollbar: [https://rollbar.com/gitcoin/gitcoin/items/826/](https://rollbar.com/gitcoin/gitcoin/items/826/)

```
Traceback (most recent call last):
  File "/home/ubuntu/gitcoin/coin/app/dashboard/management/commands/sync_geth.py", line 61, in handle
    bounty = get_bounty(bounty_enum, network)
  File "/home/ubuntu/gitcoin/coin/app/dashboard/utils.py", line 197, in get_bounty
    bounty_data_str = ipfs_cat(bountydata)
  File "/home/ubuntu/gitcoin/coin/app/dashboard/utils.py", line 126, in ipfs_cat
    response = ipfs_cat_requests(key)
  File "/home/ubuntu/gitcoin/coin/app/dashboard/utils.py", line 144, in ipfs_cat_requests
    response = requests.get(url)
  File "/home/ubuntu/gitcoin/gitcoin-3/lib/python3.6/site-packages/requests/api.py", line 72, in get
    return request('get', url, params=params, **kwargs)
  File "/home/ubuntu/gitcoin/gitcoin-3/lib/python3.6/site-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/ubuntu/gitcoin/gitcoin-3/lib/python3.6/site-packages/requests/sessions.py", line 508, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/ubuntu/gitcoin/gitcoin-3/lib/python3.6/site-packages/requests/sessions.py", line 618, in send
    r = adapter.send(request, **kwargs)
  File "/home/ubuntu/gitcoin/gitcoin-3/lib/python3.6/site-packages/requests/adapters.py", line 508, in send
    raise ConnectionError(e, request=request)
[...]

PR #4368, opened 2019-05-07
Pin the slackclient and websocks version to solve conflicts.

### LINKED ISSUE #4367: ModuleNotFoundError: No module named 'slackclient'
**Describe the bug**
`ModuleNotFoundError: No module named 'slackclient'`

**To Reproduce**
pull the newest code, and run 
```
docker-compose up --build
```
 to run the code, it throws out an error:
```
ModuleNotFoundError: No module named 'slackclient'
```

**Expected behavior**
it should not throw out this error.

Skyge:
Fix #4367

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -22,9 +22,9 @@ PyJWT==1.5.3
 PyPdf2
 python-twitter==3.2
 sendgrid==5.6.0
-slackclient
+slackclient==1.1.1
 eth-tester
-websockets
+websockets==5.0.1
 web3==4.5.0
 eth-abi==1.1.1
 mailchimp3


Skyge:
Thanks for @usmanmuhd help, at now, I use the version of `1.1.1`, It works, but I am not sure whether it is the best version. @owocki any idea? 
BTW, maybe we should pin all modules version.

danlipert:
> Thanks for @usmanmuhd help, at now, I use the version of `1.1.1`, It works, but I am not sure whether it is the best version. @owocki any idea?
> BTW, maybe we should pin all modules version.

I thinking pinning all modules isn't a bad idea, gives us some more control over these types of issues at the expense of having to be more vigilant about dependency security updates. @Skyge have you tested the Kudos functionality after pinning the websockets version?

Skyge:
@danlipert Not yet, when I try to solve the incompatible version problem of `slackclient`, I find there is a tip for websockets version, so I make a change at the same time.

Skyge:
> gives us some more control over these types of issues at the expense of having to be more vigilant about dependency security updates.

Yeah, that is a truth, we really need to update to use the new version to prevent potential risks, and we also should test to prevent incompatible problem caused by new version. so how to know we should update our dependency modules in time? write a shell to monitor? or others ways?

Skyge:
@danlipert As a matter of fact, after I solve the imcompatible version problem, I meet another problem : #932, I can not connect to ipfs, I am not sure wether it has impact on testing Kudos functionality.

danlipert:
@Skyge do you have that problem if you switch to the master branch? I wonder if it is related to the websocket change somehow

Skyge:
@danlipert Yeah, I have already switch to the master branch, and do not specify the websocket version, but it also throw out that error.

Skyge:
Even though I go back to a earlier working well version now, it will not work again.

PR #4375, opened 2019-05-12
restructure: move all lib js files into js/lib/*

thelostone-mc:
##### Description

Moving all imported js lib files into it's own folder in an attempt to make our folder structure a lil neater


danlipert:
Whats the thinking behind moving some of the files to `/lib/` while leaving others alone? I do agree our js directories could use some cleaning up for sure 👍

thelostone-mc:
@danlipert Ah so all those are files which aren't written by us -> aka library files 
Figured I'd start by taking the step to moving all of them to it's own folder so that's it's evident what we are using at the moment from outside.

Step 2: would be to restructure js files to make them more organized

app/kudos/templates/kudos_details.html
```diff
@@ -5,7 +5,7 @@
     {% include 'shared/head.html' %}
     {% include 'shared/cards_pic.html' %}
     <link rel="stylesheet" href="{% static "v2/css/kudos/styles.css" %}" />
-    <script src="{% static 'v2/js/tweenlite.js' %}"></script>
+    <script src="{% static 'v2/js/lib/tweenlite.js' %}"></script>
```
octavioamu: can we move this to the footer?

octavioamu:
Left a comment but LGTM

octavioamu:
Nice move, this will be very good also if we decide to jump into a bundle approach, like concat + minify js files having libs.js and  scripts.js called and not all separated files.

thelostone-mc:
@octavioamu addressed comments + resolved conflicts

PR #4423, opened 2019-05-16
set image size

octavioamu:


thelostone-mc:
addressed over at https://github.com/gitcoinco/web/pull/4422 :P

Issue #4308, opened 2019-04-29
As a bounty hunter, i want the opportunity to tell the system when bounties are not relevant to me, so I dont tune out the new bounty emails


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)

As a bounty hunter, i want the opportunity to tell the system when bounties are not relevant to me, so I dont tune out the new bounty emails

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)

No ability to opt out of bounties.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Buttons that allow you to change your matching preferences.

<img width="722" alt="Screen Shot 2019-04-29 at 4 13 01 PM" src="https://user-images.githubusercontent.com/513929/56930403-c76f2000-6a99-11e9-9e6d-f2ce423e2f7e.png">

These buttons will all go to https://gitcoin.co/settings/matching ; which should be extended to allow:
1. user to subscribe / unsubscribe to matching emails 
2. to receive bounties for any programming language in a list of programming languages 
3. to receive bounties only within a certain price range.
4. to whitelist / blacklist repositories.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
see above

### Data Requirements
Increased short and long term open / click rate on the new bounty emails

### Reach
All contributors that see this email

### Impact
Increase accuracy of match, High Impact 

### Confidence 
80%

owocki:
@danlipert @thelostone-mc @PixelantDesign i wrote this after our planning session today.

willsputra:
Thoughts? @PixelantDesign 

Email with new buttons
![email](https://user-images.githubusercontent.com/19514207/58721214-88b4e880-83fe-11e9-9e50-b496e919e55b.jpg)

Matching settings
![matching](https://user-images.githubusercontent.com/19514207/58721231-98ccc800-83fe-11e9-82aa-05ab61ba7e7e.png)

octavioamu:
@owocki what should be the action when you click "not interested" ? Because can be the user is not interested in some particular keywords, or organization/repo bounties.

PixelantDesign:
@thelostone-mc thoughts on this? ^^

[...]

PR #4426, opened 2019-05-17
Exclude bounties reserved for specific users from new bounties email

acolytec3:
##### Description

The current new bounties email includes bounties reserved for specific users which is annoying because reserved bounties aren't actually actually available to the general Gitcoin population.  This PR adds an exclusion rule to the query that pulls relevant bounties to ensure that only bounties a user can actually apply for are included.

##### Refers/Fixes

Addresses one element of #4308 

##### Testing

Adds new test for new_bounties_email.py that incorporates test for new exclusion of reserved bounties from email.


app/marketing/management/commands/new_bounties_email.py
```diff
@@ -53,8 +53,8 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         if settings.DEBUG:
-            print("not active in non prod environments")
-            return
+           print("not active in non prod environments")
```
danlipert: Looks like an accidental change here

app/marketing/tests/management/commands/test_new_bounty_email.py
```diff
@@ -0,0 +1,76 @@
...
+            idx_status='submitted',
+            bounty_owner_email='jack@bar.com',
+            current_bounty=True,
+            network='mainnet',
+        )
+
+
+    def test_get_bounties_for_keywords(self):
+        """Test get_bounties_for_keywords function to confirm a bounty reserved for a specific user is excluded."""
+        new_bounties, all_bounties = get_bounties_for_keywords('Python',24)
```
danlipert: I like to use `_` as the name for a "throwaway"/unused variable name here, basically saying to the reader that they can ignore this variable.

app/marketing/management/commands/new_bounties_email.py
acolytec3: Yep, corrected.

app/marketing/tests/management/commands/test_new_bounty_email.py
acolytec3: I added the underscore as requested.  No issues here.  :-)

danlipert: Ah I meant just using `_` instead of `_all_bounties` - just a style thing though so no worries :) more info here: https://www.geeksforgeeks.org/underscore-_-python/

acolytec3:
Ah, okay!  Learn something new every day.  Thanks!

PR #4429, opened 2019-05-17
Check for auth on importing linkedin data

danlipert:
##### Description

This PR checks if the user is authenticated before accepting the updated LinkedIn data for their job opportunity profile.

##### Refers/Fixes

This issue was referred to us by a security researcher.

##### Testing

After this change I am still able to update my own profile. If I change to change the profile handle in the URL I now receive a bad request error


danlipert:
@thelostone-mc I've added the `@login_required` decorator - thanks for the suggestion!

Issue #4424, opened 2019-05-16
Matching - Keyword not updated on Users Profile and Directory


willsputra:
**Describe the bug**

I've updated my skills keywords on Manage Settings > Matching, but it's not updated on Users Directory

**To Reproduce**

1. Go to Manage Settings > Matching and update keyword (e.g. 'design')
![Screenshot 2019-05-17 00 48 07](https://user-images.githubusercontent.com/19514207/57872160-a8b0ad80-783d-11e9-980f-672e5af131a1.png)

2. Search for the keyword on Users Directory(e.g. 'design'). Not found.
![Screenshot 2019-05-17 00 48 14](https://user-images.githubusercontent.com/19514207/57872194-bf570480-783d-11e9-93a0-a06e1c52b45c.png)

3. If I search for my username, it still displays old keywords (e.g. 'Shell', 'CSS', etc.)
![Screenshot 2019-05-17 00 48 42](https://user-images.githubusercontent.com/19514207/57872238-d8f84c00-783d-11e9-8b74-b76660ae9b4f.png)

(octavio edit)
Also onboard keywords seems not saved anywhere 
![image](https://user-images.githubusercontent.com/1358093/63597873-c73ce800-c594-11e9-9356-4109b9925b74.png)

**Expected behavior**
Keywords should be updated on Users Directory.

Skyge:
It seems like they use different ways to show, that is, you use this way to change keywords, but the code does not use that to show, it uses another way, so no matter what you change, it will not display.

owocki:
Matching SEttings updates the `EmailSubscriber` object not the `Profile` object.  It should probably update both, and we should perhaps have some other way of making sure that those two objects keywrods are kept in sync.

Skyge:
@owocki  Is it necessary to keep both same fields at the same time? Since they represent the same meaning, why not just keep one? If keeping one, which one? `Profile` or `EmailSubscriber`?

owocki:
not every emailsubscriber has a profile, but other than that.. thats a valid fix path too IMHO

owocki:
just submitted a PR to fix this ^^

PixelantDesign:
settings matches automatic email - when changes are made, it's not reflected in the user directory.
Need to unify everything and fix filters in user directory.

PR #4474, opened 2019-05-23
sync email subscriber and profiles keywords field

owocki:
##### Description

sync email subscriber and profiles keywords field per https://github.com/gitcoinco/web/issues/4424

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4424

##### Testing

tested on my local


owocki:
its not clear to me that a cronjob is the right place for this... maybe the profile <> emailsubscriber sync could happpen in a presave hook.

SaptakS:
I was thinking more of a post_save hook in both the models.

app/dashboard/management/commands/sync_es_profile.py
```diff
@@ -0,0 +1,63 @@
...
+                counter += 1
+                profile.keywords = keywords
+                profile.save()
+
+        # pull keywords by emailsubscriber into profile
+        for profile in Profile.objects.filter(keywords=[]):
+            es = EmailSubscriber.objects.filter(email=profile.email).exclude(keywords=[]).first()
+            if es:
+                print("2", es.keywords, profile.handle)
+                profile.keywords = es.keywords
```
danlipert: Don't we want to append these keywords rather than overwrite them?

owocki: fe9bdb63292ea26247c79b18ba2f42b17c6e0226

owocki:
just addressed dan's feedback

app/dashboard/management/commands/sync_es_profile.py
```diff
@@ -0,0 +1,63 @@
...
+                if history.data.get('keywords', None):
+                    keywords = history.data['keywords'][0].split(',')
+            if keywords:
+                print("1", counter, profile.handle, keywords)
+                counter += 1
+                profile.keywords = keywords
+                profile.save()
+
+        # pull keywords by emailsubscriber into profile
+        for profile in Profile.objects.filter(keywords=[]):
```
danlipert: Here we are iterating over the profiles that don't have any keywords, but the for loop above assigns keywords to each profile that lacks them if there is a search history object for that user. In this loop, we are combining the email subscriber keywords with the profile keywords, but we are only iterating over profiles that have no keywords.

I think we need to combine the two `for profile in Profile.objects.filter(keywords=[]):` loops so we can get both search history and email subscriber keywords applied. In the current code, adding the search history keywords to a profile essentially skips adding email subscriber keywords to that same profile.

owocki: 32a1a057affae79ff37beebcf5a6914c9d82f085

Issue #4398, opened 2019-05-15
User Directory - Sort "Default view" by Contributor number


PixelantDesign:
### User Story
As a funder I'd like to see the only relevant contributors to fulfill my bounty

### Definition of Done
- Display recently worked on your bounties first.
- Update default view (prior to entering a search) sorting to show contributors by leaderboard number
- Review with core team
- Post screenshot of implementation

- BE sort queries based on above
- add conditions when filtering

### Why
- We'd like to help funders easily find the best contributors to work with

kuhnchris:
* Isn't Leaderboard a weekly measurement? Maybe we should take some long-term measurement into account, like bounties completed, participated, payed, etc.

vs77bb:
@kuhnchris Leaderboard is a [persistent here](gitcoin.go/leaderboard) -- but agree that it'd be nice to include other things. 

@frankchen07 I wonder how difficult it'd be to come up with a weighted average of a few things to see what it does. 

- number of Github contributions [20%] 
- number of bounties completed [40%]
- leaderboard position ($$$ amount made) [40%] 

The reason number of Github contributions matters to me is because I think there may be some people with Gitcoin profiles who haven't done issues, but might be willing to do them in the future, who are clearly active on Github and we should target activating them by inviting them to great bounties

kuhnchris:
The question is: do we have github contributions in access? Since this would be a rather big burden on the github API that'd be something that needs to be stored on-site if we need to evaluate it.
Regarding the other measurements:
SUM(GitHub Contrib) * 0.2 + (1 + (25 / Leaderboard Position)) * (Number of Bounties Started / Number of Bounties Completed).
Reason why I add the number of bounties started vs. completed is to weight someone who starts many bounties but only completes like 2-3 extremly good ones (like the Hackathons) would be ranked way higher than someone who consistently does good work and commits but isn't on the leaderboard (yet).

But I'm open to suggestions ;-)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.29 ETH (71.83 USD @ $247.69/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4398/3097).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/4398/3097).
[...]

PR #4475, opened 2019-05-23
Sorts user directory by previously worked and leaderboard

SaptakS:
##### Description

- Display recently worked on your bounties first.
- Update default view (prior to entering a search) sorting to show contributors by leaderboard number
##### Refers/Fixes
Fixes https://github.com/gitcoinco/web/issues/4398


app/dashboard/views.py
```diff
@@ -758,25 +758,38 @@ def users_fetch(request):
...
+    order_by = request.GET.get('order_by', 'leaderboard_ranks__rank')
     bounties_completed = request.GET.get('bounties_completed', '').strip().split(',')
     leaderboard_rank = request.GET.get('leaderboard_rank', '').strip().split(',')
     rating = int(request.GET.get('rating', '0'))
     organisation = request.GET.get('organisation', '')
 
     user_id = request.GET.get('user', None)
     if user_id:
-        profile = Profile.objects.get(id=int(user_id))
+        profile = User.objects.get(id=int(user_id))
```
danlipert: Lets not call an instance of a `User` object `profile` so we can avoid confusion here

app/dashboard/views.py
```diff
@@ -758,25 +758,38 @@ def users_fetch(request):
...
 
     context = {}
     if not settings.DEBUG:
         network = 'mainnet'
     else:
         network = 'rinkeby'
 
-    user_list = Profile.objects.prefetch_related('fulfilled', 'leaderboard_ranks', 'feedbacks_got').order_by(order_by)
+
+    user_list = Profile.objects.prefetch_related(
```
danlipert: similarly here, can we can this `profile_list`?

app/dashboard/views.py
SaptakS: Yeah. makes sense. It's only that in the loop below we were using the variable `user` to loop through the user list. I will rename this as `current_user`

app/dashboard/views.py
SaptakS: Hmm. I mean it's called User Directory. So I thought user_list is good. 🤔

danlipert:
@SaptakS looks like Travis is failing for these changes

app/dashboard/views.py
```diff
@@ -827,13 +847,12 @@ def users_fetch(request):
             profile_json['avatar_url'] = user_avatar.avatar_url
         count_work_completed = Activity.objects.filter(profile=user, activity_type='work_done').count()
         count_work_in_progress = Activity.objects.filter(profile=user, activity_type='start_work').count()
-        previously_worked_with = 0
-        if profile:
+        if current_user:
             previously_worked_with = BountyFulfillment.objects.filter(
-                bounty__bounty_owner_github_username__iexact=profile.handle,
+                bounty__bounty_owner_github_username__iexact=current_user.username,
```
danlipert: Why use the username here instead of handle? If we are expecting a github username, then I think handle should be used, since non-github users could have username that is someone else's github handle (theoretically, at least...)
https://github.com/gitcoinco/web/blob/master/app/dashboard/models.py#L2352

SaptakS: Because the built-in User model in django has only username and no handle column, hence.

danlipert: @SaptakS I understand that, but we can easily access the profile from the `current_user` object, right? So why not use the handle? Is this a performance tradeoff?

Issue #4416, opened 2019-05-15
Propose Project: Gitcoin Users


vs77bb:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a funder, I want to propose working with top Gitcoin contributors on projects on my Github issue board. 

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: Right now, I can only engage with Gitcoin contributors if I already posted a bounty. Alternatively, if I see the right contributor, I can propose a project we could work alongside together. 

### Description
![Propose Project](https://user-images.githubusercontent.com/23297747/57810985-1de58b00-7737-11e9-9374-b27e9140d3e1.png)

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
gitcoin.co/users provides ability to 'Invite to Bounty' only to funders who already have bounties. 

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
gitcoin.co/users provides ability for **any funder** to invite a gitcoiner to work with them on an open source repo. 

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
- 'Propose Project' button implemented next to Gitcoin Bounty Hunters (people with profiles who actually have done bounties in the past). 
- When clicked, 'Propose Project' takes you to the [fund issue](gitcoin.co/new) page with the bounty either **reserved** for the hunter or with the hunter **invited** (<-- I'd advocate for invited, so other hunters can also apply)
- Use Gitcoin standard button colors
- Review with core team
- Post a screenshot or video of the build

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)
- How many new bounties are funded via this flow? 
- How many hunters are engaged through this flow? 
- How many new hunters are activated? 
- Platform value from this feature

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

octavioamu:
So I see some problems with this task:

1. If Im a contributor and see a funder I use to work I will think I can propose that funder to fund an issue for me, so I can work on it and get paid. 

2. "invite to bounty" only appears for funders with open issues.

3 . "propose a project" is using terms that are not intuitive, what is "project"?  

[...]

PR #4477, opened 2019-05-24
Fix/data changes

octavioamu:
##### Description

- remove sensitive data
- api only return needed data
- fix resume / job status condition #4391
- don't show hidden profiles

##### Refers/Fixes

fix #4391
##### Testing

Tested the diff user data configurations

PR #4478, opened 2019-05-24
Propose work

octavioamu:
##### Refers/Fixes

#4416


app/dashboard/views.py
```diff
@@ -815,23 +815,39 @@ def users_fetch(request):
...
+        profile_json['keywords'] = user.keywords
+        profile_json['organizations'] = user.organizations
         profile_json['position_contributor'] = user.get_contributor_leaderboard_index()
         profile_json['position_funder'] = user.get_funder_leaderboard_index()
         profile_json['work_done'] = count_work_completed
         profile_json['work_inprogress'] = count_work_in_progress
-
-        profile_json['job_status'] = user.job_status_verbose if user.job_search_status else None
         profile_json['verification'] = user.get_my_verified_check
         profile_json['avg_rating'] = user.get_average_star_rating
```
thelostone-mc: ^ would it be cleaner to redefine the whole bit as 

```
profile_json = {
  id: user.id,
  ...
}
```

octavioamu: @thelostone-mc actually is done here https://github.com/gitcoinco/web/pull/4477

Issue #4432, opened 2019-05-17
As a Gitcoin operator, I want to patch up security issues related to uploading.


frankchen07:
### User Story

As a Gitcoin operator, I want to patch up security issues related to uploading.

### Why Is this Needed

Unrestricted file uploads are a security risk.

### Current Behavior

Unrestricted file uploads
### Expected Behavior

Restricted file uploads using `libmagic`

### Definition of Done
- [ ] restricted file uploads

kuhnchris:
Uh, differently asked: where the F are files uploaded and reused? Avatar? CV?

frankchen07:
@kuhnchris - this ticket was deliberately left vague as to not explicitly let the public know of the vulnerability itself (maybe overly cautious, but better safe than sorry). It's a core ticket that has been brought to @danlipert - feel free to reach out to him directly on Slack if you'd like to the lowdown!

PR #4479, opened 2019-05-24
Validate uploaded file type and size

danlipert:
##### Description

This PR checks uploaded files' type directly via libmagic. This is useful to narrow the potential attack vectors due to accepting user-uploaded files.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4432

##### Testing

Tested both the profile resume upload and NDA upload and reupload forms by submitting invalid files that satisfied the client-side file type (file extension) checks. Response from server was as expected with appropriate alert shades. I then manually submitted the correct form of file and was able to create a bounty, apply to work, submit work, and payout. I tested with PDF, DOCX and DOC files for the NDA.


app/dashboard/views.py
```diff
@@ -1777,6 +1783,29 @@ def profile_job_opportunity(request, handle):
...
 
+def invalid_file_response(uploaded_file, supported):
+    response = None
+    if not uploaded_file:
+        response = {
+            'status': 400,
+            'message': 'No File Found'
+        }
+    elif uploaded_file.size > 31457280:
+        # 30MB max file size
```
thelostone-mc: would 30 MB be too much?

danlipert:
Eh, I just picked a number that I think is not too high and not too low -
PDFs can be suprisingly big sometimes but more than 30 MB or so and I think
we could open up an attack vector.

-Dan Lipert

Issue #4415, opened 2019-05-15
BE - Divide Gitcoin Users List into two separate personas for designation


PixelantDesign:
### User Story

As Gitcoin, we'd like to be able to divide our users list into two separate personas: funders & contributors.

### Why Is this Needed
We want to be able to be more targeted about our experiences and messaging.

### Definition of Done

- Work with Frank - figure out the logic - what defines bounty hunger, funder active/inactive
- Translate into Django code
- Write migration to set these values for each user.
- Write a cron job that will run every day or so to recalculate whether logic still applies to them.
- Review with @frankchen07 @PixelantDesign @danlipert 

### Data Requirements
TBD

### Additional Information

https://metabase.gitcoin.co/question/272

owocki:
@willsputra has done some great work on a dashboard design that allows uesrs to toggle between funder mode and coder mode..

PR #4481, opened 2019-05-24
Funder and hunter personas in user profile, sync personas to individual mailchimp lists

danlipert:
##### Description

This PR adds two additional boolean values to user profiles, whether a user is a bounty hunter, and/or a bounty funder. In addition, a migration is added that sets these flags for each user. The SQL query is used directly to maintain compatibility between the query used here and the query used on Metabase created by @frankchen07 . Also, two endpoints are created to export these lists easily for mailing list purposes. The hunter and funder flags are set daily via cronjob for all users.

The hunter and funder flags are only set to true for a given user if the SQL query lists the user as `active`.

The different personas are then synced to 3 new mailchimp audiences on the new mailchimp account: one for funders, one for hunters, and one for everyone.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4415

##### Testing

Tested persona assignment locally by applying migration manually and checking the export endpoints. Mailchimp integration tested via unit test.


danlipert:
@frankchen07 Correct, this would only classify you as a given persona if you were active in the last three months. I will change this to use the updated metabase question so that `bounty_hunter` has `persona_is_hunter` set to True, and `funder` has `persona_is_funder` set to True. Once a given user is marked as a hunter or funder, then they won't later be have these persona flags removed.

frankchen07:
@danlipert - also, Alisa made a good point to make sure that the github handles are also in mailchimp, since when we send it out, we'll be using their handle to address them. I'm not sure if it's already in there, but the updated SQL query also has a handle listed.

frankchen07:
@danlipert - sorry caught one small change on metabase yesterday: https://metabase.gitcoin.co/question/423

line 120 - changed to `db.web3_created created_on,` instead of `db.created_on`

app/dashboard/models.py
```diff
@@ -2076,6 +2079,27 @@ def tips(self):
...
+        network = settings.ENABLE_NOTIFICATIONS_ON_NETWORK
+        designation = None
+        with connection.cursor() as cursor:
+            try:
+                cursor.execute(PERSONA_SQL, [network, self.id, network])
+                result = cursor.fetchone()
+                _, _, designation, _, _ = result
+            except Exception as e:
+                logger.info(
+                    'Exception calculating persona for user %s: %s' % (self.id,
```
thelostone-mc: should we keep this logger info in one line ?

app/dashboard/sql/persona.py
```diff
@@ -0,0 +1,322 @@
...
+        b.join_date
+    from
+        result_raw a
+    left join
+        user_join_dates b
+    on
+        a.profile_id = b.profile_id
+)
+select * from result;
+""".strip()
```
thelostone-mc: did you generate this bit / write up the raw query yourself  ?  me curious

danlipert: As noted in the description, this query is one that Frank wrote up and uses on metabase, so I use it directly here. This allows us to adjust it easily rather than reconverting to ORM code each time we change it on metabase.

app/dashboard/models.py
danlipert: I know it looks a bit weird, but its pep8 style: https://www.python.org/dev/peps/pep-0008/

danlipert:
> @danlipert - also, Alisa made a good point to make sure that the github handles are also in mailchimp, since when we send it out, we'll be using their handle to address them. I'm not sure if it's already in there, but the updated SQL query also has a handle listed.

Isn't the github handle out of scope for this ticket? It can be added at some point (as a first name, or something else) but I want to make sure this change is in sync with the other related tickets, such as @octavioamu 's. We haven't been capturing this previously in mailchimp. @PixelantDesign @frankchen07

PR #4492, opened 2019-05-27
Clear local variable accounts data after sending tip

danlipert:
##### Description

This small PR clears the account data used for creating a proxy address tip.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4469

##### Testing

Tested locally and verified 1) the fix removed the sensitive data from local storage after the tip was sent 2) the tips still work as expected


owocki:
why closed / unmerged?

danlipert:
@owocki I haven't tested it yet and accidentally set it as a live PR instead of a draft

octavioamu:
@danlipert I was checking and clear_metadata() is just 
```
var clear_metadata = function() {
  document.account = undefined;
  document.hash1 = undefined;
};
```
I am missing something?

danlipert:
@

> @danlipert I was checking and clear_metadata() is just
> 
> ```
> var clear_metadata = function() {
>   document.account = undefined;
>   document.hash1 = undefined;
> };
> ```
> 
> I am missing something?

I dont think so.... its pretty simple! This just clears some local memory that contained the private key for the proxy address that was used to send the proxy tip to, so it could be exploited by an XSS attack. Setting it to undefined after its been used stops this from happening.

octavioamu:
> @
> 
> > @danlipert I was checking and clear_metadata() is just
> > ```
> > var clear_metadata = function() {
> >   document.account = undefined;
> >   document.hash1 = undefined;
> > };
> > ```
> > 
> > 
> > I am missing something?
> 
> I dont think so.... its pretty simple! This just clears some local memory that contained the private key for the proxy address that was used to send the proxy tip to, so it could be exploited by an XSS attack. Setting it to undefined after its been used stops this from happening.

But is just cleaning a variable not localstorage right?

danlipert:
Needs some additional fixes after chatting with @octavioamu

Issue #4545, opened 2019-05-31
I used the ‘Share’ feature on Issue Details but the email never made it to the bounty hunter.


PixelantDesign:
They are flagged as 'dangerous emails' and are not making it through to hunters.

PixelantDesign:
@octavioamu any thoughts here?

PR #4549, opened 2019-05-31
fix spoofing email sender on bounty invite

danlipert:
##### Description

Share bounty/invite user to bounty emails currently put the user who invited's email address as the sender, which is basically spoofing as them. This is bad for many reasons, but mostly it lowers our email reputation.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4545

##### Testing

Untested


danlipert:
@thelostone-mc I just tried testing this but the email still had my address as the sender... not sure why lets discuss before the deploy - would love to get this fix in

PR #4580, opened 2019-06-05
makes grant info calculation imdipident

owocki:
##### Description

makes grant info calculation imdipident

this makes debugging / correcting thte grant detail information much easier, since you don't have to reason about when the callback functions were calculated.  you just construct the info from state everytime you calc it.

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1559671421017900

##### Testing

tested it locally.   http://bits.owocki.com/424bf1731bfb/Image%202019-06-04%20at%208.02.04%20PM.png


owocki:
https://github.com/gitcoinco/web/pull/4582

PR #4582, opened 2019-06-05
makes grant info calculation imdipident

owocki:
##### Description

makes grant info calculation imdipident

this makes debugging / correcting thte grant detail information much easier, since you don't have to reason about when the callback functions were calculated.  you just construct the info from state everytime you calc it.

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1559671421017900

##### Testing

tested it locally.   http://bits.owocki.com/424bf1731bfb/Image%202019-06-04%20at%208.02.04%20PM.png


app/grants/models.py
```diff
@@ -659,14 +661,17 @@ def get_converted_amount(self):
             return None
 
     def get_converted_monthly_amount(self):
-        converted_amount = self.get_converted_amount()
+        if self.num_tx_approved == 1:
```
danlipert: Whats this line for? Checking if its a one-time contribution?

app/grants/models.py
```diff
@@ -700,6 +705,23 @@ def successful_contribution(self, tx_id):
         return contribution
 
 
+@receiver(pre_save, sender=Grant, dispatch_uid="psave_grant")
+def psave_grant(sender, instance, **kwargs):
+    
+    instance.amount_received = 0
+    instance.monthly_amount_subscribed = 0
+    print(instance.id)
```
danlipert: leftover print

app/grants/models.py
```diff
@@ -700,6 +705,23 @@ def successful_contribution(self, tx_id):
...
+    for subscription in instance.subscriptions.filter(error=False):
+        value_usdt = subscription.get_converted_amount()
+        for contrib in subscription.subscription_contribution.filter(success=True):
+            if value_usdt:
+                instance.amount_received += Decimal(value_usdt)
+
+        if subscription.num_tx_processed <= subscription.num_tx_approved and value_usdt:
+            instance.monthly_amount_subscribed += subscription.get_converted_monthly_amount()
+        print("-", subscription.id, value_usdt, instance.monthly_amount_subscribed )
```
danlipert: another print :) if we want these, lets use the logger 👍

PR #4583, opened 2019-06-05
sentry: fix undefined error

thelostone-mc:
##### Fixes

- [Cannot read property 'length' of undefined](https://sentry.io/organizations/gitcoin/issues/1038441468/)
- [Cannot read property 'now' of undefined](https://sentry.io/organizations/gitcoin/issues/1038910274/)
- [Cannot read property 'style' of null
](https://sentry.io/organizations/gitcoin/issues/1058239597/)
- ['Web3' is undefined](https://sentry.io/organizations/gitcoin/issues/915316423)
- [compiledSplitter0 is not defined](https://sentry.io/organizations/gitcoin/issues/1038512681) - on grants index page

### DIFF
--- a/app/assets/v2/js/base.js
+++ b/app/assets/v2/js/base.js
@@ -136,12 +136,14 @@ $(document).ready(function() {
     element.toggleClass('active');
     let panel = element[0].nextElementSibling;
 
-    if (panel.style.maxHeight) {
-      panel.style.maxHeight = null;
-      panel.style.marginBottom = 0 + 'px';
-    } else {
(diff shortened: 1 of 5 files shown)


thelostone-mc:
@danlipert -> checking on which page the error appears -> recreate the same env and fix the error
All of them add an extra if check which would otherwise through undefined if it continued execution

danlipert:
@thelostone-mc Sounds good but I worry that in squashing these alarms, we may introduce additional issues so I think it needs some solid manual testing. I'm down to do it in pre-deploy regression tests though, what do you think?

PR #4585, opened 2019-06-05
Explore Users / Explore Bounties + Featured Bounty recoloring

kuhnchris:
##### Description

As per request from @PixelantDesign 

##### Refers/Fixes

No related issues or PRs have been opened yet.

##### Testing

Tested, see example(s) below:
grafik
grafik


kuhnchris:
@danlipert could you do me a favor and check this out? Or at least throw it around internally? Thanks!

PixelantDesign:
Looks good to me, let's see how it performs.

danlipert:
@kuhnchris code looks fine to me - whats your concern? If you are happy with it lets merge it

kuhnchris:
For all I care we could merge it since it's not a critical issue, if it's fine by you we can merge it. 👍

danlipert:
@kuhnchris - what I mean is, was there some concern that you have that you wanted me to double check by testing manually on my local instance? Typically we don't test contributor PRs unless they are particularly complicated for the developer is unsure of their changes - we just read through the code to look for errors.

kuhnchris:
That's fine by me then. There isn't anything particular that could "break" .
The thing is just, that all of these templates are so fragile, if something goes bad or breaks, then prod is broken, since it's not just "oh a part of the sidebar is now broken", - no, if something in the sidebar breaks, /explorer is done for until we fix it. 
Hence my question.
But it workedon my local test instance, so I'm OK with a merge. :-)

app/dashboard/templates/dashboard/sidebar_search.html
```diff
@@ -17,6 +17,11 @@
 {% load i18n static %}
 <link rel="stylesheet" href={% static "v2/css/sidebar_search.css" %}>
 <div class="sidebar_search font-body">
+  <div>
+      <a href="{% url "users_directory" %}"><img style="width: 100%" src="{%  static 'v2/images/promote_userdir.png' %}"></a>
```
octavioamu: can we use w-100 class instead of the inline style?

app/dashboard/templates/dashboard/users.html
```diff
@@ -112,6 +112,11 @@
               </div>
             </div>
             <button class="btn blue" @click="params = {}; searchUsers()">Reset Filters</button>
+
+            <div>
+              <a href="{% url "explorer" %}"><img src="{%  static 'v2/images/promote_bounties.png' %}" style="width: 100%"></a>
```
octavioamu: can we use w-100 class instead of the inline style?

thelostone-mc:
^ @kuhnchris

thelostone-mc:
I think for now we are moving to another direction, closing for now! 
Might need to revisit the design ! 

Sorry about the delay @kuhnchris 😓

Issue #4544, opened 2019-05-31
There are people already working on bounties in 'Open'. How do we move this to another state?


PixelantDesign:
### User Story
As a Gitcoin contributor I'd like to be able to filter/ see/ work on bounties which no one has worked on yet.

### Defintion of Done
Add a dropdown on the left rail filters .....that allows user to filter by # of applicants.
0 Applicants
1-5 Applicants

- [ ] Please Implement this drop down

![image (4)](https://user-images.githubusercontent.com/238235/59164970-6d269d80-8ae2-11e9-859f-807c3d76ef93.png)
- [ ] Review with @octavioamu @PixelantDesign

PixelantDesign:
Are contributors just starting work without being accepted? What do you mean @vs77bb ?

We could try something like this?

![No Applicants](https://user-images.githubusercontent.com/238235/58762403-c9575380-851d-11e9-9776-798af7b3561a.png)

rafalkowalski:
I can work on this issue.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.32 ETH (73.96 USD @ $231.12/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4544/3099).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/4544/3099).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $77,594.24 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #4613, opened 2019-06-10
Add filtering by applicants in bounties explorer

rafalkowalski:
Fixes issue #4544 

Kindly review
@thelostone-mc 

image


PixelantDesign:
What is the default state?
We'd like the default to be 0 and the second option to be 1-5.

Thanks!

rafalkowalski:
@PixelantDesign I've set a default selection to 0

PixelantDesign:
Thank you!

octavioamu:
Hey @rafalkowalski did you tested this in hackathon explorer? I think the filter will apply also there.

rafalkowalski:
@octavioamu Applicants filter isn't visible in hackathon explorer
image

app/dashboard/router.py
```diff
@@ -252,6 +254,14 @@ def get_queryset(self):
             statuses = self.request.query_params.get('status__in').split(',')
             queryset = queryset.filter(idx_status__in=statuses)
 
+        if 'applicants' in param_keys:
```
danlipert: When the user selects "ALL" from the dropdown, the query parameter is set to "ALL", correct? How is this handled in the backend? If the idea is to just skip the filter, then we shouldn't add an annotation as it is bad for performance (especially with a `Count()`)

danlipert:
> @octavioamu Applicants filter isn't visible in hackathon explorer
> image

@rafalkowalski Even though the filter isn't visible, the backend code that handles the request is the same. Since we store the filter in localStorage, we must make sure the same query parameter isn't submitted with the request on the hackathon page (I think this is what @octavioamu is referring to, we've had issues with this in the past)

app/dashboard/router.py
rafalkowalski: @danlipert I've fixed it. Could you check?

octavioamu:
@rafalkowalski still failing in travis, there are some js lint errors run `npm run eslint` in your local to check it. Also update the branch with the last master changes.

app/dashboard/router.py
danlipert: Looks good - thanks!

rafalkowalski:
@danlipert I've rebased it :)

app/dashboard/router.py
```diff
@@ -267,6 +269,17 @@ def get_queryset(self):
...
+            applicants = self.request.query_params.get('applicants')
```
octavioamu: Can we do this more dynamic? since if we decide to add diff values in the future this will be parsed to the query?  something like https://github.com/gitcoinco/web/blob/master/app/dashboard/views.py#L762

app/dashboard/templates/dashboard/sidebar_search.html
```diff
@@ -153,6 +153,19 @@
...
 
+    <div class="col">
+      <div class="col-12 subheading">
+        Applicants
+      </div>
+      <div class="col-12 options">
+        <select class="w-100 js-select2" name="applicants" id="applicants_box">
+          <option value="0">0</option>
+          <option value="1-5">1-5</option>
+          <option value="ALL">ALL</option>
```
octavioamu: can we setup the default to "all"  so people will still see all bounties by default. Will be hard for users to understand if we show only 0 applicants by default and some of them will feel this as a bug.

rafalkowalski: @PixelantDesign what do you think about "all" as default value?

app/dashboard/router.py
rafalkowalski: done

app/dashboard/templates/dashboard/sidebar_search.html
thelostone-mc: I second @octavioamu !

PixelantDesign: Let's set default to all. Thanks @rafalkowalski!

rafalkowalski: done

PixelantDesign:
Hey guys! I just looked at some data and think it's best if we filter by 0 applicants as a default. I don't want to post the data here, but hit me up if you guys want to chat.

@rafalkowalski could we make the change to default to 0? Thank you!

app/dashboard/templates/dashboard/sidebar_search.html
PixelantDesign: Hey guys! I just looked at some data and think it's best if we filter by 0 applicants as a default. I don't want to post the data here, but hit me up if you guys want to chat.

@rafalkowalski could we make the change to default to 0? Thank you!

rafalkowalski: Ok i've changed to 0 value as a default

app/dashboard/router.py
```diff
@@ -267,6 +269,16 @@ def get_queryset(self):
             statuses = self.request.query_params.get('status__in').split(',')
             queryset = queryset.filter(idx_status__in=statuses)
 
+        applicants = self.request.query_params.get('applicants')
```
owocki: did you guys do any testing on performance here?  this is a multi table JOIN between bounty + interest tables, and it increases the load time for the API on the explorer from 500ms to 10s+

owocki: note the diff between the `applicants=0` calls [here](http://bits.owocki.com/69cbd5ff245e/Image%202019-07-10%20at%201.44.09%20PM.png) vs the ones that dont have that call in them...

owocki: the explorer page defaults to applicants=0, thus making most explorer lookups 10+s

rafalkowalski: @owocki I will fix it

PR #4639, opened 2019-06-13
token warnings

owocki:
##### Description

warns the user upon posting a bounty if their `approve()` is not enough

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1560434810014200

##### Testing

still need to test more
probably also need to tell the user if theres not enough balance.


thelostone-mc:
@octavioamu isn't this already fixed in https://github.com/gitcoinco/web/pull/4742 ?

PR #4644, opened 2019-06-14
Activities in Navigation Bar

kuhnchris:
##### Description
Per @PixelantDesign 's direct request

##### Refers/Fixes
n/a

##### Testing
grafik


PixelantDesign:
This is nit picky but could you change it to **Activity**? :)

kuhnchris:
...but only because you asked so nice. ;-)
Added it

PR #4646, opened 2019-06-17
feat: override org name + photo

thelostone-mc:
#### Description

- updated models Profile & Bounty -> override name + url
- added provision in admin to enable admin overrides
- updated template to use newly added property
- update existing prop to use override name + url

#### Steps

In this example -> assume we'd be updating metamask name to xyz and their logo

##### Update Profile (/profile/xyz)

- head over to admin profile section and click on profile you need to update 
- update `Admin override name` to the new name (xyz) 
- upload  photo `Admin override avatar:`

^ This would ensure that the profile page of the org shows the overridden name + photo

##### Update Bounty (only for bounty + hackathon)

This is the tricky part.

- You'll have to select all the bounties created by the org
- For each bounty -> navigate to the bounty admin
- update `Admin override org name` to the new name (xyz) 
- upload  photo `Admin override org logo`

^ This would ensure that the bounty details page of the org shows the overridden name + photo
These are temp fixes which will be removed out once orgs is built out as a feature

note: do not use this for non-hackathon bounties ( especially name change as the filters extra map to the actual Github name) 

Demo: Check [HERE](https://embed.vidyard.com/share/HzeozMUDdxCXKqWCQcE1pY?)


danlipert:
Very nice! I think the URL issue is okay since its based on their github name, so its not "wrong" per se, just not perfect - and changing it would be a big headache. @vs77bb @alexvotofuture what do y'all think?

app/dashboard/models.py
```diff
@@ -310,6 +310,13 @@ class Bounty(SuperModel):
     admin_mark_as_remarket_ready = models.BooleanField(
         default=False, help_text=_('Admin override to mark as remarketing ready')
     )
+    admin_override_org_name = models.CharField(max_length=255, blank=True) # TODO: Remove POST ORGS
```
danlipert: What is this TODO referring to?

app/dashboard/models.py
```diff
@@ -2390,13 +2416,21 @@ def absolute_url(self):
 
     @property
     def username(self):
-        handle = ''
```
danlipert: Nice change! Is there a case where both if statements return false and we end up returning `None` here?

app/dashboard/templates/dashboard/sidebar_search_hackathon.html
```diff
@@ -51,10 +51,10 @@
 
             
         {% for org in orgs %}
-          <div class="form__radio option {{org}}">
-            <input name="org" id="{{org}}" type="radio" value="{{org}}" val-ui="Turing-Chain">
```
danlipert: what the heck is "Turing-Chain" ? 🤔

app/dashboard/models.py
thelostone-mc: Oh not that I've seen of but it's safe to add it! nice catch 🙌

app/dashboard/models.py
thelostone-mc: So ideally if we built out ORGS the right way -> we shall remove this at that point 🙌 
we would do it right way :P 

I leave TODOS in code so that I can fix it up later when the time is right :P

app/dashboard/templates/dashboard/sidebar_search_hackathon.html
thelostone-mc: I have no clue :P I was tempted to remove it but then thought maybe check it out later :P

danlipert: 😂

app/dashboard/models.py
danlipert: 👍

thelostone-mc:
@gitcoinco/engineers So look like the hackathon event association  does get ported over in merge_bounty 🤷‍♂ 
False alarm! Added the overridden fields there and tested it out.
Overrides stay on blockchain interaction steps 


Screenshot 2019-06-19 at 10 26 34 PM

app/dashboard/models.py
thelostone-mc: handled @danlipert  ^_^

danlipert:
@thelostone-mc we'll need to rename the migration to 0035 and fix, I have one for dashboard as well (lots of migrations this deploy)

PR #4672, opened 2019-06-24
fix coupon syncing

danlipert:
Fix bounties with coupons not showing in explorer


thelostone-mc:
closing this out as it was already pushed into stable

PR #4685, opened 2019-06-25
sentry: Quill not defined

thelostone-mc:
##### Description

so when you switch tabs on grants  -> owocki added a change where page gets reloaded and only that particular tab content would get rendered in the HTML

So when you go to any other tab except description -> Quill tries to find the root element to which is gets instantiated with ->  which is never found in the generated HTML

##### Testing

Tested locally 
- still can update the description 
- the error disappears from the console


thelostone-mc:
Closing this out ! will re-write it better and update it

PR #4708, opened 2019-06-28
newsletter 062819

danlipert:


thelostone-mc:
@danlipert is this needed / can we close it out?

PR #4716, opened 2019-06-29
Fix reserve bounty to not require approval

### LINKED ISSUE #4395: Reserve Bounty Should Not Require Approval
### User Story
As a funder of a bounty, when selecting reserve bounty I'd like to not have to approve a bounty for a contributor I trust and have worked with before.

### Definition of Done
- Implement a fix to this so that when a funder selects reserve approval is not required.
- Uncheck auto approve after 72 hours when reserve bounty is selected.

### Why
- This will make the experience more intuitive, and creates less processing time for a funder to understand all of the inputs on the funder form, increasing number of funded bounties.

rafalkowalski:
##### Description

I've added an event listener on "reserved for" select box. If user will select user then automatically permissionless radio will be set.

##### Refers/Fixes
#4395 

##### Testing

https://drive.google.com/file/d/1DXeudMtQ3Xz6sQXAXytM6aX1FNRoBzgz/view


octavioamu:
humm not sure about this approach @rafalkowalski because then someone with the link will be able to get approved to work right? 
I believe the check need to happen in the moment the hunter applies, in the detail bounty page,  then check if reserved for the user applying , if true then get automatic approved.

danlipert:
> humm not sure about this approach @rafalkowalski because then someone with the link will be able to get approved to work right?
> I believe the check need to happen in the moment the hunter applies, in the detail bounty page, then check if reserved for the user applying , if true then get automatic approved.

Yeah this is a good point - would this allow another user to "snipe" the bounty?

app/dashboard/views.py
```diff
@@ -307,7 +307,12 @@ def new_interest(request, bounty_id):
     approval_required = bounty.permission_type == 'approval'
     if approval_required:
         msg = _("You have applied to start work.  If approved, you will be notified via email.")
-
+    elif not approval_required and bounty.bounty_reserved_for_user != profile:
```
thelostone-mc: Hmm this could work ! You let others apply for work letting funders know there are others wanting to work on it  + encourages the reserved user to make a decision quicker on whether they'd be working on it

octavioamu:
@rafalkowalski resolve the isort problem and we will be good to merge!

rafalkowalski:
@octavioamu done

PR #4719, opened 2019-06-29
bug: trigger alert notifying grant ownership process

thelostone-mc:
##### Description

Bug reported:  Grant Ownership not working

Actually what happened: 
- Alert wasn't popping up when grant owner attempted to transfer ownership
- The code was present but the conditional statements weren't get triggered (which is fixed )
- Despite the alert not showing up bug -> the functionality still works 

An email gets sent to the new potential owner and only when they accept it will ownership change (as shown in the video recording )

##### Testing

Check out [HERE](https://embed.vidyard.com/share/n5HGQfSVc5tpVrfnMhVVXm?)


thelostone-mc:
> What does change ownership actually do

Yup! It allows change owner to simply edit grant metadata. That's about

PR #4721, opened 2019-07-01
[NOT MERGE] Graphql Setup

octavioamu:
##### Description

Experimenting with graphql 

*Why this is cool?* 
https://hasura.io generate a UI from the db structure and models from django and let you setup a Graphql API, if this is successful we are going to be able to develop features faster without need to setup complex queries in the BE.
Actions like order/filtering will be super easy from the frontend.

*Ok wanna play too, how?*
docker-compose up -d 
localhost:9000 
use password mysecretkey
Have fun 
image

### DIFF
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -56,6 +56,7 @@ from kudos.router import router as kdrouter
 from .sitemaps import sitemaps
 
 urlpatterns = [
+    path('validate_request', dashboard.views.inspectUser, name='inspectUser'),
     path('jsi18n/', JavaScriptCatalog.as_view(), name='javascript-catalog'),
 
     # create bounty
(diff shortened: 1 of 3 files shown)


thelostone-mc:
Just checked it out locally -> this is pretty damn nice

owocki:
i dont have much experience with GraphQL! so this is a replacement for the API in some sense, because the frontend can query the backend right?  seems like it also might replace the admin models in that you can look at the DB data in a more native UI..

how do we handle things like permissions here?  like i want to make sure that NO ONE can access my github key, twitter key, etc.  i want to give limited access to my email address across the network, etc..

thelostone-mc:
@octavioamu closing this ! Not deleting the branch (in case we ever need to revisit this )

PR #4732, opened 2019-07-03
Load codefund weekly roundup sponsor context via API

danlipert:
##### Description

This change to the rendering of the weekly roundup allows for an optional settings entry for a URL that will return the appropriate context data for the codefund sponsor in the weekly roundup email. This will make it easier for the newsletter data to be incorporated.

##### Refers/Fixes

Discussed with @coderberry as a result of our weekly sync discussions

##### Testing

Not yet tested

### DIFF
--- a/app/retail/emails.py
+++ b/app/retail/emails.py
@@ -993,16 +993,24 @@ Back to shipping,
         'link_copy': 'View more',
     }, ]
 
-    sponsor = {
-        'name': 'Blockchain Job Kit',
-        'title': 'Blockchain Developer Job Kit',
-        'image_url': 'https://s3.us-west-2.amazonaws.com/gitcoin-static/jDSk7ZTfpY19PWdwwsk8puNd.png',
-        'link': 'http://bit.ly/EthDevKit',
-        'cta': 'View now',
-        'body': [
-            'Blockchain Developer Job Kit - See who’s hiring, salary info, and dev skills required'
-        ]
-    }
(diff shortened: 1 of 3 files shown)


app/retail/emails.py
```diff
@@ -993,16 +994,24 @@ def render_new_bounty_roundup(to_email):
...
-        'name': 'Blockchain Job Kit',
-        'title': 'Blockchain Developer Job Kit',
-        'image_url': 'https://s3.us-west-2.amazonaws.com/gitcoin-static/jDSk7ZTfpY19PWdwwsk8puNd.png',
-        'link': 'http://bit.ly/EthDevKit',
-        'cta': 'View now',
-        'body': [
-            'Blockchain Developer Job Kit - See who’s hiring, salary info, and dev skills required'
-        ]
-    }
+    sponsor_response = (requests.get(settings.CODEFUND_NEWSLETTER_URL)
```
owocki: note that this will run 25k times per mgmt command (and will grow bounded to the size of the email list).  it might be better performance and reliability-wise to call it once, and then to store the results in memory per-management command execution...

danlipert: Yeah good point - we should probably render the entire template once for better performance since theres no user-specific info in it.

thelostone-mc:
Yo @owocki / @danlipert we still want this in ?

thelostone-mc:
Closing this out

PR #4742, opened 2019-07-05
Check funds before popup metamask

### LINKED ISSUE #4618: Metamask requests fee payment when funds aren't available
**Describe the bug**

If a user doesn't have enough funds in their account to pay for the bounty, the user is still prompted to pay the fee associated with posting.

https://gyazo.com/80f9acc058a4348aca9eea296664a597

**Expected behavior**

If a user does not have enough funds in their account Metamask should not prompt the user and they should instead be asked only to refill their account, to avoid users paying fees for failed bounties.

octavioamu:
##### Description

Add a balance checker before pops metamask to avoid people paying fees and then get stuck for the next payment.
Also it add featured to the math to check if the funder have enough funds. 
Removes loading button state when fail to let the funder make changes and resubmit 
https://embed.vidyard.com/share/FQJZYZDoX8q8wFgwMuBQSA?

##### Refers/Fixes

Fix #4618
##### Testing

Tested in rinkeby and mainnet (without completing the TX)  
Tested with variations of using coupons with normal and 0 fees
Tested with featured bounties activated/deactivated 
Tested with ETH and DAI


frankchen07:
cool, maybe in v2 we can increase the number of sig figs in the decimal amount (since it said you have 0.4 ETH but need 0.4 ETH (technically 0.4011 or something).

danlipert:
@octavioamu I merged master into this branch and some random tests started failing :( not sure why

Issue #4674, opened 2019-06-24
As a funder on the Gitcoin platform, I'd like a central place to check-in to see how my bounties are doing, and know the actions I need to take next.


PixelantDesign:
### User Story

As a funder on the Gitcoin platform, I'd like a central place to check-in to see how my bounties are doing, and know the actions I need to take next.

### Why Is this Needed

No Dashboard APIs in Gitcoin

### Description

_This task is to set up the backend only._

- Set up the dashboard api
- Set up a per-bounty api that gives the applicant details shown In mockup
- Pull in Contributor Icon and github name
- Pull in high level bounty stats up top
- Break up page in 3 sections
- Open Bounties, Submitted Bounties, Expired Bounties (see mockup)

**Bounties Section**
- Provide data end points for elements that are displayed
- _Bounty:_ order by most recent applicants
- _Applicants:_ order by the time each applied
- Pull in data for all open bounties. 
- Allow ability to accept/reject
- Display all bounties
- Allow clicking to user directory in case the right contributors don’t apply.

**Submission Section**
-  _Bounty:_ Order by most recent applicant submission
- _Applicants:_ order by the time each contributor submitted, recent first.
- Provide User name eth address, pr info, commit count
- Allow ability to accept and reject

**Expired bounty**
- Pull in relevant information
- Order based on recently expired bounty
- Clicking extend expiration takes user to issue details with expiration date modal open.

### Current Behavior
No Dashboard APIs

### Expected Behavior
Dashboard APIs outlined above are set up and working

### Definition of Done

- Dashboard APIs outlined above are set up and working

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)

### Additional Information
[See the Latest design ](https://consensys.invisionapp.com/share/NYSOGHX2D8R#/371051245_Dashboard_-1-)

owocki:
latest design => https://consensys.invisionapp.com/share/NYSOGHX2D8R#/screens/370588236_Funder_Dashboard_Spec

PixelantDesign:
[See the Latest design ](https://consensys.invisionapp.com/share/NYSOGHX2D8R#/371051245_Dashboard_-1-)

PR #4743, opened 2019-07-05
Dashboard backend

danlipert:
##### Description

This PR provides a backend for the new user dashboard. There are two endpoints - one that gives the id for each relevant bounty in the dashboard. The second gives per-bounty data for each bounty such that they can be loaded individually.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4674

##### Testing

Tested locally with the following steps:
1- Sync rinkeby bounties to local instance
2- Assign all bounties to my test account
3- Manually set up fulfillments, interests, etc. in database
4- Manually check API endpoints (see screenshots below)

`/funder_dashboard/`
Screen Shot 2019-07-08 at 7 08 26 PM

An open bounty:
Screen Shot 2019-07-08 at 10 31 56 PM

A submitted bounty:
Screen Shot 2019-07-08 at 9 55 59 PM

An expired bounty:
Screen Shot 2019-07-08 at 9 56 43 PM


octavioamu:
@danlipert I was checking this with the design, the problem I see is `funder_dashboard/` only returns Id's but to get the data and show the title I will need to fetch all in load time `funder_dashboard/bounties/<int:bounty_id>/` , I believe the optimised way should be  the first load bring the first view state information:
-title and bounty details like amount, type, gh link, expiration
image

then the other endpoint will be fetched when click in "View applicants" the data in request will be the one related to interest and profile:
image

I know you started before the design was ready so maybe that is the reason, this is the last design:
 
image

app/dashboard/views.py
```diff
@@ -2812,3 +2812,91 @@ def get_hackathons(request):
...
+        return JsonResponse(
+            {'error': _('You must be authenticated via github to use this feature!')},
+            status=401)
+
+    profile = request.user.profile
+
+    open_bounties = list(Bounty.objects.filter(
+        Q(idx_status='open') | Q(idx_status='open'),
+        current_bounty=True,
+        bounty_owner_profile=profile
```
octavioamu: Owner profile doesn't get filled a lot of times making the query to don't show nothing

octavioamu: Also I have an idx_status open, current_bounty=True, filled bounty_owner_profile=the logged user and the query returns the json without data. Not sure why.

octavioamu: Forget that last one, was a problem between the chair and the screen.
But the first one still a problem because sometimes the owner profile doesn't get filled in bounty.

danlipert:
> @danlipert I was checking this with the design, the problem I see is `funder_dashboard/` only returns Id's but to get the data and show the title I will need to fetch all in load time `funder_dashboard/bounties/<int:bounty_id>/` , I believe the optimised way should be the first load bring the first view state information:

I was thinking that we'd paint each row individually instead of painting them all at once, so the `funder_dashboard/` request basically gives you the information to set up the individual requests. Something like skeleton rendering

app/dashboard/views.py
danlipert: Do you think we should fix that the owner profile isn't getting filled in the bounty as expected?

octavioamu:
Closing this one as was solved in other PR

Issue #4706, opened 2019-06-28
Favorites REST API (just Backend)


octavioamu:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a user I want to be able to favorite bounties, kudos, users and grants to follow what interest me 
### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
Today is hard to track something that interest me if I didn't interacted with that in the past, even interacting, something need to change to receive a notification.
This will be integrated into the dashboard data (#4153) and will generate diff tickets for each UI integrations for each product areas (bounties, kudos, users/orgs and grants)

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature
Create different endpoints to interact with favorites table. This table will be a table of relations between the favorite and the sourced data ID

**[Endpoints docs](https://documenter.getpostman.com/view/18800/S1a626JH?version=latest)**
`GET api/v0.1/favorites/:type`
`POST api/v0.1/favorites/:type/:id`
`api/v0.1/favorites/:type/:id`

**Documentation**
The documentation of the endpoints is in postman https://documenter.getpostman.com/view/18800/S1a626JH?version=latest

### Current Behavior
[comment]: # (Describe what actually happened.)
Doesn't exist

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Having an REST API ready to integrate

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
API tested 

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)
When implemented the UI this can make easy the life of users to follow what interest them most, finding bounties or people will be an easy task
Increment use of the platform
Decrement frustration
Make gitcoin more personal

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)
https://documenter.getpostman.com/view/18800/S1a626JH?version=latest
This task is the very beginning and will be a final feature after design and front end work get done. 
With favorites new features will be possible, for example suggest favorite users when creating a bounty, follow up organizations new bounties, etc.

usmanmuhd:
@octavioamu I'm interested in this one.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

[...]

PR #4749, opened 2019-07-07
#4718 - Fixes broken Kudos API

kuhnchris:
##### Description
"price" does not exist in Kudos' Token model. It's price_finney

##### Refers/Fixes
Fixes: #4718

##### Testing
Locally done.

PR #4751, opened 2019-07-08
Favorite backend

usmanmuhd:
##### Description

Adds backend for the favorites API

##### Refers/Fixes

Fixes: #4706 

##### Testing

No tests present currently.

CC: @octavioamu @danlipert


danlipert:
@usmanmuhd Is this ready for review? If not, please use the "draft pull request" feature here on Github or just add [WIP] to the title of the PR 👍

octavioamu:
@usmanmuhd how is going with this one? we need this ASAP

kuhnchris:
@usmanmuhd asked me to take over this, so I'm on it today

kuhnchris:
This now has a subdependency on #4749 - price/price_finney causes problems with the serializer

kuhnchris:
@usmanmuhd the PR is ready as far as I can tell - GET, POST and DELETE works for /favorites/bounty, kudos, grants and user.
@octavioamu as discussed: please review. :-)

thelostone-mc:
Could we get the travis to pass ?

kuhnchris:
@thelostone-mc I asked @octavioamu he said he didn't want them squished. I can squish them no problemo.

kuhnchris:
fyi the build failed thanks to slack... we may want to remove that check...

octavioamu:
@kuhnchris I don't think the problem is with slack https://travis-ci.org/gitcoinco/web/builds/568469967#L2754
could be related to another pr merged

app/favorites/views.py
```diff
@@ -0,0 +1,114 @@
...
+from rest_framework.decorators import api_view, authentication_classes, permission_classes
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from .models import Favorite
+
+
+@api_view(['POST', 'DELETE'])
+@authentication_classes((SessionAuthentication, BasicAuthentication))
```
octavioamu: @kuhnchris I don't think we are sharing basicauth and sessinauth with the front end just a http cookies, that means there is not possible to get http cookies from javascript. 
That is the way if you do a jwt auth or something like that and store the tokens on login to use later in any auth request.

octavioamu: unless is taking from the request tokens, I think that is the case, anyway get works perfectly but post give me 403

octavioamu: image

app/favorites/views.py
```diff
@@ -0,0 +1,114 @@
...
+        elif reqTypes[favType] == "GRAN":
+            for obj in Grant.objects.filter(id=fav.obj_id):
+                ser = GrantSerializer(obj)
+                jsonRetData.append(ser.data)
+        elif reqTypes[favType] == "USER":
+            for obj in Profile.objects.filter(user=fav.obj_id):
+                ser = ProfileSerializer(obj)
+                jsonRetData.append(ser.data)
+
+    return Response(json.dumps(jsonRetData, indent=4, sort_keys=True, default=str))
```
octavioamu: The responses are not json formatted 
image

octavioamu: return Response(**json.loads**(json.dumps(jsonRetData, indent=4, sort_keys=True, default=str)))
can solve it

app/favorites/views.py
octavioamu: you can test it on the console with something like:
```
var postFav = fetchData('api/v0.1/favorites/user/2', 'POST')
$.when(postFav).then((response, status, statusCode) => {
  console.log(response, status, statusCode)
})
```

kuhnchris: Well if you are authorized (e.a. you have access to admin) then it does indeed work. (as i tried it myself) - I'm not 100% sure how we handle the github auth - do we have any other paths where we require authentification where I could steal the code from? Else we may need to check how to protect the data. (I tested it via Boomerang REST Extension for Chrome)

app/favorites/views.py
kuhnchris: My bad. I had to play around with recursion and avoiding errors that I probabbly forgot about that. I'll add your fix in

app/favorites/views.py
kuhnchris: Also, why would I use Javascript for auth? that's a pyhton file you are looking at - it checks on the backend (django) if you are auth - and if you are authed on the entire app it checks the User Object's permission and executes the function if the permission is sufficient. I do not think there are any additional API requirements, but we'd need to check RESTful API's documentation (django-rest-framework) if they may impose a special permissions...

kuhnchris: After a quick check - not sure but maybe I can use the django.contrib.auth.decorators.login_required.

It SHOULD provide a request.user if it does, and a request.user.profile (a gitcoin Profile object) if it is authenticated.

I can try that out if it's fine with you.

See dashboard/views.py and dashboard/utils.py - helper_handle_access_token and new_bounty method, those are taking those assumtions.

```
from django.contrib.auth.decorators import login_required
...
@login_required
def new_bounty(request):
...
        user=request.user if request.user.is_authenticated else None,
...
```

kuhnchris: OK I got what you mean now - i do get a 403, but the 403 tells me I need the CSRF token set. We do not have access to the CSRF Token tho. This only happens if we use SessionAuth. Which we do. So we either need to CSRFExempt the API calls (apparently not really viable, or easy doable), or use a different auth scheme like tokens, as you mentioned before.

octavioamu: yes but anyway try to comment the auth part and do a post with the command I sent to you.  Get are working with the same auth validator but this endpoint with Post doesn't work also without that

octavioamu: @kuhnchris how is going? any problem? Ping me if you have any question about my comment.

kuhnchris: well, all my attempts to implement this right now failed, from trying to bypass csrf to try to use another scheme (which we cannot) to trying to bypass the csrf somehow again. up to now success until now. any ideas/pointers? The only "idea" i could get at this point is to redo everything in a token-auth based thing but that'd break explorer and co. :-(

thelostone-mc:
Closing this out as this feature ended up not being on the things we'd wanna build out ! 
Thanks @usmanmuhd for the PR  :D

octavioamu:
Reopening seems we want to use this for townsquare favorite posts #6121

octavioamu:
Reopening seems we want to use this for townsquare favorite posts #6121

Issue #4620, opened 2019-06-11
as a poweruser, i want to cache my profile info so that it loads in < 0.1S


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a poweruser, i want to cache my profile info so that it loads in < 0.1S

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
kevin ( https://gitcoin.co/profile/owocki ) is a poweruser of gitcoin.  because of that he has lots of activity.  because of that, his profile loads pretty slow.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
feature

### Current Behavior
[comment]: # (Describe what actually happened.)
kevin's profile takes 8s to load and loadtime is correleated to the amount of activity

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
kevins profile takes 0.1s to load.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
you may extend the pattern created at https://github.com/gitcoinco/web/pull/4316

the basic approach here is to make sure that the profile is cached in the JSONStore object.  
we will also have to make sure that anytime the profile changes, that the profile cache object (the JSONSTore) is invalidated and updated.  you may do this by creating a `profile.rebuild_profile_cache()` object that is called as a `post_save` hook on the `dashboard.models.Profile` object.

the PR should not create any regressions in the quality of the profile page.

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)
n/a

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)
please only start work on this issue if you can submit a PR that implements this by June 16th 2019.
🖖

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1.701 ETH (412.08 USD @ $242.26/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4620/3106).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/4620/3106).
[...]

PR #4762, opened 2019-07-09
more profile speed improvments

owocki:
##### Description

more profile speed improvments (see comments inline for desc)

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4620

##### Testing

tested locally

### DIFF
--- a/app/dashboard/models.py
+++ b/app/dashboard/models.py
@@ -2641,8 +2641,7 @@ class Profile(SuperModel):
 
         """
         eth_sum = 0
-
-        if not bounties:
+        if bounties is None:
             if sum_type == 'funded':
                 bounties = self.get_funded_bounties(network=network)
             elif sum_type == 'collected':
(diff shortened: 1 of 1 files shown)


app/dashboard/models.py
```diff
@@ -2641,8 +2641,7 @@ def get_eth_sum(self, sum_type='collected', network='mainnet', bounties=None):
 
         """
         eth_sum = 0
-
-        if not bounties:
+        if bounties is None:
```
owocki: if `bounties.exists()` was `0`, then this condition was hit and we'd lookup the `get_funded_bounties()` again.. thus hurting performance

app/dashboard/models.py
```diff
@@ -2655,9 +2654,7 @@ def get_eth_sum(self, sum_type='collected', network='mainnet', bounties=None):
 
         try:
             if bounties.exists():
-                eth_sum = bounties.aggregate(
-                    Sum('value_in_eth')
-                )['value_in_eth__sum'] / 10**18
+                eth_sum = sum([amount for amount in bounty.values_list("value_in_eth", flat=True)])
```
owocki: this is much faster.. 0.45 vs 0.02s

PR #4781, opened 2019-07-11
Fix invite to work on a bounty detail and user directory

### LINKED ISSUE #4780: Invite to work to a bounty in bounty detail
**Describe the bug**

the endpoint will fail when the bounty is dupe.
https://github.com/gitcoinco/web/blob/master/app/dashboard/views.py#L1146
query shouldn't by by issue url.
**To Reproduce**

https://gitcoin.co/issue/crytic/crytic-compile/2/3220
https://gitcoin.co/issue/crytic/crytic-compile/2/3220
are the same issue url so trying to invite someone will fail
**Expected behavior**

invite to a duplicated bounty should invite to the current one (query by bounty id) 
**Screenshots**
<!-- If applicable, add screenshots to help explain your problem.)

- OS:
- Browser:
- Browser Version:

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

octavioamu:
##### Description

Change the query to be by bounty id instead of github url, to avoid errors when the issue is duplicated 
##### Refers/Fixes

Fix #4780
##### Testing

Tested locally in bounty details + users directory


app/assets/v2/js/pages/bounty_share.js
```diff
@@ -21,7 +21,7 @@ $('.modal-link').click(function(e) {
 const sendInvites = (users) => {
   let usersId = [];
   let msg = $('#shareText').val();
-  const url = document.issueURL;
+  let bountyId = document.result.pk;
```
thelostone-mc: Doesn't look like bountyId gets updated within this ! 
should we just leave it as const ? 

Also should we be extra safe and do `const bountyId = document.result && document.result.pk` ?

PR #4792, opened 2019-07-13
F fix incorrect token price conversion.

### LINKED ISSUE #4186: Incorrect Token Price Conversion
**Describe the bug**

Teck funded $5k in BNB but it displays 2625 DAI: https://gitcoin.co/grants/87/kitsune-wallet

**To Reproduce**

Try funding any grant in BNB.

The root cause for this (and its going to keep happening) is that CryptoCompare lists the BNB => USDT conversion rate as $9 not $19.

**Expected behavior**

A similar price to the one found on Coinmarketcap should be used for BNB and other tokens. This might require introducing a price feed that has another non-USDT trading pair for BNB or, if CryptoCompare has one, using that pair instead (e.g. ETH-BNB then converting the ETH value).

Skyge:
I think maybe not all tokens have the value to USD, so I would like to use a pair like `TOKEN/ETH` and then convert that value to USD so that we get a correct market price.
Fix #4186


Skyge:
By the way, when I test to get trade pairs, I get some suggestions from the devs:  `EtherDelta is broken at the moment and I suggest you do all this at ` [forkdelta](https://forkdelta.app) `They use the same smart contract.` So is there a necessary to change this?

owocki:
@Skyge mind opening up a seperate ticket for this?  hopefully its just as easy as just changing the URLs of the requests

Skyge:
OK, I will open a new issue later, I am not sure if we just need to change the URL, so I will test later.

Skyge:
@owocki I asked developers whether I should replace the web socket address to get market pairs, one dev did not give a explicit answer, 
```
ForkDelta started as a fork of EtherDelta when EtherDelta was sold to new owners 
who introduced questionable development decisions. 
```
and if we want to change, maybe this is helpful: https://github.com/forkdelta/backend-replacement/tree/master/docs/api

PR #4808, opened 2019-07-17
adds tabs to the profile to reduce information overload & complexity; also introduces scalable way (tabs) to build profiles out further

owocki:
##### Description

adds tabs to the profile to reduce information overload & complexity

##### Refers/Fixes

self

##### Testing

# my profile

Screen Recording 2019-07-17 at 10 23 AM

# org profile

Screen Shot 2019-07-17 at 10 23 08 AM

# not found profile

Screen Shot 2019-07-17 at 10 22 58 AM


owocki:
current bounties, past bounties, who you've worked with.. could all be interesting tabs.

vs77bb:
> current bounties, past bounties, who you've worked with.. could all be interesting tabs.

@owocki Agree with all these. Testimonials, in the future, maybe? 

Re Org view, breaking it down into 'Current Bounties' / 'Past Bounties and clearing out a lot of the noise on all of the user actions (applications / start work / submit work) would make a big difference and would solve a funder need per this note from Arweave

image

danlipert:
@owocki whats up with the wonky commit history? Was this branched from master or stable? Looks like it maybe has the wrong base

owocki:
ive got some work to do withi product on definining the direction on profiles.. will re-open this if/when that needs to be done

thelostone-mc:
Closing this out cause our profile is evolved into something way cooler :D

Issue #4576, opened 2019-06-04
DESIGN - As a contributor, I want an easy way to view all info about an organization


willsputra:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a contributor, I want an easy way to view all about an organization

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
There's currently no central place where I can find out everything relating to an organization: important links, new bounties, activities, contributors, etc. 

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Current Behavior
[comment]: # (Describe what actually happened.)

1. Our current 'org profile' page shows some of these info, but could be improved
https://gitcoin.co/profile/MetaMask
2. There have been attempts to create an org page, but are currently being somewhat deprioritised
https://github.com/gitcoinco/web/issues/1798
https://github.com/gitcoinco/web/issues/1799
This ticket is a continuation of these two above.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

Proposed design draft (still a very rough draft, can be improved!)
![org view - inside copy](https://user-images.githubusercontent.com/19514207/58880537-b5247980-870a-11e9-887e-b8659edd7a6c.jpg)

Some of the features:
1. Different tabs showing different info about the org:
Bounties: the org's current bounties sorted by Status (Open Bounties should be on top)
Activity: we could use the same Activity component of the current profile
Core Team: who are the core team members of that org?
Contributors: who have contributed (completed bounties) to this org?
Jobs: job openings of that org, if hiring
Stats: some stats that might be important for funders or contributors (e.g. monthly funds paid out, how active the repo is, etc.)

2. Follow
Allow contributors to Follow an org, after which they'll be informed via email or notifications whenever the org posts new bounties.

3. Top Contributors For This Org
A mini-leaderboard that rank contributors based on the number/value of that org's bounties completed. May be useful for contributors because:
• Encourages contributors to participate in more bounties by gamification
• Top contributors will be able to signal to the org that they're worthy candidates for hiring
• Top contributors will be able to be discovered by other orgs with similar codebase/language

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
Explore design and features of org page that will be useful for both funders and contributors

### Data Requirements
• How many Follows, see if Follows = Retention
• See if having a leaderboard affects how active contributors are

### Additional Information
[...]

PR #4818, opened 2019-07-18
massively simplifies the activity tabs on profile

owocki:
##### Description

massively simplifies the activity tabs on profile to just:

- All Activity
- Bounties
- Kudos
- Grants
- Tips

are there any of the old tabs that should be kept?

before:

Screen Shot 2019-07-18 at 9 29 58 AM
Screen Shot 2019-07-18 at 9 28 55 AM

after:
Screen Shot 2019-07-18 at 9 28 48 AM

##### Refers/Fixes

self 
##### Testing

tested locally


thelostone-mc:
Hmm this is definitely cleaner but this was how it was originally done ! Product found out that splitting it up gave more usefulness information ! 

Now that we have a dashboard coming up, this LGTM but just letting you know the context on why the split was made earlier

owocki:
i asked @willsputra  on https://github.com/gitcoinco/web/issues/4576 whether this is the plan for the profile go forward or not.. if it is, that changes what we do for the existing profile..

PR #4836, opened 2019-07-21
show 'joined Gitcoin' as an item on /activity

owocki:
##### Description
show 'joined Gitcoin' as an item on /activity, and on profiles

this lets us eyeball who is joining and what they're into just by looking at /activity -- also makes the network more active

##### Refers/Fixes

self

##### Testing

tested locally


app/dashboard/models.py
```diff
@@ -2804,14 +2805,7 @@ def get_various_activities(self, network='mainnet'):
                 Q(tip__github_url__istartswith=url)
             )
 
-        all_activities = all_activities.filter(
```
owocki: this is as much for performance / simplicity as it is for the 'joined' activity

Issue #4724, opened 2019-07-01
As a Gitcoin operator, I'd like to know how the new and legacy grant experiments are going


frankchen07:
### User Story

As a Gitcoin operator, I'd like to know how the new and legacy grant experiments are going

### Why Is this Needed

Get data to figure out how the grant experiments are doing

### Description

vital information necessary:

```
donation date
from address
to address
from github handle?
to github handle?
token
token amount
token_amount_usdt
```

on legacy and new grant donations

### Current Behavior

no data in our db on how it's performing

### Expected Behavior

run analyses on aggregate grant donation numbers

### Definition of Done
- [ ] data specifications built out in postgres to the above

danlipert:
@frankchen07 When you say "grant donations" - are you talking about the optional donations (5%, 10%, etc.) that are being sent to us? Or is this additional information for the main grant subscriptions?

frankchen07:
@danlipert

> grant donations

thanks for the clarifying question. I'm talking about the optional donations sent to us for both the new grants created (first solution), and the legacy grants (second add on solution)

PR #4838, opened 2019-07-22
model setup for donation tracking

danlipert:
##### Description

This PR adds the model to track the optional donations made through new and legacy grants.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4724

##### Testing

Just the model setup, so no testing (other than the syntax is clean and the server boots)


thelostone-mc:
@danlipert still a WIP ?

danlipert:
Some notes for the follow on PRs:

**New Grants**

For new grants, the donations are made in two different ways: first, for a one-time contribution to a grant, the one-time donation is made in the frontend via the splitter contract. For this use case, we'll just need to POST the donation data when the one-time contribution is made.

For recurring subscriptions, donations are made with each individual contribution. For this use case, we'll need to add code to the subminer that records the donation data each time a subscription is mined.

**Legacy Grants**

For legacy grants, the donations made through recurring subscriptions and one-time contributions are treated the same - both are made via single transactions and will be recorded via POSTing the donation data when the one-time contribution is made.

danlipert:
> @danlipert still a WIP ?

Dang you're fast! Should be good to review now, but will hold off on adding the migration until we get closer to merging.

danlipert:
> > @danlipert still a WIP ?
> 
> Dang you're fast! Should be good to review now, but will hold off on adding the migration until we get closer to merging.

nevermind on holding off the migration, I see that it missing causes the tests to fail

Issue #4846, opened 2019-07-23
Add Email Capture to Main Gitcoin.co landing page


PixelantDesign:
### User Story
As Gitcoin core, we'd like to Add Email Capture on bounties landing page to main landing page

### What 

- [x] Build the funder/contributor email capture and add it to the landing page below the before & after graphic.
- [x] Make sure the inputs are properly hooked up to mailchimp

<img width="1644" alt="Screen Shot 2019-07-22 at 11 23 02 PM" src="https://user-images.githubusercontent.com/238235/61680490-c1bf5a00-acd7-11e9-9d05-ecf3b58d9d39.png">

PR #4847, opened 2019-07-23
feat: add email capture to landing

thelostone-mc:
##### Description

Add the email capture functionality onto the landing page
cc @octavioamu would like you to double-check this

##### Refers/Fixes

closes https://github.com/gitcoinco/web/issues/4846

##### Testing

Screenshot 2019-07-23 at 10 12 41 AM

### DIFF
--- a/app/retail/views.py
+++ b/app/retail/views.py
@@ -206,6 +206,9 @@ def index(request):
         'know_us': know_us,
         'press': press,
         'articles': articles,
+        'hide_newsletter_caption': True,
+        'hide_newsletter_consent': True,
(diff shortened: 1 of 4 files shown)


app/retail/views.py
```diff
@@ -206,6 +206,9 @@ def index(request):
         'know_us': know_us,
         'press': press,
         'articles': articles,
+        'hide_newsletter_caption': True,
```
octavioamu: what is this 2 for?

thelostone-mc: newletter use those 2 to show more label on the newsletter
If we are not using it -> we can remove it

danlipert: Was this tested? Why make these changes to the newsletter?

PR #4861, opened 2019-07-24
fix activities not loading

danlipert:
Fixes issue with activities not loading due to a bug with the "Joined Gitcoin" activity changes


app/dashboard/views.py
```diff
@@ -1970,7 +1970,7 @@ def profile(request, handle):
         if page:
             page = int(page)
             activity_type = request.GET.get('a', '')
-            all_activities = profile.get_various_activities(network)
```
owocki: thanks.  sorry, i thought i had gotten all the argument cleanups here.. hmm

Issue #4848, opened 2019-07-23
As a Gitcoin Operator, We want to explore/research opening up Issue explorer for Google


PixelantDesign:
### User Story

As a Gitcoin Operator, We want to explore/research opening up Issue explorer for Google so that we can increase traffic to our site.

Currently there is a Lot of javascript blocking the view.

### What 

- [ ] Research ways to open up the issue explorer vs Bounties network way (maybe varnish) or rendering metadata
- [ ] Provide recommendations (low, medium, high effort and tradeoffs of each path
- [ ] Review with Squad

### Why

We arent receiving traffic we deserve google cant see the 5000 bounty pages.
Impact: Extra 10,000 visitors/month. Current = 2,000 sessions/mo from SEO

danlipert:
Largely, this involves making the site's content more visible to Google - for reference, we can check the google search console. @octavioamu has already done some good research on this 🙌

octavioamu:
After some research I think the blocker is the rules inside a robots.txt in cloudfront https://s.gitcoin.co/robots.txt

octavioamu:
Updates on this task.
Now the robots.txt are correct, scripts are not being blocked.

But there is a bigger problem, googlebot uses chrome 41 to render pages, with that there is no support for ECMAScript 2015 (ES6). What that means?
Means there are things in javascript we have done that googlebot can't understand. (arrow functions, let, const ...) 
### What are the options?
1. Use babel and setup a pipeline to bundle our js in a proper way so the code will be "translated" to a way googlebot can understand (best solution)
2. Refactor any ES6 code to be supported by chrome 41 (really bad solution also a high effort)
3. Create a rule to generate a static page with bounties just for bots (not very good, because then will need to change a lot of code, big investment on efforts, need to maintain this page when do changes and also will never be realistic as the users version) 

@gitcoinco/engineers @PixelantDesign @frankchen07

PixelantDesign:
What are the trade-offs of each option? 
What is the level of effort (estimate) of each option?

octavioamu:
The trade offs are in each point, about the estimation I believe is something we need to discuss with the eng team

PixelantDesign:
No recommendations from Engineer right now from engineering....

All options are high lift

Refactor javascript
Rewrite javascript. At least 2 sprints for 1 engineer

Third is lower lift for low results

Cannot bounty.

owocki:
[...]

PR #4862, opened 2019-07-24
SEO add strict to support let const in chrome 41

octavioamu:
##### Description

googlebot use chromium 41 and they way to accept let/const is adding strict 
this is part of the needs to make the site more SEO friendly, read more here:
https://github.com/gitcoinco/web/issues/4848#issuecomment-514764592


danlipert:
@octavioamu @thelostone-mc well this is great - how does this affect our need to transpile down with babel? Also, the big concern for this change is testing... is there a way we can add a lint check to our build process to ensure there are no undeclared variables, `with`, etc.? Also, how can we ensure that going forward all our JS files have `use strict`?

PR #4879, opened 2019-07-29
add field to change font color

octavioamu:
##### Description

Add field to change font color in hackathon explorer

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CJWQWG6J2/p1564417276108500
##### Testing

with no selection 
image

image
image


app/dashboard/models.py
```diff
@@ -3268,7 +3268,8 @@ class HackathonEvent(SuperModel):
     logo_svg = models.FileField(blank=True)
     start_date = models.DateTimeField()
     end_date = models.DateTimeField()
-    background_color = models.CharField(max_length=255, null=True, blank=True, help_text='hexcode for the banner')
+    background_color = models.CharField(max_length=255, null=True, blank=True, help_text='hexcode for the banner, default to white')
```
danlipert: For the help text here, do we want to make a note about html colors also being allowed as in your example?

app/dashboard/templates/dashboard/index.html
```diff
@@ -35,17 +35,15 @@
     </div>
     {% if hackathon %}
       <link rel="stylesheet" href={% static "v2/css/hackathons/explorer.css" %}>
-      <div id="{{ hackathon.identifier }}" class="row pt-5 pb-5 text-center banner"
-        {% if hackathon.background_color %} style="background: {{ hackathon.background_color }}" {% endif %}
-      >
+      <div id="{{ hackathon.identifier }}" class="row pt-5 pb-5 text-center banner" style="background: {% firstof hackathon.background_color or 'white' %}; color: {% firstof hackathon.text_color or 'black' %};">
```
ririen: In the files above I have not seen any verification about the fact that the entered data is actually a color. How will it be verified if the entered data is not a fake string of characters that can perform a CSRF attack or break out of the default css styles?

octavioamu: But is actually only a backend field accessible by admin. Do you think is a problem?

app/dashboard/models.py
octavioamu: I don't think so, just people using hex will be fine. I don't know if the user will know the entity right names also.

app/dashboard/templates/dashboard/index.html
ririen: Think about it differently. 

Your approach is normal, but a bit dangerous. If all fields in the admin panel are created in this way, it means that if someone finds 1 vulnerability (administrative access) then he will be able to use some new vulnerabilities by interfering with the infrastructure. However, if the administrator's panel fields is also properly secured, someone who wrongly gains access will have less opportunities to abuse it.

Not securing something because access is only available to administrators may in the future cause a lot of problems. Increases the risk of escalating 1 bug to critical levels with the possibility of using other bugs.

octavioamu: Agree with your point but also this is only called in an inline style tag in html. So even being an open field and anyone having access I think will not open any hack opportunity.

ririen: Hmm, maybe you're right. I'm just sensitive to reporting such things. I am a bit of a pedant and that I deal with security, how do I see something like that I report :D 

Seeing that I composed in my head automatically a potential route of attack from past experience

Input via color field: `black; background-image('URL');`

I was able to create a CSRF vulnerability, where in inline html I was able to put a marker that downloads a given url. And then perform administrative actions from the user level by including the appropriate url in it.

Issue #4853, opened 2019-07-23
As a bounty funder of a hackathon I'd like for the entire Gitcoin contributor community to see the bounties i've funded for the hackathon.


PixelantDesign:
### User Story
As a bounty funder of a hackathon I'd like for the entire Gitcoin contributor community to see the bounties i've funded for the hackathon.

### Why
Currently, the hackathon explorer is separate from the normal hackathon explorer.
This limits contributor traffic and applications.

### Definition of Done 

- [ ] Open up hackathon explorer to regular explorer by displaying hackathon issues on the normal issue explorer.
- [ ] Allow featured bounties on the hackathon explorer to show on the normal featured bounties carousel.
- [ ] Review with squad and @alexvotofuture

PR #4894, opened 2019-07-30
Open hackathon bounties to normal explorer 

octavioamu:
##### Description

change query to show hackathon bounties into normal explorer 
##### Refers/Fixes

Ref #4853
##### Testing

image
image


danlipert:
Seems legit to me - did you try booting it up in docker just to double check?

octavioamu:
> Seems legit to me - did you try booting it up in docker just to double check?

ohh I didn't restarted docker, Just made the change and checked the explorer. Let me send a print for tests

danlipert:
> This might come back to bite us
> If you click on the hackathon bounty and then click on back to explorer link -> you'll end up on the hackathon explorer which would confuse users a bit.
> a
> Might be worth updating UI to add both Hackathon + Explorer link for hackathon bounties ?
> @octavioamu

I agree this would be nice to fix but I don't think its a blocker - we can add that later if we decide we want to

Issue #4901, opened 2019-07-30
persona is wrong


owocki:
looks like [my profile](https://gitcoin.co/_administrationdashboard/profile/3221/change/?_changelist_filters=q%3Dowocki) is set as a hunter, and that's def wrong.  i'm a stronger funder than i am a hunter

per this comment ( https://github.com/gitcoinco/web/pull/4727#issuecomment-516389160 ) we should look at why the cronjob is setting it to be wrong

owocki:
whats with this SQL?  (a) why long SQL?  (b) it says i'm a hunter and not a funder :(

```
with email_full as (
    select distinct
        a.profile_id,
        a.email,
        'user' as status
    from
        marketing_emailsubscriber a
),
actions as (
    select
        du.profile_id,
        du.created_on,
        null bounty_id,
        du.action
    from
        dashboard_useraction du

    union all

    select
        da.profile_id,
        da.created created_on,
        da.bounty_id,
        da.activity_type as action
    from
        dashboard_activity da
    order by
        profile_id,
        created_on
),
bounty_hunters_int as (
    select distinct
        du.created_on,
        du.profile_id,
        row_number() over (partition by du.profile_id order by du.created_on desc) rank,  -- last instance of contributor action
        row_number() over (partition by du.profile_id order by du.created_on asc) rank_asc
    from
        actions du
    where
        du.action in ('worker_applied', 'start_work', 'work_submitted', 'stop_work', 'work_done')
    group by
        du.profile_id,
        du.created_on
    order by
        du.profile_id
),
not_active_bounty_hunters_action as (
    select
        a.profile_id,
        'not active bounty hunter' as status
    from
        bounty_hunters_int a
    where
        a.rank = 1
        and
            a.created_on <= date(now()) - interval '3 months'
),
active_bounty_hunters_action as (
    select
        a.profile_id,
        'active bounty hunter' as status
    from
        bounty_hunters_int a
    where
        a.rank = 1
        and
            a.created_on > date(now()) - interval '3 months'
),
active_bounty_hunters_int as (
    select distinct
        date(dbf.created_on) created_on,
        dbf.profile_id,
        dbf.fulfiller_email,
        dbf.fulfiller_github_username,
        row_number() over (partition by dbf.profile_id order by date(dbf.created_on) desc) rank,  -- last instance of completed work
        row_number() over (partition by dbf.profile_id order by date(dbf.created_on) asc) rank_asc
    from
        dashboard_bountyfulfillment dbf
[...]

PR #4905, opened 2019-07-30
massively simplifies (and corrects) Personas calc logic

owocki:
##### Description

massively simplifies (and corrects) Personas calc logic

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4901

##### Testing

tested locally by running `./manage.py set_user_personas` locally


danlipert:
What are y'all thinking with this one? Looks good to me @frankchen07 will be quite a bit faster to process as well

owocki:
i'm in favor of this for simplicity reasons..  we can still use frank's query on the metabase side.

frankchen07:
just so I understand, we're using

```
is_hunter |= self.interested.exists()
is_hunter |= self.received_tips.exists()
is_hunter |= self.grant_admin.exists()

is_funder |= self.fulfilled.exists()
is_funder |= self.sent_tips.exists()
is_funder |= self.grant_contributor.exists()
```

we're using these parameters to determine the persona (so hunters would be interested in bounties / receive tips, or start a grant, versus funders who fulfill bounties, send tips, or contribute to a grant. Aren't contributors also ones fulfilling bounties and contribute to a grant? I'm less concerned with contributing to a grant and more concerned with `fulfilled` - maybe I'm getting the definition wrong. 

Otherwise, I'm always a fan of simplifying using the data model.

owocki:
> Aren't contributors also ones fulfilling bounties and contributing to a grant?

i think of those who are contributing to a grant as being funder.  but we should decide if we want to treat funders of grants as funders of bounties, or if we cant to have more of a matrix of personas (one for each product)

owocki:
57deb18b60576ce553c8f57f7129bd1220ee33e9 fixes the `fulfilled` thing, good catch @frankchen07 !

frankchen07:
dope - my last comment will be, is it possible to make a comparison between this new simplified version and the metabase version? I think as long as we have generally > 90% similarity, it'll be LGTM. Of course, there might be some lift on that but easiest would be to pull profile IDs and just make a comparison.

thelostone-mc:
@owocki ^

frankchen07:
sorry missed this LGTM

owocki:
cool;  @danlipert should i merge, or just leave it for u guys in your next release cycle?

owocki:
actually this has been merged into https://github.com/gitcoinco/web/pull/5093 ; so we'll just manage it via that pr/branch

PR #4912, opened 2019-07-31
crowdfund contribs + USD estimates for them were completely jacked

owocki:
##### Description

https://gitcoin.co/issue/ethereum/EIPs/2221/3296
this bounty (a) shows $4.91 USD as the USD amount when there is 1.189 ETH on the main bounty and 1.12 ETH on the crowdfund.  (b) not show the 1.12 ETH in crowdfund.

fix before / after:
Screen Shot 2019-07-31 at 3 50 31 AM

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1564565072006400

##### Testing

tested locally

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -1347,6 +1347,10 @@ function normalizeAmount(amount, decimals) {
   return Math.round((parseInt(amount) / Math.pow(10, decimals)) * 1000) / 1000;
 }
 
+function round(amount, decimals) {
+  return Math.round(((amount) * Math.pow(10, decimals))) / Math.pow(10, decimals);
+}
+
 function newTokenTag(amount, tokenName, tooltipInfo, isCrowdfunded) {
   const ele = document.createElement('div');
   const p = document.createElement('p');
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/shared.js
```diff
@@ -1347,6 +1347,10 @@ function normalizeAmount(amount, decimals) {
   return Math.round((parseInt(amount) / Math.pow(10, decimals)) * 1000) / 1000;
 }
 
+function round(amount, decimals) {
```
owocki: @gitcoinco/engineers 

heres what the programmer who built the crowdfund contributions functionality didnt know (and i guess didnt test):

`normalizeAmount()` != `round()` 

`normalizeAmount()` was originally meant to take a token (which is stored as an int which represents a float value * 10**decimals) and transfer from `amount` from that int into that float value.  

i guess this was a case of bad documentation and bad method naming.

danlipert: Makes sense - theres a lot we can communicate in terms of code readability just by picking descriptive method and variable names @thelostone-mc @octavioamu something to keep in mind :)

PR #4923, opened 2019-07-31
remove duplicate function

thelostone-mc:
### Description

- Fixes advanced payout 
- Function is already declared in shared.js https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/shared.js#L1350-L1352 causing advanced payout to fail


danlipert:
@owocki the change on stable to the rounding function caused advanced payout to fail - this PR fixes it

owocki:
thanks good catch

Issue #4786, opened 2019-07-12
"Edit info" button on Grant page isn't working


lrettig:
**Describe the bug**

When I click "Edit info" on my created grant page, nothing happens:

![image](https://user-images.githubusercontent.com/3316532/61139500-68655880-a497-11e9-9aec-7a9453fca8be.png)

**To Reproduce**

Log in. Click "Edit info" on grant page.

**Expected behavior**

That I can edit info.

**Screenshots**
<!-- If applicable, add screenshots to help explain your problem.)

**Desktop (please complete the following information):**

- OS: OS X
- Browser: Chrome, Brave (tried both)
- Browser Version: Chrome Version 75.0.3770.100 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.25 ETH (68.77 USD @ $275.09/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4786/3227).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/4786/3227).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $131,324.23 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [andylim99](https://gitcoin.co/profile/andylim99) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/4786/3227?mutate_worker_action=approve&worker=andylim99) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/4786/3227?mutate_worker_action=reject&worker=andylim99))_.**

Good good goodplan plan plan plan

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4786/3227).

[...]

PR #4928, opened 2019-07-31
Fix edit info on grant page

PierrickGT:
##### Description
As a grant owner, I wish to be able to edit infos of my grant.
This PR fix a bug who made it impossible to edit infos of a grant if you where not doing it from the first tab, it was due to the Quill editor not being conditioned to be loaded only on the first tab.

##### Screencast
Here is a recording of the fix
fix screencast

##### Refers/Fixes
Bounty: https://gitcoin.co/issue/gitcoinco/web/4786/3227
Issue: https://github.com/gitcoinco/web/issues/4786


thelostone-mc:
@danlipert / @octavioamu Tested it out locally to ensure it's working
Untitled

PR #4936, opened 2019-08-01
allow and check for regex in email suppression

danlipert:
##### Description

This change allows regexes to be used in the email suppression list, as well as checks for these regexes when syncing emails to and from mailchimp and our database.

##### Refers/Fixes

none

##### Testing

Tested locally


danlipert:
@owocki here's a fix that allows us to use the current suppression list, but also use regexes if we need to

PR #4930, opened 2019-08-01
Show Tooltips On Bounty Details Page

bakaoh:
##### Description

I added `inner-tooltip` class in base.css so we can reuse it in other pages.
Set `inner-tooltip` class in the p tag will help show the tooltips.

But...

Only bounties with additional funding (i.e: https://gitcoin.co/issue/gitcoinco/skunkworks/89/2631) show usd tooltips with content
Screenshot from 2019-08-01 10-05-02

Bounties without additional funding stucks here
https://github.com/gitcoinco/web/blob/master/app/assets/v2/js/pages/bounty_details.js#L236
so they show empty usd tooltips.

##### Refers/Fixes

Fix #4910

PR #4938, opened 2019-08-01
fixes tooltips on bounty detail pages

owocki:
Screen Recording 2019-08-01 at 09 13 AM

piggybacks on https://github.com/gitcoinco/web/pull/4930

### DIFF
--- a/app/assets/v2/js/pages/bounty_details.js
+++ b/app/assets/v2/js/pages/bounty_details.js
@@ -112,7 +112,10 @@ var heads = {
 var callbacks = {
   'github_url': link_ize,
   'value_in_token': function(key, val, result) {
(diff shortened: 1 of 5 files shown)


app/assets/v2/js/pages/bounty_details.js
```diff
@@ -112,7 +112,11 @@ var heads = {
 var callbacks = {
   'github_url': link_ize,
   'value_in_token': function(key, val, result) {
-    return [ 'amount', token_value_to_display(val) + ' ' + result['token_name'] ];
+    var title = token_value_to_display(val) + ' ' + result['token_name'];
+    var title_expand = title + ' in funding from original funder.';
```
thelostone-mc: @owocki I'm updating this to const just to be on the safer side + to avoid hoisting

PR #4981, opened 2019-08-07
stable -> master

thelostone-mc:
##### Description

- Exact duplicate of stable being merged into master


thelostone-mc:
Closing this as we are getting this via remote push ! and thanks for that @owocki , we'll get run that

PR #4986, opened 2019-08-07
Ambassador Check-In API extension for alpha

kuhnchris:
##### Description
Added new fields for ambassador check-in API

##### Refers/Fixes
n/a

##### Testing
on ambassador check-in tool


app/dashboard/models.py
```diff
@@ -429,6 +438,39 @@ def get_natural_value(self):
...
+    def escalated(self):
+        """Filter results by bounties that have been escalated for review."""
+        return self.prefetch_related('activities') \
+            .filter(
+                activities__activity_type='bounty_abandonment_escalation_to_mods',
+                activities__needs_review=True,
+            )
+
+
+
```
thelostone-mc: @kuhnchris extra lines need to be removed

kuhnchris: I wonder, since usually lint should hit my shins and makes travis fail whenever I do more than 2 lines? 😕 I can remove those two line tho if you want. I add them usually to indicate blocks, but that one looks just... stupid, correct. I'll fix that, sec.

PR #4991, opened 2019-08-08
cast first arg to string for sup email

danlipert:
fix login


app/marketing/utils.py
```diff
@@ -205,7 +205,7 @@ def get_or_save_email_subscriber(email, source, send_slack_invite=True, profile=
     # Prevent syncing for those who match the suppression list
     suppressions = EmailSupressionList.objects.all()
     for suppression in suppressions:
-        if re.match(suppression.email, email):
+        if re.match(str(suppression.email), email):
```
thelostone-mc: doesn't `suppression` return a string ? or are we just being extra safe ?

PR #5001, opened 2019-08-09
fix repeated contributor email.

### LINKED ISSUE #4691: If a grant gets a contribution, you get two mails
**Describe the bug**
When someone contributes to your grant you get 2 mails

**To Reproduce**
Get contributed at

**Expected behavior**
One mail

**Screenshots**
grafik

Skyge:
##### Description

When someone contributes to your grant you will get only 1 mail.

##### Refers/Fixes

Fix #4691 

##### Testing

Try to contribute a project.

### DIFF
--- a/app/assets/v2/js/abi.js
+++ b/app/assets/v2/js/abi.js
@@ -60,7 +60,7 @@ var get_etherscan_url = function(id, network, type = 'tx') {
     case 'ropsten':
       return 'https://ropsten.etherscan.io/' + type + '/' + id;
     case 'kovan':
-      return 'https://kovan.etherscan.io/'  + type + '/' + id;
+      return 'https://kovan.etherscan.io/' + type + '/' + id;
     case 'rinkeby':
       return 'https://rinkeby.etherscan.io/' + type + '/' + id;
     case 'custom network':
--- a/app/assets/v2/js/grants/fund.js
+++ b/app/assets/v2/js/grants/fund.js
@@ -409,7 +409,9 @@ const splitPayment = (account, toFirst, toSecond, valueFirst, valueSecond) => {
       'sub_new_approve_tx_id': $('#sub_new_approve_tx_id').val()
     };
     console.log('confirmed!');
-    saveSubscription(data, true);
+    if (confirmationNumber == 0) {
+      saveSubscription(data, true);
+    }
   });
 };
 


Skyge:
Besides fix this bug, I find at line 265: 
``` javascript
const linkURL = get_etherscan_url(transactionHash);
```
and line 396:
``` javascript
const linkURL = get_etherscan_url(transactionHash);
``` 
we use the function `get_etherscan_url()`, we, however, just provide one parameter, but it also can receive one parameter `network` more, so is there a necessary to provide that?

Skyge:
Because the `web3.eth.sendTransaction()` will return `PromiEvent` that means promise combined event emitter, and fired from confirmation 0 on, which is the block where its minded. so sometimes, it will give two mined block, so the code will run twice in the 
```js
web3.eth.sendTransaction().on('confirmation')
```
then, that will trigger to send email twice.

Skyge:
Ohhh, it crossed my mind that is there a necessary to wait for more confirmation number? such as some exchanges will have at least 12 blocks to confirm a transaction?

danlipert:
The reason a subscription gets saved multiple times is that we save the subscription when the person submits the contribution form, and then again when we get the web3 callback from metamask. To fix this I think we actually need to check when the email gets sent out from the backend rather than changing the frontend logic.

Skyge:
Yeah, but when I look through the code, I find only at the frontend logic will trigger the backend to send email.

danlipert:
@Skyge Yes - what I'm saying is that when that backend logic is triggered, a check should be added to see if the subscription is in the correct state to send the email

Skyge:
@danlipert  I think it is [here](https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L514):
```ts
if 'signature' in request.POST:
    ……
    new_supporter(grant, subscription)
```
So, only when the request.POST contains the key word `signature`, it will send email, maybe I should find more details about it.

Skyge:
@danlipert So how to judge the subscription is in the correct state?

Skyge:
@danlipert Any advice?

Skyge:
If backend logic is triggered, it should send an email, so for correct state, I think it should be contained in the backend logic.

danlipert:
@skyge I think the correct state is where the signature and such is in place - if you watch the database as a grant is created, it first gets created with many fields missing, and then once the signature is done it gets updated - once its in its final updated state, then we should fire the email

thelostone-mc:
Seems like this has been fixed during triage ! 
Closing this out !

Thanks @Skyge  ❤️

PR #5002, opened 2019-08-10
bounty/new: add new token model

thelostone-mc:
### Description

- refactors a good chunk of code to use Promises
- moved token auth check from shared to `new_bounty.js`
- added modal which asks funder to go to token page to enable token
- modal is triggered when funder changes token in dropdown to token which isn't enabled
- modal is also triggered if funder ignores that and attempts to fund an issue 

##### Fixes

closes https://github.com/gitcoinco/web/issues/4984

##### Testing

- Created bounty with ETH & succeeded 
- Created bounty with an enabled ERC20 token & succeeded
- Attempted to create a bounty with an ERC token which wasn't enabled and wasn't able to 

Recording [HERE](https://embed.vidyard.com/share/E9f8jsKpmoM3hLfPgB9WsM?)


app/assets/v2/js/pages/new_bounty.js
```diff
@@ -223,9 +283,7 @@ $(function() {
 
 
   setTimeout(setUsdAmount, 1000);
-  waitforWeb3(function() {
```
thelostone-mc: This was removed as it's not needed ! 
And having this here cause the occasional bug where the modal would popup before the page was ready 🤷‍♂

app/assets/v2/js/shared.js
```diff
@@ -1000,48 +1000,6 @@ window.addEventListener('load', function() {
...
-};
-
-var promptForAuth = function(event) {
-
-  var success = function(denomination, tokenAddress) {
-    $('.alert').remove();
-  };
-
-  callMethodIfTokenIsAuthed(success, promptForAuthFailure);
-};
```
thelostone-mc: moved the above into new_bounty.js and refactored to make it into 2 neat functions

PR #5007, opened 2019-08-13
Don't expose ports unless neccessary

kuhnchris:
##### Description
Since I keep getting CERT ABUSE messages regarding open Redis servers (as seen in this image) we should only expose ports that are neccessary for operation(s). The normal developer probabbly won't ever use much besides the web server anyways (at least I haven't had to check postgres, redis or IPFS in the last months I worked on the project).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,8 @@ services:
   db:
     image: postgres:10.1-alpine
     restart: unless-stopped
-    ports:
-      - "5432:5432"
+#    ports:
+#      - "5432:5432"
     volumes:
       - pgdata:/var/lib/postgresql/data
     deploy:
@@ -22,11 +22,11 @@ services:
     volumes:
       - 'ipfsexport:/export'
       - 'ipfsdata:/data/ipfs'
-    ports:
-      - "4001:4001"
-      - "4002:4002"
-      - "5001:5001"
-      - "8080:8080"
+#    ports:
+#      - "4001:4001"
+#      - "4002:4002"
+#      - "5001:5001"
+#      - "8080:8080"
     deploy:
       replicas: 1
       resources:
@@ -50,7 +50,7 @@ services:
       - .:/code
     ports:
       - "8000:8000"
-      - "3030:3030"
+ #     - "3030:3030"
     depends_on:
       - db
     stdin_open: true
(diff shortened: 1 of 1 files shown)


danlipert:
@kuhnchris can you post the image you mentioned?

kuhnchris:
I mentioned an image? I mean not sure, but I can post the mail if you'd want to.
image

octavioamu:
I believe the open ports are part of our production infra. 
I don't get the context of you getting that email, is about a copy of gitcoin you have online? Because we have a very specific production architecture so not sure if this applies. 
Can you explain more?

kuhnchris:
> I believe the open ports are part of our production infra.
> I don't get the context of you getting that email, is about a copy of gitcoin you have online? Because we have a very specific production architecture so not sure if this applies.
> Can you explain more?

Not hard to understand.
Developer runs gitcoin code on a machine connected to the internet to test stuff (syncing, mails, APIs, ...)
As soon as ```docker-compose up``` is used without adapting the docker-compose.yml and you are hosted in germany you get above mentioned mail about open redis server within a couple of days.

Easy as that.

octavioamu:
@kuhnchris actually was hard to understand is the reason why we asked for more explanation. 
Anyway I just to do all that form my localhost. 
You mean you get that email when you deployed the site in some place to test?

kuhnchris:
> @kuhnchris actually was hard to understand is the reason why we asked for more explanation.
> Anyway I just to do all that form my localhost.
> You mean you get that email when you deployed the site in some place to test?

Yes, I work on windows which is a objectivly bad base for development.
Hence I use my Linux VM server.
That one is hosted with Keywebs (a german Hoster).
Whenever I leave my docker-instance up over night I get this mail by the abuse team of keywebs (which just forwards it from CERT)

octavioamu:
Got it, the difference is we don't use docker on deploy we have diff instances running so not sure how this could affect production. Maybe not at all since each instance have his own security group. 
But anyway I don't think is something happening in production (amazon is very strict about open ports left open with vulnerabilities, they sent you an email and after some tries they shot down the instance)

kuhnchris:
again, this isn't an issue that is connected to your prod environment. This is purely a dev env problem.

thelostone-mc:
Discussed with the other engineers and we feel it makes sense to leave it in there as it doesn't do any harm!
Closing it out 
Thanks as always :D

kuhnchris:
well, just as a quick note:
Danlipert keeps pointing out that this docker image has no relation to the productive environment.
So the argument "prod needs that" is invalid.
It just means that we cannot do public test sessions because automated scans by providers may 0-route you due to abuse messages from scan services.

Either way, since I barely do any dev nowadays on the server anyways I'll let this topic rest in peace.

Issue #4744, opened 2019-07-05
Allow Ability to Resurface Bounties to the Top When Expiration Date is Extended


PixelantDesign:
### User Story

As a Gitcoin Admin, I'd like the ability for issues I've re-marketed to surface to the top of the issue explorer as a newly opened issue.

As a Contributor, I'd like to see re-marketed issues at the top of the issue explorer so I'll know that it is open and available to be worked on.

### What 
When an issue is remarketed, it should surface to the top of the issue explorer and show up as a new issue in notifiers-gitcoin on slack.

### Why
This will help us move issues through the platform.

### Definition of Done
- [ ] Build a feature that allows for surfacing to the top of the issue explorer when expiration has been extended.
- [ ] Post a video or screenshots of the feature built
- [ ] This issue passes all of the tests and revisions from the Gitcoin Core engineering team.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 125.0 DAI (125.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/4744/3210).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/4744/3210).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $137,130.09 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

PixelantDesign:
Reserved for @rafalkowalski

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [rafalkowalski](https://gitcoin.co/profile/rafalkowalski) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/4744/3210?mutate_worker_action=approve&worker=rafalkowalski) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/4744/3210?mutate_worker_action=reject&worker=rafalkowalski))_.**

I would like to work on resurfacing bounties

[...]

PR #4876, opened 2019-07-26
#4745: Surface re-marketed issues to the top of the issue explorer

vince0656:
##### Description

Changes meet the requirements detailed in issue #4745. The new code adds the ability to re-market issues that have been abandoned or had all workers rejected. When issues are re-marketed, the new code surfaces the issue to the top of the issue explorer under a new default sorting option ‘Last Marketed’. In addition, the new code will:
-Not allow a funder to market an issue more than twice
-Send a slack notification to #notif
-Re-order buttons on the bounty details page into a more convenient drop down

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4745

##### Testing

Appropriate unit tests have been included

##### Video demonstration

https://drive.google.com/open?id=1J_eGGdcepsupj6yx3KYV1WqHUmcFR2Lv

Please bear in mind that my laptop slowed down during recording which is why the first re-market took a while to kick in :)

PR #5012, opened 2019-08-13
Re-surface an issue to the top of the issues explorer when its expiration has been extended

vince0656:
##### Description

After extending the expiration of a bounty, this new code triggers the re-marketing functionality which in turn surfaces the bounty to the top of the issue explorer

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4744

##### Testing

Unit tests have been appropriately added and local development testing has been done as per the demonstration video.

##### Video demonstration

https://drive.google.com/open?id=1x8mCrCLeynqUodhrpYOLnc4mmhuMYjPq

My laptop slowed down during recording but you can see that when expiration is extended, the issue is re-marketed to the top of the issue explorer


thelostone-mc:
@vince0656 you might have to remove the first commit as that's already ! Could you double check that ?

vince0656:
@thelostone-mc not sure why the PR didn't automatically pick up the commit from the merge of #4876 into master but it has now picked it up and is no longer showing that commit

app/dashboard/utils.py
```diff
@@ -824,3 +828,57 @@ def get_nonce(network, address):
     JSONStore.objects.create(key=key, view=view, data=[new_nonce])
 
     return new_nonce
+
+
+def re_market_bounty(bounty, auto_save = True):
+    remarketed_count = bounty.remarketed_count
+    if remarketed_count < 2:
+        now = timezone.now()
+        one_hour_after_remarketing = bounty.last_remarketed + timezone.timedelta(hours=1) if bounty.last_remarketed else now
```
danlipert: Lets take the remarketing time limit and maximum count out of the function and have them get set as environment variables through settings.py - this way we can easily adjust them if we need to in the future, and we won't have any kind of "magic number" confusion: https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

vince0656: Hey @danlipert I’ve made the changes you’ve requested😊Thanks, Vincent!

vince0656:
@danlipert, this is ready to go again!

app/dashboard/tests/test_dashboard_utils.py
```diff
@@ -134,3 +138,130 @@ def test_get_ipfs_with_bad_host():
...
+        assert bounty.remarketed_count == 0
+        result = re_market_bounty(bounty, False)
+        assert result['success'] is True
+        assert bounty.remarketed_count == 1
+
+        result = re_market_bounty(bounty, False)
+        assert result['success'] is False
+
+        import re
+        msg_regex = '^As you recently remarketed this issue, you need to wait [0-9][0-9] minutes before remarketing this issue again.$'
```
danlipert: Can we make this check just match the string directly by using the environment variable? That way, if we set it to 120 minutes or something like that it will still work

vince0656: All done and ran unit tests to make sure that it works :)

vince0656:
@danlipert Good to go 👍

PR #5017, opened 2019-08-14
bug: remarket issue

thelostone-mc:
##### Description

Set the last_remarket to current-time when it's None

@vince0656 

##### Testing
- locally tested


thelostone-mc:
There is also something neater we could do along the lines of 

```
from django.utils.timezone import now
created_date = models.DateTimeField(default=now, editable=False)
```

vince0656:
I think the solution submitted in the PR is better than the second one suggested in the comment a we then only set last_remarketed when the bounty is actually being re_marketed :)

PR #5027, opened 2019-08-16
Bounty Payout Milestone 1-Review

### LINKED ISSUE #4969: Team Payout -Build - Milestone I - As a funder on Gitcoin - I'd like the ability to manage permissions and payouts with my team members.
### User Story
As a funder on Gitcion - I'd like the ability to manage permissions and payouts with my team members.

# Why
- Overwhelming number of funders requesting this feature.
- Make funder workflow easier
- We can package and sell in a subscription

# Assumptions
Human - Teams will be more than two people. 
Business - If we build this funders will improve our platform experience for funders.
Tech - Team members will not share private keys with each other. We will use modified version of standard bounties v2 (we currently use v1). Payout is what we are focusing on here. Other options like management are less important (extend, cancellation - later).

# Definitions

1. mirroring github actions + payout

2. when standard bounties contract v2 is released, when the bounty is created on the blockchain, we add a list of people onto the bounty that can interact with the bounty 

### Definition of Done

- The Gitcoin platform syncs repo data from Github related to collaborators and team members, mirroring their permissions model. TBD: what level of access will allow users to be added as "Issuers", who have full control, and "Approvers", who can accept submissions and issue payout.

- [x] process set up to sync github action data with our db
- [x] new data model in gitcoin + syncing process

androolloyd:
##### Description

New Model:

	Organization
	- name
	- groups (Many to Many relationship to auth.Group)

Model Updates:

	Profile
	- orgs (Many to Many relationship to dashboard.Organization)

sync_profile updates:

	- updated to account for change to orgs referenced above

New Management Command:

	- app/management/sync_orgs_repo.py
	- no arguments presently
	- designed to be run as a long running cron job

##### Refers/Fixes

#4969


app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,113 @@
...
+class Command(BaseCommand):
+    help = 'Synchronizes Organizations and Repo roles to members'
+
+    # def add_arguments(self, parser):
+
+    def handle(self, *args, **options):
+
+        try:
+            print("Loading Users....")
+            all_users = User.objects.all()  # potentially want to add some filters here around what users are synced
```
danlipert: What filters are you thinking?

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,113 @@
...
+from django.contrib.auth.models import User, Group
+from dashboard.models import Profile, Organization
+from git.utils import get_organization, get_user, get_repo
+from app.utils import sync_profile
+
+
+class Command(BaseCommand):
+    help = 'Synchronizes Organizations and Repo roles to members'
+
+    # def add_arguments(self, parser):
```
danlipert: Leftover stub

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,113 @@
...
+            # memoize the process so we only ever sync once per user
+
+            synced = []
+
+            def recursive_sync(lsynced, handle):
+                try:
+                    if handle not in lsynced:
+                        print(f'Syncing User Handle: {handle}')
+                        profile = sync_profile(handle)
+                        remove = [x for x in profile.orgs.all() if x.name not in []]
```
danlipert: Whats happening here? isn't this the same as `remove = profile.orgs.all()` since `x.name not in []` is always true?

app/dashboard/models.py
```diff
@@ -121,41 +121,41 @@ def needs_review(self):
         """Filter results by bounties that need reviewed."""
```
danlipert: In this file, can you just include the logic changes and ditch all the style changes? I'm guessing you have some kind of automated tool in your IDE or something like that, and while I appreciate the style fix up, different contributors have different style fixing preferences in their IDEs so we get a lot of PRs where style changes are just going back and forth which makes it hard to review.

androolloyd: Noted, will revert those changes, and just add the logic.

app/app/management/commands/sync_orgs_repos.py
androolloyd: Looks like I forgot to put my conditional back in when i was testing, should be checking to ensure that the old orgs are actually not in the profile.organizations that is synced from the profile check.

androolloyd: This is how I do reconciliation per user.

app/app/management/commands/sync_orgs_repos.py
androolloyd: inactive users, or some other key that would be used to determine what users we shoudl sync, potentially an argument

app/dashboard/migrations/0049_merge_20190823_0222.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.2.3 on 2019-08-23 02:22
+
+from django.db import migrations
```
octavioamu: Im worried about the amount of migrations, maybe we will need to recreate it ??

androolloyd: i can recreate them if we want, repo syncing to a repo model was added after the work was originally complete.

app/dashboard/models.py
```diff
@@ -2121,6 +2142,8 @@ class Profile(SuperModel):
     )
     keywords = ArrayField(models.CharField(max_length=200), blank=True, default=list)
     organizations = ArrayField(models.CharField(max_length=200), blank=True, default=list)
+    orgs = models.ManyToManyField(Organization, blank=True)
```
octavioamu: should this have a more descriptive name to avoid confusions with "organizations" ? Maybe something more descriptive as orgs_owner or orgs_perm. Because there is so many orgs organizations already on the app.

app/dashboard/migrations/0053_auto_20190920_2018.py
```diff
@@ -0,0 +1,51 @@
+# Generated by Django 2.2.4 on 2019-09-20 20:18
```
octavioamu: this one is conflicting with master, but maybe just recreate it just before merged to avoid having to recreate it many times.

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,172 @@
...
+
+                        return lsynced
+                except Exception as exc:
+                    print("here")
+                    print(exc)
+
+            for user in all_users:
+                # get profile data now creates or gets the new organization data for each user
+                try:
+                    synced = recursive_sync(synced, user.profile.handle)
```
octavioamu: Getting this on my local
image

androolloyd: noted, will add some strict checking around profile having a handle value before continuing, thats strange that you don't have a profile, anytime you sync with github it should create that.

app/dashboard/models.py
androolloyd: will do

app/git/utils.py
```diff
@@ -590,11 +591,40 @@ def get_interested_actions(github_url, username, email=''):
     return actions_by_interested_party
 
 
-def get_user(user, sub_path=''):
+def get_user(user, sub_path='', auth=_AUTH):
     """Get the github user details."""
     user = user.replace('@', '')
     url = f'https://api.github.com/users/{user}{sub_path}'
```
octavioamu: The request need to be https://api.github.com/user/orgs?access_token=XXX
otherwise is not going to give you the private orgs.
By default when you create or add an Org to you profile it gets private not public and is not even visible in `https://api.github.com/users/:username/orgs?access_token=XXX` you need to access with scope `https://api.github.com/user/orgs?access_token=XXX`

octavioamu: How I see my profile
image
How is showed 
image

So just with the scoped endpoint you will able to see it 
Similar with   with https://api.github.com/orgs/gitcoinco/members?access_token=XX Will give you everyone private and public but in that case without access_token just the publics ones (I believe the access_token need to be from an admin)

app/app/management/commands/sync_orgs_repos.py
octavioamu: Yes I had a user without, and also the superuser (I don't think this last one was the problem)  I had removed the one without profile because also attaching the one with a profile was getting some missing data.

octavioamu: If you want to test this case you can create a user manually (one with the same name of the fixtures)

app/git/utils.py
androolloyd: `get_user` works the same as it was previously, my addition lets the auth token be set in the calling function as an override.

organizations are fetched inside the `sync_profile` function and set to the profile object there.

I can make the changes to the sync step to account for fetching private organizations, and look into the oAuth scope for that.

octavioamu: yes but note that the scope calls are diff to be able to get the private info, even with token
`users/:username/orgs`  need to be `/user/orgs`

app/app/management/commands/sync_orgs_repos.py
androolloyd: i updated the query to get the user scoped endpoints if the new scope variable is set.

app/git/utils.py
androolloyd: i updated the query to check the `user` scope vs the `users` scope

app/dashboard/models.py
androolloyd: Ideally this would be `organizations = models.ManyToManyField(Organization, blank=True)` and replace the current organizations column value with a list from the Organization model.

There is a reference to `profile.organizations` in one of the popover javascript templates, yet upon further inspection it seems that the the `.organizations` variable is not set from the Profile.organizations model data that we're changing here.

The reason I did this was to revert that change, though I couldn't find any breaking changes as a result of the switch, it doesn't mean there aren't any.

I purpose profile_organizations.

app/git/utils.py
```diff
@@ -590,11 +591,45 @@ def get_interested_actions(github_url, username, email=''):
     return actions_by_interested_party
 
 
-def get_user(user, sub_path=''):
+def get_user(user, sub_path='', scope='', auth=_AUTH):
```
octavioamu: _AUTH is the GITHUB_API_TOKEN (application token) but need to be the profile.access_token and the request:
`https://api.github.com/user/orgs?access_token=XXXX` or by header like `get_github_primary_email`
```
headers = dict({'Authorization': f'token {oauth_token}'}, **JSON_HEADER)
response = requests.get('https://api.github.com/user/emails', headers=headers)
```

octavioamu: I will add you to an org so you to an org for testing "whatcomesnext" so you can reproduce 
yo need to set up it as private (so no visible on you profile for others) 
image

octavioamu:
def sync_profile(handle, user=None, hide_profile=True) is getting the profile hidden since is by default

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,187 @@
...
+                                continue
+
+                            orgs_synced.append(org)
+                            db_org = Organization.objects.get_or_create(name=org)[0]
+
+                            print(f'Syncing Organization: {db_org.name}')
+                            profile.profile_organizations.add(db_org)
+                            org_members = get_organization(
+                                db_org.name,
+                                '/members'
```
octavioamu: I think there is a missing ","  here

octavioamu:
> def sync_profile(handle, user=None, hide_profile=True) is getting the profile hidden since is by default

@androolloyd did you found a way to look around this?

app/dashboard/migrations/0054_auto_20190930_1525.py
```diff
@@ -0,0 +1,51 @@
+# Generated by Django 2.2.4 on 2019-09-30 15:25
+
+from django.db import migrations, models
```
octavioamu: seems to me there is a problem with this migration, since already exist on master but for activities

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,189 @@
...
+                        profile.profile_organizations.add(db_org)
+                        org_members = get_organization(
+                            db_org.name,
+                            '/members',
+                            (handle, access_token)
+                        )
+
+                        if 'message' in org_members:
+                            print(org_members['message'])
+                            continue
```
thelostone-mc: ^ do we need these 3 lines

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,189 @@
...
+                                db_group = Group.objects.get_or_create(name=f'{db_org.name}-role-{role}')[0]
+                                db_org.groups.add(db_group)
+                                member_profile_obj = Profile.objects.get(
+                                    handle=member['login'],
+                                    user__is_active=True
+                                )
+                                member_profile_obj.user.groups.add(db_group)
+                                members_to_sync.append(member['login'])
+                            except Exception as e:
+                                # print(f'An exception happened in the Organization Loop: handle {member["login"]} {e}')
```
thelostone-mc: remove ?

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,189 @@
...
+
+                        org_repos = get_organization(
+                            db_org.name,
+                            '/repos',
+                            (handle, access_token)
+                        )
+
+                        if 'message' in org_repos:
+                            print(org_repos['message'])
+                            continue
```
thelostone-mc: ^ remove the if block ?

app/app/management/commands/sync_orgs_repos.py
```diff
@@ -0,0 +1,189 @@
...
+                                    continue
+
+                        for x in members_to_sync:
+                            try:
+                                lsynced = lsynced + recursive_sync(lsynced, x)
+                            except Exception as e:
+                                # print(f'An exception happened in the Members sync Loop: handle: {handle} {e}')
+                                continue
+                    except Exception as e:
+                        # print(f'An exception happened in the Organization Loop: handle {handle} {e}')
```
thelostone-mc: ^ do we wanna remove the commented out statements

app/app/management/commands/sync_orgs_repos.py
danlipert: I think these are for logging errors in the cronjob log

androolloyd: correct we're logging out whats happening, if the returning object has a message key at all from at all its an error, you would want to notify and skip to the next element in the array.

app/app/management/commands/sync_orgs_repos.py
androolloyd: no, same as above, anytime we see a message key in the returning object, we want to log it and skip to the next element, the error's are vague but are only caused by a couple of things, permissions errors are the most prevelant

app/app/management/commands/sync_orgs_repos.py
androolloyd: I realized they are redundant for production but helpful for development, either they get changed in terms of their messaging or removed, they will appear a lot when syncing and getting a user that is outside of the current dataset of users inside gitcoin. 

we only synchronize profiles for users who are currently a gitcoin user.

Issue #5014, opened 2019-08-13
Open user directory with custom skill filters applied 


octavioamu:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a user I what to click on a link and get into the user directory with a custom skills filter applied.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:

 we can use that link to help funder clicking in the email link and get related contributors
### Description
link gitcoin.co/users?skills=python,css 
get users with keywords python and css. 

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:

### Current Behavior
[comment]: # (Describe what actually happened.)
no prefilter by url params

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

prefilter by url params
### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
prefilter by url params

fix skill in put, currently doesnt take more than one skill.

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

PR #5047, opened 2019-08-20
feat: open users with skill filter

thelostone-mc:
##### Description

http://localhost:8000/users?skills=html,PHP activates the filters and it works as expected

##### Refers/Fixes

closes https://github.com/gitcoinco/web/issues/5014

##### Test
https://embed.vidyard.com/share/7ksCTEBQJc6C3BEZGXF4mb?

### DIFF
--- a/app/assets/v2/js/users.js
+++ b/app/assets/v2/js/users.js
@@ -119,7 +119,6 @@ Vue.mixin({
           _alert('The invitation has been sent', 'info');
         }
       });
-
     },
     closeModal() {
       this.$refs['user-modal'].closeModal();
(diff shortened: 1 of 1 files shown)


octavioamu:
@thelostone-mc the problem was in  `vm.params.skills = skills.split(',');` as that deep object assignment is not reactive. So the solution is in that scenarios using set `vm.$set(vm.params, 'skills', skills.split(','));` and problem solved.
I made a few changes to make this feature available to all the filters so now can be set by url like `http://localhost:8000/users?skills=css,javascript&bounties_completed=0,5`

thelostone-mc:
Ahh got it from https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats! Niceee 🙌

PR #5000, opened 2019-08-09
Kevin/site perf2

thelostone-mc:
##### Description

Opening against master instead of stable 
https://github.com/gitcoinco/web/pull/4800

PR #5051, opened 2019-08-21
Tagmanager

octavioamu:
##### Description

Add tag manager and prepare prod for tagging


octavioamu:
@thelostone-mc I think we must merge first #5000

Issue #5045, opened 2019-08-20
Bulk Invite on User Directory


PixelantDesign:
### User Story
As a funder of Gitcoin I'd like to be able to invite many contributors at once from the user directory.
As an operator of Gitcoin I'd like to be able to invite many contributors at once from the user directory.

### Assumptions
Human - Funders want this feature for instances where they have no applicants or applicants drop from their bounty.
Business - We assume that implementing this will increase task success on the platform
Tech - We assume that we will look for the lightest lift way to implement to solve the above problem.

### Definition of Done
- [ ] ~~add a check or filter to mass invite contributors only~~
- [ ] add a check to prohibit mass inviting all of gitcoin
- [ ] This feature should allow funder/admins to select all contributors and do a bulk invite.
- [ ] ~~If user unchecks a certain number, then the system should assume that the user wants to select their own contributors from the list, and bulk invite will be deactivated.~~
- [ ] Please refer to the ux that is being worked on in sprint 17

danlipert:
We will probably need to make the "select all" admin only at first, and then have some sort of UX that prevents spam from happening (why wouldn't a funder select all for every bounty they create and invite everyone). Maybe we limit that as a potential paid feature, limit how many uses per month, etc.

willsputra:
If it's admin only, based on yesterday's discussion, we might not need the checkboxes. We can also build the 'Description' part on the modal at a later stage to make things simpler for now.

![bulk](https://user-images.githubusercontent.com/19514207/63878326-96750c80-c9f3-11e9-8cc2-533e9bf0f892.jpg)

PixelantDesign:
- Allow Gitcoin Admins to do use this feature.
- Do not include the description field (for mvp).

Issue #5048, opened 2019-08-20
Invite Groups of Contributors Based on Skills


PixelantDesign:
### User Story
As a Gitcoin Funder I'd like to invite groups of contributors to my bounty so that I have a lot of candidates to choose from.

### What 
Add form field to allow ability to invite groups of users based on skill required on this bounty

### Assumptions
Human - We assume that funders wnat to invite by groups/skills
Business- We assume that if we build this the number of contributor applications woould increase  maybe 3x and would satisfy funders.
Tech - We assume minimal changes too the backend.

### Definition of Done
- [ ] On the funder form add the ability to invite contributors based on skills in groups.
- [ ] Update Invite email subject line to include name of person that invited
- [ ] Update copy as outlined below
- [ ] Open the Invite section of the form by default

### Design

![image](https://user-images.githubusercontent.com/238235/63379955-6da1a700-c363-11e9-9abf-2450af19eedb.png)

![image](https://user-images.githubusercontent.com/238235/63379968-772b0f00-c363-11e9-8e52-46ff67a231ab.png)
Update the above text to:

Hi Guys!
I'm looking for some help on this new bounty! Give it a look and see if you're interested.

Thanks,
@github user

Bounty Name
Value
View Bounty

danlipert:
@thelostone-mc what do you think of this as a start for the backend https://github.com/gitcoinco/web/pull/5060

thelostone-mc:
@danlipert I think we can actually use what we already have from the backend !
So if a funder selects skills as HTML + CSS
Ideally we can hit https://gitcoin.co/api/v0.1/users_fetch/?skills=html%2Ccss and get the results and populate the Invite Contributor field!

The logic to fetch all names from that field on funding an issue is already in place so I believe we can do away without it !

cc @octavioamu

danlipert:
@thelostone-mc Thats good but then won't we need to do a ton of requests to actually send each invite? I think its better to submit the skills to the backend to fire the email rather than submit each user_id to fire each email.

PR #5060, opened 2019-08-23
Invite users to a bounty based on skills

danlipert:
##### Description

image

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5045

##### Testing

Tested locally with all filters, getting emails with invites perfectly according to the query


thelostone-mc:
left a comment over at https://github.com/gitcoinco/web/issues/5048#issuecomment-524438767

app/dashboard/views.py
```diff
@@ -1130,6 +1130,53 @@ def social_contribution_modal(request):
...
+        Http403: The exception is raised if the user is not authenticated or
+                 the args are missing.
+
+    Returns:
+        Http200: Json response with {'status': 200, 'msg': 'email_sent'}.
+
+    """
+
+    inviter = request.user if request.user.is_authenticated else None
+    skills = request.POST.get('skills')
```
thelostone-mc: might be safer  to do a `lower()` cause skills are in lower case

app/dashboard/views.py
```diff
@@ -1130,6 +1131,56 @@ def social_contribution_modal(request):
...
+    Returns:
+        Http200: Json response with {'status': 200, 'msg': 'email_sent'}.
+
+    """
+    if not request.user.is_staff:
+        return JsonResponse({'status': 401,
+                             'msg': 'Unauthorized'})
+
+    inviter = request.user if request.user.is_authenticated else None
+    skills = request.POST.get('skills')
```
octavioamu: I think this need to be not only skills, we need to cover all the others filters and check if at least there is one skill selected as a validation.
Because if not, the user will use the other filters think is sending just for some users but really will be sending for more users than that.

app/dashboard/views.py
```diff
@@ -1130,6 +1131,56 @@ def social_contribution_modal(request):
...
+    bounty_id = request.POST.get('bountyId')
```
octavioamu: we need to change this to be the bounty url, I think we have some ready function to fetch the id from that url.

octavioamu:
@danlipert  @thelostone-mc  When reviewing check the users_fetch_filters also, for some reason github is not showing diff there, but the def was introduced in this PR

app/dashboard/views.py
```diff
@@ -821,6 +794,54 @@ def users_fetch(request):
...
+    else:
+        network = 'rinkeby'
+    if current_user:
+        profile_list = Profile.objects.prefetch_related(
+                'fulfilled', 'leaderboard_ranks', 'feedbacks_got'
+            ).exclude(hide_profile=True)
+    else:
+        profile_list = Profile.objects.prefetch_related(
+                'fulfilled', 'leaderboard_ranks', 'feedbacks_got'
+            ).exclude(hide_profile=True)
```
thelostone-mc: Oh why did we move this from `def users_fetch_filters` ? to keep that more generic ?

app/assets/v2/js/users.js
```diff
@@ -120,6 +120,53 @@ Vue.mixin({
...
+
+      vm.errorIssueDetails = undefined;
+
+      if (url.indexOf('github.com/') < 0) {
+        vm.issueDetails = null;
+        vm.errorIssueDetails = 'Please paste a github issue url';
+        return;
+      }
+      vm.issueDetails = undefined;
+      let getIssue = fetchData(apiUrldetails, 'GET');
```
thelostone-mc: `const` ?

app/assets/v2/js/users.js
```diff
@@ -120,6 +120,53 @@ Vue.mixin({
...
+        } else {
+          vm.$refs['user-modal'].closeModal();
+          _alert('The invitation has been sent', 'info');
+        }
+      });
+
+    },
+    getIssueDetails: function(url) {
+      let vm = this;
+      let apiUrldetails = `/actions/api/v0.1/bounties/?github_url=${encodeURIComponent(url)}`;
```
thelostone-mc: `const`

app/dashboard/views.py
```diff
@@ -1145,6 +1164,77 @@ def social_contribution_modal(request):
     return TemplateResponse(request, 'social_contribution_modal.html', params)
 
 
+@csrf_exempt
```
danlipert: should we add the `@staff_member_required` decorator here to restrict access?

app/dashboard/views.py
```diff
@@ -1145,6 +1164,77 @@ def social_contribution_modal(request):
...
+
+    if len(profiles):
+        for profile in profiles:
+            bounty_invite = BountyInvites.objects.create(
+                status='pending'
+            )
+            bounty_invite.bounty.add(bounty)
+            bounty_invite.inviter.add(inviter)
+            bounty_invite.invitee.add(profile.user)
+            # emails.append(profile.email)
```
danlipert: leftover line

app/dashboard/views.py
```diff
@@ -1145,6 +1164,77 @@ def social_contribution_modal(request):
...
+            try:
+                msg = request.POST.get('msg', '')
+                share_bounty([profile.email], msg, inviter.profile, invite_url, False)
```
danlipert: I wonder how many emails we can send within the request-response window before the request times out? I'm guessing this will be somewhat buggy, may have to refactor to use celery soon but for now we'll see since its internal only.

app/dashboard/views.py
octavioamu: To be DRY instead of copy paste the same filters, both users filters and bulk invite user `users_fetch_filters` to filter and have the same result

app/dashboard/views.py
octavioamu: humm can be, Im restricting it using request.user.is_staff, seems to me there are plans to do this a feature for users so not sure if worst changing it.

Issue #5035, opened 2019-08-19
As a Funder, I'd like Gitcoin to tell me when I don't have contributors on my bounty, so that I can take action and invite contributors from the users directory


PixelantDesign:
### User Story

As a Funder, I'd like Gitcoin to tell me when I don't have contributors on my bounty, so that I can take action and invite contributors from the users directory

### Assumptions
Human - We assume that funders would want to be notified when there are no applicants on the email.
Business - We assume that building this would help funders be more active on their bounties, and increase the number of bounties that get submitted.
Tech - 

### Definition of Done
- [x] Build this email that triggers and sends out daily when a funder doesn't have a contributor on a bounty.
- [x] Follow content described in email mockup below
Hey we noticed you dont have any contributors on your bounty. Or You have 0 applicants on your bounty.
CTA - Go to user directory (show recommendation for the bounty on dashboard).
- [ ] Set Triggers: 0 applicants for 3 days, 7 days

![discoverability email](https://user-images.githubusercontent.com/19514207/63269777-60d07500-c2c9-11e9-87e3-0b23a22eb88b.jpg)

PixelantDesign:
@willsputra could you attach the source file please? Thanks!

PixelantDesign:
Front End Sprint 17
Chron Job Sprint 18

willsputra:
sketch file:
https://github.com/gitcoinco/creative/blob/188e4785587cbae36f47c6a1f3284724ad2bfa0d/Product/Discoverability%20Email/Discoverability%20Email.sketch

thelostone-mc:
@octavioamu I'll build out the email template and hand it off over the weekend  🙌

PR #5062, opened 2019-08-23
chore: email setup no applicant bounty

thelostone-mc:
##### Description

- [x] Build out email + txt template
- [x] write in test URL + logic for filtering URL
- [ ] [validate logic ](https://github.com/gitcoinco/web/pull/5062#discussion_r317259451) - @octavioamu 
- [x] set up cron + mgmt command to trigger mail sending - @danlipert 

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/5035

##### Testing

Video URL : https://embed.vidyard.com/share/ikLCZWvSek5HCNjTya5BZZ?

Navigate to http://localhost:8000/_administration/email/no_applicant_reminder to view mail 

image

Dan: Tested 3/7 day selection query on prod


app/retail/emails.py
```diff
@@ -322,6 +321,16 @@ def render_funder_payout_reminder(**kwargs):
     return response_html, response_txt
 
 
+def render_no_applicant_reminder(bounty):
+    params = {
+        'bounty': bounty,
+        'directory_link': '/users?skills=' + bounty.keywords.lower()
```
thelostone-mc: @octavioamu / @PixelantDesigns so if the bounty has keywords `Python, javascript, MySQL , Moloch` 
The button would redirect to all those filters being applied => and no users would be shown as we would be searching for user who has all those 4 skills (especially Moloch)

We could match this against an exhaustive list at our end but even then we would end up with 3 skillm `Python, javascript, MySQL` and we might not have users with all 3 filters

app/marketing/management/commands/no_applicants_email.py
```diff
@@ -26,7 +26,7 @@
 
 class Command(BaseCommand):
 
-    help = 'sends feedback emails to bounties that were closed in last xx hours, but only if its the first time a bounty has been accepted by taht persona'
+    help = 'sends reminder emails to funders whose bounties have 0 applications'
```
thelostone-mc: lol :P nice

PR #5071, opened 2019-08-26
Show Job Details Bug Fix

androolloyd:
When users opt out of having their job details displayed, the API did not adhere to this. 

dashboard.views
 - users_fetch
   - added a check to determine if the user has opted out
   - removing the job related keys from the output


thelostone-mc:
@androolloyd looks like the tests need to be updated too

PR #5094, opened 2019-08-29
breaks out profiles into reasonably-sane modules that can be worked on seperately

owocki:
breaks out profiles into reasonably-sane modules that can be worked on seperately

@danlipert this is one of the first big changes to go into `profile/v2` so PRing it seperately so we can reason about it here.

### DIFF
--- a/app/dashboard/templates/profiles/header_details.html
+++ b/app/dashboard/templates/profiles/header_details.html
@@ -1,0 +1,66 @@
+{% load i18n static avatar_tags add_url_schema email_obfuscator %}
+                <h1 class="profile-header__handle">
+                  {{ profile.name }}
+                  {% if verification.verified %}
+                    <button class="btn btn-sm animate-verify" data-container="body" data-toggle="popover" data-html="true" data-placement="bottom" data-trigger="hover click" data-content='
+                    <p class="h6 my-2 text-left">Gitcoin Verified <img width="18" src="{% static "v2/images/badge-verify.svg" %}"></p>
+                    <p>This contributor has completed the Gitcoin verification process, which means that the Gitcoin Core trusts their work.</p>
+                    <p>Please note - use your judgement on whether this is the right contributor for your task.</p>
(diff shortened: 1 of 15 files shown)


app/dashboard/templates/profiles/scorecard_funder.html
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+      <div class="card-body font-subheader">
+        <ul>
+          <li><b>{{ funded_bounties_count }}</b> {% trans "bounties funded" %}</li>
+          <li><b>{{ sum_eth_funded|floatformat:2 }} ETH</b> {% trans "funded" %}</li>
+          <li class="works_with_list">
+            {% if works_with_funded|length %}
+              <span class="font-body"><i class="fab fa-ethereum"></i> {% trans "Funds" %}:</span>
+              {% for profile, num_times in works_with_funded.items %}
+                {% if forloop.counter < 6 %}
```
danlipert: can use a template tag here `|slice:":5"` instead of the if check

owocki: hmm... this is code that i didnt write (i just moved location).. will take a look tho

app/dashboard/templates/profiles/header_details.html
```diff
@@ -0,0 +1,71 @@
...
+                  <i class="fa fa-briefcase mr-2" aria-hidden="true"></i>{{ profile.job_status_verbose }}
+                  {% if profile.resume %}<a href="{{ MEDIA_URL }}{{profile.resume}}" download>View Resume</a>{% endif %}
+                </div>
+              {% endif %}
+              {% if profile.data.bio %}
+                <div class="profile-header__bio my-2">
+                  {{ profile.data.bio }}
+                </div>
+              {% endif %}
+            </div>
```
thelostone-mc: ^ @owocki would you be open to setting up a linter into the editor ? 😅

owocki: i have anaconda setup. i think thats python only.  i guess `make fix` doesnt lint html?

PR #5096, opened 2019-08-29
adds resume tab to profile

owocki:
Screen Recording 2019-08-29 at 02 12 AM

### DIFF
--- a/app/dashboard/templates/profiles/tab_resume.html
+++ b/app/dashboard/templates/profiles/tab_resume.html
@@ -1,0 +1,56 @@
+{% load i18n static %}
+
+
+    {% if not hidden and not profile.is_org and show_resume_tab %}
+    <div class="container tab4" >
+      <div id="resume-section">
+        <p>
+          <a href="/onboard/profile/?steps=job"> Edit this info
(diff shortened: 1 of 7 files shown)


app/dashboard/templates/profiles/tab_resume.html
```diff
@@ -0,0 +1,56 @@
...
+            {% endfor %}
+          </ul>
+        </p>
+
+        <p>
+          Locations: 
+          <ul> 
+            {% for loc in profile.job_location %}
+              <li>
+                <a href="{{loc.url}}">{{loc.name}}</a>
```
ririen: **XSS here**. When saving the profile, it is possible to intercept the request and change the parameters of _job_location_ url to contain javascript code. I just did it with my profile. 

2 Problems:
  - There is no validation of data entered from maps when saving the profile. It is possible to enter payloads.
  - There is no validation of data retrieved from job_location. It is possible to load and run payloads.

Possible consequences include stealing data, taking over the account, and downloading private keys containing tips.

Zrzut ekranu z 2019-08-29 13-09-55

Issue #5036, opened 2019-08-19
As a Funder, I'd like to see the latest information on who has submitted or applied to work on my bounty.


PixelantDesign:
# User Story
As a Funder,  I'd like Gitcoin to clearly surface information I need to decide who I should choose for my bounty.

As a Funder, I'd like to see the latest information on who has submitted work on my bounty.

# Why
It’s too hard for funder to vet quality applicants  because I don’t know who are good people to work on my bounties.

# Assumptions

1. We are assuming that funders want to look for and select their contributors.

2. We are assuming that having the appropriate information to select contributors will keep our funders satisfied and therefore, positively affect retention through an enhanced user experience.

3. Tech - use same logic that exists today

# Definition of Done 
- [ ] Move Accept and Submit sections to the top of the issue details - only Funders and people in their org would see this on the issue details (it can still display publicly on github).
- [ ] Build in a success rate (number of bounties started / bounties completed)
- [ ] Display flyout with additional information - for performance.
- [ ] Review with squad

# Data
* ideally the data tracked should be which contributors were surfaced
* which contributors they selected to approve/accepted submission

# Additional Notes
Gitcoin Match will not be working until keywords and matching are fixed.
Related https://github.com/gitcoinco/web/issues/4888

![discoverability (1)](https://user-images.githubusercontent.com/238235/63287149-9eaca980-c287-11e9-8698-3cb2385a2359.jpg)

danlipert:
Looks good - note that the "people in their org" mention of the first item in the Definition of Done will change based on Andrew's work with syncing the Github permissions for shared manangement

PixelantDesign:
Sounds good. Just to confirm - backend work will change in a future sprint once Andrews part is complete, correct?

PixelantDesign:
https://github.com/gitcoinco/web/issues/4714

androolloyd:
While debugging the addition of Organizations, I noticed that there is a mention to the profile.organizations object, originally indicating it was apart of the profile table, it seems to be that the organizations object it references in user_popover.js is an array of interactions on the site, bounties, etc, and the concept is to showcase the logos(i think) of each org that they have worked with.

octavioamu:
@danlipert @thelostone-mc I have some technical questions about the data on that popover. Right now the data is not right because is using activity data:

![image](https://user-images.githubusercontent.com/1358093/63899311-ccee5f80-c9d2-11e9-86f4-bad5f47edb1a.png)

[...]

PR #5104, opened 2019-08-30
add simple success rate calculation

danlipert:
##### Description

Adds success rate calculation based on already existing calculation found in quarterly stats function

##### Refers/Fixes

exposes success rate for https://github.com/gitcoinco/web/issues/5036

##### Testing

Tested locally:
Screen Shot 2019-08-30 at 5 52 28 PM

### DIFF
--- a/app/dashboard/models.py
+++ b/app/dashboard/models.py
@@ -2366,6 +2366,17 @@ class Profile(SuperModel):
         return self.user.is_staff if self.user else False
 
     @property
+    def success_rate(self):
+        network = self.get_network()
+        num_completed_bounties = self.bounties.filter(
+            idx_status__in=['done'], network=network).count()
+        terminal_state_bounties = self.bounties.filter(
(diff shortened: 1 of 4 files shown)


app/dashboard/models.py
```diff
@@ -2365,6 +2365,17 @@ def is_staff(self):
         """
         return self.user.is_staff if self.user else False
 
+    @property
+    def success_rate(self):
+        network = self.get_network()
+        num_completed_bounties = self.bounties.filter(
+            idx_status__in=['done'], network=network).count()
+        terminal_state_bounties = self.bounties.filter(
+            idx_status__in=Bounty.TERMINAL_STATUSES, network=network).count()
```
octavioamu: don't know how to solve the cooperative bounty case, since I can be paid out but the bounty still open.

octavioamu: Is kind of the same situation on the dashboard, should we fetch fulfillments also?

PR #5110, opened 2019-08-31
Feature/celery 4976-In Review

### LINKED ISSUE #4976: [Celery Integration] As a Gitcoin user, I want invite-to-bounty requests to respond quickly, so I can work efficiently.
### User Story

As a Gitcoin user, I want invite-to-bounty requests to respond quickly, so I can work efficiently.
As a Gitcoin dev, I want to have a Celery task queue to use, so that asynchronous tasks can be performed outside the request-response cycle.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*: There are many places in the app where long-running processes take place inside of a request-response cycle. Also, there are many long-running processes that sync things with the blockchain and external services, tally leaderboards, etc. These processes are currently ran inside of cron, which is not very scalable.

### Description

*Type*: Feature

### Current Behavior
No task queue.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Sending bounty invite emails uses the task queue.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
Celery is set up in the app, and as the first upgrade, bounty invite emails are sent via celery tasks. A Celery task-runner (and Redis) are also booted as part of our local docker development containers so these features can run locally.

### Data Requirements
Response time for this specific endpoint goes down. In the future, response times will go down for many other endpoints, and our "cronbox" gets replaced with a more scalable queue/worker setup.

androolloyd:
##### Description

Celery Task Scheulder and Worker Setup
- dashboard/tasks.py
  - bounty_email task
    - locks on the invite_url
requirements/base.txt
  - django_celery_beat depdency added
  - celery dependency

RedisService added
Settings.py
  - celery specific env variables added

docker-compose.yml
- worker and scheduler added

##### Refers/Fixes

#4976 

##### Testing

Testing Still needed


androolloyd:
Some questions remain around sending emails, I noticed there are some administrator features around sending bounty emails, all aspects of sending bounty emails will need to be tested end 2 end.

danlipert:
@androolloyd whats left to do on this one so we can get it merged, hopefully by Wednesday?

danlipert:
@androolloyd any updates on this? Also, I'm wondering why we need the scheduler here since we aren't actually doing scheduling? Is it just so we can space out the emails over time?

androolloyd:
The scheduler could be turned off for now i suppose, it executes what gets defined in the database for periodic tasks. 

IIRC its not required for the email job to be function.

danlipert:
> The scheduler could be turned off for now i suppose, it executes what gets defined in the database for periodic tasks.
> 
> IIRC its not required for the email job to function.

cool, in that case lets remove it for now so we don't prematurely grow the codebase

androolloyd:
This is ready for deployment

androolloyd:
The workers are running on the celery server, once the app has been updated it should start queueing jobs to redis. The worker box should get switched back to master and get updated to the latest version as apart of the deploy.

octavioamu:
@androolloyd I noticed the file app/assets/v2/js/lib/secrets.min.js was changed to unminified file, probably another version. is that ok? Also shouldn't be good to put a minified version?

Issue #5116, opened 2019-09-02
GitCoinBot document is missing in GitCoin doc


think-in-universe:
The GitCoinBot link https://docs.gitcoin.co/app/gitcoinbot/README.md in https://docs.gitcoin.co/#integrating-gitcoin is 404. 

![image](https://user-images.githubusercontent.com/46699230/64086468-1809c900-cd6b-11e9-93a8-1cfa8169dc5f.png)

think-in-universe:
Fixed in https://github.com/gitcoinco/web/pull/5117

PR #5117, opened 2019-09-02
Fix GitCoin Bot document link

think-in-universe:
##### Description

Fix GitCoin Bot document link

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5116

##### Testing

preview markdown

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -148,7 +148,7 @@ This widget will help you advertise that you support Gitcoin bounties, so that y
 
 Gitcoinbot will allow you to add issues straight from github.
 
-[Check out the gitcoinbot readme to learn how](https://github.com/gitcoinco/web/blob/master/app/gitcoinbot/README.md)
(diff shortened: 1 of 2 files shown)


think-in-universe:
Considering maintenance cost, I would agree that we keep the relative path. 

However, the current link (https://docs.gitcoin.co/app/gitcoinbot/README.md) to the gitcoinbot page is broken, and I would prefer we have some way to make the link available in https://docs.gitcoin.co/#integrating-gitcoin

What's the tool that is used to generate the https://docs.gitcoin.co ? Maybe we can create a mirror/copy to the gitcoinbot page under /docs folder in this repo.

thelostone-mc:
Ah we use pydocmd 🙌 !
mind checking that out and updating it there ?

think-in-universe:
I added a page https://docs.gitcoin.co/md_gitcoinbot for GitcoinBot to fix the link

PR #5127, opened 2019-09-04
add missing events and fixes for gtm

octavioamu:
##### Description

Add and fix GTM events
together with gtm changes for tracking
Add data-gtm as auto tracking attr 
##### Testing

locally, intensive

### DIFF
--- a/app/retail/templates/settings/settings.html
+++ b/app/retail/templates/settings/settings.html
@@ -73,6 +73,7 @@
       <div class="form_container">
         <ul class="nav nav-tabs">
             {% for nav in navs %}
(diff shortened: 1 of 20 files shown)


app/retail/templates/settings/settings.html
```diff
@@ -73,6 +73,7 @@
       <div class="form_container">
         <ul class="nav nav-tabs">
             {% for nav in navs %}
+            <!-- {{nav.href}} {{active}} -->
```
danlipert: What does this change do?

octavioamu: ohh I miss that one, was a test

PR #5131, opened 2019-09-04
stable -> master

thelostone-mc:
##### Description

Putting master truly ahead of stable


app/kudos/kudos.yaml
```diff
@@ -5117,4 +5117,11 @@
   - Social
   - Friends
   - Love
-  image: 3box_butterfly.svg
```
thelostone-mc: @owocki deleted it by mistake ?

PR #5135, opened 2019-09-05
feat: refactoring popover on bounty details

thelostone-mc:
##### Description

- refactored code + updated UI to match mock
- wired up BE query 

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5036
fix https://gitcoincore.slack.com/archives/CAXQ7PT60/p1567628057033200

##### Testing

Screenshot 2019-09-06 at 8 31 50 PM

Screenshot 2019-09-06 at 5 38 51 PM


thelostone-mc:
I got this error twice  `redis.exceptions.ResponseError: BUSY Redis is busy running a script. You can only call SCRIPT KILL or SHUTDOWN NOSAVE.` after I threw in the bounty query!

@danlipert left a bunch of TODOS I need help with

octavioamu:
@thelostone-mc I think I found a way to solve the queries. 
One important think I changed the query logic to be based on the current bounty skills, not the profile skills.

app/dashboard/views.py
```diff
@@ -1788,31 +1788,61 @@ def profile_details(request, handle):
...
     if bounties :
         for bounty in bounties:
+
             _bounty = {
                 'title': bounty.title,
-                'url': bounty.get_absolute_url()
+                'url': bounty.get_absolute_url(),
+                'id': bounty.id,
+                'org': bounty.org_name,
+                'rating': [feedback.rating for feedback in bounty.feedbacks.all().distinct('bounty_id')],
```
thelostone-mc: This would show both funder + contributor rating but is more efficient! 🤔 
would it make sense to add `.filter(receiver_profile__isnull=False)` ?
Also does it need to be an array  ?

app/dashboard/views.py
```diff
@@ -1785,29 +1785,57 @@ def profile_details(request, handle):
...
 
+    if not settings.DEBUG:
+        network = 'mainnet'
+    else:
+        network = 'rinkeby'
+
+    keywords = request.GET.get('keywords', '')
+
+    bounties = Bounty.objects.current().prefetch_related('fulfillments', 'interested', 'interested__profile', 'feedbacks') \
+        .filter(interested__profile=profile, network=network,) \
```
danlipert: I will have to show you some efficient Python style sometime :) you don't need to resort to the `\` https://www.python.org/dev/peps/pep-0008/#maximum-line-length

PR #5161, opened 2019-09-10
Feature/setup clr round 3

### LINKED ISSUE #4369: Collusion Resistance: Capping Contributions by Gitcoin Grants Security Rating Tiers
### User Story

Related to #3520, as a grantee I would like to ensure the system isn't being gamed and I have a fair shot of receiving a reasonable amount of matching funding. 

### Assumption
Human - Without a collusion resistance mechanism, the probability of someone gaming the system is higher.
Business - If we did this, CLR contributions should be much more fair. 
Tech - We can find and obtain a 3rd party provider that is willing to work with a crypto company that isn't really doing KYC (phone). Lessen these types of issues but no way to eliminate.

### Description

One way to achieve this is to limit the total amount that individuals can contribute based on a 'security rating'. Similar to the idea of profile completion, a user would gain more privileges the more 'complete' their security rating. 

### Current Behavior

There is no limit to how much a given funder can donate to a given project.
People are creating multiple accounts to get more match (game the system).

### Expected Behavior

There is a limit, in tiers, defined by set security ratings. We could begin by implementing #3520 style verification as a way to create one additional tier of security. A second option could be to implement a more formal request for phone numbers or other PIA.

### Definition of Done
- [ ] Mechanism to set a maximum grant contribution limit
- [ ] Verification of user profiles according to #3520
- [ ] Phone number as a hidden option of user profiles 
- [ ] A 'bar' on each user profile that describes how high their 'security rating' is

Estimate: XL
Split between overal surfacing and calculation of security rating on something simple.
Intergrate account age.
Intergrate Platforma activity
Intergrate Phone nUmber.

### Data (requirement)

Measurement

androolloyd:
##### Description

New Management Command
- setup_clr_round_3
- creates a new group (Grants-CLR-Round-3)
- enables the add_clrmatch permission
- adds any users who have an account that was created before Feb 15, 2019

##### Refers/Fixes

#4369 

##### Testing

There is no testing associated with this ticket.


danlipert:
Sweet! Can you add a cronjob for this that runs every day? That way we don't have to re-run the command as new users come aboard

PR #5165, opened 2019-09-11
Users favorite

octavioamu:
##### Description

Following work from #4751
-add interface for favorite users
-fix be 

TODO
- fix data retrieval in user directory and fix server messages codes. 
- user.fav should be true or false I had tried here without success  `profile_json['fav'] = ['true' for i in favorites if user.id in range(0,i.obj_id)]` 

image


app/dashboard/views.py
```diff
@@ -883,6 +885,10 @@ def previous_worked():
         profile_json['work_inprogress'] = count_work_in_progress
         profile_json['verification'] = user.get_my_verified_check
         profile_json['avg_rating'] = user.get_average_star_rating
+        # profile_json['fav'] = [user.id for user in favorites]
+        # profile_json['fav'] = [i for i in favorites if user.id in [i_user.id for i_user in range(0,i.obj_id)]]
+        profile_json['fav'] = ['true' for i in favorites if user.id in range(0,i.obj_id)]
```
octavioamu: This is wrong, I wasn't able to compare the right data and add True/False

thelostone-mc:
@octavioamu what's the status on this ? we deprioritizing it ?

thelostone-mc:
Closing this out ! Keeping the branch alive in case we decide to re-prioritize

Issue #5109, opened 2019-08-31
Receiving Tip does not show as ETH Contributed or Bounty Completed


edsonayllon:
**Describe the bug**

Received tip amount from a completed bounty does not show as bounty completed or ETH rewarded on a profile. 

**To Reproduce**

1. Go to https://gitcoin.co/profile/edsonayllon

**Expected behavior**

Should show tip received from https://gitcoin.co/issue/bZxNetwork/Bounties/2/3297. Transaction found here: https://etherscan.io/tx/0x6f537357dd8d38e701d5427789899d74fcecda61da49877097fb29decf77ece0. 

**Screenshots**

<img width="180" alt="Screen Shot 2019-08-30 at 8 42 38 PM" src="https://user-images.githubusercontent.com/34966228/64057082-c264cf00-cb66-11e9-84c6-5df3a783307b.png">

**Desktop (please complete the following information):**

- OS: MacOS and Windows 10
- Browser: Brave Browser
- Browser Version: Version 0.68.132 Chromium: 76.0.3809.132 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

owocki:
note to whomever works on this; `profile.get_eth_sum` is the method

@edsonayllon you want to take a stab at this? if not ill probably bounty it

edsonayllon:
@owocki possibly. Would I receive the award that would go to the bounty?  `get_eth_sum` in `./app/dashboard/models.py` correct?

owocki:
yes you would.

yes thats right

edsonayllon:
Ok. Will start work on it

owocki:
awesome!

edsonayllon:
@owocki I did a pull request pushing a fix. https://github.com/gitcoinco/web/pull/5243

However, this issue has been resolved with the new profile released recently. I'm good to close this issue.

owocki:
kewl ill pay out

owocki:
oh wait theres no bo8unty on this ? derp..

> @edsonayllon you want to take a stab at this? if not ill probably bounty it

can tip u per this

edsonayllon:
@owocki Great. My address is `edsonayllon.eth`. Also, my Gitcoin profile is also my current Github username, @edsonayllon.

gitcoinbot:
[...]

PR #5191, opened 2019-09-13
Fix tips not showing as collected on profile

edsonayllon:
##### Description

Previously, payments received as a tip (partial reward from a bounty using advanced fulfillment options from the funder) did not display on a user's profile. This PR fixes that. Now tips show as ETH collected. 

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5109

##### Testing

Two bounties were created, funding at 0.01 ETH each. Two tips were delivered to own profile for ~50% each, summing to 0.01 ETH collected and 0.02 ETH funded.

Screen Shot 2019-09-13 at 12 42 43 AM


app/dashboard/models.py
```diff
@@ -2993,7 +2996,7 @@ def to_dict(self, activities=True, leaderboards=True, network=None, tips=True):
             works_with_org = self.get_who_works_with(work_type='org', bounties=orgs_bounties)
 
         total_funded = funded_bounties.count()
-        total_fulfilled = fulfilled_bounties.count()
+        total_fulfilled = fulfilled_bounties.count() + self.tips.count()
```
owocki: i wonder if we shouldnt update `total_funded` too while were at it...

edsonayllon: `total_funded` should already be correct. `total_funded` is communicated on the profile as the number of bounties the funder has opened and is funded. If we were to add how much they tipped per bounty, this number would be higher than the number of bounties they opened, as they can tip multiple people per bounty. However, since receiving a tip as a contributor does not count as a completed bounty by the `Bounty` object, `total_fulfilled` needed to be updated with tips count.

edsonayllon:
Note: if this PR still has a bug, such as collected ETH showing for the person who gave the tip instead of the person who received it, it would be dealing with `self.tips`. A method such as `self.get_fulfilled_bounties` would have to be created for querying the `Tips` object, replacing `self.tips` in this PR.

thelostone-mc:
@edsonayllon `unsupported operand type(s) for +: 'decimal.Decimal' and 'float'`

could you add in null checks and ensure they are of the same format ! 
we deployed this in live and it gave the following error 

commented out that piece of code for now 
could you debug it and throw in a fix ?

edsonayllon:
@thelostone-mc Yes. Sorry. On it

PR #5243, opened 2019-09-24
Fix typings for addition in eth_sum

edsonayllon:
##### Description

This provides a fix, converting tips to type float before adding to `eth_sum`.

`eth_sum` did not include tips received by default. When tips were added, they were kept in a type incompatible for addition with a non-empty `eth_sum`. No error was thrown when a profile only had tips or only had completed bounties, but threw an error for profiles with both tips and full bounty rewards. This provides a fix to that error.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5109

https://github.com/gitcoinco/web/pull/5191

##### Testing

Most testing was done with `print()` in the `eth_sum` function in `models.py` by loading a page calling `eth_sum` while docker was in the foreground. Types were logged with `print()`. `eth_sum` was filled with a float variable to test for addition compatibility.


edsonayllon:
@thelostone-mc This PR provides a fix to the previous PR. However, this issue doesn't seem to be a problem with the new profiles released. I added the PR regardless in case `eth_sum` is used in the future.

danlipert:
@edsonayllon why not use the list comprehension as before? Seems like the right way to go, just some issues with the typing

edsonayllon:
@danlipert Both do the same thing. However, having it all on one line resulted in syntax errors when messing with multiple parenthesis, which made debugging a bit frustrating. So for legibility, I expanded the logic.

edsonayllon:
@danlipert I returned it to a list comprehension, and reduced the if statement.

octavioamu:
Again failing on production 😞

edsonayllon:
@octavioamu 

Same error? That would mean `bounties[n].get_value_in_eth` is type `decimal.Decimal`, and `self.tip[n].value_in_eth` is of type `float` already.

```
eth_sum = eth_sum + sum([ float(amount.value_in_eth) for amount in self.tips ])
```

Would need to be changed to:

```
eth_sum = eth_sum + sum([ Decimal(amount.value_in_eth) for amount in self.tips ])
```

Issue #4695, opened 2019-06-26
Bounty request if nothing is found


octavioamu:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a contributor using the explorer if nothing is found the "nothing found" message should have a link to https://gitcoin.co/requests
### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
 
*Summary*:
To increment the number of interesting bounties and be more assertive
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
just a nothing found message 
### Expected Behavior
[comment]: # (Describe what you expected to happen.)
![image](https://user-images.githubusercontent.com/1358093/60196828-6acb7f80-9814-11e9-80a0-38b659d8c49a.png)
People click the button, fill the form and then an email is sent to the organisation with all the data of the request.  

![image](https://user-images.githubusercontent.com/1358093/63968525-5b67ed00-ca76-11e9-853d-7537bc606c37.png)

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- [ ] Edit empty state to add a the link as the image
- [ ] Create a new field in bounty request "org email"
- [ ] fetch the organisation issue email (ex: https://api.github.com/orgs/gitcoinco)
- [ ] Send an email to the organisation with the request data (simple email template will be provided for this task) Data to include into the email are: 
user profile, 
link to the issue, 
amount, 
comment,
link to create an issue with parameters (this already works like this with issue url and user id: https://gitcoin.co/bounty/new?invite=USERID&source=ISSUEURL )

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)
This will increment the number of funders
This will increment the number of bounties
Bounty will likely get done faster and with better quality.

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)
This feature will be used in the future to be able to request bounties from other sites like organization directory. 

### Security
- Be sure to fill the org email in the server side to prevent abuse of spammers
- Be sure to fill userid checking if the user is logged in
- Be sure the url field coming from the form is actually a valid one (github url)

rafalkowalski:
@octavioamu I can work on this issue

gitcoinbot:
[...]

PR #5250, opened 2019-09-24
Nothing found

vince0656:
##### Description

As a contributor using the explorer if nothing is found the "nothing found" message should have a link to https://gitcoin.co/requests

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4695

##### Testing

I have done a lot of dev testing including the testing done for the video demonstration. As this was simply a case of storing a new field and using it to change the 'to' address of an email, there was not much scope for a new unit test that would achieve a lot.

##### Video demo

https://drive.google.com/open?id=1-puxbFUQpHwboNNjNI5syHKIo3WWFcOA


vince0656:
My understanding is that this meets the following definition of done (from #4695): 
-Edit empty state to add a the link as the image
 -Create a new field in bounty request "org email"
 -fetch the organisation issue email (ex: https://api.github.com/orgs/gitcoinco)
 -Send an email to the organisation with the request data (simple email template will be provided for this task) Data to include into the email are:
user profile,
link to the issue,
amount,
comment,
link to create an issue with parameters (this already works like this with issue url and user id: https://gitcoin.co/bounty/new?invite=USERID&source=ISSUEURL )

As well as meeting the following security requirements:
-Be sure to fill the org email in the server side to prevent abuse of spammers
-Be sure to fill userid checking if the user is logged in
-Be sure the url field coming from the form is actually a valid one (github url)

vince0656:
Let me know if you guys spot something that isn't right!

octavioamu:
Yay! I will take a look @vince0656

octavioamu:
Testing on my local I get error: "Invalid JSON." seems there is some problem in forms.py
I had created an org here is an issue you can use to test if you need. 
 https://github.com/whatcomesnext/web/issues/1

octavioamu:
I believe the problem could be on the js this is the form data
```
{"github_url":"https://github.com/whatcomesnext/web/issues/1","eth_address":"0xeda95ed3e3436c689376889f9ed0a8f4ba23e866","amount":"30","comment":"I want to work on that","github_org_name":""}: 
```

octavioamu:
Update, maybe was cache now is sending that data but seems the problem is with the email.
Sometimes the org can choose not to display or have an email, we need to cover that. 
If there is not email from the gh Api just follow the old behaviour (send the email to vivek) 
Also if there if everything ok, send to the org email and vivek email.

app/marketing/mails.py
```diff
@@ -1090,15 +1090,19 @@ def setup_lang(to_email):
 
 
 def new_bounty_request(model):
-    to_email = 'vivek.singh@consensys.net'
+    to_email = model.github_org_email if model.github_org_email != '' else 'vivek.singh@consensys.net'
```
octavioamu: Send to the org and also vivek

app/bounty_requests/views.py
```diff
@@ -60,7 +61,16 @@ def bounty_request(request):
             return JsonResponse({'error': 'Invalid JSON.'}, status=400)
 
         model = result.save(commit=False)
+
+        gh_org_api_url = 'https://api.github.com/orgs/' + model.github_org_name
+        gh_org_api_resp = requests.get(url=gh_org_api_url).json()
+
+        gh_org_email = ''
+        if gh_org_api_resp is not None and gh_org_api_resp['email'] is not None:
```
octavioamu: If the org didn't setup an email is failing. 
If the org doesn't have an email just keep the old flow and send just to vivek

app/marketing/mails.py
```diff
@@ -1090,15 +1090,19 @@ def setup_lang(to_email):
...
         subject = _("New Bounty Request")
-        body_str = _("New Bounty Request from")
-        body = f"{body_str} {model.requested_by}: "\
-            f"{settings.BASE_URL}_administrationbounty_requests/bountyrequest/{model.pk}/change"
+        body_str = _("New Gitcoin bounty funding request from")
+        body = f"{body_str} {model.requested_by.github_url}: \n\n"\
+               f"Link to the issue: {model.github_url}. \nThe user proposes funding of {model.amount} (USD) "\
+               f"and has attached the following comment along with the funding request:\n \"{model.comment}\".\n\n "\
+               f"Should you wish to fund the aforementioned issue, you can follow this URL:\n\n "\
+               f"https://gitcoin.co/bounty/new?invite={model.requested_by}&source={model.github_url}"
```
octavioamu: Use the domain from the system so in development will be localhost:8000

octavioamu: is sending the username but need to be the user ID

vince0656: Can you clarify where the user id is? Is it in the User model pk?

vince0656: Ah yes, good point

octavioamu: Is the profile id when you get into something like http://localhost:8000/bounty/new?invite=13 will open the bounty fund form with the invite filled with the user profile id 13

octavioamu:
Just that fixes and this will be done, good work!

thelostone-mc:
Oh nice :D damn neat

octavioamu:
@vince0656 how is going?? When do you think we can get this in?

vince0656:
@octavioamu sorry I will look into this tomorrow. Hopefully won't be much longer - thanks for your patience

vince0656:
image

vince0656:
@octavioamu all good to go as per the above screenshot

app/marketing/mails.py
```diff
@@ -1107,6 +1111,18 @@ def new_bounty_request(model):
             from_name=_("No Reply from Gitcoin.co"),
             categories=['admin', 'new_bounty_request'],
         )
+
+        if model.github_org_email != '':
```
octavioamu: image
I don't know why if I update my org email (or hide it) the email always keep going to the old one (but correctly save on the table)

octavioamu: Also can we put from the server email? To avoid scam gmail message

octavioamu: oohh wait I think is just because is from my local is already the server email right?

app/marketing/mails.py
```diff
@@ -1107,6 +1111,18 @@ def new_bounty_request(model):
...
+            to_email = model.github_org_email
+            setup_lang(to_email)
+            send_mail(
```
octavioamu: image
This is what I get, the email is correctly set here but then is sent to octavioamuchastegui not oc.tavioamuchastegui

octavioamu: image
never mind is overwrite when running local.

octavioamu:
@vince0656 great work, Just commented on my comments, I was getting "wrong" email just because is overwrite by default to avoid sending emails everywhere. 
The only missing piece is apply the email header / footer you can check any of the other emails (maybe invite one)

vince0656:
@octavioamu ok so to confirm I just need to add the header and footer seen in emails like the invite? Will
You still tip as it has taken longer than expected. Thanks

octavioamu:
> @octavioamu ok so to confirm I just need to add the header and footer seen in emails like the invite? Will
> You still tip as it has taken longer than expected. Thanks

Correct, will be amazing if we can get this merged today for this week deployment.

vince0656:
@octavioamu ok give me max an hour if possible

vince0656:
@octavioamu this is taking a bit longer than expected as I'm doing this for the first time

octavioamu:
Are you having any issue? you can use emails/template.html and just put the content inside check app/retail/templates/emails/start_work_approved.html for example

vince0656:
@octavioamu yes I've managed to do that but I'm working through why my local machine is saying that django doesn't recognise the trans block in my template

octavioamu:
Commit the code and I can take a look

vince0656:
image

vince0656:
@octavioamu almost there! Just have to fix the invite link

vince0656:
image

Looks like it's working - will push this up now

octavioamu:
Cool just a small thing 
can we put the link to gitcoin profiles?  <a href="domain/profile/username>@username</a> 
and add a margin bottom at the end ?

vince0656:
image

@octavioamu how does this look? :)

vince0656:
@octavioamu code pushed

octavioamu:
Yay, excellent @vince0656 I will ask the other team members to review.

app/bounty_requests/migrations/0004_bountyrequest_github_org_email.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.7 on 2019-09-24 20:57
```
danlipert: Can you combine these migrations?

vince0656:
I know that I've followed the pattern used within the other html files for Gitcoin emails but are you guys aware that emails should really be built exclusively in HTML tables? It ensures the best compatibility across multiple email clients. However, it's quite time consuming to do it that way

danlipert:
Thanks @vince0656 - as far as tables go its probably something we should get to at some point but we haven't received much in the form of complaints

vince0656:
@danlipert fair enough. I'm quite knowledgable so if you guys ever needed assistance, you're more than welcome to get in contact

thelostone-mc:
@vince0656 mind resolving the conflicts ?

vince0656:
@thelostone-mc done😊

PR #5252, opened 2019-09-25
Add 'see prize' when lower value

### LINKED ISSUE #5238: As a hackathon participant, 0.001 ETH on all of our prizes is inaccurate and misleading
### User Story

As a hackathon participant, 0.001 ETH on all of our prizes is inaccurate and misleading

### Problem

Previously, the 0.001 ETH display on all of our hackathon prizes was because the sponsor didn't necessary have the actual funds, the incorrect currency was being used, or Gitcoin  didn't support the correct currency at the time. As we tighten up our hackathon experience, it doesn't make sense to see all the prizes being an insignificant amount of ETH, as it subtracts from the experience.

### Description

image

### Current Behavior
prizes are all 0.001 ETH when they shouldn't be

### Expected Behavior
hackathon participants should be able to see what the prize is, straight up.

### Solutions
1. longer solution: restrict sponsors to using DAI or ETH so the issues can be posted correctly
2. stop gap solution: override amount numbers without syncing from the blockchain
3. stop gap solution: administrative controls on what to sync to the blockchain

### Definition of Done
TBD

octavioamu:
##### Description

Add 'see prize' when low value
##### Refers/Fixes
Fix #5238

##### Testing

image

### DIFF
--- a/app/retail/templates/shared/result.html
+++ b/app/retail/templates/shared/result.html
@@ -66,7 +66,7 @@
         </div>
       [[/if]]
       [[if rounded_amount <= 0.001]]
-      <div class="tag token"><span>See Prize</span></div>
+        <div class="tag token"><span>See Prize</span></div>
       [[else]]
         <div class="tag token">
           <p>
--- a/app/retail/templates/shared/result.html
+++ b/app/retail/templates/shared/result.html
@@ -65,28 +65,32 @@
           </p>
         </div>
       [[/if]]
-      <div class="tag token">
-        <p>
-          [[:rounded_amount]]
-          <span>[[:token_name]]</span>
-          [[if tokens]]
-            <i class="fas fa-plus mx-1" style="font-size: 0.5rem; position: relative; top: -1px;"></i>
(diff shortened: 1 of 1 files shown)


app/retail/templates/shared/result.html
```diff
@@ -65,28 +65,32 @@
...
-        [[if tokens ]]
-          <button class="btn-expand-token" onclick="toggleExpandableBounty(event, '.[[:network]].i[[:standard_bounties_id]]')">
-            {% trans 'See All' %}   <i class="fa fa-angle-down arrow"></i>
-          </button>
-        [[/if]]
-      </div>
-      [[if value_in_usdt]]
-        <div class="tag usd">
+      [[if rounded_amount <= 0.001]]
+      <div class="tag token"><span>See Prize</span></div>
```
thelostone-mc: indent this by 1?

kuhnchris:
Uhhh hey, quick shot here, can we please also check if that's a hackathon bounty? Else this is a general "usable feature" - not sure if that is really what we want?

danlipert:
nice!

octavioamu:
@kuhnchris is general, if the value is 0.001 = there is no value and the rules of what you get should be on the description.

PR #5272, opened 2019-09-26
allows anyone to upload a kudos; no deploy + minting required

owocki:
##### Description

allows anyone to upload a kudos

##### Refers/Fixes

biz dev requests

##### Testing

tessted on http://localhost:8000/kudos/new


app/kudos/models.py
```diff
@@ -544,6 +545,58 @@ def __str__(self):
...
+    def __str__(self):
+        """Define the string representation of a conversion rate."""
+        return f"{self.name} on {self.network} on {self.created_on}; approved: {self.approved} "
+
+
+    def mint(self):
+        """Approve / mint this token."""
+        from gas.utils import recommend_min_gas_price_to_confirm_in_time
+        from kudos.management.commands.mint_all_kudos import mint_kudos
+        from kudos.utils import KudosContract
```
danlipert: I assume there is some circular reference funkiness here?

app/kudos/views.py
```diff
@@ -755,3 +757,59 @@ def receive_bulk(request, secret):
...
+        'msg': None,
+    }
+
+    if not request.user.is_authenticated:
+        login_redirect = redirect('/login/github?next=' + request.get_full_path())
+        return login_redirect
+
+    if request.POST:
+        required_fields = ['name', 'description', 'priceFinney', 'artist', 'platform', 'numClonesAllowed', 'tags', 'to_address']
+        validtion_passed = True
```
danlipert: validtion -> validation

app/kudos/views.py
```diff
@@ -755,3 +757,59 @@ def receive_bulk(request, secret):
...
+        if validtion_passed:
+            #upload to s3
+            img = request.FILES.get('photo')
+            session = boto3.Session(
+                aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+                aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+            )
+
+            s3 = session.resource('s3')
+            key = f'media/uploads/{random.randint(999999999,99999999999)}_{img.name}'
```
danlipert: What do you think about using `uuid.uuid4` from the standard library? https://docs.python.org/3/library/uuid.html

app/kudos/views.py
```diff
@@ -755,3 +757,59 @@ def receive_bulk(request, secret):
...
+                platform=request.POST['platform'],
+                numClonesAllowed=request.POST['numClonesAllowed'],
+                tags=",".split(request.POST['tags']),
+                to_address=request.POST['to_address'],
+                artwork_url=artwork_url,
+                network='mainnet',
+                approved=False,
+                metadata={
+                    'ip': get_ip(request),
+                    'email': request.POST.get('email'),
```
danlipert: if this email is used for anything sensitive we should just get it from the user's profile

app/kudos/models.py
owocki: yes let me try to remove as many as possible now

app/kudos/views.py
owocki: updated

app/kudos/views.py
owocki: great idea

app/kudos/views.py
owocki: hmmm... lets do one better + just store both :P

app/kudos/views.py
owocki: updating

Issue #5223, opened 2019-09-19
Gitcoin Grants CLR Round 3 v2 Improvements


frankchen07:
### User Story

Gitcoin Grants CLR Round 3 v2 Improvements

### Why Is this Needed

Product experiment asks from Vitalik, v2 improvements post v1 launch.

### Description

1. CLR live pop-up design update (@willsputra / @PixelantDesign stated that using a navy colored background would be better)

<img width="610" alt="Screen Shot 2019-09-19 at 11 25 35 AM" src="https://user-images.githubusercontent.com/7516920/65276890-9a211e00-dadd-11e9-9667-01a5a92d8b0a.png">

Is it possible to make it pop up only when there's a number entered in amount field?

2. How are we calculating non-pairs? Currently it makes sense that if a grant only has one contribution, it would be paired by itself, resulting in a CLR contribution that is either equal to the contribution itself or the threshold. (@thelostone-mc to check). This will also fix the 0 DAI matching for those grants that don't have a matching CLR pair yet.

3. Add phantom funding to the CLR estimate (@thelostone-mc & @danlipert)

4. Ensure that dropped txns make its way into the formula. As of 2019-09-24, contribution IDs (1207, 1208, 1209, 1210, 1211, 1212, 1213) were dropped and replaced due to gas issues

### Current Behavior

v1 grants clr

### Expected Behavior

add v2 improvements

### Definition of Done
- [x] update design
- [x] ensure grant non-pairs calculation makes sense
- [x] add phantom funding to the CLR estimate

thelostone-mc:
<img width="310" alt="Screenshot 2019-09-26 at 8 54 59 PM" src="https://user-images.githubusercontent.com/5358146/65702016-ec46de00-e09f-11e9-818d-d540349c392b.png">

made it a lil cleaner like in the landing page

PR #5277, opened 2019-09-27
feat: account for single contribution for grant in clr

thelostone-mc:
##### Description

Pairs the contributor with itself and ensures grants get matching clr even if it has one contributor

If grant has only 1 contributor. 
Assume with `contributor_profile : 1` who contributed `5 DAI` :

- while generating pair we assume another pseudo contribution who makes `1 DAI` 
  (I'm assuming this is the lowest possible contribution )
- pair generated would : `1` & `_1` with contributions `5 DAI` and `1 DAI`  

If a new profile contributes,
the pesudo contributor_profile's (`_1`) contibution (`1 DAI`) is ignored and algorithm works as usual

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/5223

##### Testing

https://share.vidyard.com/watch/rTqnXvuiRuswAnbvK91GbR?

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -64,6 +64,10 @@ def generate_grant_pair(grant):
             else:
                 unique_contributions[profile] = amount
 
+    if len(unique_contributions) == 1:
+        profile = next(iter(unique_contributions))
+        unique_contributions['_' + profile] = 1
+
     profile_pairs = list(combinations(unique_contributions.keys(), 2))
     contribution_pairs = list(combinations(unique_contributions.values(), 2))
 


app/grants/clr.py
```diff
@@ -64,6 +64,10 @@ def generate_grant_pair(grant):
             else:
                 unique_contributions[profile] = amount
 
+    if len(unique_contributions) == 1:
+        profile = next(iter(unique_contributions))
```
danlipert: nice! some mystical python here :)

PR #5286, opened 2019-10-01
Security fixes

ririen:
##### Description

This is a PR eliminating the vulnerability reported in the bug bounty program. The name of the report submitted to the address: `engineering@gitcoin.co` is: `Gitcoin-CWE-79-1`.

In addition, some functionalities related to cleaning the fetched data have been improved. 

ping @danlipert 

##### Refers/Fixes

- XSS in activities.
- corrected `sanitizeAPIResults` function. 
_[Although this function existed and is assigned to many downloaded api, it did not work properly due to the wrong construction and allowed the download of dangerous strings (it did not clean it).]_

##### Testing
- After making these changes, crafted issues no longer trigger XSS in activities.
- Strings downloaded from api are more secure.


danlipert:
@ririen thanks for this - looks like Travis is failing though, can you check out whats up?

ririen:
@danlipert now everything should be okay. There was a problem with tabs instead of spaces and import orders. I added a commit called "normalization".

app/assets/v2/js/shared.js
```diff
@@ -159,10 +159,16 @@ var sanitizeDict = function(d, keyToIgnore) {
 };
 
 var sanitizeAPIResults = function(results, keyToIgnore) {
-  for (var i = 0; i < results.length; i++) {
-    results[i] = sanitizeDict(results[i], keyToIgnore);
+  if (results.length >= 1) {
+    for (var i = 0; i < results.length; i++) {
```
thelostone-mc: could we use `let` ?

app/assets/v2/js/shared.js
```diff
@@ -634,7 +642,7 @@ var retrieveIssueDetails = function() {
     result = sanitizeAPIResults(result);
     if (result['keywords']) {
       var keywords = result['keywords'];
-
+
```
thelostone-mc: could we remove this ? ^_^

app/assets/v2/js/shared.js
ririen: I think "let" is a good solution :D

app/assets/v2/js/shared.js
ririen: If I remove this eslint throw error: 

`644:7  error  Expected blank line after variable declarations  newline-after-var`

so this empty line in this case is required.

app/assets/v2/js/shared.js
```diff
@@ -159,10 +159,16 @@ var sanitizeDict = function(d, keyToIgnore) {
...
-    results[i] = sanitizeDict(results[i], keyToIgnore);
+  if (results.length >= 1) {
+    for (let i = 0; i < results.length; i++) {
+      results[i] = sanitizeDict(results[i], keyToIgnore);
+    }
+    return results;
   }
-  return results;
+
+  results = [results];
```
danlipert: whats up with the list wrapping here?

danlipert:
@ririen Looks pretty good - did you test creating non-exploit bounties as well? I'm wondering if bleach will affect any of the github flavored markdown people like to use

app/assets/v2/js/shared.js
ririen: For proper operation in all cases and backward compatibility, it was necessary to make an array/object also for individual data entries (e.g. issue title from github).

Then there is a reference to `sanitizeDict` function which checks if the input is an object and only if it is an object removes unwanted characters from them. Someone has previously created an oversight, and therefore downloaded issues titles from GitHub or other data that were not in the array were not cleaned and skipped and could be abused.

`sanitizeDict` creates a loop that cleans every element from the array one by one, therefore it was necessary to replace elements without arrays (i.e. single ones) into arrays to keep the function and backward compatibility and at the same time make single data also be cleaned.

ririen:
@danlipert my PR does not affect or operate on any places where markdown is used. Conditions and cleaning have been added to the elements which by default should not contain any specific characters including html and markdown.

* The email address is hidden if it is not an email address and someone manages to insert unwanted characters there.
* The bounty / issue title is cleared, which should not have markdown or html in the default.
* Keywords are cleared if they do not resemble keywords.
* Activity titles are cleaned if they are not ordinary strings.
* Meta description / title tags which should also contain only pure texts are cleared of unnecessary characters.

app/assets/v2/js/shared.js
danlipert: So this code is to support the case where the `results` argument passed in is not an array. But on line 169 you wrap `resutls` into an array, but then call `santizeDict()` on the first element in the array-wrapped `results` object, which is the same as what it was before it was wrapped, right? I guess I'm just asking whats the difference between the code in this PR lines 169-171 and something like:

`return sanitizeDict(results, keyToIgnore);`

danlipert:
> @danlipert my PR does not affect or operate on any places where markdown is used. Conditions and cleaning have been added to the elements which by default should not contain any specific characters including html and markdown.
> 
> * The email address is hidden if it is not an email address and someone manages to insert unwanted characters there.
> * The bounty / issue title is cleared, which should not have markdown or html in the default.
> * Keywords are cleared if they do not resemble keywords.
> * Activity titles are cleaned if they are not ordinary strings.
> * Meta description / title tags which should also contain only pure texts are cleared of unnecessary characters.

Sounds good 👍 - just to be clear, you did test this with non-exploit bounties in your local?

app/assets/v2/js/shared.js
ririen: Jeez, you're absolutely right! I didn't even notice that I had too much of it and used a few phrases when you can do all the same in one line you mentioned xD Corrected in the latest commit :D

ririen:
> > @danlipert my PR does not affect or operate on any places where markdown is used. Conditions and cleaning have been added to the elements which by default should not contain any specific characters including html and markdown.
> > 
> > * The email address is hidden if it is not an email address and someone manages to insert unwanted characters there.
> > * The bounty / issue title is cleared, which should not have markdown or html in the default.
> > * Keywords are cleared if they do not resemble keywords.
> > * Activity titles are cleaned if they are not ordinary strings.
> > * Meta description / title tags which should also contain only pure texts are cleared of unnecessary characters.
> 
> Sounds good +1 - just to be clear, you did test this with non-exploit bounties in your local?

Like I have checked it before, but for peace of mind and safety, I decided to thoroughly test it again.

It turned out that my small add-on, which is not required but only stylized, affected the markdown so I removed it in the latest commit. Markdown works fine :)

PR #5294, opened 2019-10-02
WIP - no desc specificity removal

owocki:
no desc specificity removal

still need to test - do not merge


thelostone-mc:
^ we'll add this in after testing if needed 🙌

Issue #5292, opened 2019-10-02
chore: hide kudos badges on mobile


thelostone-mc:
#### Description 

Navigate to gitcoin.co/profile/thelostone-mc on mobile view
You'll notice the kudos badges are overflowing.

Ensure they are hidden on mobile so that the profile seems cleaner

<img width="382" alt="Screenshot 2019-10-02 at 8 52 07 PM" src="https://user-images.githubusercontent.com/5358146/66057641-b057c100-e556-11e9-851e-27196c0a4220.png">

PR #5295, opened 2019-10-02
Hides kudos badges on mobile #5292

aewens:
##### Description

Hiding the kudos badges on mobile at the point where overflowing occurs.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5292

### DIFF
--- a/app/dashboard/templates/profiles/profile.html
+++ b/app/dashboard/templates/profiles/profile.html
@@ -47,7 +47,7 @@
       <div class="container profile-card mt-md-n5 mh-150">
         <div class="row">
           {% if total_kudos_count %}
-            <div id=kudos_header>
+            <div id="kudos_header" class="d-md-block d-none">
               {% for kudos_group in my_kudos %}
                 <img src="{{ kudos_group.kudos_token_cloned_from.preview_img_url }}" title="{{ kudos_group.kudos_token_cloned_from.name|humanize_name }}" class="img-thumbnail border-transparent kd-shadow" width="70">
               {% endfor %}


aewens:
@octavioamu That works, too. My latest commits employ using the classes on the element as opposed to a new CSS rule.

app/dashboard/templates/profiles/profile.html
```diff
@@ -47,7 +47,7 @@
       <div class="container profile-card mt-md-n5 mh-150">
         <div class="row">
           {% if total_kudos_count %}
-            <div id=kudos_header>
+            <div id="kudos_header" class="d-md-block d-none">
```
owocki: per https://getbootstrap.com/docs/4.3/utilities/display/  shouldn t this be `.d-sm-none .d-md-block` to just hide on mobile?

owocki: better yet why dont we just hide 2 of the kudos so it displays/aligns on mobile?

octavioamu: actually no, because bootstrap is mobile first. this will showed on min like this:
image
Smaller than that will be hidden 
image

octavioamu: @owocki or maybe just :
image

aewens: Just to make sure I am on the same page here, are you now wanting a redesign of how kudos badges are displayed rather than just hiding them?

octavioamu: Let attach to the original ticket, unless @owocki have any suggestion. 
Just cleanup the changes not related to the ticket on the commit and I believe we are good.

aewens:
@octavioamu 
> Just cleanup the changes not related to the ticket

I assume you are referring to the last line of the other two files? If so, I'll try to remove those, though I am not 100% sure why they show up in the first place since I did not touch those lines and it does not look like anything was actually changed.

app/grants/clr.py
```diff
@@ -67,6 +67,10 @@ def generate_grant_pair(grant):
             else:
                 unique_contributions[profile] = amount
 
+    if len(unique_contributions) == 1:
+        profile = next(iter(unique_contributions))
+        unique_contributions['_' + profile] = 1
+
```
octavioamu: 🤔

aewens: I tried cleaning up using a rebase and used the wrong commit. I'm working on it and should have it fixed here in a moment.

aewens:
Well, I now know way more about rebasing in git than I did before this all started, but I now have it all cleaned up.

aewens:
Does anything else need to be done before it is merged?

thelostone-mc:
Haha thanks @aewens ! Nah we are good on this ^_^
If we have iterations -> we'll open up a new ticket

PR #5306, opened 2019-10-08
popover success rate >= 100% fix

kuhnchris:
##### Description
as reported by fronk

##### Refers/Fixes
No issue reported

##### Testing
Please test.


thelostone-mc:
Hmm I'm on two minds about this ! 
This does fix up the issue but our logic itself might be broken! It was all good until we got the gitcoin profile updated 🤔 

@danlipert what do you think ?

danlipert:
Yeah I'd rather figure out what the actual issue is instead of just hiding it - also @kuhnchris the testing portion of the PR form is to describe what the author has done in terms of testing

owocki:
the root issue is that my profile v2 PR treated success_rate as an int; whereas the code that dan put together it was a float.

PR #5354, opened 2019-10-22
opens grants creation to anyone

owocki:
##### Description

opens up gitcoin grants to anyone...  reasons:
1. reduce the sense of elitism (and complaints) i get
2. make these projects shill their gitcoin grant; enabling a long tail of virality

instead any new grant will be hidden by default, and us admins will get an email asking us to approve the grant.

when this is merged @ceresstation we should email everyone on the typeform and let them know they can create a grant.

##### Refers/Fixes

convo that @ceresstation and i had on slack..

##### Testing

tested locally

### DIFF
--- a/app/grants/models.py
+++ b/app/grants/models.py
@@ -280,13 +280,14 @@ class Grant(SuperModel):
             from grants.abi import abi_v1
             return abi_v1
 
-
     @property
     def url(self):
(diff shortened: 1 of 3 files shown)


thelostone-mc:
Oh so we are doing this I see :D *opens the gates*

PR #5376, opened 2019-10-24
Proposed fix for Kudos marketplace search bar

ososco:
##### Description

When you add whitespace in search bar value, for example: "Limited edition", the query parameters will be "Limited+edition", so when results come back in the view, the search bar get its value from query parameters (queryParam.q), so it will also show up in the search bar as Limited+edition. The simplest solution is to replace the plus signs with whitespace:\

$('#kudos-search').val(queryParam.q.replace(/\+/g, " "));

I think it does look better.

##### Testing

simplescreenrecorder-2019-10-28_10 27 32

### DIFF
--- a/app/assets/v2/js/pages/kudos-search.js
+++ b/app/assets/v2/js/pages/kudos-search.js
@@ -5,7 +5,7 @@ $(document).ready(function() {
   var clearBtn = $('.clear-search');
 
   if (queryParam && queryParam.q) {
-    $('#kudos-search').val(queryParam.q);
+    $('#kudos-search').val(queryParam.q.replace(/\+/g, " "));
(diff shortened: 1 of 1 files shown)


danlipert:
@ososco nice! Thanks for the PR! In the PR description, theres a section to add how you tested your changes - here a screenshot or video would be good, or just a brief description on how you verified the changes are working. Thanks again!

ososco:
@danlipert 

simplescreenrecorder-2019-10-28_10 27 32

ososco:
Actually there is a better way to fix this with decodeURIComponent, i will commit the new fix after testing it.

octavioamu:
@ososco I changed the base branch to be 'master'

Issue #5387, opened 2019-10-25
XSS protection causes tags in issue explorer to be inproperly escaped


kuhnchris:
**Describe the bug**
C#, C++ and anything else with "special characters" is getting stripped.

**To Reproduce**
dashboard->bounties->load more bounties->redirection to explorer

**Expected behavior**
C# and C++ being honored

**Screenshots**
![grafik](https://user-images.githubusercontent.com/761911/67588489-88730c00-f756-11e9-8d4c-a095b35f34e3.png)

![grafik](https://user-images.githubusercontent.com/761911/67588483-8446ee80-f756-11e9-86c0-97ce3905060d.png)

danlipert:
@frankchen07 we should get this in the next sprint

PR #5388, opened 2019-10-25
bug: update scrub logic

thelostone-mc:
##### Description

update scrub logic to show languages 

##### Refers/Fixes

fixes https://github.com/gitcoinco/web/issues/5387

##### Testing

Screenshot 2019-10-25 at 5 59 59 PM

### DIFF
--- a/app/assets/v2/js/pages/dashboard.js
+++ b/app/assets/v2/js/pages/dashboard.js
@@ -108,7 +108,7 @@ function debounce(func, wait, immediate) {
   };
 }
 
-const scrub = value => value.replace(/[\W]+/g, '');
+const scrub = value => value.replace(/[!@#$%^&*(),.?":{}|<>]+/g, '');
 
 /**
  * Fetches all filters options from the URI


app/assets/v2/js/pages/dashboard.js
```diff
@@ -108,7 +108,7 @@ function debounce(func, wait, immediate) {
   };
 }
 
-const scrub = value => value.replace(/[\W]+/g, '');
+const scrub = value => value.replace(/[!@#$%^&*(),.?":{}|<>]+/g, '');
```
danlipert: I'm wondering if this opens us up at all to security risks and what the minimum characters we should allow is - I'm thinking `+` and `#` - any other weirdness out there y'all can think of?

thelostone-mc: nope ! just those hence added those in
@octavioamu  ?

octavioamu: lgtm!

PR #5394, opened 2019-10-26
Stored XSS - [ In Review ]

### LINKED ISSUE #5245: Stored XSS
Placeholder for stored XSS issue

androolloyd:
##### Description

	- Adding a check to the `invalid_file_response` helper method to ensure that <script> tags aren't defined inside the content of the uploaded file buffer for uploaded user content.

##### Refers/Fixes

#5245 

##### Testing

Tested through onboarding, and through /settings/job

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -2279,6 +2279,7 @@ def profile_job_opportunity(request, handle):
 
 def invalid_file_response(uploaded_file, supported):
     response = None
+    forbidden_content = ['<script>']
     if not uploaded_file:
         response = {
             'status': 400,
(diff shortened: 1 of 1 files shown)


androolloyd:
a solution that works with file chunks would work better here instead of loading the entire buffer into memory at one time, I tried a few different approaches with mmap, save having to write the files to the hard disk before loading I haven't been able to solve it using that approach(chunks + mmap)

thelostone-mc:
Oh this is sweet :D

androolloyd:
I have updated the check to include a forbidden content array that can be populated/managed to ensure that content that would violate terms of service or otherwise cause harm can be scrubbed and denied.

Issue #5322, opened 2019-10-16
As a hackathon sponsor, I don't have a place to see live, ongoing projects, which makes me think the hackathon is not as active.


frankchen07:
### User Story

As a hackathon sponsor, I don't have a place to see live, ongoing projects, which makes me think the hackathon is not as active.

### Why Is this Needed

While sponsors can catch up on their issues via Github, having a central location where they can see the ideas and workplans of their own projects and other projects brings forth a feeling of energy about the hackathon.

### Description

Currently, the release is: https://hackathons.gitcoin.co/grow-ethereum/results/, in order to show the status of submitted projects.

Are there other ideas to increase the liveliness of hackathons?

### Current Behavior

The template exists, but we want to ensure that the results page is updated with projects that have been submitted, and sponsors can easily find out and access the results page.

### Expected Behavior

The results page of every hackathon is the central location that sponsors go to see the results of each hackathon, for their issues and other issues.

### Definition of Done
TBD

### Data Requirements
- [ ] regular GA analytics on the results page

### Additional Information

owocki:
idea mentioned on the call: embed gitcoin.co/actvity (but only hackathon activity) in the site

alexvotofuture:
This also affects onboarding for new hackers. When they join a hackathon they often ask how to get involved in a team. Having projects all visible in one place, with contact info to get in touch with the team lead can help people jump in and get involved more quickly, which I hope will help lead to more submissions and participation from registrants.

alexvotofuture:
In addition to helping sponsors see all projects underway this view additionally provides a way for hackers to find each other to form teams and collaborate during the hackathon.

The view can also showcase final submissions, as the "results page" currently does, and can feature winning submissions.

Some related tickets:
- #5038 (start/submit work form, with additional info on a project, ideally editable)
- #5321 (sponsors have had issues paying out because they don't know who is on a project's team, and there's no way to formally add team members when submitting so they will be paid out as well. The shared payout issue will be solved by Bounties v2 but teammates who work/submit together should also be reflected on project pages in this view. Question for this - who decides who is on the team and gets a cut of payout? The submitter?)

Sponsors should be able to filter by their org to see all project underway for their prizes. This will help them get information at a glance, rather than view submissions within each prize individually, and will provide more info than the activity feed within a prize bounty's page currently provides.

frankchen07:
[...]

PR #5414, opened 2019-10-29
Add hackathon projects

octavioamu:
##### Description

- Projects board (w/ filters, pagination (9 x page)  winners and admin managing)
- Add project (for users only available on start work and work submission)
- Edit project (everyone on the project team can edit in start work, work submission or project board)
- Handle multiple projects for same bounty (you can edit your current project or add a new one)
- Flow prompt when starting work and when finish fulfilment (after submission is minted)
- Handle image preview and square verification. 
- You can't remove yourself from a project

Also:
- Update select2 version
- Fix a problem with select2 where `required` select wasn't working/ 

Project card
image

Project Modal add/edition
image

Multiple Projects handling 
image

##### Refers/Fixes

Ref: #5322 #5420

##### Testing

Tested locally with a lot of diff situations, and multiple bounties, hackathon and projects
https://share.vidyard.com/watch/h5d5nv916eRiAeYK5ErnHL?


app/dashboard/views.py
```diff
@@ -3442,7 +3442,137 @@ def hackathon_onboard(request, hackathon=''):
...
+
+    return TemplateResponse(request, 'dashboard/hackathon/projects.html', params)
+
+
+@csrf_exempt
+def hackathon_get_project(request, bounty_id, project_id=None):
+
+    profile = request.user.profile if request.user.is_authenticated and hasattr(request.user, 'profile') else None
+    try:
+        bounty = Bounty.objects.get(id=bounty_id)
```
octavioamu: @danlipert @thelostone-mc is this a problem here with the "current" bounty? 
If the bounty is duplicated I will lost the reference to the current one? 
Im adding current() to the query but not sure this will solve that (if there is something to solve)

octavioamu: answering my self: Yes is a problem, changing the query.

danlipert:
looks great but the vidyard link doesn't work :( @octavioamu

octavioamu:
> looks great but the vidyard link doesn't work :( @octavioamu

fixed, is the same video I posted in the squad channel https://share.vidyard.com/watch/h5d5nv916eRiAeYK5ErnHL?

app/assets/v2/js/hackathon-projects.js
```diff
@@ -0,0 +1,99 @@
...
+  let reader = new FileReader();
+
+  reader.onloadend = function() {
+    let imageURL = reader.result;
+
+    getImageSize(imageURL, function(imageWidth, imageHeight) {
+      if (imageWidth !== imageHeight) {
+        elem.val('');
+        preview.src = elem.data('imgplaceholder');
+        return alert('The image format need to be square');
```
danlipert: Can we have the alert say "Please use a square image"

app/dashboard/templates/dashboard/index.html
```diff
@@ -53,6 +53,11 @@
           {% endif %}
           <span class="d-block font-title-lg">
             {{ hackathon.start_date|date:"M j, Y" }} - {{ hackathon.end_date|date:"M j, Y" }}
+            {% if hackathon.end_date|timeuntil >= "1 min" %}
```
danlipert: Does the comparison here actually work? I don't think something like "8 hours" >= "1 min" works with string comparisons, but if im wrong I will be amazed at Django's usefulness :) If it turns out we need better comparison logic, according to the Django docs `timeuntil` will return the string "0 minutes" if the end date here is already passed so we can just check `{% if hackathon.end_date|timeuntil != "0 minutes" %}` https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#timeuntil

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
     }
-    return TemplateResponse(request, 'dashboard/hackathon_onboard.html', params)
+    return TemplateResponse(request, 'dashboard/hackathon/onboard.html', params)
+
+
+def hackathon_projects(request, hackathon=''):
+    q = request.GET.get('q', '')
+    order_by = request.GET.get('order_by', '-created_on')
+    filters = request.GET.get('filters', '')
+    print(q)
```
danlipert: leftover print here

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
+
+    page = request.GET.get('page', 1)
+    try:
+        hackathon_event = HackathonEvent.objects.filter(slug__iexact=hackathon).latest('id')
+        profile = request.user.profile if request.user.is_authenticated and hasattr(request.user, 'profile') else None
```
danlipert: If the user is logged out, will this fail? I think `request.user` will be `None` so we may get an error like `Nonetype has no attribute "is_authenticated"` or something like that

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
+        'filters': filters,
+        'query': q.split
+    }
+
+    return TemplateResponse(request, 'dashboard/hackathon/projects.html', params)
+
+
+@csrf_exempt
+def hackathon_get_project(request, bounty_id, project_id=None):
+    # TODO FIX cache problem
```
danlipert: I think this TODO is solved, right?

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
+
+    profile = request.user.profile if request.user.is_authenticated and hasattr(request.user, 'profile') else None
+    try:
+        bounty = Bounty.objects.current().get(id=bounty_id)
+        projects = HackathonProject.objects.filter(bounty__standard_bounties_id=bounty.standard_bounties_id, profiles__id=profile.id).nocache()
+        print(projects)
```
danlipert: extra print

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
+
+
+@csrf_exempt
+def hackathon_get_project(request, bounty_id, project_id=None):
+    # TODO FIX cache problem
+
+    profile = request.user.profile if request.user.is_authenticated and hasattr(request.user, 'profile') else None
+    try:
+        bounty = Bounty.objects.current().get(id=bounty_id)
+        projects = HackathonProject.objects.filter(bounty__standard_bounties_id=bounty.standard_bounties_id, profiles__id=profile.id).nocache()
```
danlipert: For the filter query here, could we just do `HackathonProject.objects.filter(bounty=bounty)`?

app/dashboard/views.py
```diff
@@ -3442,7 +3443,145 @@ def hackathon_onboard(request, hackathon=''):
...
+
+    bounty_obj = Bounty.objects.current().get(pk=bounty_id)
+
+    kwargs = {
+        'name': request.POST.get('name'),
+        'hackathon': bounty_obj.event,
+        'logo': request.FILES.get('logo'),
+        'bounty': bounty_obj,
+        'summary': request.POST.get('summary'),
+        'work_url': request.POST.get('work_url')
```
danlipert: Is there a danger someone may put something malicious in this args? We probably need to sanitize them

app/dashboard/views.py
octavioamu: image
nop, is not failing if unlogged

octavioamu: because `is_registered = HackathonRegistration.objects.filter(registrant=profile, hackathon=hackathon_event) if profile else None`  so if profile is  `None` also will be `is_registered`

app/assets/v2/js/hackathon-projects.js
```diff
@@ -0,0 +1,99 @@
+// document.result.pk
+const projectModal = (bountyId, projectId) => {
+  $('#modalProject').bootstrapModal('hide');
+
+  let modalUrl = `/modal/new_project/${bountyId}/`;
+
+  if (projectId) {
+    modalUrl = `/modal/new_project/${bountyId}/${projectId}/`;
+  }
```
thelostone-mc: `const modalUrl = projectId ? /modal/new_project/${bountyId}/${projectId}/ : /modal/new_project/${bountyId}/` ?

app/dashboard/views.py
thelostone-mc: why not just force login ? `@login_required`

app/dashboard/views.py
```diff
@@ -3442,7 +3443,141 @@ def hackathon_onboard(request, hackathon=''):
...
+
+    params = {
+        'active': 'hackathon_onboard',
+        'title': 'Hackathon Projects',
+        'hackathon': hackathon_event,
+        'is_registered': is_registered,
+        'projects': projects_paginated,
+        'order_by': order_by,
+        'filters': filters,
+        'query': q.split
```
thelostone-mc: do we wanna do a check if `q` exists ?

app/dashboard/views.py
octavioamu: because is not login_required, actually I will remove this part is not used anymore.

app/dashboard/views.py
octavioamu: is already request.GET.get('q', '') and I use on the view if exist to show the terms is being query

app/dashboard/templates/dashboard/index.html
danlipert: thats crazy! magic django!

app/dashboard/views.py
danlipert: ah I see, makes sense!

PR #5416, opened 2019-10-29
quiet common sentry error when syncing orgs w/o tokens

danlipert:
##### Description

This handles a common sentry error when attempting to sync an org where we don't have a valid github token for that org's profile

##### Refers/Fixes

Triage issue

##### Testing

Untested

### DIFF
--- a/app/app/utils.py
+++ b/app/app/utils.py
@@ -242,7 +242,8 @@ def sync_profile(handle, user=None, hide_profile=True):
 
         profile.keywords = keywords
         profile.save()
-
+    except UserSocialAuth.DoesNotExist:
+        pass
     except Exception as e:
         logger.error(e)
         return None


danlipert:
cherry-picked onto stable

owocki:
if this only happens on sync mgmt comands, why does sentry report it on the URL `/complete/{backend}/`

owocki:
my volley at fixing this: 6d027f67d8aa417403595fb31ce1cff3d569aab3

app/app/utils.py
```diff
@@ -242,7 +242,8 @@ def sync_profile(handle, user=None, hide_profile=True):
 
         profile.keywords = keywords
         profile.save()
-
```
owocki: @danlipert shouldnt we add a try catch around
```
profile.organizations =
```
??

if we dont, then none of the code between that and profile.save() will be called, and the profile will not sync!

danlipert: Its already wrapped in a try catch, right? Just no good exception handling. As far as profile.save() is concerned, if getting the orgs fails, then I think the fallout as it stands is that the keywords won't save, but everything else should be saved correctly - assuming we pass on the exception.

danlipert:
> if this only happens on sync mgmt comands, why does sentry report it on the URL `/complete/{backend}/`

A recent change we made in the last deploy is that we now sync manually on login to prevent cases where a new user logs in and tries to take an action that requires their profile to be synced - in particular hackathon registration.

PR #5421, opened 2019-10-30
bypass github perms error on sync profile

danlipert:
##### Description

This PR fixes the case when a user attempting to login or have their profile synced, wherein they did not give us the required perms to access user orgs, would fail. Here we catch this expected error and allow the sync to complete.

##### Refers/Fixes

https://sentry.io/organizations/gitcoin/issues/1009676218/?project=1398424&referrer=slack

##### Testing

Tested locally


danlipert:
Alternatively we can use @owocki 's stable commit here: https://github.com/gitcoinco/web/commit/6d027f67d8aa417403595fb31ce1cff3d569aab3

danlipert:
This was actually fixed by one of @owocki 's stable fixes

PR #5427, opened 2019-10-30
Gitcoin Quests is a pageview generator. Monetize it with Ethical Adds to support OSS>

owocki:
Gitcoin Quests is a pageview generator. Monetize it with Ethical Adds to support OSS>

I, Kevin Owocki, hereby declare that if monetization of Gitcoin Quests is successful, then I will rev share 50/50 with Quest Creators the monies made from their quest each quarter.  successful = to the point in which we owe more than $20 to any one creator..

Demo:
Screen Shot 2019-10-30 at 1 05 20 PM
Screen Shot 2019-10-30 at 1 05 07 PM

Open Questions:
- How does this affect the UX of the Questee?
- How does this affect the UX of the Quest Creator?
- Ottherwise, comments?


frankchen07:
for the in-game ad, I think it depends on frequency.

for the ad on the gitcoin.co/quests page, the ad might be better after the quests and quest boxes, and right before the leaderboard title. It feels better "phrased" there, but honestly I don't think it makes a huge difference either way.

owocki:
@gitcoinco/engineers i opneed this  against the wrong branch. whoops; leaving it open so tohers can comment.  to comment on the code.. this is the sole commit https://github.com/gitcoinco/web/pull/5427/commits/89505f328e9dc91f9af05fc1e2da192a33b84b6d

owocki:
maybe if u win enough quest u can disable the ads?

frankchen07:
^ interesting thought there, we could try that

ososco:
In-game ads before the quest started sounds good, the ad in the second screenshot for example is not distracting, generally this might encourage quest creators to share quests online more.

calumtomeny:
I like the idea and I think, as you mentioned @owocki, that it would be good if you get given the option for disabling adds if you've funded or completed a bounty. Maybe if the advert was shown after the quest there would be less chance someone would skip the quest increasing the drop off.

DistributedDoge:
Having a quiz monetized by ads reminds me of https://beta.freerice.com/ ~ I think it could be pretty cool if you could make an "endless quest" (like that proposed hangman-game mechanic) monetized by code-fund advert next to each question that donates proceeds to UNICEF, or other charitable cause.

thelostone-mc:
@owocki opened up a PR porting over commit  + resolving conflicts over at #5567

Issue #5317, opened 2019-10-14
As a bounty funder - web3 not recognized, cannot pay bounties


rachelhamlin:
As a user, I want Gitcoin to recognize MetaMask so that I can pay out my bounties. 

It appears that Gitcoin is not recognizing web3/MetaMask. I've tried with MetaMask on two desktop browsers and in Status. On all 3 browsers tested, I receive the `Web3 locked. Please unlock Metamask` error. 

**Expected behavior**
Gitcoin recognizes web3 & allows me to fund and pay out bounties.

**Screenshots**

Brave
![image](https://user-images.githubusercontent.com/15470435/66740945-43083200-ee6c-11e9-9caa-ade2367b4b4f.png)

Status
<img src="https://user-images.githubusercontent.com/15470435/66741040-819dec80-ee6c-11e9-9e9d-6a94dfb0478e.PNG" width=250/>

**Desktop (please complete the following information):**

- OS: macOS High Sierra 10.13.6
- Browser: Brave 
- Browser Version: 0.69.135

- Browser: Opera
- Browser Version: 64.0.3417.54

**Smartphone (please complete the following information):**

- Device: iPhone XS Max
- OS: iOS 13.1.2
- Browser: Status
- Version: v1.0 develop build

**Additional context**
I first noticed this last week, 9 or 10/10/19.

kuhnchris:
Could you do me a favor and try it in Chrome?
Opera has some own wallet implementation, not sure if that may interfere, same goes for Brave.

StatusSceptre:
@kuhnchris I think that's exactly what it is. I'm now on Chrome using our Gitcoin gh account, and it works. I'd just set up a wallet in Brave when this issue first occurred.

owocki:
WIP for this : https://github.com/gitcoinco/web/issues/4209

StatusSceptre:
Re Status - our desktop browser is only chat right now @owocki. Our mobile betas should allow you to test, so if you want/need to [download](https://status.im/get/) for iOS let me know and I can ensure you're our testflight list.

owocki:
this should correct the issue across all non metamask web3 browsers https://github.com/gitcoinco/web/pull/5432

owocki:
my PR was merged and deployed today..  You will now no longer see complaints that 'metamask is unlocked' as you use Gitcoin in opera :)

Hat tip to the engineers for helping me get this across the finish line! 

[...]

PR #5432, opened 2019-10-30
MVP Opera support

owocki:
##### Description

MVP of Opera ssupport - tested on Opera + Chrome

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5317#issuecomment-547836980

##### Testing

Tested on Opera + Chrome

If approahc makes  sense I will do one last batch of testing and merge


owocki:
this one is for you @yorkerhodes3 !

app/assets/v2/js/shared.js
```diff
@@ -705,10 +705,12 @@ var currentNetwork = function(network) {
           $('#current-network').text(gettext('Metamask Locked'));
           $('#navbar-network-banner').html(info);
         } else {
-          info = gettext('Metamask not connected. ') +
-            '<button id="metamask_connect" onclick="approve_metamask()">Click here to connect to metamask</button>';
-          $('#current-network').text(gettext('Metamask Not Connected'));
-          $('#navbar-network-banner').html(info);
+          if (window.ethereum._metamask) {
```
thelostone-mc: can we make this into an `else if` in `L107` ?

app/assets/v2/js/shared.js
```diff
@@ -746,10 +748,12 @@ var currentNetwork = function(network) {
...
```
thelostone-mc: can we make this into an `else if` in `L749`?

thelostone-mc:
approach looks alright to me !  If we can get one more ronud of testing in + address those comments -> we can get this in this week ❤️

app/assets/v2/js/shared.js
owocki: cf5de1e0c91ed3befd9c8967589360048b7cba4c

app/assets/v2/js/shared.js
owocki: cf5de1e0c91ed3befd9c8967589360048b7cba4c

owocki:
just regression tested kudos send/ grants creation/tip sending/bounty creation on metamask and all went fine.. can i have a review @danlipert @thelostone-mc @octavioamu

octavioamu:
Im trying to send  a tip but I got stuck here, Is something more I need to do? 
image
image
when I played with this I remember the problem in that was 
```
web3.eth.sendTransaction({
            to: destinationAccount,
            value: amountInDenom,
```
since `from:` is missing

octavioamu:
Yes I added the from and now is working I will commit @owocki 

image

owocki:
thanks @octavioamu - can u push it back to my branch??

octavioamu:
Just tested kudos and tips working perfectly. Going to test a little more, excited with this

owocki:
@octavioamu thanks for f631fb5a91daf80a3c74f832ce71bc866ac664fb ; im concerned that there might be other places in the codebase where we implicitly expect `from:` to exist .. thereby peoples first opera experiences with gitcoin may fail.. 

since i wasnt able to reproduce the issue.. im not sure how to test for that tho.. what opera version r u on?

octavioamu:
Im with the last, tested with the desktop browser. Yes I think we can add the other places in a  progressive  way.

PR #5444, opened 2019-11-04
shows quests on ajax instead of right away

owocki:
##### Description

shows quests on ajax instead of right away

this is for sleep reasons..loading the quests is the most performance intensive part of the quests LP load

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5436

##### Testing

tested locally


app/quests/views.py
```diff
@@ -220,51 +220,66 @@ def index(request):
...
+
+        print(f" phase1.3 at {round(time.time(),2)} ")
+        # new quests!
+        new_quests = Quest.objects.filter(visible=True, created_on__gt=(timezone.now() - timezone.timedelta(hours=hours_new))).order_by('-ui_data__success_pct')
+        if new_quests.exists():
+            quests.append(('New', get_package_helper(new_quests, request)))
+
+        print(f" phase1.4 at {round(time.time(),2)} ")
+        # popular quests
+        popular = Quest.objects.filter(visible=True).order_by('-ui_data__attempts_count')[0:5]
```
danlipert: Since this queryset is being sliced, I think we can skip the `exists()` check. Actually, if there are less than 5 quests, I think this statement will fail with an index error

app/quests/views.py
```diff
@@ -220,51 +220,66 @@ def index(request):
...
```
danlipert: same concerns about slicing here

app/quests/views.py
danlipert: Hmm I figured slicing would force queryset evaluation but apparently not for simple slicing, nevermind!

PR #5454, opened 2019-11-06
makes the profile urls.py entry way more  restrictive

owocki:
##### Description

makes the profile urls.py entry way more  restrictive

##### Refers/Fixes

see octavio msg https://gitcoincore.slack.com/archives/CC24APWGN/p1572989388030500

and likely other profile handle URL things we haven't figured out already!

##### Testing

tested against http://localhost:8000/media/avatars/72f4f6d365732c50762730b1bc630379/octavioamu.png
http://localhost:8000/owocki
http://localhost:8000/owocki/resume

### DIFF
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -610,9 +610,9 @@ if settings.DEBUG:
     ]
 
 urlpatterns += [
-    re_path(r'^([a-z|A-Z|0-9|\.])+/([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
+    re_path(r'^([a-z|A-Z|0-9|\.]+)/([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
 
-    re_path(r'^([a-z|A-Z|0-9|\.])+/?$', dashboard.views.profile, name='profile_min'),
+    re_path(r'^([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
 ]
 
 handler403 = 'retail.views.handler403'
--- a/app/app/urls.py
+++ b/app/app/urls.py
@@ -610,8 +610,9 @@ if settings.DEBUG:
     ]
 
 urlpatterns += [
-    re_path(r'^(.*)/(.*)?', dashboard.views.profile, name='profile_min_by_tab'),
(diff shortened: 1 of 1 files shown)


app/app/urls.py
```diff
@@ -610,8 +610,9 @@
     ]
 
 urlpatterns += [
-    re_path(r'^(.*)/(.*)?', dashboard.views.profile, name='profile_min_by_tab'),
-    re_path(r'^(.*)', dashboard.views.profile, name='profile_min'),
+    re_path(r'^([a-z|A-Z|0-9|\.]+)/([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
+
+    re_path(r'^([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
```
thelostone-mc: still open to dropping the short url to profile :P
just saying

thelostone-mc:
@owocki mind if I switch this to master ? We are deploying in less than 24 hours

app/app/urls.py
```diff
@@ -610,8 +610,9 @@
     ]
 
 urlpatterns += [
-    re_path(r'^(.*)/(.*)?', dashboard.views.profile, name='profile_min_by_tab'),
-    re_path(r'^(.*)', dashboard.views.profile, name='profile_min'),
+    re_path(r'^([a-z|A-Z|0-9|\.]+)/([a-z|A-Z|0-9|\.]+)/?$', dashboard.views.profile, name='profile_min'),
```
danlipert: the name should be `profile_min_by_tab` here right?

octavioamu: tested in my local working fine with /username /username/area profile/username and profile/username/area

owocki: @danlipert yes

note to self : figure out what non alphanumeric chars we want to support so that usernames like http://localhost:8000/virtual-face work.. maybe just anything thats not a forward slash?

danlipert: https://github.com/shinnn/github-username-regex `/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/i`

octavioamu: true slash profiles will fail here

octavioamu: Have a fix commiting here

octavioamu: https://regex101.com/r/wSi6ou/1

PR #5478, opened 2019-11-11
if not funding_org fill with org name

octavioamu:


thelostone-mc:
@octavioamu Is this still needed ?

Issue #5425, opened 2019-10-30
As a Gitcoin grant contributor, I want better, more accurate live calculations, because it makes me feel like I'm contributing something worthwhile


frankchen07:
### User Story

As a Gitcoin grant contributor, I want better, more accurate live calculations, because it makes me feel like I'm contributing something worthwhile.

### Why Is this Needed

We want to improve on post v1 live calculations implementation.

### Description

Better leverage curves calculation with the new Vitalik formulas and feedback.

### Current Behavior

current work: https://github.com/thelostone-mc/clr/blob/master/clr.py
current post: https://vitalik.ca/general/2019/10/24/gitcoin.html

### Expected Behavior

faster, more accurate live CLR measures

### Definition of Done
- [ ] faster leverage curves (better calculation)
- [ ] fix any small bugs and data inconsistencies

### Data Requirements
- [ ] ensure the calculation matches what we have in metabase

PR #5494, opened 2019-11-13
chore: update clr formula

thelostone-mc:
Based on the changes made by @frankchen07 after discussing with vitalik

@frankchen07 I haven't commented out the one off contribution -> let me know if that is something you would still want to remove. Cause in that case if a grant has no contributors -> it'll appear in the UI that contributing any amount would have a matching 0 as pair could not be formed

ref: https://github.com/thelostone-mc/clr/commit/ef292e4088a907811eda2aff36a524bb0a409cfe

fixes: https://github.com/gitcoinco/web/issues/5425

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -66,9 +66,10 @@ def generate_grant_pair(grant):
             else:
                 unique_contributions[profile] = amount
 
-    if len(unique_contributions) == 1:
-        profile = next(iter(unique_contributions))
-        unique_contributions['_' + profile] = 1
+    # NOTE: comment out single pair contribution
+    # if len(unique_contributions) == 1:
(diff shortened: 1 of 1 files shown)


frankchen07:
@thelostone-mc - I think for the sake of making an accurate QV statement, if a grant only has 1 donator, the CLR match will be 0, since that's how we're performing pairwise matching. 

The downside risk here is the "potential match" is marketing as well, just tough to say how effective it was compared to blasting it out on twitter and getting Vitalik's help.

I say we trend towards being true and accurate to QV in the front as we are in the behind the scenes calculations, and if we want to later add something saying "get a friend to donate too and you'll get a match!" we can salvage the marketing situation

thelostone-mc:
@frankchen07 done ^_^

PR #5500, opened 2019-11-15
Check and fix for bounties, tips and kudos

zoek1:
##### Description

Test of gitcoin bounties, tips, kudos and grants. The elements were tested on rinkeby network. Some problems occured in opera, that can produce a conflict for example:
1. some times the etherProvider or web3 are not set, need to close and open the browser
2. Sometimes comunication between the wallet and browser never conclude to update the transaction listener
3. Sometimes the transaction never arribe to the wallet to being accepted.

Scope:
1. kudos send, receive, airdrop: Patched and working
2. tip send, tip receive: Patched and working
3. bounty new, increase, payout, accept: Patched and working
4. grant fund, grant create, grant cancel: Pending

##### Refers/Fixes

#5466


app/retail/templates/shared/footer_scripts.html
```diff
@@ -16,6 +16,8 @@
 
 {% endcomment %}
 {% load static %}
+<script type="text/javascript" src="https://cdn.jsdelivr.net/gh/ethereum/web3.js@0.20.6/dist/web3.min.js"></script>
```
danlipert: Can you move this into the repo itself? We already have `0.20.7` in the repo itself at `app/assets/v2/js/lib` if that will work for this patch.

octavioamu: shared/shared_scripts.html is included in this file at the bottom, why adding it here again?

zoek1: Yep, you're right i forgot remove this line after finish some tests.

zoek1:
@octavioamu yes, i removed this line, but it's right back. This was when i making tests on grant.  Also one note, when i test on opera developer edition the `from` parameter was not required, but when i test on stable opera requires it :thinking:. Also i pushed a fix for cancel bounty related to this.

thelostone-mc:
@octavioamu given that you fixed up the opera stuff ! What are your thoughts on getting this in ?
It's got checks which is anyways a plus IMO

PR #5501, opened 2019-11-15
WIP - site wide search

owocki:
my theory is that we can index all the kudos/bounties/users/grants/pages on the site + this will be where people can easily find anything they’re looking for

Screen Recording 2019-11-15 at 06 33 AM


app/dashboard/management/commands/create_search_results
```diff
@@ -0,0 +1,39 @@
...
+from grants.models import Grant
+from kudos.models import Token
+
+
+class Command(BaseCommand):
+
+    help = 'creates earnings records for deploy of https://github.com/gitcoinco/web/pull/5093'
+
+    def handle(self, *args, **options):
+        for obj in Bounty.objects.all():
```
octavioamu: Should we filter out the hidden/non visible objects here?

PR #5540, opened 2019-11-25
fix bounty request email getter when expecting None

danlipert:
Fixes Sentry issue: https://sentry.io/organizations/gitcoin/issues/1255090196/?project=1398424&query=is%3Aunresolved+requests&statsPeriod=14d&utc=true


octavioamu:
Just tested with an org with email and without email and is working fine. Also when this was written I tested a lot I don't think this is the problem

octavioamu:
Tested with https://github.com/whatcomesnext/web/issues/1
first enabled email > received the email
then I removed the org email 
Request was sent > don't received the email (that is right) 
No error 500

PR #5287, opened 2019-10-01
Add support for Fortmatic

fvictorio:
This is a followup to #5163. I opened a new PR because I did a lot of cleanup and it seemed better to start over.i

I tested the fund, submit work and (basic) payout scenarios, both with Fortmatic and MetaMask, and it seems to work. I will leave some inline comments explaining some of the changes to make reviewing this easier.

Some comments:
- I had to include the web3 library because, if Metamask is not installed, there's no `window.Web3` available.
- The Fortmatic API key has to be included as an environment variable. I'm not familiar with Django, so please let me know if the approach I took is not the way to do it.
- Since this is not a SPA, and since you need to know between page changes if the user has connected Fortmatic, a `fortmatic` variable is set in local storage. Plus, this variable has the address of the Fortmatic wallet, since this simplifies things a lot.

PR #5163, opened 2019-09-10
Add support for Fortmatic

fvictorio:
### Description

Add option for connecting with Fortmatic instead of MetaMask.

image

This isn't 100% complete: it has been tested mainly with the "Fund issue" scenario. The PR is mainly to validate that the implementation approach is OK, and to check if this is a feature you are willing to add.

Some things that are pending:

- As mentioned, this works for the "Fund issue" scenario. In other scenarios, there is a message asking only to connect to MetaMask.
- If this is merged, there are several palces where "MetaMask" should be replaced with the more generic "Wallet".

PR #5568, opened 2019-11-29
[duplicate]: fortmatic support

thelostone-mc:
### Description 

Same https://github.com/gitcoinco/web/pull/5287 except 
conflicts  have been resolved + rebased against master + minor refactoring along the way 

@fvictorio


app/retail/templates/shared/head.html
```diff
@@ -76,9 +76,13 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
 {% if meta_title %}<meta name="title" content="{{meta_title}}" />{% endif %}
 {% if meta_description %}<meta name="description" content="{{meta_description}}" />{% endif %}
+<script>
+  window.fortmatic_api_key = '{{ fortmatic_api_key }}'
+</script>
+<script async src="{% static "v2/js/lib/watch.js" %}"></script>
+<script async src="{% static "v2/js/lib/web3-0.20.7.min.js" %}"></script>
```
octavioamu: we are using 0.20.6 on shared_scripts 20.7 is with some issues.

octavioamu: also we already load it on shared_scripts.html

thelostone-mc: @fvictorio ^

fvictorio: Hi, I cannot push to this branch, but assuming `web3` is present in all the relevant pages, this line and the file it includes could just be removed (I did it locally but couldn't test it because i'm getting a `No module named 'celery'` error when I start docker compose).

thelostone-mc: @fvictorio I'll push the changes over to your branch and you can take it forward from there

thelostone-mc:
@fvictorio it doesn't look like you've enabled repo owners to update your PR
So I am unable to push my commits onto your branch !
Could you pull this in and update your PR and make your changes on top of that 

I'll close this PR once you update yours

app/retail/templates/shared/head.html
owocki: > i'm getting a No module named 'celery' error when I start docker compose).

you need to pip install the latest requirements file

fvictorio: Shouldn't this be done when the image is built?

fvictorio: Oh, my bad. I forgot that you need to do `docker-compose build` explicitly.

owocki: it is; but i think the upstream branch got updated + this was pulled in after you built

fvictorio:
@thelostone-mc I tested it locally and I think it wasn't working correctly. I want to test everything again with the requested change (and the latest merges/modifications), but it will take me some time and I won't be able to do it for a couple of days. I'll update the original PR after that and mention you.

fvictorio:
@thelostone-mc So, I looked into this and I think the problem is that in bfb63dcd8d612b65956156d717a7233b91659501 a line was added that creates a web3 instance using Infura if there isn't one already. This means that the `typeof web3 == 'undefined'` checks are always false. A check like this is used to show the `no_metamask_error` page, where you can enable Fortmatic. So, with that change, you can't connect Fortmatic if you don't have MetaMask installed (which, arguably, is the main scenario for using Fortmatic). You can still use it when you are asked to connect MetaMask.

I'm not sure how to fix this without breaking anything else, since I'm not that familiar with the project. I guess that the `typeof web3 === 'undefined'` check could be replaced with checking if the current provider is Infura. Alternatively, when the infura instance is created, you could set some global flag indicating this, but this is very ugly and you need to reset that flag in every place where a different kind of web3 instance is created.

With respect to the PR, I think we can continue working on this one, since it has all the changes that the previous one had, and a little more. But if you prefer to work on the other one for some reason, we can continue the discussion there.

fvictorio:
Oh, right, the problem is I can't push on this one :man_facepalming: I re-opened #5163. The "Allow edits from maintainers" option is checked, so you would be able to push there too.

PR #5569, opened 2019-11-29
Chat 1.2 - General Availability

### LINKED ISSUE #5522: the big 'gitcoin chat' feedback thread
post yer feedabck here

androolloyd:
##### Description

Chat/tasks.py - tasks to communicate needs to Mattermost
dashboard/views.py - after a bounty interest is created connect hunter and funder together into the bounty channel in the hackathon team

##### Refers/Fixes

#5522 #5614


app/chat/tasks.py
```diff
@@ -0,0 +1,48 @@
+from django.conf import settings
+
+from app.redis_service import RedisService
+from celery import app, task
+from celery.utils.log import get_task_logger
+
+from mattermostdriver import Driver
+
+mm_driver = Driver({
+    'url': settings.CHAT_URL,
```
owocki: is there a doc on how to setup mattermost locally for local gitcoin dev? is it just another docker command?

app/app/settings.py
```diff
@@ -541,8 +544,11 @@
 GITHUB_APP_NAME = env('GITHUB_APP_NAME', default='gitcoin-local')
 
 # Chat
-CHAT_URL = env('CHAT_DRIVER_USER', default='')  # location of where mattermost is hosted
-
+CHAT_PORT = env('CHAT_PORT', default=443)  # port of where mattermost is hosted
+CHAT_URL = env('CHAT_URL', default='https://chat.gitcoin.co')  # location of where mattermost is hosted
```
danlipert: will this default cause any issues when people run the docker containers locally?

app/chat/management/commands/sync_users_to_chat.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def handle(self, *args, **options):
+        try:
+
+            users = Profile.objects.filter(user__is_active=True).prefetch_related('user')
+
+            tasks = []
+
+            for profile in users:
+                # if profile.chat_id is None:
```
danlipert: leftover comment

app/chat/management/commands/sync_users_to_chat.py
```diff
@@ -0,0 +1,79 @@
...
+from marketing.utils import should_suppress_notification_email
+
+
+class Command(BaseCommand):
+    help = "create users to Gitcoin chat, creates the user if it doesn't exist"
+
+    def handle(self, *args, **options):
+        try:
+
+            users = Profile.objects.filter(user__is_active=True).prefetch_related('user')
```
danlipert: lets call this `profiles` instead of `users` - I think its very confusing to conflate them since they are two separate and specific models

app/chat/management/commands/sync_users_to_chat.py
```diff
@@ -0,0 +1,79 @@
...
+                    "username": profile.handle,
+                    "first_name": profile.user.first_name,
+                    "last_name": profile.user.last_name,
+                    "nickname": profile.handle,
+                    "auth_data": f'{profile.user.id}',
+                    "auth_service": "gitcoin",
+                    "locale": "en",
+                    "props": {},
+                    "notify_props": {
+                        "email": "false" if should_suppress_notification_email(profile.user.email, 'chat') else "true",
```
danlipert: Can we just turn this off for now?

app/chat/management/commands/sync_users_to_chat.py
```diff
@@ -0,0 +1,79 @@
...
+                }, params={
+                    "tid": settings.GITCOIN_HACK_CHAT_TEAM_ID
+                }))
+            job = group(tasks)
+
+            result = job.apply_async()
+            for r in result.get():
+                if r is not None:
+                    if 'username' in r and 'id' in r:
+                        profile = Profile.objects.filter(handle=r['username'])[0]
```
danlipert: This will throw an exception instead of returning `None` if the handle cannot be found, right?

app/chat/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@app.shared_task(bind=True, max_retries=3)
+def create_channel(self, options, retry: bool = True) -> None:
+    """
+    :param options:
+    :param retry:
+    :return:
+    """
+    with redis.lock("tasks:create_channel:%s" % options['channel_name'], timeout=LOCK_TIMEOUT):
```
danlipert: Why the need for locks here?

app/chat/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+def add_to_channel(self, options, retry: bool = True) -> None:
+    """
+    :param options:
+    :param retry:
+    :return:
+    """
+    with redis.lock("tasks:add_to_channel:%s" % options['channel_id'], timeout=LOCK_TIMEOUT):
+        chat_driver.login()
+        try:
+            for x in options['profiles']:
```
danlipert: What is x here? Please use variable names that make sense to increase readability.

app/chat/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+    :return: None
+    """
+
+    if update_opts is None:
+        return
+
+    with redis.lock("tasks:update_user:%s" % user.profile.handle, timeout=LOCK_TIMEOUT):
+
+        try:
+            if user.profile.chat_id is None:
```
danlipert: This will throw an exception if `user.profile` doesn't exist

app/dashboard/models.py
```diff
@@ -2314,6 +2314,7 @@ class Profile(SuperModel):
     last_calc_date = models.DateTimeField(default=get_time)
     email = models.CharField(max_length=255, blank=True, db_index=True)
     github_access_token = models.CharField(max_length=255, blank=True, db_index=True)
+    chat_id = models.CharField(max_length=255, blank=True, db_index=True)
```
danlipert: Do we need to index on this field?

app/dashboard/tasks.py
```diff
@@ -15,6 +15,34 @@
 LOCK_TIMEOUT = 60 * 2
 
 
+@app.shared_task(bind=True, max_retries=3)
+def bounty_on_create(self, team_id, new_bounty, retry: bool = True) -> None:
+    # what has to happen that we want to chain data from one another together
```
danlipert: What do these comments mean?

app/dashboard/views.py
```diff
@@ -110,16 +110,26 @@
...
         "login": active_user_profile.handle,
         "email": active_user_profile.user.email,
         "name": active_user_profile.user.get_full_name(),
         "handle": active_user_profile.handle,
-        "id": active_user_profile.user.id,
+        "id": f'{active_user_profile.user.id}',
+        "auth_data": f'{active_user_profile.user.id}',
+        "auth_service": "gitcoin",
+        "notify_props": {
+            "email": False if should_suppress_notification_email(active_user_profile.user.email, 'chat') else True,
```
danlipert: @owocki is there any place else we should check to make sure we don't get any GDPR complaints from these emails?

app/dashboard/views.py
```diff
@@ -351,6 +361,118 @@ def new_interest(request, bounty_id):
         if interest.pending:
             start_work_new_applicant(interest, bounty)
 
+        if bounty.event:
+            from chat.tasks import add_to_channel, create_channel, create_user, chat_driver
+            from django.utils.text import slugify
```
danlipert: Do we need to have these imports here?

app/dashboard/views.py
```diff
@@ -351,6 +361,118 @@ def new_interest(request, bounty_id):
...
+            try:
```
danlipert: In general, we should limit the use of `try` blocks to places where we expect exceptions to happen, rather than wrapping huge chunks of code in try blocks "just to be safe"

app/dashboard/views.py
```diff
@@ -351,6 +361,118 @@ def new_interest(request, bounty_id):
...
+                        bounty.save()
+                else:
+                    bounty_channel_id = bounty.chat_channel_id
+
+                funder_profile = Profile.objects.filter(handle=bounty.bounty_owner_github_username)[0]
+
+                if funder_profile is not None:
+                    if funder_profile.chat_id is None:
+                        result = create_user.apply_async(args=[{
+                                "email": funder_profile.user.email,
```
danlipert: extra indent here

app/dashboard/views.py
```diff
@@ -351,6 +361,118 @@ def new_interest(request, bounty_id):
...
+                        bounty_channel_id = bounty_channel_id_response['id']
+                        bounty.save()
+                    else:
+                        bounty_channel_id = channel_lookup['id']
+                        bounty.chat_channel_id = bounty_channel_id
+                        bounty.save()
+                else:
+                    bounty_channel_id = bounty.chat_channel_id
+
+                funder_profile = Profile.objects.filter(handle=bounty.bounty_owner_github_username)[0]
```
danlipert: If we expect a single object, we should use `.get()` instead of `.filter()[0]`

app/chat/management/commands/sync_users_to_chat.py
androolloyd: ah perhaps checking on the array position, i will add a check if not none

app/chat/tasks.py
androolloyd: if 2 hunters requested to start work before a channel has been created, 2 messages could get queued to create the same channel, but if we lock with redis we dont have to worry about potentially trying to create 2.

app/chat/tasks.py
androolloyd: its a chat_id, will update

app/chat/management/commands/sync_users_to_chat.py
androolloyd: We can set the props to just false, so they will have to manually opt into emails from chat.

app/dashboard/views.py
androolloyd: so you'd rather multiple blocks? or less generalized Exceptions?

app/app/settings.py
androolloyd: Yes, will update it to a default for local dev instead.

app/dashboard/views.py
danlipert: both are good

app/dashboard/views.py
androolloyd: for sanity we have changed this to be false always, users have to opt in in the chat client to receive emails.

We'll revist making this dynamic once we have more testing completed.

app/chat/tasks.py
thelostone-mc: @androolloyd you probably missed this ! Could you shoot a follow up PR?

app/dashboard/tasks.py
thelostone-mc: @androolloyd is this creating a channel on bounty creation ?

PR #5575, opened 2019-12-02
Update views.py

Solexplorer:
##### Description

@alexvotofuture  changed his username in Github but the change wasn't reflected in gitcoin.co/about#team . I only updated the record there

##### Testing

ezgif com-video-to-gif (1)


thelostone-mc:
Yup I would not have seen this :P

PR #5587, opened 2019-12-04
responsive fixes for chat

### LINKED ISSUE #5522: the big 'gitcoin chat' feedback thread

androolloyd:
##### Description
CSS updates to provider a more responsive wrapper, originally was going to get the permission work added in but after reviewing and looking at @owocki's  comments around usage for templates etc, changing to permissions in this release would be too rushed so it was held back.

##### Refers/Fixes
#5522 1.1 ticketing notes.

### DIFF
--- a/app/chat/templates/embed.html
+++ b/app/chat/templates/embed.html
@@ -51,7 +51,7 @@
       }
     </style>
     <div class="chatWrapper">
-          <iframe frameborder="0" src="https://google.com" allowfullscreen></iframe>
+          <iframe frameborder="0" src="{{ chat_url }}" allowfullscreen></iframe>
     </div>
     <footer class="container-fluid no-gutters footer">
 <div class="row">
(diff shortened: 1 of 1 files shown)


app/chat/templates/embed.html
```diff
@@ -29,6 +29,13 @@
       {% include 'home/nav.html' %}
     </div>
     <style type="text/css">
+
+      @media only screen and (max-width : 320px) {
+        .chatWrapper {
+          padding-bottom: 150% !important;
```
octavioamu: not sure what is the exact intention here, but you can use 100vh (100% of the viewport height ) or even use calc(100 - size of element)

androolloyd: How the wrapper works to ensure the internal contents are responsive is by extending the bottom of the wrapper div with padding, 39% gives 16:4 with a footer, but when it comes to mobile we want to extend beyond the current max height by 50%.

how would 100vh in this context?

octavioamu: but then is not adding scroll to the screen?

octavioamu: maybe just adding `d-none d-md-block` class to the footer so get hidden on mobile

PR #5590, opened 2019-12-04
add missing migration + isort fix

thelostone-mc:


Sirlupinwatson1:
Hi @thelostone-mc

À description would be great !

PR #5599, opened 2019-12-05
Changes to avoid self fund on same addres at grants

### LINKED ISSUE #5459: Grant owners cannot self-fund grants via recurring subscription
**Describe the bug**
If a grant owner attempts to self fund a grant, the subscriptions will fail in the subminer due to a check in the grant contract code that the origin and destination addresses are not the same: https://github.com/danlipert/grants1337/blob/specify-relayer/contracts/Subscription.sol#L186

**To Reproduce**
Create a grant and then with the same recipient address create a recurring subscription. Run the subminer and the metatx will fail.

**Expected behavior**
Given that this code is already in the deployed contracts, changing it now may not be wise. I suggest we popup a message or disable recurring subscriptions in this case or something similar instead of changing the contract. @owocki what are your thoughts?

zoek1:
##### Description
When a grant owner try to fund a grant with the recurring option selected, one message shol=uld prompt the user that switch the account.

Fund Grant _ Gitcoin

Also at the backen side there one rule to redirect to grant detail page when the grant admin address and the contributor address are the same.

##### Refers/Fixes
#5459


app/assets/v2/js/grants/fund.js
```diff
@@ -199,6 +199,10 @@ $(document).ready(function() {
         let approvalSTR = realApproval.toLocaleString('fullwide', { useGrouping: false });
 
         web3.eth.getAccounts(function(err, accounts) {
+          if (data.recurring_or_not == 'recurring' && grant_admin_addres === accounts[0]){
+            _alert({ message: gettext('Grant owners cannot self-fund grants. Please change your account.') }, 'error');
```
thelostone-mc: update it to state grant owners cannot use recurring option while self funding

app/grants/templates/grants/fund.html
```diff
@@ -354,6 +354,8 @@ <h5 class="form__label">{% trans "Gas Settings" %}</h5>
     {% include 'grants/shared/shared_scripts.html' %}
 
     <script>
+      const grant_admin_addres = '{{ grant.admin_address }}';
```
thelostone-mc: `grant_admin_addres` -> `grant_admin_address`

app/assets/v2/js/grants/fund.js
thelostone-mc: `grant_admin_addres` -> `grant_admin_address`

zoek1:
> would it be possible to set the default to One Time Payment in the UI
if the grant owner logs in with the same wallet address as the grant admin address?

Sure, should the page requests the account address and update to one time payment as soon the page loads or should wait until the user click the button, display the alert and then change the option to one time payment?

@thelostone-mc

zoek1:
> would it be possible to set the default to One Time Payment in the UI
if the grant owner logs in with the same wallet address as the grant admin address?

@thelostone-mc i added this behavior to the code, when is detected a self fund change inmediately to 'once opton'.

app/grants/views.py
```diff
@@ -518,7 +517,7 @@ def grant_fund(request, grant_id, grant_slug):
             'grant': grant,
             'text': _('You already have an active subscription for this grant.')
         }
-        return TemplateResponse(request, 'grants/shared/error.html', params)
+        # return TemplateResponse(request, 'grants/shared/error.html', params)
```
thelostone-mc: Why was this commented out ?

thelostone-mc:
@zoek1 it's almost wrapped up ! just had one minor comment which I've 
Otherwise LGTM to me

app/grants/views.py
zoek1: :scream: Sorry, I commented this to reuse some grants, but I comment it again!

app/assets/v2/js/grants/fund.js
```diff
@@ -199,6 +200,11 @@ $(document).ready(function() {
         let approvalSTR = realApproval.toLocaleString('fullwide', { useGrouping: false });
 
         web3.eth.getAccounts(function(err, accounts) {
+          if (data.recurring_or_not == 'recurring' && grant_admin_address === accounts[0]) {
```
octavioamu: shouldn't we check the user instead? I believe the address for the owner can be set to any address. Or also another grant team member can still go and auto fund.

danlipert: @octavioamu self funding with a different address is okay - theres actually a limitation in the contract itself that causes an error when you self fund with the grant admin's address

zoek1: @octavioamu @danlipert is it okay the current validation?

thelostone-mc:
Hey @zoek1 so we had to block grant owners from funding their own grant as explained here
https://github.com/gitcoinco/web/pull/5807
Also we had to hotpatch a fix in to ensure that  one time payment loads first by default as users preferred using that

zoek1:
Definetly improves the flow, happy for that!  🙂

PR #5600, opened 2019-12-05
feat: /1 crosschain etc integration 

thelostone-mc:
##### Description

- introduce new endpoint for ETH + EC20 bounties
- splits code for ETH and ETC bounty creation to ensure ETH
  bounties go normal flow while ETC bounties hit new endpoint
- sync_etc url + create API are restricted to admin
- update bounty details condition to ensure for ETH bounties
  actions needs user to be logged in and have the right address

##### Pending 

- sync up with @willsputra to figure out how display pill button for ETC bounties 
- check TODOs to evaluate duplicate/redundant fields in bounty models
- Wire in email creation for `/create` endpoint
- Wire in activity updates for `/create` endpoint
- Add in more strict checks for `/create` endpoint

##### Testing

image

Screenshot 2019-12-11 at 4 17 56 PM


app/dashboard/views.py
```diff
@@ -2866,6 +2867,74 @@ def sync_web3(request):
...
+        retry = True
+        blockNumber = 0
+        confirmations = 0
+
+        # TODO: figure out retry / exponential backoff
+        while retry:
+            transaction = get_etc_txn_status(txnid, network)
+            if not transaction:
+                logging.error('blockscout failed')
+            elif transaction['has_mined']:
```
thelostone-mc: Other options:
- we could off load the retry mechanism to the frontend if that makes sense !
- we could invoke the blockscout API from the frontend and then shoot a POST request to our backend once txn is mined

app/app/urls.py
```diff
@@ -62,8 +62,8 @@
     # oauth2 provider
     url('^o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
     path('jsi18n/', JavaScriptCatalog.as_view(), name='javascript-catalog'),
-    # create bounty
-    url(r'^create?', dashboard.views.new_bounty, name='create_bounty'),  # TODO: Remove
```
danlipert: This definitely isn't used anywhere? or by some weird 3rd party integration that we forgot about or something?

app/dashboard/views.py
```diff
@@ -4172,3 +4238,158 @@ def save_tribe(request,handle):
...
+
+    bounty = Bounty()
+
+    bounty.bounty_owner_profile = profile
+    bounty.bounty_state = 'open'
+    bounty.title = request.POST.get("title", '')
+    bounty.token_name = request.POST.get("token_name", '')
+    bounty.bounty_type = request.POST.get("bounty_type", '')
+    bounty.project_length = request.POST.get("project_length", '')
+    bounty.estimated_hours = request.POST.get("estimated_hours", None)
```
danlipert: with `.get` if you don't specify the 2nd argument the default is `None` so you don't need to add it

app/app/urls.py
thelostone-mc: Yeah ! This was introduced as a means to have multiple ways to access the same way !
the 3rd part were with -> `/new` OR `/fund` (cause one of those were added cause of that )

I don't know if that 3rd part integration even exists (separate story all together )

app/dashboard/views.py
thelostone-mc: Oh  noice :O

app/assets/v2/js/pages/new_bounty.js
```diff
@@ -271,6 +273,21 @@ const tokenAuthAlert = (isTokenAuthed, tokenName) => {
   }
 };
 
+
+const updateViewForToken = (token_name) => {
+
+  if (token_name == 'ETC') {
+    $('#featuredBounty').prop('checked', false);
```
octavioamu: is not the opposite?  true for ETC

app/retail/templates/shared/nav.html
```diff
@@ -20,12 +20,6 @@
   <a class="navbar-brand pt-0 pb-0" href="/bounties/funder">
     <img id="logo" src="{% static "v2/images/top-bar/bounties_desktop.svg" %}" height="35" alt="Logo" />
   </a>
-  {% if request.path|matches:"^\/((explorer)|(issue.*)|(.*new.*))$" %}
```
octavioamu: we are not longer show this for metamask?

app/assets/v2/js/pages/new_bounty.js
thelostone-mc: @octavioamu ah so that's taken care [here](https://github.com/gitcoinco/web/pull/5600/files#diff-bda00bc127f28e4f8ec2648a72152af7R33)

But pointing that out -> it makes sense to not even keep it there !  
removing it from there

app/assets/v2/js/pages/bounty_details.js
```diff
@@ -483,8 +483,17 @@ const isAvailableIfReserved = function(bounty) {
 };
 
 const isBountyOwner = result => {
-  if (typeof web3 == 'undefined' || !web3.eth ||
-      typeof cb_address == 'undefined' || !cb_address || !result) {
+  if (document.is_bounties_network) {
```
octavioamu: how this handle web3 browsers?

app/retail/templates/shared/nav.html
thelostone-mc: it would be removed in  pages (where it's not needed like retail + explorer )
+ bounties (where we have the metamask checks which block user from taking any interaction on the page ) 
=====
It would still show up in other products 
AKA kudos, quests, grants, tips, ens (as the navbar code is duplicated there )

PR #5601, opened 2019-12-05
Update emails.py

connoroday:


danlipert:
cherry-picked! @connoroday if you want to do emojis, just use the emoji itself, not the unicode bytes

PR #5606, opened 2019-12-06
mint kudos token requests (and redeem bulk kudos) in celery via a queue with retries, not on request/response cycle

owocki:
##### Description

mint token requests (and redeem kudos) in celery via a queue with retries, not on request/response cycle

##### Refers/Fixes

my own todo list + request form @alexvotofuture ; no ticket

##### Testing

tested and working (both functions)


owocki:
@androolloyd any idea what this exception is about? i followed the docs but no luck.... tried googling but then got out of my depth pretty quickly

owocki:
@danlipert just fixed the issues on this one + submitted it for reivew

app/kudos/utils.py
```diff
@@ -380,8 +380,12 @@ def _get_contract(self):
             obj: Web3py contract object.
 
         """
-        with open('kudos/Kudos.json') as f:
-            abi = json.load(f)
+        try:
+            with open('kudos/Kudos.json') as f:
+                abi = json.load(f)
+        except:
+            with open('app/kudos/Kudos.json') as f:
```
danlipert: Is this some kind of prod/local difference?

owocki: no; i think it has to do with different base paths and therfore relative paths from celeryworkers vs gunicorn/manage.py

androolloyd: Hmm, the paths should all be the same, but perhaps there is something preventing them or we're not starting from the right context. 

Interesting to note, the failure for the auto merge is because it can't find anything in the app namespace.

PR #5608, opened 2019-12-06
Faster CLR Calculation

frankchen07:
### Description

Added three methods:

1. method that translates grant dictionary data into a list of lists
2. method that aggregates and calculates the by pair totals
3. method that continually tests thresholds until the proper total CLR amount is found, and calculates the result

### Testing

I was going through the curve calculation part, and I believe it may break it. My concern is around line 224, but it may be that it's alright. Let's pair up and take look at that @thelostone-mc.

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -29,7 +29,6 @@ from django.utils import timezone
 from grants.models import Contribution, Grant, PhantomFunding
 from perftools.models import JSONStore
 
-CLR_DISTRIBUTION_AMOUNT = 100000.0
 LOWER_THRESHOLD = 0.0
 CLR_START_DATE = dt.datetime(2019, 9, 15, 0, 0)
 
@@ -51,12 +50,12 @@ CLR_START_DATE = dt.datetime(2019, 9, 15, 0, 0)
     Returns:
         [[grant_id (str), user_id (str), contribution_amount (float)]]
 '''
-def translate_data(grants_data):
+def translate_data(grants):
     grants_list = []
-    for g in grants_data:
-        grant_id = g.get('id')
-        for c in g.get('contributions'):
-            val = [grant_id] + [list(c.keys())[0], list(c.values())[0]]
+    for grant in grants:
+        grant_id = grant.get('id')
+        for contribution in grant.get('contributions'):
+            val = [grant_id] + [list(contribution.keys())[0], list(contribution.values())[0]]
             grants_list.append(val)
     return grants_list
 
@@ -72,9 +71,6 @@ def translate_data(grants_data):
 
     Returns:
         {grant_id (str): {user_id (str): aggregated_amount (float)}}
-
-        and
-
         {user_id (str): {user_id (str): pair_total (float)}}
 '''
 def aggregate_contributions(grant_contributions):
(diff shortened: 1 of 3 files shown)


owocki:
how much faster is this?

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
-        'id': grant_id,
-        'profile_pairs': profile_pairs,
-        'contribution_pairs': contribution_pairs,
-        'sqrt_of_product_pairs': sqrt_of_product_pairs
-    }
-
-    return grant
+def translate_data(grants_data):
+    grants_list = []
+    for g in grants_data:
```
thelostone-mc: `g -> grant`

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
-
-    grant = {
-        'id': grant_id,
-        'profile_pairs': profile_pairs,
-        'contribution_pairs': contribution_pairs,
-        'sqrt_of_product_pairs': sqrt_of_product_pairs
-    }
-
-    return grant
+def translate_data(grants_data):
```
thelostone-mc: `grants_data -> grants`

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
+    grants_list = []
+    for g in grants_data:
+        grant_id = g.get('id')
+        for c in g.get('contributions'):
```
thelostone-mc: `c -> contribution`

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
+def aggregate_contributions(grant_contributions):
+    contrib_dict = {}
+    for proj, user, amount in grant_contributions:
+        if proj not in contrib_dict:
+            contrib_dict[proj] = {}
+        contrib_dict[proj][user] = contrib_dict[proj].get(user, 0) + amount
+
+    tot_overlap = {}
+    for proj, contribz in contrib_dict.items():
+        for k1, v1 in contribz.items():
```
thelostone-mc: should we rename k1 , v1 into something more meaningful

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
+            if k1 not in tot_overlap:
+                tot_overlap[k1] = {}
+            for k2, v2 in contribz.items():
```
thelostone-mc: should we rename k2 , v2 into something more meaningful

app/grants/clr.py
```diff
@@ -47,125 +47,113 @@
...
+            tot = 0
+            for k1, v1 in contribz.items():
+                for k2, v2 in contribz.items():
+                    if k2 > k1:  # remove pairs
+                        # # pairwise matching formula
+                        # tot += (v1 * v2) ** 0.5 * min(1, threshold / pair_totals[k1][k2])
+                        # vitalik's division formula
+                        tot += ((v1 * v2) ** 0.5) / (pair_totals[k1][k2] / threshold + 1)
+            bigtot += tot
+            # totals.append((proj, tot))
```
thelostone-mc: if it's not needed -> let's remove it

thelostone-mc:
@owocki so 
@frankchen07 did a run here https://github.com/thelostone-mc/clr/tree/frank-dev

Ran both the algo against the previous clr dataset
algorithm: ran 1 time
data: previous clr data set from live
speed: ~5x faster

Upside: 
- this is 5x faster

Reality: 
- it's not still fast enough to be able to run it on demand and has to run has a cron
- it can be run ~30times as opposed to ~6 times per day which makes the results a lil more accurate but not something the user will notice
- the end results of both algo are the same 


cc @gitcoinco/engineers

PR #5629, opened 2019-12-13
Removes edit this page link from every page

### LINKED ISSUE #4943: Gitcoin documentation Update Bounty
This is a bounty that is designed to help the core team make sure the documentation at https://github.com/gitcoinco/web/tree/master/docs is up to date.

How it works:
1. Find somewhere where the documentation is lacking or incorrect.
2. Click 'start work' and let me know where it is / why.
3. Submit a PR that updates the documentation according to your work plan from step 2.  If it is accepted, I will pay out 15 DAI for that update.

agbilotia1998:
##### Description

The PR removes the `Edit this page` button from the documentation pages which redundantly redirects to a 404 page.

Screenshot_2019-12-13 Home - Gitcoin Developer Documentation

##### Refers/Fixes

#4943

##### Testing

No tests required.

### DIFF
--- a/pydocmd.yml
+++ b/pydocmd.yml
@@ -1,6 +1,6 @@
 site_name: "Gitcoin Developer Documentation"
 repo_name: 'gitcoinco/web'
-repo_url: 'https://github.com/gitcoinco/web'
+edit_uri: ''
 
 # This tells pydocmd which pages to generate from which Python modules,
 # functions and classes. At the first level is the page name, below that


molecula451:
IMO, the button should be able to redirect to a valid link. But not to be removed somehow.

agbilotia1998:
@molecula451 the markdown files are generated from the python modules by configuring in pydocmd and the generated files are then rendered on the server. These files probably are not uploaded or updated on git repositories so it won't be possible to provide an edit link. The thing that can be done is error reporting, by which the user can open an issue on github regarding the errors in the documentation, but that should not be the part of `Edit this page` default button provided by MkDocs. Can we do something else here?

agbilotia1998:
@owocki @thelostone-mc @danlipert @octavioamu please review.

pydocmd.yml
```diff
@@ -1,6 +1,6 @@
 site_name: "Gitcoin Developer Documentation"
 repo_name: 'gitcoinco/web'
-repo_url: 'https://github.com/gitcoinco/web'
```
danlipert: what does removing the `repo_url` do here?

danlipert:
@agbilotia1998 The core engineers are automatically notified when PRs are opened, so please don't tag us for reviews as it spams our inboxes - thanks!

pydocmd.yml
agbilotia1998: @danlipert MkDocs automatically depending upon the theme takes this parameter `repo_url` and creates an edit link over every page.

danlipert: @agbilotia1998 I see - does removing the `repo_url` affect anything else in the docs?

agbilotia1998: @danlipert seems nothing else gets effected.

danlipert: @agbilotia1998 great - how did you test that nothing else was affected?

agbilotia1998: A basic sanity for the redirections to pages and check if the url is present at other places. May be you can apply this patch and check if I've missed anything.

agbilotia1998: A place where additionally this repo url is present is at the header where the number of stars and forks for the repo are present. This gets removed on deleting the repo url.

thelostone-mc: can't we set the edit_uri to '' and leave the repo url ?

agbilotia1998: @thelostone-mc some themes take the repo url and automatically checks hosts like github.com, if present, it automatically creates an edit link by appending `/edit/master/docs` to the repo_url for every page.

danlipert: @agbilotia1998 would be good to keep the stars and forks info.. is there some way we can disable the edit links without removing the metrics?

agbilotia1998: @danlipert @thelostone-mc this PR in pydocmd package fixes the problem by adding the `edit_uri` parameter in the mkdocs configuration https://github.com/NiklasRosenstein/pydoc-markdown/pull/80
This change will be included in v2.0.6 which is not released yet, currently, we are using v2.0.4 and v2.0.5 also does not provide a fix for this. For now, maybe we can just comment the repo_url parameter stating to be used with v2.0.6 or other thing that can be done is cloning this package and adding `edit_uri` parameter at https://github.com/NiklasRosenstein/pydoc-markdown/blob/54c33c8845e5936a25667ebb20d10a9913793910/pydocmd/__main__.py#L75

danlipert: Hmm... looks like v2.0.5. was released over a year ago... in that case I'm fine with having the metrics removed for now since it doesn't look like v2.0.6 is coming any time soon.

PR #5637, opened 2019-12-14
feat: update grants landing page

thelostone-mc:
##### Description

- new field added to grant model -> grant_type
- new field added to pledge model -> pledge_type
- admin view for tips + pledges + grants updated tabular
- view updated to distinguish between grant types
- fixed buggy search + shareable URI with filters

##### Testing

DEMO: https://share.vidyard.com/watch/ZhWopgeHmmTrxd5G46vhD6?

ADMIN TABULAR 

Screenshot 2019-12-14 at 1 12 38 PM

image

image

PR #5639, opened 2019-12-15
clr: update estimate_clr command to query by grant type 

thelostone-mc:
##### Description

- estimate_clr.py -> new arguments clr_type, network
- clr logic updated to query grants by network + grant_type + not hidden
- grant str updated to also print grant_type
- crontab updated to run both clr for media and tech grants

##### Refers/Fixes

Gets in after #5637 

##### Testing

Screenshot 2019-12-15 at 2 57 30 PM

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -243,14 +243,22 @@ def calculate_clr_for_donation(donation_grant, donation_amount, total_pot, base_
     print('error: could not find grant in final grants_clr data')
     return (None, None)
 
-def predict_clr(random_data=False, save_to_db=False, from_date=None):
+def predict_clr(random_data=False, save_to_db=False, from_date=None, clr_type=None, network='mainnet'):
     # setup
     clr_calc_start_time = timezone.now()
 
     # get all the eligible contributions and calculate total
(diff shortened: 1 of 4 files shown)


app/grants/clr.py
```diff
@@ -243,14 +243,22 @@ def calculate_clr_for_donation(donation_grant, donation_amount, total_pot, base_
...
+def predict_clr(random_data=False, save_to_db=False, from_date=None, clr_type=None, network='mainnet'):
     # setup
     clr_calc_start_time = timezone.now()
 
     # get all the eligible contributions and calculate total
     contributions = Contribution.objects.prefetch_related('subscription').filter(created_on__gte=CLR_START_DATE, created_on__lte=from_date)
     debug_output = []
-    grants = Grant.objects.all()
+
+    grants = Grant.objects.none()
```
danlipert: is this line some kind of fallback? It seems the following if else block would assign grants to something without needing this line

thelostone-mc: Yeah ! Just kept it as fallback -> Ideally we would never need this

danlipert: @thelostone-mc I think from a programming logic standpoint there is no case where this would be undefined after the if else block, so I think it should be removed

thelostone-mc: removed ^_^

Issue #5647, opened 2019-12-17
Error after publishing the grant


ninabreznik:
I published a grant but I am getting this error (which I can't click on)
![Screenshot_2019-12-17_16-03-07](https://user-images.githubusercontent.com/1076427/71016247-79da1d00-2127-11ea-9ef3-3a39ae3a8dfb.png)

owocki:
looks like the msg no longer pulls the grant admin address on the frontend anymore..

should be easy to fix this!  adding a bounty on this now

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.3 ETH (37.87 USD @ $126.23/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/5647/3819).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/5647/3819).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $71,941.66 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [molecula451](https://gitcoin.co/molecula451) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/5647/3819?mutate_worker_action=approve&worker=molecula451) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/5647/3819?mutate_worker_action=reject&worker=molecula451))_.**

I am already working on this issue

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/5647/3819).

gitcoinbot:
@molecula451 Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] reminder (3 days)
* [ ] escalation to mods (6 days)

[...]

PR #5650, opened 2019-12-18
Fix for Admin Address Pull on Grant

molecula451:
##### Description

Expected Admin Grant Address <code> #grant_contract_owner_address </code> on the Error Message but I am seeing on the file it's appending the value as <code> #contract_owner_address </code>

71016247-79da1d00-2127-11ea-9ef3-3a39ae3a8dfb

Syntax:

<code> $('selector expression').text('content if applied'); </code>

##### Refers/Fixes

Line fixing

#5647 

##### Testing

Brave Browser
This should not break any other components

### DIFF
--- a/app/assets/v2/js/grants/detail.js
+++ b/app/assets/v2/js/grants/detail.js
@@ -2,7 +2,6 @@ const editableFields = [
   '#form--input__title',
   '#form--input__reference-url',
   '#contract_owner_address',
-  '#grant_contract_owner_address',
   '#grant-members',
   '#amount_goal'
 ];
--- a/app/assets/v2/js/grants/detail.js
+++ b/app/assets/v2/js/grants/detail.js
@@ -13,10 +13,10 @@ $(document).ready(function() {
   setInterval (() => {
     notifyOwnerAddressMismatch(
       $('#grant-admin').text(),
-      $('#contract_owner_address').html(),
+      $('#contract_owner_address').text(),
       '#cancel_grant',
       'Looks like your grant has been created with ' + 
-      $('#grant_contract_owner_address').html() + '. Switch to take action on your grant.'
+      $('#grant_contract_owner_address').text() + '. Switch to take action on your grant.'
     );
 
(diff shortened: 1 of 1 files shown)


app/assets/v2/js/grants/detail.js
```diff
@@ -13,10 +13,9 @@ $(document).ready(function() {
   setInterval (() => {
     notifyOwnerAddressMismatch(
       $('#grant-admin').text(),
-      $('#grant_contract_owner_address').text(),
+      $('#contract_owner_address').text(),
```
molecula451: <code>  formData.append('contract_owner_address', $('#contract_owner_address').val()</code>

molecula451:
> @molecula451 could you
> 
> * undo the indentation you had done for HTML
> * also how does this fix the issue ?
>   grant_contract_owner_address & contract_owner_address have the same text value
>   So ideally if it errors out for one wouldn't the same happen for the other ?

It is suppose we are getting the text <code> .val() </code> from the $selector <code> contract_owner_address </code> which is taking the owner address on the 

**function** <code> web3.eth.getAccounts(function(err, accounts) {
    $('#input-admin_address').val(accounts[0]);
    $('#contract_owner_address').val(accounts[0]);
  }); </code> 

Why would we use <code> grant_contract_owner_address </code> instead when there is no passed <code> .append </code> value on this selector? it might be the actual case why it is not outputing the <code> .text() </code>

molecula451:
@thelostone-mc I can't even see a <code>id=type</code> in our template <code>grants/new.html</code>  for <code>grant_contract_owner_address</code> so far there is one  <code> input type="hidden" id="contract_owner_address" name="contract_owner_address" </code>

molecula451:
Hi @thelostone-mc <code> grant_contract_owner_address" class="hidden"</code> has a css property of <code>.hidden display:none</code> in<code>base.css</code> even though the #selector it's getting the address value it is being preventing from being display it

PR #5653, opened 2019-12-18
restarts celery on deploy

owocki:
##### Description

restarts celery on deploy

##### Refers/Fixes

deploy channel

##### Testing

still working out kinks.  see deploy channel

### DIFF
--- a/scripts/deploy.bash
+++ b/scripts/deploy.bash
@@ -87,7 +87,7 @@ if ! [ "$JOBS_NODE" ]; then
       done
     else
       echo "- celery"
-      sudo systemctl restart celery
+      sudo systemctl restart celery.service
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/deploy.bash
```diff
@@ -84,7 +84,8 @@ if ! [ "$JOBS_NODE" ]; then
     sudo kill -1 $pid
     sleep 0.5
     done
-
```
owocki: is there another env var w3e should use to detect if were on a celery node?

androolloyd: We should add one to indicate its a celery box, rather than checking for a file system.

owocki: is that something you can do? im not sure the correct way to do so.. i just submitted this PR bc celery wasnt working in production the other day..

androolloyd:
I'll add `export CELERY_NODE=1` to /etc/profile on the celery box to match this change.

danlipert:
@androolloyd this ready to go?

PR #5683, opened 2019-12-24
Add option to send kudos from profile

zoek1:
##### Description

1. Users now seeing a button to send kudos to this user:
image
2. When the sender user clicks the button, is redirected to the page with the user to field populated with the user info:
image

##### Refers/Fixes

#5661


owocki:
thanks; does the kudos stick as they click 'pick a kudos' and browse the marketplace too?

owocki:
@zoek1 note that this PR differs slightly from the expected behaviour in the ticket in that the kudos to username specification will be lost when the user browses to select their kudos.. it should probably stick until cleared.

zoek1:
Yep, I missed that. I'm changing that to persist the username with this rules:
1. Until the kudos is sent, clear persistent username
2. Until the user removes the username manually and for a max period of one day. In such a case, the stored username will be discarded.

@owocki

zoek1:
The changes are ready @owocki, maybe three hours could be better instead of one day to remove the receiver user from local storage, what do you think?

owocki:
@zoek1 i think 3 hours is fine.. also if the sending user clicks out of that user; then i think that should remove it too.

zoek1:
By click out what do you mean?  when sending user? As I see you refer to some of these:
1. when sending user select another user from the list or 
2. select another tab to send kudos (eth address, anywhere)

owocki:
@zoek1 both

zoek1:
@owocki the users now is removed when:
> 1. When sender user selects another user from the list or
> 2. when sender user selects another tab to send kudos (eth address, anywhere)

app/kudos/templates/transaction/send.html
```diff
@@ -165,4 +173,50 @@ <h3 class="font-weight-300">{% blocktrans %}<span id="new_username">the user</sp
...
+    let data = JSON.parse(encoded_data);
+    let current = new Date();
+    let past = new Date(Date.parse(data.date))
+
+    if ((current.getTime() - past.getTime()) / (60 * 60) > 1440) {
+      localStorage.removeItem('current_receiver')
+    } else {
+      let option = document.createElement('option')
+      option.value = data.id
+      option.setAttribute('avatar_id',  data.avatar_id)
```
danlipert: Theres a couple places in this PR with some spacing issues - can you take another look and clean these up? thanks!

zoek1:
Sure, i'll check and fix it!

app/kudos/views.py
```diff
@@ -289,6 +289,22 @@ def send_2(request):
...
         raise Http404
 
+    username = request.GET.get('username', None)
+    user = {}
+
+    if username:
+        profiles = Profile.objects.filter(handle__iexact=username)
+
+        if profiles.exists():
+            profile = profiles.first()
```
danlipert: If we had two users with handles "zoek1" and "zoek1asdfasdf", and I wanted to send a kudos to "zoek1asdfasdf", would this work?

zoek1: Sure, It works because the filter method uses iexact on the handle attribute, in the previous bounty you recommended that

owocki:
pumped to get this live.. (once it passes code review).. seems like its close to me

zoek1:
@owocki also this ready!

zoek1:
fix conflicts :+1:

owocki:
pumped for this

app/dashboard/templates/profiles/header_details.html
```diff
@@ -1,7 +1,6 @@
 {% load i18n static avatar_tags add_url_schema email_obfuscator %}
 
 <h1 class="profile-header__handle">
-  @{{ profile.name }}
```
danlipert: Was this removed by mistake?

zoek1: :see_no_evil: Thanks, yes deleted by mistake. I just restored this line.

Issue #5281, opened 2019-09-29
As a hackathon hacker I want team channels created for me in Discord automatically


alexvotofuture:
### User Story
As a hackathon hacker I want to have a Discord team channel created for me automatically when I "Start Work" on a prize.

### Why Is this Needed
Hackathon participants often need a real-time communication channel for coordinating with potential and active teammates, and need a place to ask sponsors project-specific details. 

The Gitcoin team currently uses Discord to provide these communication channels for hackers. However, all channels must be manually created by Gitcoin administrators, since there's no link between "Start Work" instances and the Discord server.

### Description

*Type*: Feature

### Current Behavior
Discord servers are manually created by Gitcoin administrators. Administrators periodically check hackathon prize pages for new "Start Work" instances, then create a Discord team channel for the user who started work for every new instance. They invite this user to their team channel according to the Discord username they provide in the "Start Work" work plan form (a field recently added by @owocki).

Gitcoin administrators make the Discord team channel public by default. This allows hackers to invite collaborators easily. But some hackers prefer private channels; in these cases an administrator will simply convert the public channel to a private channel, allowing Gitcoin administrators/ambassadors and sponsors to see messages but will block other hackers from seeing them.

### Expected Behavior
When a user clicks "Start Work" on a hackathon prize bounty a public Discord channel should be automatically created and the user that started work should be invited to/tagged in the channel.

Ideally, the "Work Plan" should be automatically pasted to the channel as a message or channel description, so that other hackathon participants/sponsors/administrators can see the project summary/approach within Discord and decide whether to get involved.

### Definition of Done
Discord team channels are automatically created when someone hits "Start Work".

Should provide permissions and simple documentation to help Gitcoin administrators use and enable/disable/configure the tool.

alexvotofuture:
cc @owocki @kuhnchris @octavioamu @frankchen07 @mul1sh for input on how to approach this

mul1sh:
@alexvotofuture awesome thanks for this , i'm fixing a tough bug for the other discord bot issue, but i'll definitely be done with it tomorrow. So if this is bountied up I can get it done by Friday without further delays, thanks.

kuhnchris:
@mul1sh you're making your own bot, or do you just create the interface for getting an signal when the start work occoured? thanks

mul1sh:
[...]

PR #5684, opened 2019-12-25
Gitcoin chat bot for creating channels

Solexplorer:
##### Description

This PR makes it possible to create channels automatically when the user clicks `Start Work` on a hackathon related bounty. What I have done is the following: 

- Added a function in the `notifications.py` that builds the message as per the example below and then creates a channel:

```
 @solexplorer Work Plan:
I will update the sponsors on my progress as I go.

Use this channel to communicate with teammates and sponsors.
https://gitcoin.co/issue/hashingsystems/hash.js/13/3756
```

- Added docs regarding environment variables

- Created a class that manages the actions of the bot for the gitcoin chat

##### Refers/Fixes

It's more related to #5281 but for MatterMost

##### Testing

I didn't add tests because the setup of the gitcoin chat has not been completed yet. Here is a short of video of [how it works](https://streamable.com/awxwy)


Solexplorer:
@thelostone-mc @octavioamu @danlipert Can I have a review on this ?
CC: @androolloyd

frankchen07:
I think we'll leave Discord name in until we deprecate, and then we can remove it and make sure we have the correct hooks in to relate gitcoin usernames to mattermost chat (which should be easy).

danlipert:
@androolloyd we can close this right? its already incorporated?

Solexplorer:
Yeah, this is not needed anymore

PR #5685, opened 2019-12-25
Christmasbranch

vs77bb:
##### Description

Made some changes to automated emails 
- no more asking for physical mailing addresses, we'll send shirts online
- fixed some post-bounty retention email copy 

##### Refers/Fixes
changed the email to send after 60 days of inactivity as it current reads 'every few months'... and i think thats less spammy
 
image

##### Testing

would love to know if there are any tests i should run for email updates, might have some more fun here in the new year!

### DIFF
--- a/app/retail/emails.py
+++ b/app/retail/emails.py
@@ -462,9 +462,10 @@ hi {github_username},
 alisa and frank from Gitcoin here (CC scott and vivek too) — i see you haven't funded an issue in {time_as_str}.
 in the spirit of making Gitcoin better + checking in:
 
-- has anything been slipping on your issue board which might be bounty worthy?
+- have any issues which might be bounty worthy or projects you're hoping to build?
 - do you have any feedback for Gitcoin Core on how we might improve the product to fit your needs?
-- are you interested in joining one of <a href="https://gitcoin.co/hackathon-list/">our upcoming hackathons?</a>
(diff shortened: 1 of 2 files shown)


danlipert:
@vs77bb nice! looks like there are some newsletter changes as well? Or is that an old newsletter... I can't remember. Anyway, let me know if you meant to make the newsletter changes and if not I'll show you how to fix this PR

app/retail/emails.py
```diff
@@ -443,7 +443,7 @@ def render_admin_contact_funder(bounty, text, from_user):
     return response_html, response_txt
 
 
-def render_funder_stale(github_username, days=30, time_as_str='about a month'):
+def render_funder_stale(github_username, days=60, time_as_str='a couple months'):
```
owocki: will need to update this line in the crontab too.  this change only edits the default arguments, not the actually executed ones

```
15 10 * * * cd gitcoin/coin; bash scripts/run_management_command_if_not_already_running.bash funder_stale_email 30  >> /var/log/gitcoin/funder_stale_email.log  2>&1
```

thelostone-mc: taken care

thelostone-mc:
raised a PR removing the newsletter commit over at https://github.com/gitcoinco/web/pull/5699
@vs77bb if that looks alright we can get that in ^_^

PR #5699, opened 2020-01-02
chore: update email text for stale funder + schwag

thelostone-mc:
##### Description

duplicate of https://github.com/gitcoinco/web/pull/5685/

### DIFF
--- a/app/retail/emails.py
+++ b/app/retail/emails.py
@@ -385,7 +385,7 @@ thanks for being a member of the community.
 
 alisa / frank (gitcoin product team)
 
-PS - we've got some new gitcoin schwag on order. send me your mailing address and your t shirt size and i'll ship you some.
+PS - we've got some new gitcoin schwag on order. if interested, let us know and we can send you a code to order some :)
 
 """
         elif bounty.status == 'cancelled':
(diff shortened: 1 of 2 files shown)


app/retail/emails.py
```diff
@@ -443,7 +443,7 @@ def render_admin_contact_funder(bounty, text, from_user):
     return response_html, response_txt
 
 
-def render_funder_stale(github_username, days=30, time_as_str='about a month'):
+def render_funder_stale(github_username, days=60, time_as_str='a couple months'):
```
vs77bb: while i think this is the right code change -- @frankchen07 also mentioned in chat that emails are happening sometimes even more regularly than 30 days, so i'm not sure if this will solve the root problem 🤔 

still worth making the change to 60 days IMO

thelostone-mc: will keep an eye out for that 🙌

frankchen07: I think it's more that funders who are funding issues are still getting the emails - they usually respond with "I think something's wrong with your system, I've funded a couple issues." 

Perhaps it is just a time elapsed issue, since if I funded something 30 days back and just am waiting on work, I might get sent the funder email.

PR #5711, opened 2020-01-06
switch phantom funding round

danlipert:
##### Description

Change phantom funding round to 4

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -208,7 +208,7 @@ def grant_details(request, grant_id, grant_slug):
         subscriptions = grant.subscriptions.filter(active=True, error=False).order_by('-created_on')
         cancelled_subscriptions = grant.subscriptions.filter(active=False, error=False).order_by('-created_on')
         _contributions = Contribution.objects.filter(subscription__in=grant.subscriptions.all())
-        phantom_funds = grant.phantom_funding.filter(round_number=3)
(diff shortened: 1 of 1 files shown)


app/grants/views.py
```diff
@@ -208,7 +208,7 @@ def grant_details(request, grant_id, grant_slug):
         subscriptions = grant.subscriptions.filter(active=True, error=False).order_by('-created_on')
         cancelled_subscriptions = grant.subscriptions.filter(active=False, error=False).order_by('-created_on')
         _contributions = Contribution.objects.filter(subscription__in=grant.subscriptions.all())
-        phantom_funds = grant.phantom_funding.filter(round_number=3)
+        phantom_funds = grant.phantom_funding.filter(round_number=4)
```
owocki: hmm i actually think this should be a `grant.phantom_funding.all()` not filtered by round number.. since we'll want to show the old phantom funding items on the grant details page too

danlipert: If we do that I think we'll need to update the grant details page as well to have two seperate sections and do the $5/funding selections separately

owocki: why is that? i dont see anything on the grant details page (not grant/fund page) that denotes phantom funds seperately by round.

the phantom funds object here is just used to create the contributors + contributions list..

danlipert: All makes sense - thanks Kevin!

PR #5730, opened 2020-01-08
fix incorrect link in quests link

danlipert:
##### Description

Changes the quests nav link to point to quests index instead of grants index

##### Refers/Fixes

Reported on slack

##### Testing

Tested locally


danlipert:
dupe

PR #5763, opened 2020-01-10
adds linux and mac desktop apps; alerts users when apps not ready yet

owocki:
##### Description

adds linux and mac desktop apps; alerts users when apps not ready yet

##### Refers/Fixes

slack convo

##### Testing

tested locally

### DIFF
--- a/app/assets/v2/js/pages/chat.js
+++ b/app/assets/v2/js/pages/chat.js
@@ -1,12 +1,7 @@
 
 $(document).ready(function() {
-  $("a").click(function(e){
-    var href = $(this).attr('href');
-    if (href=='#'){
+  $("a.comingsoon").click(function(e){
       alert("Coming soon!");
       e.preventDefault();
-      return;
-    }
   })
-
 });
--- a/app/assets/v2/js/pages/chat.js
+++ b/app/assets/v2/js/pages/chat.js
@@ -1,7 +1,7 @@
 
 $(document).ready(function() {
-  $("a.comingsoon").click(function(e){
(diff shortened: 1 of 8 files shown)


app/assets/v2/js/pages/chat.js
```diff
@@ -0,0 +1,12 @@
+
+$(document).ready(function() {
+  $("a").click(function(e){
```
octavioamu: this is goiing to block any link on that page, why we just don't comment out the download section?

owocki: it wont, only ones that have an `#` as their href will have `e.preventDefault()`

owocki: see the condition on line 5

androolloyd: I feel this selector is a bit too global, giving that there are other "#" links

Screen Shot 2020-01-10 at 1 40 42 PM


There would be more than just the links we want to say coming soon caught under this, selector.

What about doing something like this. 

`$(".my-3 a[href='#']").on('click', (e) => {alert("Coming Soon!"); e.preventDefault(); return;})`

Will only target the anchors inside the my-3 divs section on this page.

owocki: 325e042ff311224688c6ff4adbf86a2e6a72d290

octavioamu:
Approved, do you want to merge this one to master? not stable?

owocki:
ill cherrypick to stable!

PR #5776, opened 2020-01-11
Remove product update modal

### LINKED ISSUE #5771: login experience on gitcoin needs patching
this login experience is ridiculously bad

screencap - https://bits.owocki.com/p9u5OJlx

two modals show up and the cookie warning

then when you click out of the product updates modal; you cant actually click out of anything else.

expected result
1. no more bounty experience pop up upon first login
2. no more 'product update' pop up upon login

iamonuwa:
##### Description
Remove popups on login

##### Refers/Fixes
Fixes #5771 

##### Testing

clear localstorage and login again. it won't show again.

### DIFF
--- a/app/assets/v2/js/base.js
+++ b/app/assets/v2/js/base.js
@@ -352,24 +352,12 @@ const sendPersonal = (persona) => {
 };
 
 
-const gitcoinUpdates = (force) => {
-  let urlUpdates = `https://api.github.com/repos/gitcoinco/web/issues/5057?access_token=${document.contxt.access_token}`;
-  let today = new Date();
-  let showedUpdates = JSON.parse(localStorage.getItem('showed_updates'));
-  let lastPromp = showedUpdates ? showedUpdates.last_promp : today;
-  let lastUpdated = showedUpdates ? showedUpdates.last_updated : 0;
+const gitcoinUpdates = () => {
+  const urlUpdates = `https://api.github.com/repos/gitcoinco/web/issues/5057?access_token=${document.contxt.access_token}`;
 
-  if (!force && showedUpdates && (moment(today).diff(moment(lastPromp), 'days') < 7)) {
-    return;
-  }
+  const getUpdates = fetchData (urlUpdates, 'GET');
 
-  let getUpdates = fetchData (urlUpdates, 'GET');
-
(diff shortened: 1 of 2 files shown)


app/dashboard/templates/dashboard/index.html
```diff
@@ -177,7 +177,6 @@ <h3 class="popover-header"></h3>
     </script>
     <script src="/dynamic/js/tokens_dynamic.js"></script>
     <script src="{% static "v2/js/tokens.js" %}"></script>
-    <script src="{% static "v2/js/capture-rating.js" %}"></script>
```
owocki: is this going to break capturing of ratings on the profile page/rating tab, or in the rating flow on the bounty experience?

iamonuwa: The profile rating is different from this

app/assets/v2/js/base.js
```diff
@@ -403,10 +403,6 @@ const gitcoinUpdates = (force) => {
   });
 };
 
-if (document.contxt.github_handle) {
-  gitcoinUpdates();
-}
```
thelostone-mc: ^ can we update the `gitcoinUpdates` to remove the `force` argument + update it's invocation in `footer.html`

iamonuwa: We can do that.

iamonuwa: Removing `force` will not stop the popups. Hence I had to remove it entirely.

thelostone-mc: I meant we remove this and update the function so that it no longer needs an argument `force`

iamonuwa: Without the force, the popup will still be there

thelostone-mc: Ah it doesn't get invoked elsewhere ! worth removing the entire function itself?

iamonuwa: Noted.

iamonuwa: Its called from the footer as well as on login. I've removed the login call. Except you need me to remove the entire function and the link on footer

thelostone-mc: @iamonuwa updated the code to show what i meant !
Removed unused code in the function

PR #5786, opened 2020-01-13
udpate idx_status -> status

thelostone-mc:
##### Description

This partially retires the use of `idx_status` and instead uses the newer `bounty_state` field

Goes in after https://github.com/gitcoinco/web/pull/5785


thelostone-mc:
Closing this out for now! Will reopen if this something we wanna pursue at a later point in time

PR #5789, opened 2020-01-13
Chat/1.2.1 - bug fixes

androolloyd:
##### Description

Fixes to the bounty channel creation
Adds a mgmt command so that channels can be created or restored from the command line

### DIFF
--- a/app/chat/management/commands/sync_event_bounties_to_chat.py
+++ b/app/chat/management/commands/sync_event_bounties_to_chat.py
@@ -16,71 +16,45 @@
 
 '''
 from django.conf import settings
+from django.db.models import Q
 from django.core.management.base import BaseCommand
 from django.utils.text import slugify
 from dashboard.models import Bounty, Interest, Profile
 from chat.tasks import create_user, get_driver, create_channel, add_to_channel
+from chat.utils import create_user_if_not_exists, create_channel_if_not_exists
 import logging
 from celery import group
-from mattermostdriver.exceptions import (
-    ResourceNotFound
-)
+
 logger = logging.getLogger(__name__)
-
-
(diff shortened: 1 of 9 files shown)


app/dashboard/views.py
```diff
@@ -363,112 +364,52 @@ def new_interest(request, bounty_id):
...
-                        }]
-                        )
-
-                        chat_profile_interest_user = result.get()
-                        if 'message' in chat_profile_interest_user:
-                            raise ValueError(chat_profile_interest_user['message'])
-
-                        profile.chat_id = chat_profile_interest_user['id']
-                        profile.save()
+                    if funder_profile.chat_id is '':
```
octavioamu: many times the ones creating bounties are internals gitcoin members example https://gitcoin.co/issue/wearekickback/app/201/3765 that will create a channel with vivek and the contributor.

octavioamu: maybe we need some preconfiguration on this so we can setup the users from the bounty side that will make part of that channel? bounty_channel_sponsors_usernames=[]

octavioamu: ok as Alex confirmed in that situations they will manage to add manually the funder

app/chat/management/commands/sync_event_bounties_to_chat.py
```diff
@@ -0,0 +1,91 @@
...
+                try:
+                    funder_profile = Profile.objects.get(handle=bounty.bounty_owner_github_username)
+
+                except Exception as e:
+                    print("here")
+                    print(str(e))
+                    continue
+
+                if funder_profile is not None:
+                    if funder_profile.chat_id is None:
```
octavioamu: if not funder_profile.chat_id

PR #5807, opened 2020-01-15
feat: tweak grant fund settings

thelostone-mc:
##### Description

- grant team members cannot fund their own grant
- fund button is made hidden for grant owner + team members
- added new field link_to_new_grant which is set when grant
  is migrated
- if grant has link_to_new_grant -> fund button is replaced
  with button to link to new grant.
- funding is blocked on old grants

##### Steps

- update migrate_grant.py to with the old and new grant id + run it
- the old grant has a f-key link to the new grant
- this can be seen on admin in read only mode
- the old grant fund page would be blocked and instead users will be asked to visit the newer grant

##### Testing

Screenshot 2020-01-15 at 7 23 06 PM

Screenshot 2020-01-15 at 7 23 14 PM

Screenshot 2020-01-15 at 7 23 42 PM

Screenshot 2020-01-15 at 7 24 04 PM

Screenshot 2020-01-15 at 7 24 22 PM


thelostone-mc:
@owocki could you check out the `migrate_grant.py` file once

app/grants/admin.py
```diff
@@ -55,8 +56,9 @@ class GrantAdmin(GeneralAdmin):
         'logo_svg_asset', 'logo_asset',
         'team_member_list',
         'subscriptions_links', 'contributions_links', 'link',
+        'migrated_to'
     ]
-    list_display =['pk', 'title', 'active','grant_type', 'link', 'hidden']
+    list_display =['pk', 'title', 'active','grant_type', 'link', 'hidden', 'migrated_to']
     raw_id_fields = ['admin_profile']
```
owocki: should we add to raw_id_fields? i honestly just like the raw id field explorer more than the regular FK explorer

app/grants/views.py
```diff
@@ -568,6 +568,33 @@ def grant_fund(request, grant_id, grant_slug):
         }
         return TemplateResponse(request, 'grants/shared/error.html', params)
 
+    is_team_member = False
+    if grant.admin_profile == profile:
```
owocki: this would be good to put in a helper function so we can easily find out is_team_member in other views

thelostone-mc:
image

@owocki updated it

app/grants/views.py
thelostone-mc: donee

app/grants/admin.py
thelostone-mc: You mean it'd end up like this -> I didn't want this to be edited via admin cause I wanted to ensure this is done via migrate_grant.py script file 


right now -> the `migrated_to` displayed in table and within the grant admin page is a read only mode 

Screenshot 2020-01-16 at 2 39 27 PM

app/grants/views.py
```diff
@@ -585,6 +585,33 @@ def grant_fund(request, grant_id, grant_slug):
...
+            'title': _('Invalid Grant Subscription'),
+            'grant': grant,
+            'text': _('You cannot fund your own Grant.')
+        }
+        return TemplateResponse(request, 'grants/shared/error.html', params)
+
+    if grant.link_to_new_grant:
+        params = {
+            'active': 'grant_error',
+            'title': _('This Grant has been shifted'),
```
danlipert: migrated instead of shifted?

PR #5828, opened 2020-01-17
fixes kudos foreign keys

owocki:
##### Description

fixes kudos foreign keys, which were called 'kudos' but were actually kudostransfers
- fixes the activity feed such that the kudos are actually linked correctly in the activity feed items
- fixes the kudos details page so that you can see a newsfeed of them being funded on the kudos details

##### Refers/Fixes

twitter DM from a gitcoin user

##### Testing

tested on my local


owocki:
just fixed merge conflicts

PR #5863, opened 2020-01-22
basic edgerank implementation

danlipert:
##### Description

This PR adds a basic algorithm to identify trending activities on townsquare

##### Refers/Fixes

n/a

##### Testing

Tested locally but still a WIP


danlipert:
Screen Shot 2020-01-22 at 9 46 20 PM

app/townsquare/management/commands/calculate_activity_scores.py
```diff
@@ -0,0 +1,16 @@
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
```
owocki: should we schedule this to run every 15 mins in crontab?

owocki:
LGTM; i think we should maybe switch over the `trending=` flag to use this instead of view counts. https://bits.owocki.com/2Nur8o4A

app/townsquare/management/commands/calculate_activity_scores.py
danlipert: yup definitely!

app/townsquare/views.py
```diff
@@ -345,3 +361,12 @@ def offer_new(request):
         'nav': 'home',
     }
     return TemplateResponse(request, 'townsquare/new.html', context)
+
+def activity_score(activity):
+    if activity.activity_type in ACTIVITY_TYPE_SCORES:
+        base_score = settings.EDGERANK_SETTINGS['activity_weights'][ACTIVITY_TYPE_SCORES[activity.activity_type]]
```
thelostone-mc: So even -ve activities like stop work etc would recieve a +ve score ?

app/townsquare/management/commands/calculate_activity_scores.py
```diff
@@ -0,0 +1,16 @@
...
+from dashboard.models import Activity
+from townsquare.views import activity_score
+
+class Command(BaseCommand):
+
+        def handle(self, *args, **options):
+            recent_activities = Activity.objects.filter(created__gt=timezone.now() - timezone.timedelta(hours=72))
```
owocki: In [5]: Activity.objects.filter(created__gt=timezone.now() - timezone.timedelta(hours=72)).count()
Out[5]: 2149

about 2k scores to calculate every 72 h.. (for now)... well probably see a 10x increase during next grants round

app/townsquare/views.py
octavioamu: I think yes, even being something bad if people is doing to much to get tending, we will then need to understand what is happening

thelostone-mc:
Closing this out but keeping branch alive until reprioritized

PR #5891, opened 2020-01-27
copy old fulfillments data from old bounty to new instead of recreate…

danlipert:
##### Description

This PR fixes a bug where fulfillments data is not copied over from an old bounty because the window is closed when work is submitted or a bounty is updated

##### Refers/Fixes

reported in slack

##### Testing

untested

PR #5892, opened 2020-01-27
fix for missing is_bounties_network value

danlipert:
##### Description

Fixes a sentry error when is_bounties_network is None in bounties detail

##### Refers/Fixes
n/a

##### Testing

Tested locally


thelostone-mc:
Pinged on slack ! This should be handled by default 
https://github.com/gitcoinco/web/pull/5891/commits/e0d3b51169a6e8a9380388467ae7fed57631e779 fixes it

PR #5916, opened 2020-01-31
Kevins vision for building up a huge social graph on Gitcoin via tribes

owocki:
1. Shows the follow button on each and every profile in a standard spot
2. Shows tribes stats on the follow button
3. creates a script/post save hooks that auto-create tribes follow relationships when a new earning is created
4. allows users to opt out of creation of relationships this way.
5. names the 'join tribe' button to 'follow' which is something that people understand more intituiveily  already .

Screen Recording 2020-01-31 at 01 50 PM


octavioamu:
I like this, I also suggested the change to "follow" when we built it but I don't remember what were the arguments against that.

owocki:
@octavioamu talked to alisa about it .. sounds like shes good with th ename change.. 

also i just made the fixes u requested

PR #5924, opened 2020-02-03
Distribute GeoIP files in dist/ because they are not downloadable anymore

vporton:
…more.


danlipert:
@vporton nice! I wanted to pin the date since the changing dates tend to break tests but looks like they aren't retaining the older versions so this works too :) thanks!

PR #5931, opened 2020-02-03
Board matching

octavioamu:
##### Description

Todo fix join button
Show only tribes I didn't join yet


thelostone-mc:
@octavioamu Is this something that is needed ?

Issue #5921, opened 2020-02-03
Banners are not being loaded on the live site


walidmujahid:
Since around the 31st of January, I have been unable to update my profile's header image. It would seem that `/api/v0.1/banners` is responding with: 

`{"active": "error", "code": 500, "nav": "home", "title": "Error 500"}`

I am recieving the alert: `Failed to load list of banner images. Try refreshing your browser`

![error500](https://user-images.githubusercontent.com/1489224/73622717-3a4c8f00-4608-11ea-9f66-cda37a346bce.png)

----------------
<sup><b>Side Note:</b> I had attempted to debug this and see if I can track the issue to a specific commit, however I am unable to build the containers properly due to: #5920 and I do not wish to manually setup Gitcoin on my machine. Once, or <b>if</b>, I am able to properly debug this, I will update this issue along with making a pull request.</sup>

owocki:
> [Errno 2] No such file or directory: '/home/ubuntu/gitcoin/coin/app/static/wallpapers'

@danlipert i think maybe the new API server needs this dir created... maybe we should add the creation of this dir to the deploy script so we dont need to remember to do it anymore?

owocki:
@walidmujahid try now! i just created that dir on the new server

danlipert:
@owocki good thinking - one less thing to remember! https://github.com/gitcoinco/web/pull/5932/files

walidmujahid:
@owocki That solved the problem.

PR #5932, opened 2020-02-03
create wallpaper dir if doesnt exist

danlipert:
##### Description

This PR creates a directory to store the profile wallpaper if it doesn't exist - this will happen on a newly deployed server so now one less thing to keep track of when putting up a new box

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5921

##### Testing

untested but non-destructive

### DIFF
--- a/scripts/deploy.bash
+++ b/scripts/deploy.bash
@@ -77,6 +77,8 @@ if [ "$CREATE_CACHE_TABLE" ] && [ "$JOBS_NODE" ]; then
     python3 manage.py createcachetable
 fi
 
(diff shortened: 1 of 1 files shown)


owocki:
one thing i didnt realize from the other thread.. the wallpaper dir needs to copy the wallpapers over from /assets/wallpapers too (or collectstatic should run).. otherwise the API call will not return anything

danlipert:
@owocki thanks for the heads up - i moved the directory creation call to before collectstatic getting ran

owocki:
note that collectstatic doesnt run on all servers tho :P typically just one server..

PR #5640, opened 2019-12-15
Integrate 3Box and Gitcoin: Backup Gitcoin Profiles to 3Box [Gitcoin Global Communities Hackathon]

think-in-universe:
### Description

According to the requirements in [the GitCoin issue](https://gitcoin.co/issue/3box/3box/839/3793), we implemented 3Box integration for GitCoin:

- [x] Integrate 3Box.js into Gitcoin's frontend
- [x] UI for backing up my Gitcoin data to 3Box, inside of a "Gitcoin" space. This likely happens on a button from the "My Profile" page
- [x] When clicked, the user should be prompted to approve the messages inside their wallet (openBox() and openSpace() methods via 3Box.js). This logs them in to 3Box.
- [x] You will need to decide which data should be saved publicly, which should be saved privately (encrypted), and the keys to use for the key-value store.
- [x] Logic for updating a user's backup with new information. User can choose to update manually or automatically. 

##### Design Principles

1. Simplistic and seamless user experience for data synchronization.
1. Privacy and security: keep sensitive data safe at both client and 3Box cloud
1. Data interoperability: open / public data should be provided with context for external applications to utilize the datasets easily

##### Design Details

1. **User Experience**: Here we provide one-button click to get all the backup process done, and save the most meaningful data to user's 3Box Hub spaces. We may provide user more control over their data at different granularity, such as which fields to be exported, public or private, etc. 
1. **Data Model**: Here we used a pre-defined data models for the user profiles and contribution histories from the users, including the account info, statistics, feedbacks, grants, portfolio, active work, bounties, and activities, which are the major datasets that an external application could refer to make the best use of the value of the data. 
1. **Data Storage**: All the profiles data will be saved in a single "GitCoin" space, with public and private key-value for different fields. 
1. **Data Privacy**: When exporting data from the data source (or ETL process), we need to take care of both private fields and private items when mapping them into the GitCoin spaces key-value data structure. For a typical list of items, one solution is to divide them into 3 sets {public items with public fields, public items with private fields, private items) and saved separately with different keys. For example, private "feedback" items will be placed in a key-value pair called "feedbacks.private_items", while private "tip" fields will be saved in a key-value pair called "tips.private_fields"
1. **Automation**: Automatic update of profile data should be enabled or disabled based on the user scenario. We plan to also support the data synchronization at server side, in the case when the user would like to view data daily or at a fixed frequency. The client-driven data backup is not sufficient for certain use cases. 

##### Data Models

[...]

PR #5947, opened 2020-02-05
3box integration

thelostone-mc:
##### Description

Moved over from https://github.com/gitcoinco/web/pull/5640


owocki:
did testing of this fail?? cc @think-in-universe @thelostone-mc

think-in-universe:
@owocki yes. we met some testing failure in the production. We tested with 3Box team with some mock data in dev environment last time, so some cases are not well tested. Let's fix here. 

@thelostone-mc I have created an PR https://github.com/gitcoinco/web/pull/5955 to fix the issue introduced in #5947. But I failed to create mock bounties data in my local dev environment after tried for quite some times, and could you please help test before you merge the fix?

PR #5964, opened 2020-02-08
graph visualization of econoimc activity as first class part of the app

owocki:
##### Description

allows an admin to create 3d graph networks of economic activity in the admin. 
Screen Recording 2020-02-08 at 04 07 PM

part of the idea is that this could be useful for understanding the app (internal) and for marketing reasons (external).  not just kevin should be able to create these..

##### Refers/Fixes

self; marketing material prep

##### Testing

tested locally


androolloyd:
thats pretty dope.

owocki:
https://gitcoin.co/_administration/mesh

Issue #5865, opened 2020-01-22
Add warning for Brave users


PaulRBerg:
In v1, Brave introduced their native crypto wallet, which is rather intrusive and overrides Metamask.

Some users told me they would've loved to contribute to our [Sablier](https://gitcoin.co/grants/253/sablier) grant, but they couldn't because nothing happened after tapping the submit button. I jumped on TeamViewer to help them and it turned out that the issue was Brave - after disabling it, it went fine.

It'd be nice to display a warning for Brave users. Related: https://github.com/brave/brave-browser/issues/6053

LefterisJP:
Yeah I saw the same in the previous gitcoin round. That's why I disabled Brave's own crypto wallet.

owocki:
adding a bounty for thsi. scope is to fix grant contributions (and cancellations) in brave's native wallet

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.5 ETH (83.55 USD @ $167.1/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/5865/3922).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/5865/3922).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $75,101.84 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [iamonuwa](https://gitcoin.co/iamonuwa) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/5865/3922?mutate_worker_action=approve&worker=iamonuwa) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/5865/3922?mutate_worker_action=reject&worker=iamonuwa))_.**

Check if browser is brave browser . Then show message

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/5865/3922).

[...]

Issue #1052, opened 2018-05-01
as a user, i want to be able to save my avatar to be my twitter and github profiles


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a user, i want to be able to save my avatar to be my gitcoin profile

follow up to https://github.com/gitcoinco/web/issues/813 and as a user, i want to be able to save my avatar to be my twitter / github profile

### Why Is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
so that users can save their avatar to their profiles and show off their bad selves (and market gitcoin)

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: feature

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
- Introduce new backend python methods to:
  - [x] Output avatar SVG as PNG
  - [ ] Impose the Gitcoin watermark (asset provided) on any given svg or png asset
  - [ ] Request permissions from Twitter as the current user that allows us to update their profile picture with the avatar PNG
  - [ ] Upload PNG avatar asset directly from s3 storage backend to twitter
  - [ ] View to download the avatar png and provide instructions on manually updating the Github profile picture
- avatar is generated, with a watermark that says 'gitcoin' along the bottom, 
- avatar can be easily saved as a twitter/github profile picture (ideally in /settings, but if no API is available to do this.. then perhaps with instructions to save the image and upload it to those networks)

owocki:
@pixelantdesign any objetions if i bounty this?

PixelantDesign:
Sounds good. Is this a build task?

I have a watermarked avatar that I was testing, do you think we need Made with Gitcoin on it?

**Option 1**
![alisa_icons](https://user-images.githubusercontent.com/238235/41985600-a1c9be7a-79e8-11e8-9771-2f212d32a46e.png)

**Option 2**
![alisa_icon](https://user-images.githubusercontent.com/238235/41985610-aaf65e72-79e8-11e8-82ed-e15cf4b045af.png)

**Option 3**
<img width="467" alt="screen shot 2018-06-27 at 9 07 09 am" src="https://user-images.githubusercontent.com/238235/41985936-871f2d52-79e9-11e8-98a1-bd75d0707ba0.png">

**Option 4**
[...]

PR #5988, opened 2020-02-14
Add warning for users visiting Grants page

sebastiantf:
##### Description

Adds a warning for users about using Brave Wallet and MetaMask while visiting `https://gitcoin.co/grants`

grantswarning

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/5865

##### Testing

No issues on eslint and stylelint.

### DIFF
--- a/app/assets/v2/js/grants/fund.js
+++ b/app/assets/v2/js/grants/fund.js
@@ -12,6 +12,8 @@ document.suppress_faucet_solicitation = 1;
 
 $(document).ready(function() {
 
+  _alert({ message: gettext('Note: Brave users seem to have issues while contributing to Grants while using both Brave Wallet and MetaMask. We recommend disabling one. For more info, see this <a target="_blank" href="https://github.com/brave/brave-browser/issues/6053">issue</a>') }, 'warning');
+
   predictPhantomCLRMatch();
   predictCLRMatch();
 
--- a/app/assets/v2/js/grants/new.js
+++ b/app/assets/v2/js/grants/new.js
@@ -5,6 +5,9 @@ let description = new Quill('#input-description', {
 });
 
 $(document).ready(function() {
+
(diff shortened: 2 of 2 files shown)


thelostone-mc:
@sebastiantf 

This would probably be most effective on the /fund page when users are about to contribute to grant and the grant creation page (/grants/new)

What do you think on restricting the alert to just those pages?
Also If we can figure out if user is using brave and only show it then -> it's super effective 

a quick google led me to this 
https://stackoverflow.com/questions/36523448/how-do-i-tell-if-a-user-is-using-brave-as-their-browser

sebastiantf:
@thelostone-mc Showing the warning on `/grants/new` seems like a good idea. That can be done.

But as I mentioned on the [issue](https://github.com/gitcoinco/web/issues/5865) page, Brave browser doesn't seem to provide a Brave-specific `User Agent` which makes detecting Brave browser difficult. See [#1052](https://github.com/brave/brave-browser/issues/1052)

Since Brave Wallet is a fork of MetaMask, the `web3.currentProvider.isMetamask` returns `true`.

So it seems displaying the warning only for Brave users by detecting **Brave Browser** or **Brave Wallet** to be hard.

thelostone-mc:
@sebastiantf let's do it that way !
If I see a way to detect it -> i'll throw that in later

Danke :D

molecula451:
Good job. This was the best suggestion as per Brave removed UA a while ago

sebastiantf:
@thelostone Shall I change the warning from /grants to the other two pages mentioned?

sebastiantf:
@thelostone-mc Please check my new commits and review this PR. I have changed the warning from `/grants` to `/fund` and `/grants/new`

thelostone-mc:
@sebastiantf Looks good ^_^

PR #6014, opened 2020-02-18
add browser notifications flow

octavioamu:
##### Description

Adds flow to enable browser notifications (not push notifications)  from townsquare then a function was created so we can send a notification (if the users is on the site) 
image


thelostone-mc:
check

PR #6045, opened 2020-02-21
As a gitcoin user, I would like to see the responses to my post in the Notifications.

### LINKED ISSUE #5943: Add activity of Town Square to notifications
### User Story

As a gitcoin user, I would like to see the responses to my post in the Notifications.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide a description of the current state, what you would like to happen, and what actually happen)
*Summary*: I would like to get the answers to my post in my notifications without having to surf all the town square or having to look at my e-mail

### Description

*Type*: Feature

### Current Behavior
There is no notification for town square related activity

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Get notifications from someone pinging me in the town square or for responses to my post

zoek1:
##### Description

Get notifications from someone pinging me in the town square or for responses to my post
- I receive a notification when someone pings me on a post
- I receive a notification when someone comments on my posts
- I receive a notification when someone pings me on some comments
##### Refers/Fixes
#5943 

##### Testing

FireShot Capture 001 - Home - Gitcoin - localhost

### DIFF
--- a/app/assets/v2/js/notifications.js
+++ b/app/assets/v2/js/notifications.js
@@ -19,13 +19,9 @@ Vue.mixin({
       $.when(getNotifications).then(function(response) {
         newNotifications = newData(response.data, vm.notifications);
 
-        if (vm.notifications.length === 0) {
-          newNotifications.forEach(function(item) {
-            vm.notifications.push(item);
-          });
-        } else if (newNotifications.length > 0) {
-          vm.notifications = newNotifications.concat(vm.notifications);
-        }
+        newNotifications.forEach(function(item) {
+          vm.notifications.push(item);
+        });
 
         vm.numPages = response.num_pages;
         vm.hasNext = response.has_next;
--- a/app/assets/v2/js/notifications.js
+++ b/app/assets/v2/js/notifications.js
@@ -18,9 +18,14 @@ Vue.mixin({
 
       $.when(getNotifications).then(function(response) {
         newNotifications = newData(response.data, vm.notifications);
-        newNotifications.forEach(function(item) {
-          vm.notifications.push(item);
-        });
+
+        if (vm.notifications.length === 0) {
+          newNotifications.forEach(function(item) {
+            vm.notifications.push(item);
+          });
+        } else if (newNotifications.length > 0) {
+          vm.notifications = newNotifications.concat(vm.notifications);
+        }
 
         vm.numPages = response.num_pages;
         vm.hasNext = response.has_next;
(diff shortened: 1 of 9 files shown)


app/assets/v2/js/notifications.js
```diff
@@ -18,9 +18,14 @@ Vue.mixin({
...
         newNotifications = newData(response.data, vm.notifications);
-        newNotifications.forEach(function(item) {
-          vm.notifications.push(item);
-        });
+
+        if (vm.notifications.length === 0) {
+          newNotifications.forEach(function(item) {
+            vm.notifications.push(item);
+          });
+        } else if (newNotifications.length > 0) {
```
octavioamu: not sure why this change, this will add the notification to the end
image

app/inbox/models.py
```diff
@@ -55,6 +55,8 @@ class Notification(SuperModel):
         ('new_milestone', 'New Milestone'),
         ('update_milestone', 'Updated Milestone'),
         ('new_kudos', 'New Kudos'),
+        ('new_mention', 'New Mention'),
+        ('new_post_comment', 'New comment'),
```
octavioamu: missing new_like

octavioamu:
Really nice and supper excited with this I have another on going PR I will integrate with this to add the notifications to the browser

app/inbox/models.py
zoek1: Ok, i'll add it!

app/assets/v2/js/notifications.js
zoek1: It's strange. I just tested the changes to avoid this behavior. 😱

octavioamu: That is because if you reload the page then the result will be on the top, but if you click the notifications to open/close, that will fetch  new notifications and then will be attached to the end.

octavioamu: just tested the new changes and now is getting on the bottom, maybe when I tested my js was cached not sure what changed because I don't remember this on the past. Maybe you are right we will need to do concat or unshift

octavioamu: ok, dont worry about this part I will do a hotfix on stable about this order but not here, in the filter part

octavioamu: fixed 68777da012cea06c64105dfa2dea77bfb01bce6e

zoek1: Thanks for the fix @octavioamu 👍

PR #6070, opened 2020-02-24
Make amount, contributors, and match_amount more readable in mini-clr leaderboard

walidmujahid:
Based on: https://gitcoin.co/townsquare?tab=activity:140260

For each ranked user, it current provides the amount, contributors, and match amount in the following format: `amount x contributors = match_amount`

The concern is this reads wrong and I agree. Alternative is: `amount from contributors amount matched`

Example: ` $41.78 from 13 contributors $60.66 matched`

Screenshot:


walidmujahid:
Actually, amount from contributors := match amount be better instead of "for"?

thelostone-mc:
@walidmujahid yo i had tweaked the design on the stable branch ! ported over your code 

image

PR #6076, opened 2020-02-25
Give leaderboard have history charts (branched off master:)

IntoThe1337ness:
##### Description

this PR let leaderboard have history charts

##### Refers/Fixes

there is thread on twitter and townsquare about this. crates competitivness.

##### Testing

i tested local (See Below Comment)

##### Other Notes

we have to give the new cleanup DB script deploy first.  Give data a few Week to AccumuLte.


app/dashboard/management/commands/cleanup_db_space.py
```diff
@@ -56,8 +56,3 @@ def handle(self, *args, **options):
                 created_on__hour=1,
             ).delete()
         print(f'Stat: {result}')
-
-        result = LeaderboardRank.objects.filter(
```
owocki: hmmm m we should stop doing this now.. gonna fix that here d84147cfa3d4245589949d06a6799ba3c99b83fd

Issue #6085, opened 2020-02-26
Duplicate Activity Comments: Initial fix was incomplete


walidmujahid:
The fix introduced in PR https://github.com/gitcoinco/web/pull/6008 is incomplete.

The issue of duplicate comments, as intially described in https://github.com/gitcoinco/web/issues/6007 still exists for the `.post_comment`  button itself. 

This can be fixed by diasbling `.post_comment` button for every call of `post_comment` function.

PR #6086, opened 2020-02-26
fix: disable comment button on post

walidmujahid:
Disable the comment button when posting to avoid duplicate comments.

closes https://github.com/gitcoinco/web/issues/6085


thelostone-mc:
nice catch 🙌

PR #6088, opened 2020-02-27
put unread counts on the nav menu of town square

owocki:
##### Description

this PR puts put unread counts on the left hand side of the townsquare nav.

prevoiusly the badge showed how many threads in the last 24 hours.. which is quite random and arbitrary.  specific unread counts on a per user basis are a better way to cue ot the users which tabs they need to checkout

Screen Shot 2020-02-26 at 7 36 00 PM

when you click through each tab, it clears the badge on that tab.  quite satisfying!!!

the way it works is by storing the top ten activities of each type in the DB.  then, for each type, when you load the townsquare, we compare the latest viewed pk versus the ones in the DB.

##### Refers/Fixes

common sense

##### Testing

tested locally.  
- all tabs work
- this adds only 0.05s of performance overhead


owocki:
can i has  code review @danlipert @gitcoinco/engineers

app/perftools/management/commands/create_activity_cache.py
```diff
@@ -0,0 +1,57 @@
...
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from dashboard.models import Activity, HackathonEvent
+from perftools.models import JSONStore
+
+
+def create_activity_cache():
+
+    print('activity')
```
octavioamu: remove print

octavioamu:
@owocki seems this one got 4 conflicts in just one file 🤕

PR #6103, opened 2020-03-01
remove position-relative on clr_container 

developerfred:
### Before
Screen Shot 2020-03-01 at 02 26 55

### After
Screen Shot 2020-03-01 at 02 29 25


thelostone-mc:
Could we instead just reduce the spacing btwn the avatar and the text on either side ?
image

developerfred:
@thelostone-mc  I will make changes to see how it looks, I found a great idea.

PR #6108, opened 2020-03-02
Chat/1.2.2 - Metrics, Hackathons, Notifications.

androolloyd:
##### Description

Chat/1.2.2 - Metrics and Hackathon updates

##### Refers/Fixes

- Containerized Chat Added to Docker-compose
- Chat Data fixture to create an oauth2 application for the chat application
- Chat metrics added to stats
- Email Suppression Settings respected
- Chat Access Token added to context.py
  - notification settings long poll 30s
- Hackathon Setup Script
  - run as a cron job, will query active Hackathons and ensure that any of the channels needed are created and will add in all the users who are apart of that Hackathon to the sponsor channels as well as the default channels.

### Production env variable required
`CHAT_SERVER_URL`, can be identical to the CHAT_URL given the current setup, the difference is to allow for a different URL should it be needed to connect to the mattermost API from server server, as is the case in the docker network, where it connects via named servers, `web`, `db`, `chat` etc


app/app/fixtures/oauth_application.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+  "model": "oauth2_provider.application",
+  "pk": 1,
+  "fields": {
+    "created": "2020-02-29 21:30:30Z",
+    "updated": "2020-02-29 21:30:30Z",
+    "client_id": "VkHFfFlxnVCMYtYCWRxphbvPQUJjmER4cWtwqe7Q",
+    "client_secret": "xhLXNqXnOlkUiGCDFJLxQJjPb3CltLZNFZtzOsLgFfvwbYY60kYWRRBmqyjzfrnNyT6WWdIbfQEPpU854K2UiofmXtybFD1QkfIR9VVQyK3ttJpET7GormIBjK9CQRap",
```
octavioamu: are this keys fake?

app/assets/v2/js/base.js
```diff
@@ -439,11 +439,38 @@ const gitcoinUpdates = () => {
 
 };
 
+
+if (document.contxt.chat_access_token && document.contxt.chat_id) {
+  // setup polling check for any updated data
+  // scope our polling function so any potential js crashes won't affect it.
+  (($) => {
+    setInterval(() => {
```
octavioamu: not sure we need to do this as we are not an SPA maybe just requesting on load is enough.

app/app/fixtures/oauth_application.json
androolloyd: Yes are installed with fixtures so it can connect out of the box.

app/assets/v2/js/base.js
androolloyd: Was asked to remove it from requesting on load.

octavioamu: humm why? not sure that make sense since now is being onload and each X seconds

androolloyd: we don't check on load anymore in the context.py, we just setup this poll when the page loads.

androolloyd: We could remove the interval but than if someone is sitting on the site for any period of time and is not navigating, like on town square or something, then they won't get a dot until they navigate if we don't poll.

app/app/context.py
```diff
@@ -108,35 +121,22 @@ def preprocess(request):
...
+        chat_access_token = profile.gitcoin_chat_access_token
+        chat_id = profile.chat_id
     # handles marketing callbacks
     if request.GET.get('cb'):
         callback = request.GET.get('cb')
         handle_marketing_callback(callback, request)
 
-    chat_unread_messages = False
-
-    if profile and hasattr(profile, 'chat_id'):
```
owocki: hey @androolloyd does the new chat system replace this functionality somewhere?

should i see a notification somewhere? https://bits.owocki.com/4gumPZgN

app/app/utils.py
```diff
@@ -458,6 +459,16 @@ def get_profile(request):
     if is_authed and not profile:
         profile = sync_profile(request.user.username, request.user, hide_profile=False)
 
+    if profile is not None and (profile.chat_id is '' or profile.gitcoin_chat_access_token is ''):
+
+        try:
+            from chat.tasks import associate_chat_to_profile
+            created, profile = associate_chat_to_profile(profile)
```
owocki: shouldnt this be `associate_chat_to_profile.delay(profile)`? its gonna increase the loadtime every time we call get_profile (which is everywhere)

app/app/context.py
owocki: i see the changes in base.js but its not working in prod rn

app/app/settings.py
```diff
@@ -568,10 +568,11 @@
 # Chat
 CHAT_PORT = env('CHAT_PORT', default=8065)  # port of where mattermost is hosted
 CHAT_URL = env('CHAT_URL', default='localhost')  # location of where mattermost is hosted
+CHAT_SERVER_URL = env('CHAT_SERVER_URL', default='chat')  # location of where mattermost is hosted
```
owocki: dan says

@Owocki @thelostone-mc @octavioamu I'm going to merge @androolloyd’s chat PR into stable - it requires additional ENV configs to run locally: CHAT_URL=localhost and CHAT_SERVER_URL=chat

should we add these to local.env?

app/app/utils.py
androolloyd: It should only do it if you don't have a chat_id or a chat_access_token yet, and its not a delayed function because we want to return the profile back to the scope updated.

app/app/settings.py
androolloyd: great idea.

app/app/utils.py
owocki: hmm for some reason when i run this on my profile in prod i get this

> HTTPError: 500 Server Error: Internal Server Error for url: https://chat.gitcoin.co:443/api/v4/users/mqe31q5oo3bctd9qzqgtdco1kw/tokens

so as it stands this will run every time i run `get_profile()` as i never successfully get a chat access token

app/assets/v2/js/base.js
androolloyd: refactoring this to check for new notifications on page load and to setup an interval

PR #6124, opened 2020-03-03
merge stable => master

owocki:
merge stable => master


owocki:
the only thing i didnt merge was 'my feed' mostly because i dont know what it is or how it compares to my threads

PR #6141, opened 2020-03-05
feat: encrypt address

### LINKED ISSUE #6090: Address isn't 100% private
**Describe the bug**

Even if a user set the `Hide my wallet address` option to `true` under the profile privacy settings, the address can be easily unveiled through a GET request

**To Reproduce**

You can get any user ETH address by doing a GET request to `https://gitcoin.co/tip/address/<user>`

**Expected behavior**

Remove the `/tip/address/<user>` endpoint and change the logic for sending tips on `/app/assets/onepager/js/send.js` so user address lookup is performed on the backend when sending a POST request

developerfred:
##### Description
frontend receives the encrypted address, and when the payment is sent, we decrypted address before sending

##### Refers/Fixes
#6090


danlipert:
@developerfred I'm not sure I understand what this PR does - also, @owocki left a comment on the original issue with the desired solution, so please check that out and if you disagree let us know why your solution is better - thanks! Also dont forget to fill out the testing section

Issue #6132, opened 2020-03-04
status update, attach youtube video doesnt work anymore.


owocki:
**Describe the bug**

status update, attach youtube video doesnt work anymore.

**To Reproduce**

put a youtube video in status update form

**Expected behavior**

video preview should show before posting, andn then the video should display inline after posting

owocki:
looks like the API key is over quote

we get this results:

```
{
 "error": {
  "errors": [
   {
    "domain": "usageLimits",
    "reason": "dailyLimitExceeded",
    "message": "Daily Limit Exceeded. The quota will be reset at midnight Pacific Time (PT). You may monitor your quota usage and adjust limits in the API Console: https://console.developers.google.com/apis/api/youtube.googleapis.com/quotas?project=1064792076301",
    "extendedHelp": "https://console.developers.google.com/apis/api/youtube.googleapis.com/quotas?project=1064792076301"
   }
  ],
  "code": 403,
  "message": "Daily Limit Exceeded. The quota will be reset at midnight Pacific Time (PT). You may monitor your quota usage and adjust limits in the API Console: https://console.developers.google.com/apis/api/youtube.googleapis.com/quotas?project=1064792076301"
 }
}
```

owocki:
@zoek1 any ideas of what to do here? is the API key in the code one u registered for us?  maybe its worth putting the key into settings.py (where we keep our secrets).

i wonder if someone has stolen the key out of our codebase and is using it for their own reasons..

zoek1:
I checked the dashboard but theres no strange behavior. For the moment i límited the usage to gitcoin.co domain. I'll make the PR to move this key to settings. @owocki

zoek1:
@owocki the key is different from what i provided. @octavioamu could you check the api usage to see if someone stolen the key? One important thing is verify if the key is restricted to gitcoin domain: 
![image](https://user-images.githubusercontent.com/660973/75958322-93b31280-5e81-11ea-8ec2-e75da0f808d4.png)

owocki:
@octavioamu do u have the key dashboard? can u share with me?

PR #6146, opened 2020-03-05
Fix attach youtube video

zoek1:
##### Description
The youtube api key was replaced with the `YOUTUBE_KEY` variable in `settings.py`.

##### Refers/Fixes

#6132


octavioamu:
ohh we totally forgot about that one. those are created on the google console right?

zoek1:
Yep, in the Google console :thumbsup:

PR #6153, opened 2020-03-05
bug: fix up grant category

thelostone-mc:
##### Description

`/grants/new`

- [x] select multiple categories
- [x] update categories based on grant type

`/grants/`

- [x] show right catergories under grant type
- [x] ensure filters work as expected with grant type & category
- [x] ensure grant category appears on the grant card at the bottom

`/grants/<:grant-id>`

- [x] grant categories appears as read only tags for non team members
- [x] team members have editable & non editable view of category (works like team section) 
- [x] team member can add new categories

##### Demo

https://share.vidyard.com/watch/3WXLB5bzS9Dc4NGFyWLKwc

Refs: https://github.com/gitcoinco/web/pull/5615

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -94,7 +94,7 @@ def grants(request):
     keyword = request.GET.get('keyword', '')
     grant_type = request.GET.get('type', 'tech')
     state = request.GET.get('state', 'active')
-    category = request.GET.get('category', '')
+    category = request.GET.get('category')
     _grants = None
 
     show_past_clr = False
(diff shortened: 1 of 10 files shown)


app/grants/views.py
```diff
@@ -103,7 +103,6 @@ def grants(request):
     sort_by_clr_pledge_matching_amount = None
     if 'match_pledge_amount_' in sort:
         sort_by_clr_pledge_matching_amount = int(sort.split('amount_')[1])
-        sort_by = 'pk'
```
thelostone-mc: this was removed cause it was not used anywhere

PR #6187, opened 2020-03-09
Fix #5923

### LINKED ISSUE #5923: Website is slow and sluggish in Firefox
**Describe the bug**

In Firefox everything is extremely slow. Buttons are not clickable. Context menu does not appear. To click a button I have to close the support window and wait like 20 seconds. Network console shows hammering on Infura servers. Although it is not clear what exactly causes the slowness.

**To Reproduce**
Disable Metamask and load https://gitcoin.co/issue/alialiwa2005/Maple/1/3976 in Firefox.

- OS: Fedora 31
- Browser: Firefox
- Browser Version: 72.0.1

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

sebastiantf:
Fixes #5923

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -1025,7 +1025,7 @@ attach_change_element_type();
 
 if (typeof is_bounties_network == 'undefined' || is_bounties_network) {
   window.addEventListener('load', function() {
-    setInterval(listen_for_web3_changes, 1000);
+    setInterval(listen_for_web3_changes, 5000);
   });
 }
 


owocki:
note from the other thread :

There is great improvement in the performance. You could identify the drop in performance if you try scrolling the page up and down continuously. The scrolling of the site would hang for a second whenever the listen_for_web3_changes fires. When the delay is 1s, the site hangs every second, making it unusable. When it is increased to 5s, the site would hang only in 5-second intervals which is far better that hanging every second.

If it's fine, we could further increase the delay which would further improve the performance.

thelostone-mc:
@sebastiantf so we've updated this to ensure it loads only on pages which need metamask interaction ! Else it's removed 
Gonna close this out ! Moving this to 5000 is a nice way to fix it but when I tried it a year back , users complained that their address change wasn't being detected (cause it was happening a few seconds later and not immediately )

sebastiantf:
Ok @thelostone-mc. Good to see the new solution.

PR #6213, opened 2020-03-11
New bounty detail

octavioamu:


octavioamu:
Good catch @owocki

thelostone-mc:
Closing this out cause we'd be moving away from std bounties so best to keep both separate views

PR #6238, opened 2020-03-16
Handle Prod Timeout Worker Approve Timeout Errors

owocki:
##### Description

Uses a [solution found here](https://stackoverflow.com/questions/492519/timeout-on-a-function-call) to prevent production timeout errors.

adds a re-usable decorator for managing those timeouts

##### Refers/Fixes

Production timeout errors on approving users
Screen Shot 2020-03-16 at 11 15 13 AM

##### Testing

```
from app.utils import timeout
import time

@timeout(2)
def foo():
    time.sleep(10)

foo()
```

will yield

```
---------------------------------------------------------------------------
TimeoutError                              Traceback (most recent call last)
<ipython-input-1-f4b9f2b0c14b> in <module>
      6     time.sleep(10)
      7
----> 8 foo()

/code/app/app/utils.py in func_wrapper(*args, **kwargs)
    508             async_result = pool.apply_async(item, args, kwargs)
    509             # raises a TimeoutError if execution exceeds max_timeout
--> 510             return async_result.get(max_timeout)
    511
    512         return func_wrapper

/usr/lib/python3.6/multiprocessing/pool.py in get(self, timeout)
    638         self.wait(timeout)
    639         if not self.ready():
--> 640             raise TimeoutError
    641         if self._success:
    642             return self._value

TimeoutError:

In [2]:
```

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -1776,10 +1776,7 @@ def helper_handle_approvals(request, bounty):
 
                 maybe_market_to_github(bounty, 'work_started', profile_pairs=bounty.profile_pairs)
                 maybe_market_to_slack(bounty, 'worker_approved')
-                try:
-                    maybe_market_to_user_slack(bounty, 'worker_approved')
-                except TimeoutError:
-                    pass
+                maybe_market_to_user_slack(bounty, 'worker_approved')
                 record_bounty_activity(bounty, request.user, 'worker_approved', interest)
             else:
                 start_work_rejected(interest, bounty)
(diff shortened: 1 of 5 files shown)


app/dashboard/views.py
```diff
@@ -1776,7 +1776,10 @@ def helper_handle_approvals(request, bounty):
 
                 maybe_market_to_github(bounty, 'work_started', profile_pairs=bounty.profile_pairs)
                 maybe_market_to_slack(bounty, 'worker_approved')
-                maybe_market_to_user_slack(bounty, 'worker_approved')
+                try:
```
owocki: longer term it may be a good idea to move these jobs to celery instead of pageload

owocki:
cc @danlipert for an async review. merging this bc its prod and ppl cant approve certain workers

PR #6240, opened 2020-03-16
fix etc payout flow

octavioamu:
##### Description

change pending by default after payout
close modal
refetch bounty data.

### DIFF
--- a/app/assets/v2/js/pages/bounty_details2.js
+++ b/app/assets/v2/js/pages/bounty_details2.js
@@ -169,28 +169,32 @@ Vue.mixin({
         });
       }
     },
-    fulfillmentComplete: function(fulfillment_id, amount) {
-
+    fulfillmentComplete: function(fulfillment_id, amount, event) {
       let vm = this;
-
       const token_name = vm.bounty.token_name;
       const decimals = tokenNameToDetails('mainnet', token_name).decimals;
-
       const payload = {
         amount: amount * 10 ** decimals,
         token_name: token_name,
         bounty_owner_address: vm.bounty.bounty_owner_address
       };
-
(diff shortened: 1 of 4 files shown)


app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -169,28 +169,32 @@ Vue.mixin({
...
-
       const payload = {
         amount: amount * 10 ** decimals,
         token_name: token_name,
         bounty_owner_address: vm.bounty.bounty_owner_address
       };
-
       const apiUrlBounty = `/api/v1/bounty/payout/${fulfillment_id}`;
 
+      event.target.disabled = true;
```
thelostone-mc: what's the event.target for ? to disable the button ?

app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -169,28 +169,32 @@ Vue.mixin({
...
+
       fetchData(apiUrlBounty, 'POST', payload).then(response => {
+        event.target.disabled = false;
         if (200 <= response.status && response.status <= 204) {
           console.log('success', response);
+          vm.fetchBounty();
```
thelostone-mc: do we need to fetchBounty ? wouldn't updating the state suffice ?

octavioamu: is good to check if something else changed in that same time (ex other users submitting) also because to change the UI in the activity I will need to go to that exact fulfillment and do the modifications (is a viable option)

app/assets/v2/js/pages/bounty_details2.js
octavioamu: yes

PR #6242, opened 2020-03-16
Fixes the remaining perf issues on the site.

owocki:
This PR fixes the remaining perf issues on the site.

### search results now have JOINs removed + indexes added

hide_profile add db_index
make data_type into first class Profile attribute + add index
make follower_count into first class Profile attribute + add index
make avg_feedback into first class Profile attribute + add index
make fulfilled_count into first class Profile attribute + add index
make previous_worked into its own table (maybe re-use earnings)
put chat presence in redis

##### Refers/Fixes
ongoing production issues

##### Testing

tested locally

tested by migrating your local with

```
In [3]: for pk in Profile.objects.values_list('pk', flat=True):
   ...:     print(pk)
   ...:     profile = Profile.objects.get(pk=pk)
   ...:     profile.save()
   ...:
   ...:
```

#### Other notes

in prod, i propose that we institute the following perf rulers
0. Limit calls that do time-concentrated IOPS :
no more JOINs in user facing queries
no more contains or icontains in production code if we can help it
3. no more loops where each iteration makes a DB call


app/dashboard/models.py
```diff
@@ -2608,6 +2608,19 @@ class Profile(SuperModel):
     automatic_backup = models.BooleanField(default=False, help_text=_('automatic backup profile to cloud storage such as 3Box if the flag is true'))
     as_representation = JSONField(default=dict, blank=True)
     tribe_priority = models.TextField(default='', blank=True, help_text=_('HTML rich description for what tribe priorities.'))
+
+    is_org = models.BooleanField(
```
octavioamu: should we go in this way or just replicate what we have on data from github? https://api.github.com/users/gitcoinco
adding user_type= "Organization"  or "User"

owocki: i figured a boolean was cheaper in terms of storage space.. plus theres only two possible user types

PR #6250, opened 2020-03-18
add a jitsi livestream video to any townsquare post

owocki:
##### Description

add a jitsi livestream video to any townsquare post

##### Refers/Fixes

multiple slack chatters about doing this on gitcoin

##### Testing

Screen Recording 2020-03-18 at 10 35 AM


app/assets/v2/js/activity.js
```diff
@@ -17,6 +17,23 @@ $(document).ready(function() {
...
+    var jitsi_options = {
+      roomName: roomName,
+      parentNode: document.querySelector('#' + roomName),
+      welcomePageEnabled: true,
+      userInfo: {
+        displayName: document.contxt.github_handle ? document.contxt.github_handle : 'Anonymous',
+        email: document.contxt.email ? document.contxt.email : null
+      }
+    };
+    var jitsi_api = new JitsiMeetExternalAPI(jitsi_domain, jitsi_options);
```
thelostone-mc: Do we need this in a variable ?
Also var -> let/const :P

octavioamu: yep need to be in a variable (const works). As it need to run dispose when ends the call and others related methods.

app/assets/v2/js/activity.js
```diff
@@ -17,6 +17,74 @@ $(document).ready(function() {
...
+    api.executeCommand('avatarUrl', avatarURL);
+    api.executeCommand('toggleAudio'); // default off
+    api.executeCommand('toggleVideo'); // default off
+    var participants_count = api.getNumberOfParticipants() + 1
+    var $target = $(this).parents('.activity_detail_content');
+    $target.append("<a href=# class='float-right leave_video_call'>Leave Video Call</a>")
+  });
+
+  //refresh job for live call
+  setInterval(function(){
```
octavioamu: If I get it right this pull will run for everyone on the call, that can be a problem. Can we do the POST just with the room owner?

app/assets/v2/js/activity.js
owocki: OK

app/assets/v2/js/activity.js
owocki: no because the room owner may leave!  

the calls are actually pretty performant becasue the backend is only making a redis write..  no DB writes involved.

ideally one day when jitsi releasees a python API we can query it in the backend for all active calls instead of leveraging the clients JS.

app/assets/v2/js/activity.js
owocki: done

octavioamu: still showing me 'var ' for me

octavioamu: stupid github, seems I can see the changes just here https://github.com/gitcoinco/web/pull/6250/commits/3a9290216b247d34a3c8e73ae4c3931d025d6125

PR #6251, opened 2020-03-18
clr improv

thelostone-mc:

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -31,12 +31,19 @@ from perftools.models import JSONStore
 
 CLR_START_DATE = dt.datetime(2020, 1, 6, 0, 0)
 
+# TODO: MOVE TO DB
+THRESHOLD_TECH = None
+THRESHOLD_MEDIA = None
+THRESHOLD_HEALTH = None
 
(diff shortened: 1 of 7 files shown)


app/grants/clr.py
```diff
@@ -29,9 +29,16 @@
 from grants.models import Contribution, Grant, PhantomFunding
 from perftools.models import JSONStore
 
-LOWER_THRESHOLD = 0.0
 CLR_START_DATE = dt.datetime(2020, 1, 6, 0, 0)
 
+# TODO: MOVE TO DB
+THRESHOLD_TECH = 20.0
```
owocki: should these be dicts instead of a bunch of varaiables, eg

```
matching_pots = [
{
 'slug': 'tech',
 'amount': 100000,
'config': { 'negative_voting': false },
]
```

thelostone-mc: Yup ! We don't need the -ve voting though -> that is something we'd control via the frontend 
Allow -ve contributions only to tech grants

owocki:
heres my PR https://github.com/gitcoinco/web/pull/6263

PR #6268, opened 2020-03-21
chore: final BE wiring for grants

thelostone-mc:
##### Description

- remove random data as the algorithm has changed
- filter contribution as +ve and -ve contributions
- update code to ensure -ve voting is taken into account
- update check to ensure algorithm runs when a grant has no `-ve` voting

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -1061,7 +1061,8 @@ def grant_categories(request):
 def predict_clr_v1(request, grant_id):
     '''
         {
-            amount: <float>
+            amount: <float>,
+            match_direction: <char> +, -
         }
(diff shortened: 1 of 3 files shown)


app/grants/views.py
```diff
@@ -1092,7 +1093,15 @@ def predict_clr_v1(request, grant_id):
         response['message'] = 'error: missing parameter amount'
         return JsonResponse(response)
 
-    predicted_clr_match = predict_clr_live(grant, profile, float(amount))
+
+    match_direction = request.GET.get('match_direction', '+')
```
owocki: thanks

PR #6289, opened 2020-03-24
Add a guide to run gitcoin remotely.

inspireme6:
##### Description

Guide to run Gitcoin remotely using a VPS server (DO, in this case).

##### Refers/Fixes

Refers #4943


owocki:
oooh interesting :)

PR #6313, opened 2020-03-27
core: base commit for celo + zil + refactor

thelostone-mc:
##### Description

- update mgmt command to to cater for all cross-chain pending
  fulfillments
- refactor and move all token specific logic to it's own file
  within sync/
- remove un-used url
- update bounty creation to include celo
- set fulfillments > 5 min as expired

##### NOTE

- remove erc-20 ZIL from token

###### Decimals
- ZIL:  12
- cUSD: 18
- cGLD: 18


app/app/urls.py
```diff
@@ -250,11 +250,6 @@
     re_path(r'^bounty/quickstart/?', dashboard.views.quickstart, name='quickstart'),
     url(r'^bounty/new/?', dashboard.views.new_bounty, name='new_bounty'),
     re_path(r'^bounty/change/(?P<bounty_id>.*)?', dashboard.views.change_bounty, name='change_bounty'),
-    url(
```
octavioamu, line 253: we don't need this one anymore?

app/assets/v2/js/pages/create_bounty/token.js
```diff
@@ -76,7 +76,7 @@ createBounty = data => {
     'eventTag': metadata.eventTag,
     'auto_approve_workers': data.auto_approve_workers ? 'True' : 'False',
     'web3_type': 'qr',
-    'activity': data.activity
+    'activity': data.activity,
```
octavioamu, line 79: good catch

danlipert:
@wizzfile what is that?

PR #6340, opened 2020-03-30
puts searchresults objects in elasticsearch/ makes search GA

owocki:
##### Description

puts searchresults objects in elasticsearch/ makes search GA

##### Refers/Fixes

convo with Dan last 1:1

##### Testing

tested locally. super fast :)


app/search/management/commands/create_search_index.py
```diff
@@ -0,0 +1,30 @@
...
+
+from search.models import SearchResult
+
+
+class Command(BaseCommand):
+
+    help = 'uploads search results into elasticsearch'
+
+    def handle(self, *args, **options):
+        for sr in SearchResult.objects.all():
```
danlipert: One thing thats actually really cool about Django querysets is that it is smart enough to not load all of these objects into memory if you back it with postgres: https://docs.djangoproject.com/en/dev/ref/models/querysets/#iterator

owocki: good call, so sounds like we should use `.iterator()` here to explicitly do this

PR #6348, opened 2020-04-01
Add looking for a team option to projects form 

zoek1:
##### Description
- [X] Add option and input to indicate the user is looking for team members
- [X] Add one image field add one banner to hackathons
- [X] remove discord from project form

##### Refers/Fixes
No issue

##### Testing
Sia Hackathon Projects _ Gitcoin

image


app/dashboard/views.py
```diff
@@ -3939,24 +3943,24 @@ def hackathon_save_project(request):
                 'logo': request.FILES.get('logo', project.first().logo)
             })
 
-            project.update(**kwargs)
+            # project.update(**kwargs)
```
octavioamu, line 3946: the code doesn't need to update the project?

app/dashboard/views.py
```diff
@@ -3939,24 +3943,24 @@ def hackathon_save_project(request):
...
 
-            try:
-                bounty_profile = Profile.objects.get(handle=project.bounty.bounty_owner_github_username.lower())
-                if bounty_profile.chat_id is '' or bounty_profile.chat_id is None:
-                    created, bounty_profile = associate_chat_to_profile(bounty_profile)
+            #try:
```
octavioamu, line 3948: we are removing the chat associated actions?

zoek1: Thanks, i thought i restored these lines. The chat related actions caused crash my environment so i commented this piece of code.

app/dashboard/models.py
```diff
@@ -4668,6 +4669,12 @@ class HackathonProject(SuperModel):
         choices=PROJECT_STATUS,
         blank=True
     )
+    message = models.CharField(
+        max_length=150,
```
danlipert, line 4673: I'm wondering if this limit is too short, what do you think?

zoek1: i think is ok for a short message, originally the tweets have a length of 140 characters and worked well. One example could be:

`Hi! I'm python developer and I'm looking for a fronted partner to hack on the ETC bounty "Create A DIY Hardware Wallet Using Signatory"`

150 character limit will help to synthesize messages, and having a character count helper will help a lot the user. What do you think @PixelantDesign?

PixelantDesign:
check it out ^ @alexvotofuture hopefully this helps with what we talked about last week!

app/dashboard/views.py
androolloyd: @zoek1 there are some docs for the Gitcoin Chat setup, the container will spin up automatically.

If you go through that setup then you'll stop getting any connection issues.

zoek1: Thanks @androolloyd, I'll check it later today!

PR #6349, opened 2020-04-01
test script to do grants tx validation

### LINKED ISSUE #6318: Server Side Grants Transaction Validator
# Validator 1

I would like a python script written that I can pass in a list of transactions, token amounts, and token names, and it can return a 1 or a 0 for each txn to tell me if those tokens were transfered as part of that txn

Note that Gitcoin has a tool that does this already, but we're having some problems with false negatives and with replaced transactions. So I'm interested in seeing how a bounty hunter would handle this issue so we can perhaps integrate it.

example input: https://gist.github.com/owocki/179186552ec7e4c036918c5195949a25

example output
```
success, reason
True
True
False - tx failed
False - amount was off by 0.001
````

Acceptance Criteria:
- No more than 1% false negatives
- Should handle dropped/replaced txns

Please submit a PR that can validate token transfers (both ETH and ERC20 tokens) by Sunday night to participate in this bounty.

# Validator 2
I would like some python functions that I can pass in an ethereum mainnet account and a erc20-style contract address, and it can tell me where the tokens came from.  

ie.
- if account minted the token, then it returns `0x0`
- if account received the tokens from `0xabc` it returns `0xabc`
- if account received the tokens from `0xabc` and `0xabc2` it returns [`0xabc`, `0xabc2`]

### PR DESCRIPTION
**Pull request by inspireme6**
##### Description

Script to validate grants transactions, handles dropped/replaced txs by scraping the etherscan tx page since apparently there are no api that can provide such info.

It works with the first format (the one with the four rows).

##### Refers/Fixes

Fixes #6318 

##### Testing
screen1

### CODE REVIEW

**Comment by danlipert** [comment]
@inspireme6 this is dope, but I do worry about "gettting the finger from etherscan" as you put it... @owocki what do you think? I think we may risk getting the banhammer from etherscan, which would definitely hurt

**Comment by owocki** [comment]
we have contacts at Etherscan (we have a partnership with them via codefund) so if we do get the banhammer we couold probably work with them around it

File: app/grants/grants_validator.py [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+            if ok_tx:
+                pass
+            else:
+                print("Failed, tx sent to wrong address")
+                return
+        
+        amount = float(correct_amount)
+        amount_fee = float(correct_amount_fee)
+        if (prov.value == 0): # token transaction
+            token_decimals = getTokenDecimals(to_address)
```
owocki, line 79: @inspireme6 this is an interesting way of validating the txn. i wonder if its fool proof in all cases tho.. like are there any cases where the transfer could have failed but the txn would succeed?

also how do you know to only check the first 64 chars of the input?  what would u recommend i do if i want to validate a txn like https://etherscan.io/tx/0xd874b4fceb06b0611315394ec161c2ef874d6c6a9d68af99ff8570a08bd20510 that sends two tokens

PR #6388, opened 2020-04-07
Fix profile portfolio tab error

### LINKED ISSUE #6369: Cannot remove/edit projects in profile profolio tab
**Describe the bug**
I made a typo when adding works to my profolio, and I wanted to remove/edit it, which seems impossible

**To Reproduce**
1. Go to your profile, profolio tab e.g. https://gitcoin.co/williamchong007/portfolio
2. Add a new project and make a typo
3. Stuck with that typo project forever

**Expected behavior**
Some kind of edit or remove function should be implemented?

**Screenshots**

image
The second one was the typo one, no UI to remove it.

- OS: Win10
- Browser: Chrome
- Browser Version: 80

**Additional context**
also the `Add a project ( Only visible to you  )` seems confusing, since one will doubt if the project list below is also only visible to myself

iamonuwa:
##### Description
Adds edit and delete functions to the portfolio projects section

##### Refers/Fixes
#6369 

##### Testing
https://www.loom.com/share/b1a7bf0e60f349c4a50206360cfefaec

Go to profile portfolio section.


app/dashboard/views.py
```diff
@@ -555,6 +555,23 @@ def unrated_bounties(request):
...
+@csrf_exempt
+@require_POST
+def remove_portfolio(request):
+    portfolio_id = int(request.POST.get('portfolio_id'))
+    handle = request.user.profile.handle
+
+    try:
+        profile = profile_helper(handle, True)
+        portfolio_item = PortfolioItem.objects.get(id = portfolio_id)
+        portfolio_item.delete()
```
danlipert, line 567: Instead of deleting, lets set a property on the PortfolioItem such that if the property is True, it doesn't get displayed. Deleting data is generally something we should do manually as part of archival processes. Interesting discussion here: https://stackoverflow.com/questions/2549839/are-soft-deletes-a-good-idea

app/dashboard/views.py
```diff
@@ -555,6 +555,23 @@ def unrated_bounties(request):
...
     }, status=200)
 
+@csrf_exempt
+@require_POST
+def remove_portfolio(request):
+    portfolio_id = int(request.POST.get('portfolio_id'))
+    handle = request.user.profile.handle
+
+    try:
+        profile = profile_helper(handle, True)
```
danlipert, line 565: here we already have the profile from `request.user.profile` - no need to use the profile_helper

danlipert: Also, the profile isn't actually being used for anything... but we SHOULD use it to ensure the portfolio item is actually owned by this user.

app/dashboard/views.py
```diff
@@ -2687,21 +2703,32 @@ def get_profile_tab(request, profile, tab, prev_context):
...
-                    )
-                messages.info(request, 'Portfolio Item added.')
+        portfolio_id = request.POST.get('id')
+        if portfolio_id:
+            portfolio_item = PortfolioItem.objects.get(id = portfolio_id)
+            portfolio_item.title = title
+            portfolio_item.profile = request.user.profile
+            portfolio_item.id = portfolio_id
+            portfolio_item.link = request.POST.get('URL')
+            portfolio_item.tags = request.POST.get('tags').split(',')
```
danlipert, line 2713: if there are no tags, this will return `None` and then `None.split(',')` will throw an error. To overcome this, the 2nd argument to `.get` allows you to add your own value instead of None, so lets use an empty string.

```
Python 3.7.3 (default, Oct 11 2019, 19:39:43)
[Clang 11.0.0 (clang-1100.0.33.12)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> ''.split(',')
['']
>>> None.split(',')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'split'`

**Comment by iamonuwa** [comment]
@danlipert thanks for the review. Fixed all changes

File: app/dashboard/views.py (hunk 2 above)
Comment (danlipert) [reply app/dashboard/views.py]: @iamonuwa please see my second comment here regarding permissions

Comment (iamonuwa) [reply app/dashboard/views.py]: Gotcha!

Comment (iamonuwa) [reply app/dashboard/views.py]: Fixed, please take a look

Comment (danlipert) [reply app/dashboard/views.py]: This doesn't look right - why would the user profile be equal to a PortfolioItem?

Comment (iamonuwa) [reply app/dashboard/views.py]: Fixing this now

File: app/dashboard/views.py [hunk 4]
```diff
@@ -2501,6 +2519,31 @@ def invalid_file_response(uploaded_file, supported):
 
     return response
 
+@csrf_exempt
+@require_POST
+def bounty_upload_nda(request):
```
danlipert, line 2524: We removed this functionality recently, not sure why its showing up in this PR - can you please remove?

app/dashboard/templates/profiles/tab_portfolio.html
```diff
@@ -80,25 +82,33 @@ <h5 class="font-weight-semibold">{{submission.bounty.title}}</h5>&nbsp;
...
-            <small class="tag-list__item">{{keyword}}</small>
-          {% endfor %}
+        <div class="float-right">
+          {% if pi.link %}
+            <a class="btn btn-gc-blue btn-sm" href="{{pi.link}}">View Work &gt;</a>
+            {% if is_my_profile %}
+              <button type="button" data-id="{{pi.id}}" id="edit-portfolio-item" class="btn btn-gc-blue btn-sm">Edit Work</button>
+              <button type="button" data-id="{{pi.id}}" id="remove-portfolio-item" class="btn btn-gc-blue btn-sm">Delete</button>
+            {% endif %}
+            {% endif %}
```
danlipert, line 108: fix indentation here

iamonuwa:
Will be sending in PR for this today

PixelantDesign:
@iamonuwa are you still working on this? If so could you share a video? Thanks!

iamonuwa:
> @iamonuwa are you still working on this? If so could you share a video? Thanks!

Yes, waiting for reviews

octavioamu:
There is a lot of lint errors here I will change this to draft, please change it back when ready.

iamonuwa:
Alright, I'll fix this

thelostone-mc:
Closing this out for now! Do reopen once this is ready

PR #6439, opened 2020-04-15
Hackathon Townsquare - General Cleanup and Project Panel Updates

androolloyd:
##### Description

- Adding Looking for members Filter
- Project Panel updates
- message team members from project panel
- 

##### Refers/Fixes

Issue with adding members to channel, users who have left the chat experience halt the mass adding of hackers.


app/dashboard/templates/dashboard/index-vue.html
```diff
@@ -956,16 +913,18 @@ <h6 class="font-weight-bold mb-3">Invite [[numUsers]] Users to the Bounty</h6>
...
     {{hackathon_obj|json_script:"hackathon-object"}}
+    {{is_registered|json_script:"hackathon-registration"}}
     <script>
       $('body').bootstrapTooltip({
         selector: '[data-toggle="tooltip"]'
       });
       const csrftoken = jQuery("[name=csrfmiddlewaretoken]").val();
       document.keywords = {{ keywords | safe }};
-      document.is_registered = {{ is_registered | safe }};
+      document.is_registered = JSON.parse(document.getElementById('hackathon-registration').textContent);
```
octavioamu, line 923: where is this element with id hackathon-registration ?

app/dashboard/views.py
```diff
@@ -3659,9 +3664,10 @@ def hackathon(request, hackathon='', panel='prizes'):
         'orgs': orgs,
         'keywords': json.dumps([str(key) for key in Keyword.objects.all().values_list('keyword', flat=True)]),
         'hackathon': hackathon_event,
+        'hacker_count': hacker_count,
+        'projects_count': projects_count,
         'hackathon_obj': HackathonEventSerializer(hackathon_event).data,
-        'is_registered': json.dumps(True if is_registered else False),
-        'hackathon_not_started': hackathon_not_started,
+        'is_registered': True if is_registered else False,
```
octavioamu, line 3670: this will not get into document.is_registered === 'True' (string instead of boolean?)

app/dashboard/templates/dashboard/index-vue.html
```diff
@@ -956,16 +913,18 @@ <h6 class="font-weight-bold mb-3">Invite [[numUsers]] Users to the Bounty</h6>
     <script src='{% static "v2/js/vue-components.js" %}'></script>
     {{orgs|json_script:"sponsor-list"}}
     {{hackathon_obj|json_script:"hackathon-object"}}
+    {{is_registered|json_script:"hackathon-registration"}}
```
octavioamu, line 916: ohh ok so this will get into <script id=hackathon-registration> ?
But why changeing into it instead of json.dumps(True if is_registered else False)?

octavioamu: another option also if you did it because you are using is_registered in some other django html view we and need both True and true for js then we can just do `{{ document.is_registered = is_registered|yesno:"true,false" }};`

app/dashboard/views.py
androolloyd: Nah it gets converted to json Boolean. 

This is the way django suggests now

PR #6457, opened 2020-04-17
purge: amount_goal in Grants

thelostone-mc:
##### Description

- removes amount_goal field which has been removed from UI in the previous round
- updates emails + backend + UI code which uses as all of those are dead code
- update fixtures
- removed un-used css from emails

##### Flows Tested

- Grant Creation
- One Time Payment (+Splitter)
- Recurring Payment
- Cancel Grant

TO-CHECK @danlipert + @octavioamu  :
Had to update `0034_data_for_old_grants.py` which was a custom migration created to populate grants activity when the feature was introduced


app/dashboard/migrations/0034_data_for_old_grants.py
```diff
@@ -22,7 +22,6 @@ def record_grant_activity_helper(activity_type, grant, profile, date):
     metadata = {
         'id': grant.id,
         'value_in_token': '{0:.2f}'.format(grant.amount_received),
-        'amount_goal': '{0:.2f}'.format(grant.amount_goal),
```
danlipert, line 25: This migration is cursed, would love to get rid of it somehow

PR #6466, opened 2020-04-17
Get Rid of useless IP method

x5engine:
removing the method that saves the IPs of users, for privacy sake!

### DIFF
--- a/app/kudos/views.py
+++ b/app/kudos/views.py
@@ -128,17 +128,6 @@ def marketplace(request):
     if q:
         title = f'{q.title()} Kudos'
         token_list = token_list.keyword(q)
-        # log this search, it might be useful for matching purposes down the line
-        try:
-            SearchHistory.objects.update_or_create(
-                search_type='kudos',
-                user=request.user,
-                data=request.GET,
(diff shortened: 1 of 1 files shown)


app/kudos/views.py
```diff
@@ -128,17 +128,6 @@ def marketplace(request):
     if q:
         title = f'{q.title()} Kudos'
         token_list = token_list.keyword(q)
-        # log this search, it might be useful for matching purposes down the line
-        try:
-            SearchHistory.objects.update_or_create(
-                search_type='kudos',
-                user=request.user,
-                data=request.GET,
-                ip_address=get_ip(request)
```
danlipert, line 137: @x5engine what do you think of just removing the ip_address portion?

thelostone-mc:
Closing this out ! 
@x5engine if you could get one just removing the IP -> we can get it in ^_^

app/kudos/views.py
x5engine: yes why not but the whole app has many ip addresses

x5engine:
sure  @thelostone-mc

PR #6472, opened 2020-04-19
chore: cross-chain improvs

thelostone-mc:
##### Description

- remove unused admin methods
- add activity on close bounty
- fixed broken fa icons
- updated tooltip design
- added icon for expired fulfillment to inform user
- set txn/address link based on token
- made qr code generate based on token

### DIFF
--- a/app/assets/v2/js/pages/bounty_details2.js
+++ b/app/assets/v2/js/pages/bounty_details2.js
@@ -88,7 +88,28 @@ Vue.mixin({
       }
       return url;
     },
+    getQRString: function(token_name, address) {
+      let qr_string;
 
+      switch (token_name) {
+        case 'ETC':
+          qr_string = `ethereum:${address}`;
+          break;
+
+        case 'cUSD':
+        case 'cGLD':
+          // TOOD : UPDATE
+          qr_string = `ethereum:${address}`;
+          break;
+
+        case 'ZIL':
+          // TOOD : UPDATE
+          qr_string = `ethereum:${address}`;
+          break;
+      }
+
+      return qr_string;
+    },
     syncBounty: function() {
       let vm = this;
 
--- a/app/assets/v2/js/pages/bounty_details2.js
+++ b/app/assets/v2/js/pages/bounty_details2.js
@@ -9,7 +9,6 @@ const loadingState = {
 };
 
 document.result = bounty;
-// listen_for_web3_changes();
 
 Vue.mixin({
   methods: {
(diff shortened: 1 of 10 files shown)


app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -254,33 +299,6 @@ Vue.mixin({
...
-    },
-    overrideStatus: function() {
-      let vm = this;
-      let text = window.prompt('What new status (valid choices: "open", "started", "submitted", "done", "expired", "cancelled", "" to remove override )?', '');
-
-      if (text === null) {
-        return;
-      }
-      document.location.href = `${vm.bounty.url}?admin_override_satatus=${text}`;
-    },
```
octavioamu, line 283: why are we removing these actions? that will be undefined in the html vue click

app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -370,13 +388,6 @@ Vue.mixin({
         activities.forEach(activity => {
           if (activity.metadata) {
             if (activity.metadata.new_bounty) {
-              // ETH
-              activity.metadata.new_bounty['token_value'] = activity.metadata.new_bounty.value_in_token / 10 ** decimals;
-              if (activity.metadata.old_bounty) {
-                activity.metadata.old_bounty['token_value'] = activity.metadata.old_bounty.value_in_token / 10 ** decimals;
-              }
-            } else {
-              // cross-chain
```
octavioamu, line 379: are not going to jump and use this file in eth bounties?

app/dashboard/templates/bounty/details2.html
```diff
@@ -36,9 +36,6 @@
             Admin Actions
           </span>
           <a v-for="link in quickLinks" class="btn btn-outline-gc-blue px-2 mx-2 font-smaller-4" :href="[[link.href]]" :title="[[ link.title ]]">[[ link.label ]]</a>
-          <button class="btn btn-outline-gc-blue px-2 mx-2 font-smaller-4" @click="contactFunder()">Contact Funder</button>
```
octavioamu, line 39: these removed are not part of quickLinks since are not links, are prompts and I fixed them to not send the actions if are empty (today that is broken in normal bounties if you click that prompts then the action is sent even without nothing

app/assets/v2/js/pages/bounty_details2.js
thelostone-mc: @octavioamu so those buttons don't work for ETH bounties as well!
Hence I made sure only the buttons which worked were being ported over

app/assets/v2/js/pages/bounty_details2.js
thelostone-mc: I'd want to do it via the backend where we streamline the response and then pass it to the frontend !
Activity will be in common format when it comes to frontend (irrespective of if it's ETH or cross-chain)

app/assets/v2/js/pages/bounty_details2.js
thelostone-mc: These actions don't work in live ! (haven't in a long time)
https://gitcoin.co/issue/gitcoinco/web/6486/4252

So removed them out from the JS and the html for cross-chain

thelostone-mc: I'll re-add the contactFunder and snooze bot but the override status -> let's not do that ?

app/dashboard/templates/bounty/details2.html
thelostone-mc: Added this back !

app/assets/v2/js/pages/bounty_details2.js
octavioamu: but eth bounties are not using this code

octavioamu: the override status is something they use to manage hackathon bounties sometimes I believe

thelostone-mc: lol that doesn't work :P cause that updates `idx_status` and not `bounty_state`

we'll fix this up if they ever bring it up 📦

PR #6495, opened 2020-04-23
Tribes Townsquare

androolloyd:
##### Description

Tribes Townsquare Integrated View

##### Refers/Fixes
#6495


thelostone-mc:
@androolloyd yo we got conflict

docker-compose.yml
```diff
@@ -81,19 +81,6 @@ services:
         reservations:
           memory: 512M
 
-  testrpc:
```
thelostone-mc, line 84: oh niceee

androolloyd: I can undo this, do we even use testrpc for anything? I do all my testing to rinkeby, just eats up ram if we're not using it actively.

app/dashboard/views.py
```diff
@@ -915,25 +915,33 @@ def projects_fetch(request):
...
 
-    try:
-        hackathon_event = HackathonEvent.objects.get(id=hackathon_id)
-    except HackathonEvent.DoesNotExist:
-        hackathon_event = HackathonEvent.objects.last()
+    if hackathon_id:
+        try:
+            hackathon_event = HackathonEvent.objects.get(id=hackathon_id)
+        except HackathonEvent.DoesNotExist:
+            hackathon_event = HackathonEvent.objects.last()
```
danlipert, line 922: Is this a fallback for some other page that users this endpoint?

app/dashboard/views.py
```diff
@@ -1002,8 +1012,22 @@ def users_fetch(request):
...
-    show_banner = None
+    if tribe:
+
+        profile_list = profile_list.filter(Q(follower__org__handle__in=[tribe]) | Q(organizations_fk__handle__in=[tribe]))
+
+        if user_filter and user_filter != 'all':
+            if user_filter == 'owners':
+                profile_list = profile_list.filter(Q(organizations_fk__handle__in=[tribe]))
+
+            if user_filter == 'members':
```
danlipert, line 1023: should be elif here

app/dashboard/views.py
```diff
@@ -2810,18 +2838,29 @@ def profile(request, handle, tab=None):
     context['tab'] = tab
     context['show_activity'] = request.GET.get('p', False) != False
     context['is_my_org'] = request.user.is_authenticated and any([handle.lower() == org.lower() for org in request.user.profile.organizations ])
-    context['is_on_tribe'] = False
-    if request.user.is_authenticated:
-        context['is_on_tribe'] = request.user.profile.tribe_members.filter(org__handle=handle.lower())
+    if request.user.is_authenticated and hasattr(request.user, 'profile'):
+        context['is_on_tribe'] = len(request.user.profile.tribe_members.filter(org__handle=handle.lower())) > 0
```
danlipert, line 2842: can use `exists()` here rather than checking the length I think

docker-compose.yml
danlipert: yeah lets ditch it if possible - but I want to make sure @octavioamu and @owocki don't use it either - can you put this in a separate PR?

app/dashboard/views.py
androolloyd: I believe it’s a fallback for other requests

app/assets/v2/js/pages/profile-tribes.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  if (existingScript && callback)
+    callback();
+};
+
+(function($) {
+  // doc ready
+  $(() => {
+    Vue.use(VueQuillEditor);
```
octavioamu, line 27: Is this loaded?

app/assets/v2/js/pages/profile-tribes.js
```diff
@@ -0,0 +1,272 @@
+/* eslint-disable no-loop-func */
+
+
+const loadDynamicScript = (callback, url, id) => {
```
octavioamu, line 4: this is already on tribe-edit.js also loaded on profile so you probably have it already on that page

app/dashboard/views.py
```diff
@@ -4499,13 +4546,18 @@ def join_tribe(request, handle):
     if request.user.is_authenticated:
         profile = request.user.profile if hasattr(request.user, 'profile') else None
         try:
-            TribeMember.objects.get(profile=profile, org__handle=handle.lower()).delete()
+            try:
+                TribeMember.objects.get(profile=profile, org__handle=handle.lower()).delete()
+            except TribeMember.MultipleObjectsReturned:
+                TribeMember.objects.filter(profile=profile, org__handle=handle.lower()).delete()
+
```
octavioamu, line 4553: should be situations where profile return multiple? and in that case do we want to delete all of them?

package-lock.json
```diff
@@ -3670,6 +3670,11 @@
       "integrity": "sha1-IZMqVJ9eUv/ZqCf1cOBL5iqX2lQ=",
       "dev": true
     },
+    "prettier": {
```
octavioamu, line 3673: is this change package related?

package.json
```diff
@@ -25,5 +25,8 @@
     "eslint": "^5.16.0",
     "stylelint": "^9.10.1",
     "stylelint-config-recommended": "^2.2.0"
+  },
+  "dependencies": {
+    "prettier": "^2.0.5"
```
octavioamu, line 30: is this being used do we need to add a new dependency as we don't really run node only manually when checking lint.?

app/assets/v2/js/pages/profile-tribes.js
androolloyd: i believe is loaded by the profile-vue.html

package.json
androolloyd: ah nah that can be removed.

app/dashboard/views.py
androolloyd: shouldn't ever be multiple members for a single profile and an org.

androolloyd: the issue is that if a request fails, the user could get multiple entries in the db and then can't follow/unfollow properly, this ensures that we delete any multiple entries and keep the user's ability to follow/unfollow in tact.

app/assets/v2/js/pages/profile-tribes.js
androolloyd: i don't use tribe-edit and am not actually using the load dynamic script yet, i just hard coded on to the tribes-vue, lets leave as I intend to refactor to use this.

Issue #4152, opened 2019-04-09
Add Flag to Leaderboard


aashutoshrathi:
### What 

The current leaderboard just has country names with their respective bounties count.
We can make it more user-friendly by adding a flag of the country ahead of their name.

![image](https://user-images.githubusercontent.com/21199234/55767457-a3ba4a80-5a96-11e9-8a29-7b1da9e1e30e.png)

### Why

It will be convenient for the user to search for the country and will give better UX.

hydr063n:
@thelostone-mc  Hi. I want to point out some issues I found while testing on live site.

1. I saw it working a few week back but some of the country names are not mapped correctly in the data file. I figured we could extend to a new PR so as to fine tune the render. All that's needed to is to match the country names that gitcoin uses to a country abbreviation in the data file.

2. I think another commit may have changed the frontend code during a loop so its currently not rendering the flag. I zipped the flags with the items list as I needed a parallel list but now only the items list is being traversed in the loop. I think the name may have been confused. Might be better to rename the list to much more understandable name like `items_and_flags`
![zipped list not used in loop] (https://i.imgur.com/JoEX93A.png "zipped list not used in loop")
![zipped list not used in loop] (https://i.imgur.com/JoEX93A.png "zipped list not used in loop")

I will be happy to extend the extra work into a new PR. Anyway, the pull request has been merged, will I receive a tip for this PR?

PR #6509, opened 2020-04-27
feature - add flags to leaderboard

hydr063n:
##### Description

I've added functionality to load flag icon next to the country in Community leaderboard

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/4152

##### Testing

Before: https://user-images.githubusercontent.com/21199234/55767457-a3ba4a80-5a96-11e9-8a29-7b1da9e1e30e.png

After:
1. Personally curated test data 4 countries: https://imgur.com/a/8UQcuSj

2. If the mapping of Country name to flag fails the default flag is USA e.g Incorrectly spelled Yemen displays US flag https://imgur.com/a/qMRwD15

3. Real Rinkeby data https://imgur.com/a/KhC84UK

I opted for the flag API instead of emojis as (from my personal research) Microsoft Windows doesn't support flag emojis. Idk, works on my Linux but the userbase of Linux compared to Windows is small.


app/marketing/views.py
```diff
@@ -790,6 +790,24 @@ def leaderboard(request, key=''):
             for tech in techs:
                 technologies.add(tech)
 
+    import pandas
```
danlipert, line 793: Can you move this import to the top of the file?

danlipert: Actually, we don't need to use pandas for this - you could either take that list of country codes and set it up as a python file and just import that here. I think thats easiest.

app/marketing/views.py
```diff
@@ -845,8 +863,11 @@ def leaderboard(request, key=''):
     if next_update and next_update < timezone.now():
         next_update = timezone.now() + timezone.timedelta(days=1)
 
+    dualList= zip(items[0:limit],flags)
```
danlipert, line 866: please use snake case for variable names: https://en.wikipedia.org/wiki/Snake_case

app/marketing/views.py
hydr063n: I've converted the csv into a python list and set the required flag variables inside the new file

app/marketing/views.py
hydr063n: whoops. fixed!

PR #6511, opened 2020-04-27
add uri schema for celo, zil

danlipert:
##### Description

This PR adds testnet URI schema for Celo and the expected schema for ZIL

##### Refers/Fixes

n/a
##### Testing

untested


danlipert:
Tested this with the current build of the mobile app - it only supports raw address scanning right now, will update when I know the compatible build is coming out.

PR #6545, opened 2020-05-01
Fixes for serialize issue/performance tweak

androolloyd:
…ce/lower serialization weight, if the new view fails, load the old view

##### Description
Remove Active Bounties from the TribesSerializer
move the response higher up the request as we don't need the overhead of processing data we're not using for this view.


thelostone-mc:
@androolloyd derp not sure why this didn't get in!
Could you resolve the conflicts ? 😅

androolloyd:
ill Review and update. I have another branch pushed that I believe encompasses these changes.

androolloyd:
this is apart of a newer update.

PR #6585, opened 2020-05-05
no hardcoded mainnet in Kudos search

vporton:
##### Description

Mainnet was hardcoded in Kudos search (in quests, for instance). I've fixed this. Now it works with Rinkeby.

Check it on mainnet before merging to `master`.


owocki:
while i agree this is useful for development, i think we should strongly consider the UX implications of this before merging.  

we had it so that it'd only load tokens from the network you were on before, and it created several support requests per week from people who didnt understand that it was only loading tokens from that network, and we had to explain it to them.  

not saying we shouldnt make this change, but i do think that if we do then we should consider adding some UX bumpers to let people know thats how its populating.

vporton:
> while i agree this is useful for development, i think we should strongly consider the UX implications of this before merging.
> 
> we had it so that it'd only load tokens from the network you were on before, and it created several support requests per week from people who didnt understand that it was only loading tokens from that network, and we had to explain it to them.

"the network you were on before" you mean the currently chosen (by MetaMask) network? I don't understand your wording.

> not saying we shouldnt make this change, but i do think that if we do then we should consider adding some UX bumpers to let people know thats how its populating.

We may add a noticeable note "Tokens from Rinkeby" (when on Rinkeby) in editing a quest. That would be a perfect solution.

thelostone-mc:
@vporton I'd recomment a diff way to go about this !
Define a variable which checks if the app is in live / local ->
If live -> just leave it at mainnet 
If develop -> use document.web3network


What do you think of that ?

vporton:
> @vporton I'd recomment a diff way to go about this !
> Define a variable which checks if the app is in live / local ->
> If live -> just leave it at mainnet
> If develop -> use document.web3network
> 
> What do you think of that ?

I see no sense in this. What's wrong with my way?

Everything in the Web3 interface of GitCoin is controlled with `document.web3network` variable. Why should Kudos be an exception?

danlipert:
@vporton I think there are some good points here with some good paths forward - if you disagree then close the PR, otherwise please add the additional functionality (networking warning, local/prod settings, maybe something else) and we'll merge it in.

vporton:
I reverted my Kudos network change, as it was used only for debugging. So the PR is ready to accept.

PR #6591, opened 2020-05-05
Quests improvements

vporton:
##### Description

Work in progress (now no more WIP) over https://github.com/gitcoinco/web/issues/6542:

I made quests cache updated more often (every hour; isn't this too often?) Ideally it would be updated dynamically (without a cache), but that would require:

1. Add sums to the DB (with `db_index=True`).

2. Write a Django DB migration to update sums for existing data. (This would also probably require the site to be down for probably a few seconds.)

3. Rewrite the view.

The result seems not worth the work. (Anyway, who does need quest results updated more often than hourly? We are not a real time gaming site.)

It remains to implement the other half of the issue.


vporton:
All three issues are solved (I will remove WIP status from this PR):
* no referral rewards for quests
* disallow running several instances of the same quest in parallel
* refresh the leaderboard more frequently

To solve the issues with parallel running, I only allow running questions in a fixed order. If the user opens another tab, it leads to his second quest instance to fail. (Should probably be a better error message, but do we care about informing hackers that they attempt to hijack us?) It uses the existing quest cooldown timer.

This solution is incompatible with the idea to display questions in a random order, but let's solve this issue when work on it, not now.

app/assets/v2/js/kudos-search.js
```diff
@@ -32,7 +32,7 @@ function kudosSearch(elem) {
 
           let query = {
             term: params.term[0] === '@' ? params.term.slice(1) : params.term,
-            network: 'mainnet',
+            network: document.web3network,
```
aneopsy, line 35: I'm not sure about that

app/quests/helpers.py
```diff
@@ -46,6 +46,7 @@ def record_award_helper(qa, profile, layer=1, action='Beat', value_multiplier=1)
     """
     Awards point awards (and referral rewards) to a winner
     """
+    return  # temporarily disabled (until we solve the hijacking problem)
```
aneopsy, line 49: The goal is to only disable the referral rewards, not all the rewards by commenting L90 & L91

vporton: Fixed.

vporton: Note I changed the network to `document.web3network`, because it is otherwise impossible to debug it on localhost. See also https://github.com/gitcoinco/web/pull/6585

app/assets/v2/js/kudos-search.js
owocki, line 35: note that this will yieid complaints from end users who are not prompted to change their web3 netowrk when searching...

danlipert: @vporton do you want to address Kevin's comment here?

vporton: Just revert my change.

vporton: I have reverted the change.

thelostone-mc:
Closing this out as we'd like this to be maintained by the DAO instead of the monorepo

PR #6596, opened 2020-05-06
Creating gitcoinbot profile

vporton:
##### Description

Automatically create `gitcoinbot` profile.

We use a `manage.py` command rather than `profiles.json` fixture because otherwise it could overwrite existing data in the DB!

I don't try to assign an Ethereum address to GitCoin bot, because it's different (apparently probably yet non-existing) for Rinkeby.

If you want an Ethereum address to be automatically set even for Rinkeby, then don't merge yet this PR.

##### Testing

`gitcoinbot` is created by `docker-compose up --build`. Quests do work.  (They don't work without `gitcoinbot`.)


thelostone-mc:
I'd still lean towards keeping this in fixtures so that it's cleaner !
We'll be updating fixture cause as we purge code -> so I'd lean towards that as opposed to introducing another mgmt command

vporton:
> I'd still lean towards keeping this in fixtures so that it's cleaner !
> We'll be updating fixture cause as we purge code -> so I'd lean towards that as opposed to introducing another mgmt command

"We'll be updating fixture cause as we purge code" - I don't understand this phrase.

However, I know two ways to keep this in fixtures:

1. Use very big numbers as primary keys.

2. Use instead UUIDs as primary keys.

So, what do you want?

thelostone-mc:
I'd recomment just adding it as a fixture incrementing the pk by 1 from the latest entry

vporton:
> I'd recomment just adding it as a fixture incrementing the pk by 1 from the latest entry

AFAIK, it is impossible. Fixtures support fixed PKs not increments. If I increment it by 1 from the latest entry, this PK would be the same as an existing profile (added through the UI, not through a fixture) and this profile would be overwritten! That's a data loss.

thelostone-mc:
Ah it seemed to work with the existing fixtures where if I were to login after running the fixtures -> my logging in -> it would create an entry & not an existing data cause

vporton:
> Ah it seemed to work with the existing fixtures where if I were to login after running the fixtures -> my logging in -> it would create an entry & not an existing data cause

If you create a user and _then_ run the fixtures, the user could be overwritten.

It is bad for debugging. But even worse, it would overwrite some real user on the production site as soon as the fixtures are run.

thelostone-mc:
Fixtures aren't run on production data! I would be stoked if you could just append it as another entry in the fixture ! 
Running fixtures when you have data -> would overwrite/ error out stuff in DB !

vporton:
> Fixtures aren't run on production data! I would be stoked if you could just append it as another entry in the fixture !
> Running fixtures when you have data -> would overwrite/ error out stuff in DB !

But fixtures are run by developers after new fixtures are added. Running fixtures when you have data would overwrite stuff in DB! And that's bad.

thelostone-mc:
Oops this slipped through the cracks !
Hmm fair point @vporton !
Okie could you update it the fixture -> and we can get this in

vporton:
> Oops this slipped through the cracks !
> Hmm fair point @vporton !
> Okie could you update it the fixture -> and we can get this in

I don't understand what you want from me. First you want to use fixtures or not. What should I do?

thelostone-mc:
Use fixture + set the pk has a large number

thelostone-mc:
Closing now ! Reopen if we you get time to revisit this

PR #6598, opened 2020-05-06
hardcode mainnet tokens on /settings/tokens page

owocki:
##### Description

hardcode mainnet tokens on /settings/tokens page 

we get a support request about this like twice a week.  rather than keep on trying to explain it, i think we should just show the mainnet tokens in all cases.

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1588720173226000

##### Testing

tested locally


owocki:
https://github.com/gitcoinco/web/pull/6599

PR #6626, opened 2020-05-11
ENS support for sending tips and sending / receiving kudos

erib3:
##### Description

I added support for ENS by resolving addresses on the server side, it proved to be a simpler way to implement this. ENS resolution is now possible when sending tips and kudos and also receiving kudos. 

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/5470

### DIFF
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -977,6 +977,20 @@ function getNetwork(id) {
   return networks[id] || 'custom network';
 }
 
+function resolveENS(ens_name) {
+  let test_Url = 'http://localhost:8000/resolve_ens/';
+  let prod_Url = 'https://gitcoin.co/resolve_ens/';
+
+  var request = new XMLHttpRequest();
+
+  request.open('GET', test_Url.concat(ens_name), false); // `false` makes the request synchronous
+  request.send(null);
+
+  if (request.status === 200) {
+    return (request.responseText);
+  }
+}
+
 // figure out what version of web3 this is, whether we're logged in, etc..
 var listen_for_web3_changes = async function(no_ui_updates) {
   reloadCbAddress();
--- a/app/assets/v2/js/shared.js
+++ b/app/assets/v2/js/shared.js
@@ -978,17 +978,18 @@ function getNetwork(id) {
 }
 
 function resolveENS(ens_name) {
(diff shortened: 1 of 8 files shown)


app/assets/v2/js/shared.js
```diff
@@ -977,6 +977,20 @@ function getNetwork(id) {
   return networks[id] || 'custom network';
 }
 
+function resolveENS(ens_name) {
+  let test_Url = 'http://localhost:8000/resolve_ens/';
+  let prod_Url = 'https://gitcoin.co/resolve_ens/';
+
+  var request = new XMLHttpRequest();
+
+  request.open('GET', test_Url.concat(ens_name), false); // `false` makes the request synchronous
```
thelostone-mc, line 986: code needs to be updated -> to use test in local / production

app/assets/v2/js/shared.js
```diff
@@ -977,6 +977,20 @@ function getNetwork(id) {
   return networks[id] || 'custom network';
 }
 
+function resolveENS(ens_name) {
+  let test_Url = 'http://localhost:8000/resolve_ens/';
+  let prod_Url = 'https://gitcoin.co/resolve_ens/';
+
+  var request = new XMLHttpRequest();
```
thelostone-mc, line 984: ^ let /const 
Do we wanna use `fetchData` instead of `XMLHttpRequest`?

app/assets/v2/js/shared.js
```diff
@@ -977,6 +977,20 @@ function getNetwork(id) {
...
+
+  request.open('GET', test_Url.concat(ens_name), false); // `false` makes the request synchronous
+  request.send(null);
+
+  if (request.status === 200) {
```
thelostone-mc, line 989: Could we also handle when status is not 200

erib3:
https://drive.google.com/file/d/1In00O2YNDDG_8VUU3_u63t4RhOUDH2wH/view?usp=sharing

app/assets/v2/js/shared.js
octavioamu: for sure we should be using fetch instead.

thelostone-mc:
Closing this out ! Do reopen this if comments can be addresed

PR #6641, opened 2020-05-12
limit maximum input in the comments 500 chars

x5engine:
limit maximum input in the comments otherwise, you will get something like this

https://gitcoin.co/townsquare?tab=activity:273525

image

so better allow maximum 500 chararacters


x5engine:
yeaah but where is my 1 eth :D @thelostone-mc

owocki:
https://github.com/gitcoinco/web/blob/master/docs/CONTRIBUTING.md#monetization-policy

x5engine:
okay then it means, I get nothing for contributing?

owocki:
did you read the above monetization policy

x5engine:
okay sure

PR #6658, opened 2020-05-14
chore: remove animation

thelostone-mc:
##### Description

remove rain which causes screen to scroll on its own during waiting screen


owocki:
this PR makes makes the product more bland & lame.

why not just add some CSS such that it doesnt scroll on this one page where its a problem?

PR #6692, opened 2020-05-20
adds view counts to many objects on site, so we can see what items are popular when

owocki:
##### Description

adds view counts to many objects on site, so we can see what items are popular when

this is important bc it allows us to provide analytics data to users about whats popular, and visualize the funnel

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CBWA6A75M/p1590015181146800

##### Testing

tested locally, bounties, hackathons, kudos, grants all load


app/economy/models.py
```diff
@@ -134,6 +135,43 @@ def created_human_time(self):
...
+    def content_type(self):
+        ct = ContentType.objects.get_for_model(self)
+        return str(ct)
+
+    @property
+    def view_count_redis_key(self):
+        key = f"{self.content_type}_{self.pk}"
+        return key
+
+    def increment_view_count(self):
```
owocki, line 148: @gitcoinco/engineers im consdidering shifting from storing as a total count in redis, to a `ObjectView` django database object, which would allow us to 
1. store time series data
2. also keep track of who views what (useful for segmenting purposes)
3. differentiate index page views (like on /grants) from specific viewings of individual objects (like /grants/35/sustainability fund)

any objection if i do this?  seems like the only counter-point would be DB bloat.

danlipert: There will no doubt be a performance impact in reading and tallying this data - if you want to store that data for later data analysis I think it should be done in parallel with storing the total count in redis. Only the redis count should ever be read from during a web request.

app/dashboard/tasks.py
```diff
@@ -135,3 +136,21 @@ def m2m_changed_interested(self, bounty_pk, retry: bool = True) -> None:
...
+
+@app.shared_task(bind=True, max_retries=1)
+def increment_view_count(self, pks, content_type, retry: bool = True) -> None:
+    """
+    :param self:
+    :param pk:
+    :param content_type:
+    :return:
+    """
+    redis = RedisService().redis
```
danlipert, line 149: A good opportunity here to optimize performance is to swap out the `redis_service.py` code we use for just using a proper connection pool

app/economy/models.py
owocki: good idea.

as a rule, i think we shouldnt write to the DB from most HTTP requests. factoring into a task is almost always a good idea

owocki:
82ab79fdad94dc2e4490030722c4eab7e17b9494 adds time series and user specific data to this PR.  

marking as rdy for review

owocki:
rdy for review

PR #6699, opened 2020-05-22
Fix issue with posting to integrated townsquare views.

androolloyd:
##### Description
Pinned post introduced a regression due to a variable name re-use.

shared/activity.html was including a hidden input that gets passed along, it is failing for hackathons/tribes/anything that uses the shared/activity with the exception of plain townsquare.

##### Refers/Fixes
No ticket, was found by @connoroday reported via slack

##### Testing
Please test with Prod data before deploying.

### DIFF
--- a/app/retail/templates/shared/activity.html
+++ b/app/retail/templates/shared/activity.html
@@ -572,9 +572,9 @@
     </div>
   {% if pinned %}
     <div class="pinned-by small mt-2">
-      <dic class="col">
+      <div class="col">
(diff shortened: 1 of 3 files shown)


app/retail/templates/shared/activity.html
```diff
@@ -572,9 +572,9 @@
     </div>
   {% if pinned %}
     <div class="pinned-by small mt-2">
-      <dic class="col">
```
thelostone-mc, line 575: LOL

androolloyd: :D

PR #6702, opened 2020-05-22
WIP: Bulk Grants Checkout Cart

apbendi:
This WIP PR implements the UI mechanics for Bulk Grants checkout via a "shopping cart" metaphor.

#### It includes:

* A redesigned Grant Detail screen, with an "Add To Cart" Button
* A new side cart that pops out on the Grant Details page when an item is added
* A new cart page at `/grants/cart`, implemented as VUE.js component
* All basic UI mechanics to go along with the above, i.e. add/edit/remove from cart functionality
* Collection of all relevant donation data in `localStorage` so it can
  1. Persist across user sessions
  2. Be used at final checkout to drive contract/backend interaction (TBD)

#### It currently lacks:

* Full mobile responsiveness
  * Most views are partially responsive but could use some tweaking on mobile or smaller browser windows
  * The side cart is not yet mobile responsive at all— needs to cover the grant detail screen, but currently it just get smooshed
* Backend/Contract integration — the checkout functionality, which will make contract calls and post donation data to the backend, is not implemented


owocki:
just tested, seems like its working well to me.  im leaving some product notes, will leave engineering notes to the @gitcoinco/engineers 

a few notes:
- u can get the full list of possible tokens from the economy.models.token object. might be worth putting that into the dropdown.
- for grants that are MY GRANTS, there used to be a left sidebar where you could edit your grant ( like this https://bits.owocki.com/BluZ8ZnA )  we should figure out where that goes in the new view
- if i click 'add to cart' and that grant is already in my cart, could it increase the quantity?
- how do i trigger the cart view, without clicking 'add to cart' again? i noticed that this fontawesome icon is in the designs, but not wired up https://bits.owocki.com/RBuvxvb4 - is that going on this PR, or another piece of scope?

app/assets/v2/js/grants/funding.js
```diff
@@ -0,0 +1,198 @@
...
+    if (!cartList) {
+        return [];
+    }
+
+    return JSON.parse(cartList);
+}
+
+function setCart(list) {
+    localStorage.setItem('grants_cart', JSON.stringify(list));
+}
```
thelostone-mc, line 198: ^ would it make more sense to turn these all into Vue methods ?

apbendi: hey @thelostone-mc 👋. Hmm, yes that's a good thought. I agree we could probably use a better abstraction here. Thinking about the way `localStorage` cart data is shared between the Vue cart component and the jQuery code on the details page, it probably makes sense to make these static methods on a shared helper class or something along those lines.

apbendi:
Hey @owocki, thanks for the feedback!

> u can get the full list of possible tokens from the economy.models.token object. might be worth putting that into the dropdown.

Yes! just didn't get there.

> for grants that are MY GRANTS, there used to be a left sidebar where you could edit your grant ( like this https://bits.owocki.com/BluZ8ZnA ) we should figure out where that goes in the new view

Good catch, I will have to test it out logged in as the user who created one of the test grants to make sure we didn't lose something in the UI refactor. Maybe @PixelantDesign can weigh in on where that goes in the new design?

> if i click 'add to cart' and that grant is already in my cart, could it increase the quantity?

Thought about this, but wasn't sure if it was the right course from a product perspective, since "quantity" here is really "an amount of money in one of several currencies that I'm donating." So if the user currently has 1 Dai in the cart, should it go to 2 Dai? What if they have 0.01 ETH?

> how do i trigger the cart view, without clicking 'add to cart' again? i noticed that this fontawesome icon is in the designs, but not wired up https://bits.owocki.com/RBuvxvb4 - is that going on this PR, or another piece of scope?

Ahh yes, sorry, meant to include in the initial PR that we hadn't gotten to that yet. Shouldn't be particularly hard or time consuming, we just didn't dive into the nav code yet to get to it.

apbendi:
> for grants that are MY GRANTS, there used to be a left sidebar where you could edit your grant ( like this https://bits.owocki.com/BluZ8ZnA ) we should figure out where that goes in the new view

Hey @owocki, testing this out, I'm still seeing this when viewing one of my own grants. I added some padding to make it a little cleaner. Any thoughts @PixelantDesign? Did you want to do something different with it?

Screen Shot 2020-06-01 at 11 02 29 PM

owocki:
hmmm.. that looks a little slunky. will defer to @PixelantDesign on the best path forward tho

app/app/fixtures/grants.json
```diff
@@ -1,265 +1,259 @@
 [
```
danlipert, line 1: Why are there so many changes to the fixtures?

mds1: Looks like the changes are mostly formatting changes. I think what happened is when I did a find + replace to update the Rinkeby DAI address to use the official Maker address in 8330a2076ee4cd72ae15acd60df95dc8dcb2bac8, VSCode must have auto-formatted these JSON files based on my default VSCode settings. Looks like indents were updated from 4 spaces to 2, along with other formatting changes. 

Oddly, I don't recall seeing any formatting changes when looking at the original diff, and @apbendi would have noticed these during the original PR, so I'm totally sure how these changes got in. I'm guessing they snuck in during a rebase somehow.

PixelantDesign:
> > for grants that are MY GRANTS, there used to be a left sidebar where you could edit your grant ( like this https://bits.owocki.com/BluZ8ZnA ) we should figure out where that goes in the new view
> 
> Hey @owocki, testing this out, I'm still seeing this when viewing one of my own grants. I added some padding to make it a little cleaner. Any thoughts @PixelantDesign? Did you want to do something different with it?
> 
> Screen Shot 2020-06-01 at 11 02 29 PM

Hey @apbendi could we try something like this for the edit view?
Screen Shot 2020-06-08 at 3 09 23 PM

app/app/fixtures/grants.json
danlipert: Ok, thanks for the explanation - always good to limit the diffs to the scope so that its easier to review and see where the meaningful changes are

owocki:
closing this for now, as we want to merge the work into https://github.com/gitcoinco/web/pull/6819 instead of master

PR #6704, opened 2020-05-23
Notifs

thelostone-mc:
#### Description

- Wire in GitHub notification for cross-chain

Q: not clear on when the email + GitHub notification + activity should be created for payout flow
Is it each time when funder pays out / when the funder closes the bounty -> just 1 update ?

In other words should [these be ](https://github.com/gitcoinco/web/blob/notifs/app/dashboard/views.py#L5157-L5160) be in `close_bounty_v1` or should it be in `payout_bounty_v1`  
I think it's correct the way it is but wanted to double check

##### Testing

**NOT TESTED**


danlipert:
I'm thinking one the bounty is closed is probably best, but either way is fine

PR #6745, opened 2020-05-31
re-enable free kudos for Gitcoin Quests/Grants

owocki:
A few weeks ago, i noticed that with high gas prices on the eth network, the kudos minter was burning $200 per day in gas fees.

so i introduced a change to require people to pay their own gas, which would cut down the costs.

but following up on that today, it looks like this has really hurt the number of quest players, see below:

Screen Shot 2020-05-30 at 9 54 14 PM

so i'm introducing this PR, which introduces a "not submitted" state to the kudos transfer object, wherein the kudos will be submitted to one's profile immediately, but not to their web3 wallet until the ethereum network isnt super expensive anymore.

hopefully this approach reaches the twin goal of keeping gitcoin's costs down AND also keeping quest usage up

from testing:
Screen Shot 2020-05-30 at 9 55 23 PM

### DIFF
--- a/app/kudos/views.py
+++ b/app/kudos/views.py
@@ -688,7 +688,7 @@ def receive(request, key, txid, network):
     return TemplateResponse(request, 'transaction/receive.html', params)
 
 
-def redeem_bulk_coupon(coupon, profile, address, ip_address, save_addr=False):
+def redeem_bulk_coupon(coupon, profile, address, ip_address, save_addr=False, submit_later=False, exit_after_sending_tx=False):
     try:
         address = Web3.toChecksumAddress(address)
     except:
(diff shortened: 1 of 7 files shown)


app/kudos/views.py
```diff
@@ -723,18 +723,27 @@ def redeem_bulk_coupon(coupon, profile, address, ip_address, save_addr=False):
         return None, error, None
     else:
 
-        if profile.bulk_transfer_redemptions.filter(coupon=coupon).exists():
-            error = f'You have already redeemed this kudos.'
-            return None, error, None
+        #if profile.bulk_transfer_redemptions.filter(coupon=coupon).exists():
```
danlipert, line 726: Do we actually want to remove this check or were these comments accidentally left in?

thelostone-mc: Oh that should be there :P

PR #6747, opened 2020-06-01
Change slack links to chat ones

Solexplorer:
##### Description

Changed links to redirect to chat instead of slack

##### Testing

PR so small, are tests needed?


danlipert:
@Solexplorer just so you know, the testing section the PR doesn't require you to write tets, but rather just explain how it was tested - that could be via included unit tests, or you could just write that it doesnt need testing because its so simple, or the change isn't in the application itself, or you can add a video showing the results of your changes :)

PR #6759, opened 2020-06-03
qrcode: wire in amount in qr string

thelostone-mc:
##### Description

- [x] adds amount into qr code
- [x] capability to allow user to not include amount in QR 

##### Refers/Fixes

Untitled

##### Testing

- QR code tested in Trust wallet for ETC and it works
- QR code works on Moonlet wallet


app/dashboard/templates/bounty/details2.html
```diff
@@ -500,7 +501,15 @@ <h5 class="font-body font-weight-semibold">{% trans "SUBMISSIONS" %}</h5>
...
-                                <qrcode :string="getQRString(bounty.token_name, fulfillment.fulfiller_address)"></qrcode>
+                                <qrcode :string="getQRString(bounty.token_name, fulfillment.fulfiller_address, fulfillment_context.amount)"></qrcode>
+                                <!-- TODO: WHEN CHECKBOX IS SELECTED -> re-render QR code so that amount is not added -->
+                                <!-- <p class="mt-2 mb-1 font-body font-weight-semibold">Include Amount In QR Code</p>
+                                <div class="toggle font-caption font-weight-bold">
+                                  <span>No</span>
+                                  <input type="checkbox" checked id="include_amount">
+                                  <label for="include_amount"></label>
+                                  <span>Yes</span>
+                                </div> -->
```
octavioamu: should be this commented?

app/assets/v2/css/card.css
```diff
@@ -1,9 +1,10 @@
 .toggle {
-  margin-top: 10px;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  font-weight: 600;
+  display: inline-flex;
```
octavioamu, line 2: can we create an extra class instead of overwrite?

octavioamu: or if our component gc-toggle but personally I will say the best will be using https://getbootstrap.com/docs/4.5/components/forms/#switches

thelostone-mc: donezo

PR #6761, opened 2020-06-03
[BUG]: Change email fix

### LINKED ISSUE #6561: Not able to change email address under profile setting: email is already in use error message
**Not able to change email address**
When going into https://gitcoin.co/settings/email/ I am unable to change the email address at the top. Regardless what email I use, the following error message pops up: Web3Foundation attempting to use an email which is already in use on the platform. 

I can say with confidence that the email is not in use, since it was newly created solely for this purpose. 

**To Reproduce**
1. Go to Manage setting on the profile drop down (https://gitcoin.co/settings/email/). The first thing that opens up are Email Preferences. 

2. Change email address.

3. Press save at the bottom.

4. Same site loads, no changes on it, only the yellow error message at the top. 

**Expected behavior**
Type in new email address, press save and have a new email address in use.

**Desktop:**

- OS: macOS Catalina v10.15.4 
- Browser: Chrome 
- Browser Version: 80.0.3987.16

androolloyd:
##### Description
A 500 error occurs during email change, updated to use the update_or_create functionality of the Django orm to simplify the process.

##### Refers/Fixes
#6561


danlipert:
@androolloyd you were unable to replicate the bug, correct? What convinces you that this is the correct fix?

androolloyd:
the issue was from a 500 error during the change, it happens when updating to an email you already used in the email marketing table, and lead me to the lines i changed. 

Once moving to an update or create, the error stops, and from what i can tell everything is operating as expected, i am able to change between any email associated to me.

androolloyd:
> @androolloyd you were unable to replicate the bug, correct? What convinces you that this is the correct fix?

I was able to reproduce a 500 error and that error prevented me from updating my email address, there was a comment on a slack thread about a user getting a 500, once i saw that i investigated further as I had solved a 500 issue just moments before.

Issue #6729, opened 2020-05-27
Email speed architecture discussion


owocki:
From my work on https://github.com/gitcoinco/web/pull/6728 I have realized that it takes about 3s to send a daily email.

From profiling the code ( 714b0fdd1174d274f92e71944bc63542f8e3d775 ) I have discoverd that the premailer is the priamry culprit of this issue.

This leads me to a question.. What do we do about making our email sends faster??  Right now it takes about 3s to send a daily email and 2.5s to send the weekly roundup.  at 31k active subscribers itll take 31000*2.5 = 77.5k seconds or 21 hours to send to the entire userbase.

if we can get that down to 0.1 seconds, it take roughly an hour to send to the entire userbase.

so what do we do about this?  I can see two approaches:
1. we pump all the emails through a celery queue and we run the premailer there.  and we can sclae that up horizonatally
2. we refactor the premailer to run at deploy time, instead of running at execution time.  this means that we wouldnt have to run the premailer for every single email send, but instead we'd compute the premailer information as part of the build process.

what does the group think?

owocki:
@gitcoinco/engineers @androolloyd @sebastiantf would love ur feedback on this.

owocki:
here is a very hacky example script that could compile CSS at compile time, but it'd need to be expanded to correctly follow `include` or `extend`s

```
pre = 'retail/templates/'
template = 'emails/template.html'
with open(pre + template, 'r', encoding='utf-8') as myfile:
  data = myfile.read()

from retail.emails import premailer_transform
premailer_transform(data)
```

danlipert:
It seems the biggest win is the premailer optimization. Is there ever I case that we need it to be run per-email? I don't think so. Given that, parallelizing it or offloading the recreation of the same CSS to another box seems like we are doing both refactoring and throwing hardware at it, rather than just doing the proper refactor.

owocki:
i tend to agree. the script above gets us like 20% - 50% of the way there.  will need to close the gap form there.    do we think this work should be done internally or can i bounty an external contributor to look at it?

owocki:
note: daily email is rolled out to 20% of userbase now + takes about 8 hours to run. bumping up to 40% today a849c97e7288a9a3660cd3d8f95d3b0d7cc5e810

androolloyd:
There are a few options for bulk emailing we could consider, SendGrid has a great bulk email delivery api that allows for unique templates. 

[...]

PR #6762, opened 2020-06-03
makes weekly roundup + daily update emails sent via celery

owocki:
##### Description

makes weekly roundup + daily bounty emails sent via celery

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/6729

##### Testing

tested, works locally

WARNING: in a production environment, care should be taken to make sure that the 40k celery jobs created by each management command do not throttle the processing of other important jobs.  two ways we can do that:
1. create two celery "lane"s that define emails that run with expediency + emails that can be backlogged for a few hours
2. spin up enough celery servers that its just not an issue

will leave it to the engineers to decide which path is best.

### DIFF
--- a/app/marketing/mails.py
+++ b/app/marketing/mails.py
@@ -118,7 +118,7 @@ def send_mail(from_email, _to_email, subject, body, html=False,
         mail.add_attachment(attachment)
 
     # debug logs
-    logger.info(f"-- Sending Mail '{subject}' to {to_email}")
+    logger.info(f"-- Sending Mail '{subject}' to {to_email.email}")
     try:
         response = sg.client.mail.send.post(request_body=mail.get())
     except UnauthorizedError as e:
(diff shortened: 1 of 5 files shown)


app/marketing/mails.py
```diff
@@ -118,7 +118,7 @@ def send_mail(from_email, _to_email, subject, body, html=False,
         mail.add_attachment(attachment)
 
     # debug logs
-    logger.info(f"-- Sending Mail '{subject}' to {to_email}")
+    logger.info(f"-- Sending Mail '{subject}' to {to_email.email}")
```
owocki, line 121: not a direct celery related fix, but is super nice for readability

danlipert:
@owocki I think the easiest way to create the lanes is to add rate limiting to the email tasks. We'll have to figure out the sweet spot so that the entire worker resources arent locked up. More info here: https://docs.celeryproject.org/en/stable/userguide/workers.html#rate-limits

danlipert:
Marking as WIP so we don't merge until I spin up the new celery boxes and set up rate limiting

owocki:
what settings / configuration r u thinking for the rate limiting? itd be interesting to have some boxes that rate limit some jobs, and others who rate limit other jobs.. effectively creating seperate queues. but i'd of course defer to u on the architecture.

owocki:
spun up new celery servers, gonna merge this into master via  direct commit now

owocki:
708165ec9597bb5ed1d629ddf992c06e34a0b61e

PR #6766, opened 2020-06-04
chore: add in feedback

thelostone-mc:
##### Description

Untitled


app/dashboard/templates/dashboard/hackathon/hackathons.html
```diff
@@ -25,14 +25,32 @@
...
         .container.custom-container {
           max-width: 1400px;
         }
       }
+
+
+      @media (max-width: 920px) {
+        img.card-top {
+          max-width: 100%;
+        }
```
octavioamu, line 42: same here, the problem is when we jump into a bundle this code will be loaded over the site and overwrite others components.

PR #6772, opened 2020-06-04
only use a profiles most recent location to mark a location, for leaderboards

owocki:
##### Description

only use a profiles most recent location to mark a location, for leaderboards

##### Refers/Fixes

https://gitcoin.co/leaderboard/countries?cadence=quarterly&keyword=all&product=all
why does the country list show prague still active in 2020?  could it have anything to do with the fact that the entire ehtereum community was in prague in 2019? lol

##### Testing

tested locally


thelostone-mc:
@owocki looks like CI is failing !

PR #6781, opened 2020-06-05
add animated logos

octavioamu:
##### Description

Add animated logos 

**Note**  js inside svg only runs when svg are inlined, for security reasons. So those were included as templates and called inside the navs to keep the code more maintainable.


owocki:
yesssss

thelostone-mc:
@octavioamu do we wanna remove the commented out code / wanna take it out in a follow up PR ?

octavioamu:
> @octavioamu do we wanna remove the commented out code / wanna take it out in a follow up PR ?

Kevin already cherry picked and deployed but is ok for now I think.

danlipert:
already cherry-picked into stable

PR #6800, opened 2020-06-07
WIP - POC for cleaned production backup

owocki:
##### Description

POC for cleaned production backup

##### Refers/Fixes

POC for cleaned production backup, a production backup that has been cleansed of most users data

We can do this by running a prod backup into a proxy DB. Then run commands on that box to remove any user info that should not be in the backup proxy DB (emails, password hashes, PIAA).  Then another backup can be created + uploaded to a public place.

##### Testing

Still WIP


danlipert:
should we scramble prod data or remove sensitive info in particular? I wonder how usable we can make this sample DB so that it has some parity w/ the actual prod database in terms of replicating bugs and such

owocki:
@danlipert yes, if people are ok with the approach i put into the existing backup file, then i think we should create a clean.sql file, which we maintain in this repo,  which is file that contains commands that removes most prod rows, but leaves a cleaned sample set that can be expose-able to developers on the secondary DB.

owocki:
>  I wonder how usable we can make this sample DB so that it has some parity w/ the actual prod database in terms of replicating bugs and such

we'll have to be iterative i think

thelostone-mc:
Closing this out for now but will keep branch alive if need to revisit

PR #6855, opened 2020-06-15
Add help bot to navbar

zoek1:
##### Description
image


owocki:
@zoek1 what ticket is this for?

i think its a good idea as it puts our support reuests down the github path more often (people will stop emailing founders@) but @PixelantDesign had some concerns

owocki:
can ya fix travis / merge conflicts pls! !?

the merge conflicts on stable should be mostly indentiation/formatting

PixelantDesign:
@zoek1 does the help text display on hover or is it always there?

@octavioamu @willsputra what do you think?  I'm curious if this will decrease the number of support requests.

thelostone-mc:
We already have a lot going on in the nav ! Do we really add one more to that list ?

zoek1:
@PixelantDesign only on hover

PixelantDesign:
if it's only on hover people might not know that it's for help

thelostone-mc:
If the information is useful for the user -> making it visible would defeat the purpose 
cause the user has to hover on the nav to see that and aka it would be an accidental discovery 

I think we need to find out a cleaner solution to this 🤔

danlipert:
closing this out for now

PR #6894, opened 2020-06-16
chore: account for unverified profiles

thelostone-mc:
Screenshot 2020-06-16 at 11 02 36 PM

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -215,7 +215,7 @@ def calculate_clr(aggregated_contributions, pair_totals, verified_list, v_thresh
 
             # pairwise matches to current round
             for k2, v2 in contribz.items():
-                if k2 > k1 and (k2 and k1) in verified_list:
+                if k2 > k1 and all(i in verified_list for i in [k2, k1]):
                     tot += ((v1 * v2) ** 0.5) / (pair_totals[k1][k2] / v_threshold + 1)
                 else:
                     tot += ((v1 * v2) ** 0.5) / (pair_totals[k1][k2] / uv_threshold + 1)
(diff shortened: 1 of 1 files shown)


app/grants/clr.py
```diff
@@ -95,8 +102,8 @@ def translate_data(grants_data):
 '''
 def get_verified_list(grant_contributions):
     verified_list = []
-    for proj, user, ver_stat, amount in grant_contributions:
-        if ver_stat == 'verified' and user not in verified_list:
+    for _, user, ver_stat, _ in grant_contributions:
+        if ver_stat and user not in verified_list:
```
frankchen07, line 106: hm, why do we need if `ver_stat` not in `verified_list` if the list is primarily users?

app/grants/clr.py
```diff
@@ -346,30 +354,40 @@ def fetch_data(clr_type=None, network='mainnet', clr_start_date=None, clr_end_da
...
     elif clr_type == 'matic':
         grants = Grant.objects.filter(network=network, hidden=False, active=True, grant_type='matic', link_to_new_grant=None)
-        threshold = THRESHOLD_MATIC
+        v_threshold = V_THRESHOLD_MATIC
+        uv_threshold = UV_THRESHOLD_MATIC
         total_pot = TOTAL_POT_MATIC
+
     else:
-        return None, None, None, None
+        return None, None, None, None, None, None
```
frankchen07, line 386: is the purpose of this to not have it error out if its empty?

frankchen07:
left some comments / questions that won't affect merging. last change would be to change uv_threshold to the estimate that would give a 20% difference between verified and non verified

app/grants/clr.py
thelostone-mc: Yup! Just being safe

app/grants/clr.py
thelostone-mc: I'm not clear here ! Isn't that what you intended when you wrote it up?

frankchen07: sorry wasn't thinking right on the ver_stat variable. it's boolean now, not a string. 👍

Issue #6892, opened 2020-06-16
grants - abandoned cart data


owocki:
the purpose of this task is to mirror the grant cart data to the server, every time a grant is added or removed from the cart

lets not overfcomplicated it, lets just make a grants.models.CartData object in the DB and store a JSON blob in it and a FK to the user.  and do a POST request

owocki:
@zoek1 u want?

zoek1:
Sure!

PR #6902, opened 2020-06-17
Add tracking cart functionality

zoek1:
##### Description
lets just make a grants.models.CartData object in the DB and store a JSON blob in it and a FK to the user. and do a POST request

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/6892

##### Testing
https://www.loom.com/share/fcafdd80c7124b0891010353c59d6cb7


owocki:
hmmm this seems to work, but it requires someone on the backend to reassemble the cart from the activity to make sense of it.

seems itd be easier just to have 1 helper function that takes the entire cart data, serializes it, and sends it to the backend to be stored in a blob every time something changes.

PR #6934, opened 2020-06-20
some wiki updates

molecula451:


molecula451:
Can we put this one on a deploy asap to see how it handles the issue with the timing out? @thelostone-mc 

Screenshot from 2020-06-20 08-35-31

app/app/settings.py
```diff
@@ -796,7 +796,7 @@ def callback(request):
 
 WIKI_ACCOUNT_HANDLING = True
 WIKI_ACCOUNT_SIGNUP_ALLOWED = True
-WIKI_CACHE_TIMEOUT = True
+WIKI_CACHE_TIMEOUT = 600
```
owocki, line 799: 600 s? ms?

owocki:
omg what an elegant fix!

app/app/settings.py
molecula451: > 600 s? ms?

Seconds to time out!

owocki: 600s seems like a lot. should it be more like 1s?

r u sure this will actually fix the issue?   IIRC from https://github.com/django-wiki/django-wiki/issues/1031 the issue was infinite recursion

molecula451: 600s is standard setting + the live **gitcoin.co** indefenetely expecting for the **jquery 3.3.1** timing out the wiki!

wiki1

owocki:
idk; i think the stack trace at https://github.com/django-wiki/django-wiki/issues/1031#issuecomment-597323865 shows that the root issue is otherwise..

molecula451:
> idk; i think the stack trace at [django-wiki/django-wiki#1031 (comment)](https://github.com/django-wiki/django-wiki/issues/1031#issuecomment-597323865) shows that the root issue is otherwise..

I think that traceback is showing the requests for preview button and other relevant stuff on the wiki which dependant on jquery

Screenshot from 2020-06-22 16-51-12


gItcoin.co requests for [**jquery3.3.1**](https://github.com/gitcoinco/web/pull/6934/commits/a33956c627c956c5339769aa63f793f49ceff5fc) where django-wiki uses [**jquery-3.4.1**](https://github.com/django-wiki/django-wiki/blob/master/src/wiki/static/wiki/js/jquery-3.4.1.min.js) timing out the wiki on 404 not found

molecula451:
Won't hesitate on this one. It looks like we are a lot dependant on a **Work In progress (Django-wiki)**.

It seems they have made a lot of updates for the **markdown compatibility** thats what traceback shows since the issue is from opened from around early March. Including other plugins updates. So this will still revamp the jquery plugin

owocki:
origianlly when i commented on this i thought you had found some cache timeout that would fix it.

i doubt that that frontend fix will solve the backend issue.  but i guess we'll see when its deployed. i dont think changing the jquery version will cause any harm

app/app/settings.py
molecula451: refer to [timeout settings](https://django-wiki.readthedocs.io/en/latest/settings.html#wiki.conf.settings.CACHE_TIMEOUT)

molecula451:
> origianlly when i commented on this i thought you had found some cache timeout that would fix it.
> 
> i doubt that that frontend fix will solve the backend issue. but i guess we'll see when its deployed. i dont think changing the jquery version will cause any harm

and It's a possible for sure! but it most likely that this issue shall be dependant on core **django-wiki** updates

molecula451:
### Set default filestorage in production for wiki static files

Heres what I got: storage backend for static files <code> s.gitcoin.co </code>

The Wiki works alright! the issue arises when you upload a file to the wiki and subsequently the django-wiki looks for the **MEDIA_PATH**. Since we rely on AWS Backend for static files at <code> s.gitcoin.co</code> the client end ups in an indefinetely **MEDIA_PATH** request ending in a timing out

Set default filestorage object to retrieve media static fails when in production.

I initially thought it was the jquery plugin. But this is because this shall solve in production. Not locally at all

## Look below how the wiki ends in a timing out when upload an image but works perfect on navigation, then the <code> slug/_edit </code> is collapse

wikitimeout

thelostone-mc:
@molecula451 is this still supposed to be in draft?

molecula451:
> @molecula451 is this still supposed to be in draft?

this will revamp the wiki jquery plugin version and set a timeout limit for the wiki. Push it

app/app/settings.py
danlipert: yeah 600s is way too long, and i dont think its going to fix the issue anyway? I think the issue is within django-wiki

molecula451: > yeah 600s is way too long, and i dont think its going to fix the issue anyway? I think the issue is within django-wiki

Totally! Set 1s to timeout. We will find that fix for the wiki sooner or later

thelostone-mc:
looks alright to me ! Just a lil concerned about the jquery update 
cc @octavioamu

octavioamu:
> looks alright to me just to try out and see if this fixes it !
> Just a lil concerned about the jquery update
> cc @octavioamu

seems there are not major changes on the version mostly fixes, seem safe to me to do the upgrade

thelostone-mc:
closing this out for now ! will reopen if we decide to go down this route

PR #6964, opened 2020-06-23
Fix for bottomVisible method scope.

androolloyd:
##### Description
Fixes an issue where project-directory is not calling the bottomVisible method, 

methods not defined inside the method object on vue objects are not available to the lexical scope of this.


owocki:
i cant say i really understand this code; so it gives me comfort that @danlipert approved them. 
shall we merge + hotfix them @gitcoinco/engineers  ? i know the project view on the active hackathons is broken right now?

PR #6983, opened 2020-06-25
 modifying to get_amount, …

### LINKED ISSUE #6960: grant triggers 4 calls to 'get_amount' everytime a cart item is updated
https://bits.owocki.com/z8u86rwg

can we get this down to 1? or for DAI, lets just not even make the call to the server, its wasteful. 1 DAI = $1 aand the frontend can know that.

or maybe we get the conversion rate ONCE per page load and just use it multiple times

developerfred:
##### Description

to make calls only when the token is different from the DAI 

##### Refers/Fixes
#6960


thelostone-mc:
@developerfred this would be fixed here https://github.com/gitcoinco/web/pull/6978
We've updated the API so than one call would suffice

developerfred:
@thelostone-mc Perfect! I will close this PR

PR #6991, opened 2020-06-25
for the chat application move all logger calls to info instead of deb…

androolloyd:
…ug or error

##### Description
Migrates logger.debug and logger.error to logger.info for the django chat application


thelostone-mc:
Why would we do this ? 
A few of them seem like logger.error should be used right

danlipert:
@thelostone-mc just too many chat errors in sentry

owocki:
> A few of them seem like logger.error should be used right

i mean if were actually going to be diligent about cleaning up the errors as they pop up in sentry, then lets use logger.error for the ones that are actually errors.   but right now were just needlessly burning sentry credits with errors we dont take action on.  i dont know the application specific enough to know which should be in which category tho :-/

androolloyd:
We could leave the .error, as they are connection based ones, debug errors can be info as they aren handled exceptions.

Issue #6965, opened 2020-06-23
add the ability to bulk edit the amount in carts


owocki:
i'd relaly like to have the add the ability to bulk edit the amount in grant carts.  this would save me from clicking into each of them.

owocki:
https://twitter.com/LefterisJP/status/1275552216372428801

owocki:
@mds1 would love to push on this one.

mds1:
Closed by #6965

PR #6996, opened 2020-06-25
Add bulk edit of cart amounts

mds1:
Closes https://github.com/gitcoinco/web/issues/6965

Changes:
- Adds an "Apply to all" button under each item on the /grants/cart page
- Adds an "Apply to all" button under the first item on the side cart
- When "Apply to all" is clicked, each cart entry is changed to use that amount and token. If the selected token is not available for a given grant, we convert to that amount to the equivalent amount in ETH and use that

If someone with better jQuery skills than me can add "Apply to all" under each item in the side cart that would be appreciated! I'm not sure how to split up the IDs with jQuery so it recognizes the array index

cc @apbendi @owocki 

Also, I originally planned to use this branch to also let users set their default amounts and currencies. But we'd want that saved to the user's profile, and since I'm not yet familiar with that aspect of the codebase I figured we can handle that in a separate PR and ship this one for now

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -513,6 +513,26 @@ Vue.component('grants-cart', {
       return amount.div(factor).toString(10);
     },
 
+    async applyAmountToAllGrants(grant) {
+      const preferredAmount = grant.grant_donation_amount;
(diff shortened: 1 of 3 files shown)


mds1:
Changes for this have been added to #6978

Issue #7002, opened 2020-06-26
make ONE post per grants checkout instead of N where N is the number of cart items


owocki:
we should make ONE post per grants checkout instead of N where N is the number of cart items

bc:
- reliability, it crushes our servers when someone checks out with 90 grants
- just in case one HTTP request gets lost, filterd by the load balancer
- i want to store the entire POST request in the DB in case there is some sort of 500 error upon contribution creation, so we dont lose the grant contribution creation

mds1:
Handler for current endpoint: https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L782-L850

PR #7019, opened 2020-06-30
Reduce carts post requests

mds1:
Closes https://github.com/gitcoinco/web/issues/7002

Changes/notes:
1. New endpoint of `/grants/bulk-fund`
2. Similar API to the original `/grant_id/grant_slug/fund` endpoint. The difference is that now the payload expects arrays of data for the object fields that differ for each donation.
3. Like before, when the `postToDatabase()` function of `cart.js` sends the POST request, it *does not* wait for the response before proceeding
4. `perftools.models.JSONStore` is used to save the POST payload before any processing happens

cc @owocki @apbendi

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -652,10 +652,10 @@ Vue.component('grants-cart', {
               });
           } else {
             approvalTx.send({ from: userAddress })
-              .on('transactionHash', (txHash) => { // eslint-disable-line no-loop-func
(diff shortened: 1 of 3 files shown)


mds1:
Updated to store off the payload before any processing happens, so this is now ready for review

danlipert:
@mds1 looks good but Travis is failing - can you rebase against the latest master so we can re-trigger Travis since it was against stable previously - thanks!

mds1:
@danlipert Should be all set now!

PR #7026, opened 2020-07-01
Super Responsive Quest Table

molecula451:
##### Description

Following the standard structure

respomsive

- [x] Add Quest Owner row when not in 'created' level
- [x] Change **"Play"** row to **"Stage"** 
- [x] Some icons


molecula451:
push it @thelostone-mc

owocki:
how do i actually click into the quest now?

https://bits.owocki.com/WnubmQzv

theres no play button and clicking the row doesnt do anything

owocki:
https://bits.owocki.com/OAuBokEv

users are not going to be able to to find the play button on the right
also just adding a x-scroll is not responsive.  it actually needs to fit in the users view.

the spec required that clicking anywhere on the row would take the user into playing the quest

molecula451:
> https://bits.owocki.com/OAuBokEv
> 
> users are not going to be able to to find the play button on the right
> also just adding a x-scroll is not responsive. it actually needs to fit in the users view.
> 
> the spec required that clicking anywhere on the row would take the user into playing the quest

@owocki  Totally. couldn't  reproduce this locally (You can see screenshots). Put it on a quick fix in #7065 for production. Will still dive into

PR #7040, opened 2020-07-02
celo: move to mainnet

thelostone-mc:
##### Description

- replace cGLD-> CELO
- replace testnet with mainnet
- test out flow

##### Testing

https://share.vidyard.com/watch/RAHFbh3jrXtkhvFqhGiJzQ?


octavioamu:
Just one more thing we need to do , remove the is_staff on the new hackathon form

PR #7057, opened 2020-07-03
chore: set CLR_END_DATE

thelostone-mc:
##### Description

- set CLR_END_DATE if not default to current.timezone 
- removed defaults and added null checks in place
- updated code to use CLR_END_DATE

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -35,6 +35,12 @@ from perftools.models import JSONStore
 PREV_CLR_START_DATE = dt.datetime(2020, 3, 23, 12, 0)
 PREV_CLR_END_DATE = dt.datetime(2020, 4, 7, 12, 0)
 CLR_START_DATE = dt.datetime(2020, 6, 15, 12, 0)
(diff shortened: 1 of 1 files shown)


app/grants/clr.py
```diff
@@ -35,6 +35,12 @@
 PREV_CLR_START_DATE = dt.datetime(2020, 3, 23, 12, 0)
 PREV_CLR_END_DATE = dt.datetime(2020, 4, 7, 12, 0)
 CLR_START_DATE = dt.datetime(2020, 6, 15, 12, 0)
+CLR_END_DATE = dt.datetime(2020, 7, 3, 16, 0)
+
+try:
+    CLR_END_DATE
+except NameError:
```
owocki, line 42: what is this for

owocki:
didnt test but LGTM

owocki:
at high level

app/grants/clr.py
thelostone-mc: Just to be extra safe to ensure no one comments out `CLR_END_DATE` while round is active 
That check ensure if that variable is not defined -> set it to current timezone

PR #7065, opened 2020-07-06
Text ellipsing description row

molecula451:
##### Description

this will ellipsis the data on the Description row which is overriding the style


molecula451:
@thelostone-mc Our "Description row" in our table quest it's overriding the style on long text descriptions. This will fix it. while I work on some other improvements for the whole /quests

cc: owocki

molecula451:
fixing it #7067

PR #7100, opened 2020-07-14
Integrate keycard as a login method and introduce ghost accounts

zoek1:
## If the browser is the status app, allow sign in with the keycard due some conflict with the github redirect flow
photo_2020-07-14_02-24-08

## Request permission to access the address 
photo_2020-07-14_02-19-28

## Request sign a message so we can validate the identity
photo_2020-07-14_02-19-31

## Wait the phone detect the keycard
photo_2020-07-14_02-19-33

## THe user is logged with its keycard account
photo_2020-07-14_02-19-36

## The user can configure the address to use the new sing in method
Account Settings _ Gitcoin

## To maximize the privacy option of keycards, i introduced the ghost mode so when the user activate this configuration, the system will create a fresh private account so the user can fund, or contribute in a private way. 
The user can change between private and public mode but it will require sign again a message so we get full confirmation to switch between the modes.
image
image

image

##### Refers/Fixes
https://github.com/status-im/status-keycard/issues/70


developerfred:
Awesome  @zoek1!

danlipert:
Pretty interesting - I think the status signin is cool, not sure about the ghost accounts and whether thats behavior we want to encourage... I'll pass that over to product for them to determine @PixelantDesign @frankchen07 @vs77bb

frankchen07:
I think I'd leave the use case at just making sure we can connect whichever wallets we want to support. Private ghost accounts are not a feature I'd be too concerned with. I vote to stash and revisit when appropriate.

guylouis:
@zoek1 thanks for your interest with keycard and your work. 

I don't understand exactly the end purpose of the PR. If you use status as your browser and status account is stored on keycard, you can already sign all dApp transactions with your keycard, so there is nothing to develop here. 

The gitcoin issue here https://gitcoin.co/issue/status-im/status-keycard/70/4398 is about a dApp requesting any keycard tapped on the phone to sign a transaction provided by the dApp. That's what's needed for a payment terminal for instance.

thelostone-mc:
Closing this out for now! Will reopen if this something we wanna pursue at a later point in time

PR #7105, opened 2020-07-15
clickeable row quest table

molecula451:
##### Description

click


molecula451:
@thelostone-mc push it

thelostone-mc:
@molecula451 closing this out cause the play button does the exact same thing except it was a bunch of checks like if the quest is already beaten / it's in cooldown / locked ! 

Thanks though ^_^ Hoping you can get nicer tweaks into the quests flow

PR #7134, opened 2020-07-21
sets next round start date in grants headder

owocki:
Screen Shot 2020-07-21 at 8 00 01 AM


thelostone-mc:
@owocki this PR has to be revisited cause of the self-service PR.

owocki:
ok i think the header should show all upcoming rounds & when they start.
and probably which rounds are active too.

we can either ship this PR now (most people watching grants will be expecting round 7) + do a v2, or we can fold it into the other PR + do the above scope in that PR.  im open to either.

thelostone-mc:
Cherrypicked onto master ! Hence closing

PR #7136, opened 2020-07-22
chore: self service

thelostone-mc:
#### Description

- CRUD operation for Grant Categories via admin
- CRUD operation for Grant Types via admin
- Ability To link Grant Types To Categories via admin
- Create new Grant CLR Round
- Turn CLR Round on/off as needed
- Update mgmt command to run only active CLR rounds by default 

https://share.vidyard.com/watch/QPa8xCJXAEuPYjU6GBnY5i?
https://share.vidyard.com/watch/ZspnbU1s8AVYWj85m6aECi?

##### NOTE

- There are 3 migrations which should not be squashed. 2 of these are custom 
- The most important piece is the data migration as we've removed columns + renamed columns  to accommodate self service architecture 
- The custom migrations ensure data is ported over to new columns (aka no data loss)

WHILE USING ENSURE -> There **SHOULD** be only **1 ACTIVE CLR ROUND**  for a GrantType AT **ANY GIVEN POINT IN TIME** 

##### Testing

- Overall Grant UI
- Grant Creation + Funding flow


app/grants/models.py
```diff
@@ -135,16 +134,17 @@ class Meta:
...
-        ('change', 'change'),
-        ('matic', 'matic')
-    ]
+    # GRANT_TYPES = [
+    #     ('tech', 'tech'),
+    #     ('health', 'health'),
+    #     ('media', 'Community'),
+    #     ('change', 'change'),
+    #     ('matic', 'matic')
+    # ]
```
thelostone-mc, line 143: will remove commented out code in a follow up PR

thelostone-mc:
@owocki / @danlipert the test failures are from mgmt commands which I'm not clear on how to fix. I'd just like to remove those imports but figured you guys might know better

app/grants/views.py
```diff
@@ -97,11 +98,11 @@
 
 kudos_reward_pks = [12580, 12584, 12572, 125868, 12552, 12556, 12557, 125677, 12550, 12392, 12307, 12343, 12156, 12164]
 
-if not clr_active:
```
owocki, line 100: all of these variables (line 77 onward) need to be driven by the DB now, not just commented out/ignored

thelostone-mc: They are commented out cause I'd like you to check out how we'd wanna handle this ! I'm not very clear on what is it we wanna do with this piece of code !

For example: 
How we do we handle diff routes when clr is not all active / when a few are active / etc

danlipert:
@thelostone-mc can you fix the conflicts and merge? 👍

Issue #6424, opened 2020-04-13
as a suggested tribes user, i want to be able to X out suggested tribes so i dont have to see them again


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
as a suggested tribes user, i want to be able to X out suggested tribes so i dont have to see them again

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
to allow peoiple to filter through tribes that they dont want to see

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
feature

### Current Behavior
[comment]: # (Describe what actually happened.)
suggested tribes
<img width="276" alt="screen_shot_2020-04-12_at_1 59 24_pm" src="https://user-images.githubusercontent.com/513929/79116024-9b00f200-7d44-11ea-8627-68d0e1226a36.png">

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
an X next to each suggested tribe, which when clicked
- removes that tribe from the DOM
- sends a PUT reuqest to the backend which stores an 'IgnoredSuggestedTribe' object in the DB
- refreshes the suggested tribes module with 5 new tribes
- next pageload, the X'd out tribe will not be loaded again

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
see above

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.6 ETH (91.89 USD @ $153.15/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/6424/4201).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/6424/4201).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $100,975.38 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

[...]

PR #7152, opened 2020-07-27
Add remove tribe on townsquare

zoek1:
##### Description

https://www.loom.com/share/8f3829efefc14c87a731c14bd8a2fd40

- removes that tribe from the DOM
- sends a PUT reuqest to the backend which stores an 'IgnoredSuggestedTribe' object in the DB
- refreshes the suggested tribes module with 5 new tribes
- next pageload, the X'd out tribe will not be loaded again

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/6424


app/dashboard/models.py
```diff
@@ -2788,7 +2788,7 @@ class Profile(SuperModel):
     last_validation_request = models.DateTimeField(blank=True, null=True, help_text=_("When the user requested a code for last time "))
     encoded_number = models.CharField(max_length=255, blank=True, help_text=_('Number with the user validate the account'))
     sybil_score = models.IntegerField(default=-1)
-
+    ignore_tribes = ArrayField(models.IntegerField(), blank=True, default=list)
```
danlipert, line 2791: Why use an `ArrayField` here instead of a `ManyToManyField`? Will cause problems if a tribe is later removed and we aren't cascading the deletion down to the reference.

zoek1: I used the `ArrayField` because the spec in the issue requests for one object to store the ignored tribes. I got your point, at this moment doesn't matter if a tribe is deleted because all ids in this list will be excluded

zoek1: I used the `ArrayField` because the spec in the issue requests for one object to store the ignored tribes. I got your point, at this moment doesn't matter if a tribe is deleted because all ids in this list will be excluded

zoek1: Replaced with ManyToMany

danlipert:
@zoek1 Conceptually looks good - looks like Travis doesn't like it, possibly the model setup or migrations, please take a look

zoek1:
Conflicts fixed @danlipert

PR #6940, opened 2020-06-22
feat: purge std bounty creation + use web3 modal for payout

thelostone-mc:
#### Description

-  Bounty Creation
   - [x]  update token.js to cater for new ETH flow with the right web3_type
   - [x] allow rinkeby -> (how do we control this)
   - [ ] retire ETH.js
   - [x] Do we collect funder address
   - [ ] feature bounty + fee

- [x] Start Work Flow
- [x] Approve / Reject Flow

- [x] Submit Work Flow
  - [ ] Ability to detect fulfiller address  (Do we want this)

- [ ] Payout 
  - [ ] ability to auto detect address
  - [x] async payout
  - [x] new payout flow 
  - [ ] add checks to ensure wallet is unlocked
  - [ ] ensure if bounty is mainnet -> allow payout in mainnet only

- [x] Close Bounty
- [x] Cancel Bounty
- [x] Increase Bounty
- [x] Update Bounty Details
- [x] Model Changes

PR #7184, opened 2020-08-05
remove duplicate env key `ETHERSCAN_API_KEY`

walidmujahid:
##### Description

remove duplicate .env key `ETHERSCAN_API_KEY`


app/app/settings.py
```diff
@@ -577,9 +577,6 @@ def callback(request):
 GITHUB_API_TOKEN = env('GITHUB_API_TOKEN', default='')  # TODO
 GITHUB_APP_NAME = env('GITHUB_APP_NAME', default='gitcoin-local')
 
-# Etherscan API
-ETHERSCAN_API_KEY = env('ETHERSCAN_API_KEY', default='')
```
thelostone-mc, line 581: Shouldn't this be kept ?

walidmujahid: @thelostone-mc That was the original line, however, PR #6940 added the duplicate to line 56, so I removed the original from L581 and kept then one from the PR. Same with the local.env file.

See: https://github.com/gitcoinco/web/blob/82b91f975ffb17f58f94f38dada489de06dd362f/app/app/settings.py#L56

PR #7203, opened 2020-08-06
Hackathon stats

octavioamu:
##### Description

add hackathon stats and link to create prize into sponsors dashboard 
https://share.vidyard.com/watch/zYYgmX5LrnkiD3tyD1YBsx?

image


frankchen07:
looks awesome, just gotta update metabase to hide the filters, and then maybe we test with a existing hack ID & organization so we can see that nothing weird shows up (the correct metrics should show)!

octavioamu:
> looks awesome, just gotta update metabase to hide the filters, and then maybe we test with a existing hack ID & organization so we can see that nothing weird shows up (the correct metrics should show)!

yes I checked with a hardcoded is working fine. the other test will need to be on prod with real data.

PR #7207, opened 2020-08-10
proactively stop retrying tasks before max retries exception

danlipert:
##### Description

This PR stops exceptions that are thrown when minting or redeeming airdropped kudos, either due to network congestion or some other reason that causes the celery task to be retried. After the max retries are hit, we stop attempting to retry to the task.

##### Refers/Fixes

n/a

##### Testing

untested

### DIFF
--- a/app/kudos/tasks.py
+++ b/app/kudos/tasks.py
@@ -36,7 +36,7 @@ def mint_token_request(self, token_req_id, retry=False):
             obj = TokenRequest.objects.get(pk=token_req_id)
             multiplier = 1
             gas_price = int(float(recommend_min_gas_price_to_confirm_in_time(1)) * multiplier)
-            if gas_price > delay_if_gas_prices_gt_mint:
+            if gas_price > delay_if_gas_prices_gt_mint and self.request.retries < self.max_retries:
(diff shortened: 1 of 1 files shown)


app/kudos/tasks.py
```diff
@@ -36,7 +36,7 @@ def mint_token_request(self, token_req_id, retry=False):
             obj = TokenRequest.objects.get(pk=token_req_id)
             multiplier = 1
             gas_price = int(float(recommend_min_gas_price_to_confirm_in_time(1)) * multiplier)
-            if gas_price > delay_if_gas_prices_gt_mint:
+            if gas_price > delay_if_gas_prices_gt_mint and self.request.retries < self.max_retries:
```
owocki, line 39: wont this LOC do *more* retries now that you've removed the `self.request.retries < self.max_retries` condition?  now it will infinitely retry, if im reading it right.

owocki: 416e458a43ef1415ac5594afa479b31842c3f0c8

owocki: https://gitcoincore.slack.com/archives/CAXQ7PT60/p1599159069014200

Issue #7163, opened 2020-07-29
As a grrants user, I want the ability to star/unstar a grant, so that I can have a shortlist of grants to checkout later.


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a grrants user, I want the ability to star/unstar a grant, so that I can have a shortlist of grants to checkout later.

### Why Is this Needed
Users want better curation features.

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
Users requesting better curation features.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
No ability to star / unstar a grant.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
- There is a font-awesome star in the top right of each grant card on the grants index page.
- Users can star/unstar a grant, which then creates an AJAX call to the backend to save that info for the end user.
- If users are not logged in they cannot star / unstar a grant
- If a user has starred grants a "My Stars(n)" category will appear on the nav allowing htem to look at their stars.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
See above

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)
None

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

PR #7245, opened 2020-08-18
Implement track of grants

zoek1:
##### Description

https://www.loom.com/share/f3e87e25d78e4628bb92623066137def

**note*
I changed start instead of following, I think this will be more meaningful
##### Refers/Fixes
https://github.com/gitcoinco/web/issues/7163
As a grrants user, I want the ability to star/unstar a grant, so that I can have a shortlist of grants to checkout later.


app/townsquare/models.py
```diff
@@ -394,9 +394,14 @@ def __str__(self):
...
     """Model for each favorite."""
     user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, related_name='favorites')
-    activity = models.ForeignKey('dashboard.Activity', on_delete=models.CASCADE)
+    grant = models.ForeignKey('grants.Grant', on_delete=models.CASCADE, related_name='grant_favorites', null=True)
+    activity = models.ForeignKey('dashboard.Activity', on_delete=models.CASCADE, null=True)
     created = models.DateTimeField(auto_now=True)
 
+    @staticmethod
+    def grants():
+        return Favorite.objects.filter(activity=None)
```
danlipert, line 403: why use the `activity=None` filter to find grants instead of `Favorite.objects.filter(grant__isnull=False)`? Seems unintuitive

danlipert: Also the related_name should probably just be `'favorites'` so you can access it like `grant.favorites`

app/townsquare/views.py
```diff
@@ -142,7 +142,7 @@ def get_sidebar_tabs(request):
             tabs = [new_tab] + tabs
             default_tab = 'grants'
 
-        num_favorites = request.user.favorites.all().count()
+        num_favorites = request.user.favorites.filter(grant=None).all().count()
```
danlipert, line 145: Maybe I'm misunderstanding how this works, but how does this filter give you the count of your favorite grants? Shouldn't it be `.exclude(grant=None)`?

danlipert (changes requested):
Just a few questions - also, I'm wondering if we can have the favorite count in the sidebar update as you favorite and unfavorite a grant so the count updates without having to reload or go to another page.

thelostone-mc:
Closing as this got it via #7332

PR #7067, opened 2020-07-06
Info on Description row

molecula451:
##### Description

This PR will add a "Info" button on description row so it shows heavy long description text on hover

quest

PR #7260, opened 2020-08-20
Info on Description Row Quests

molecula451:
##### Description

quest

 This was originally done in #7067 but somehow @thelostone-mc revamped and commit to stable and this one didn't take effect which I think it looks WAY cooler...


thelostone-mc:
Last time we got the change in -> it ended up breaking things in live and I had to undo the changes to revert it back to the old format so that folks could use it !

I'm not sure what exactly went wrong cause your code did look sane last time !
will pull this and check later and if it's alright locally -> will get it in

molecula451:
> Last time we got the change in -> it ended up breaking things in live and I had to undo the changes to revert it back to the old format so that folks could use it !
> 
> I'm not sure what exactly went wrong cause your code did look sane last time !
> will pull this and check later and if it's alright locally -> will get it in

Yeah. I notice that! This one is with the most recent master so it shouldn't be an issue. Tested locally. Working smooth

PR #7262, opened 2020-08-20
Integrate Circle

sanchaymittal:
##### Description

- Integration of circle to apollo hackathon only.
- Replaces the old town square with gitcoin circle community.

Demo Video
https://www.loom.com/share/7574a829191e47b8b1c428e3f122caf2

### DIFF
--- a/app/dashboard/templates/dashboard/index-vue.html
+++ b/app/dashboard/templates/dashboard/index-vue.html
@@ -198,7 +198,8 @@
                     {% trans "Townsquare" %}
                   </div>
                 </template>
-                <div class="ml-md-4 townsquare_header">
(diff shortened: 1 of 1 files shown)


thelostone-mc (changes requested):
> With this, we won't need Townsquare filters as circle got its own filters and we can create different spaces of discussions.


@sanchaymittal I'm talking about the highlighted section on the landing page:

image


Also these links have to be removed from the posts.
As clicking that will lead them to hackathon Townsquare which no longer exists 
image

sanchaymittal:
> > With this, we won't need Townsquare filters as circle got its own filters and we can create different spaces of discussions.
> 
> @sanchaymittal I'm talking about the highlighted section on the landing page:
> 
> image
> 
> Also these links have to be removed from the posts.
> As clicking that will lead them to hackathon Townsquare which no longer exists
> image

Highlighted section on the landing page shouldn't pose any problem as it redirects to townsquare only.

Also, the link onto the messages via hackathon will lead to townsquare only which exist. So it shouldn't break anything. But we can scrap out the messages.

thelostone-mc (approved):

Issue #7070, opened 2020-07-07
^register_hackathon/ endpoint should validate whether a user has already registered for a event


walidmujahid:
**Describe the bug**

Duplicate `HackathonRegistration`  instances created when multiple requests are sent to the ^register_hackathon/ endpoint.

**To Reproduce**

In my local docker build, I create a hackathon event, then I go to the onboarding page and click "Join Now". I continue to click and each click sends new request. 

**Expected behavior**

There should not be duplicate `HackathonRegistration` for the same user under the same hackathon. 

**Solutions** 

The `hackathon_registration` should check if there is aleady an `HackathonRegistration` instance for the user under the given hackathon event and if so then an error should be returned or nothing should happen.

Furthermore, on the user frontend, any only a single POST request should be sent to ^register_hackathon/ and the page should either refresh, redirect somewhere else, or show a success alert. For example, on the hackathon onboarding page, it may be prudent to disable the "Join Now" button after the first click and the page should show a success alert, redirect to the prize page, refresh to onboarding page, or a combination such as: disable button, show success alert, then automatically redirect to prize page after a few seconds if applicable, otherwise refresh the page.

PixelantDesign:
Thanks @walidmujahid! Are you interested in helping us with this one?

walidmujahid:
> Thanks @walidmujahid! Are you interested in helping us with this one?

Yes. I will be working on this unless someone had grabbed it. I should have posted yesterday my intention.

PR #7290, opened 2020-08-26
fix: duplicate registration on hackathon onboarding

walidmujahid:
To prevent duplicate hackathon registrations under an event:

- return error on POST to /register_hackathon/ if user already registerd
- if no poll on hackathon onboard page, disable 'Join Now' button on click
- if poll on hackathon onboard page, disable poll submit button on click

fixes https://github.com/gitcoinco/web/issues/7070

### DIFF
--- a/app/dashboard/templates/dashboard/hackathon/onboard.html
+++ b/app/dashboard/templates/dashboard/hackathon/onboard.html
@@ -475,6 +475,7 @@
           const data = {'name': name, 'referer': referer};
           const sendRegister = fetchData(url, 'POST', data, {'X-CSRFToken': csrftoken});
 
+          $('[data-registration]').addClass('disabled');
           $.when(sendRegister).then((response) => {
             _alert('You have now registered for this hackathon.', 'success');
             console.log(response)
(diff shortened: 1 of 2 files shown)


thelostone-mc (approved):

octavioamu:
@walidmujahid notice you are just adding a disable class not actually disabling the button + if the request fails should "reenable" the button.

walidmujahid:
> notice you are just adding a disable class not actually disabling the button

Are requests still sent after the disabled class added? I thought I checked for that. The idea was to prevent any new requests from being sent via the frontend button after the user first clicks to join an event by adding `disabled` to prevent further clicks. I will look at this later today.

> if the request fails should "reenable" the button.

I see. I completely missed this. I will be sure to fix it.

walidmujahid:
> notice you are just adding a disable class not actually disabling the button

@octavioamu Are requests still sent after the disabled class added? I thought I checked for that. The idea was to prevent any new requests from being sent via the frontend button after the user first clicks to join an event by adding `disabled` to prevent further clicks. I will look at this later today.

> if the request fails should "reenable" the button.

I see. I completely missed this. I will be sure to fix it.

walidmujahid:
Thinking back on how I did this, the click event is still running. I should have probably just set a conditional and exited -and added the class for possibly UX. Sorry for the trouble @octavioamu, I should have noticed the error before it was merged. I will try to get a fix out later today.

octavioamu:
No problem just add attr disabled I stead of class and remove it on error

PR #7307, opened 2020-08-31
updated user directory models/indexers to reflect proper production d…

androolloyd:
##### Description
Correct the UserDirectory Model and Indexers are not configured with correct types/schema requirements to match what is used in production

##### Refers/Fixes
Inability to index the current userdirectory


thelostone-mc (changes requested):
@androolloyd could you create a new migration instead updating the old one 
This has already been run on live

androolloyd:
Since we’re not managing the migrations with django it won’t affect anything, while the migration execution status is stored in the db as being called. 

No SQL is executed by it. 

These changes are to bring the python model in line with what’s actually being produced by prod data. 


I can create a migration that modifies the view but if we have an execution status already for this migration there isn’t a need to run anything again anyways.

thelostone-mc (approved):
Synced up with @androolloyd who's assured like the migrations files are not bring tracked and has no affect on the rest of the app.
Updating the migration file should be alright even though it's already in live

app/dashboard/views.py
```diff
@@ -877,7 +877,7 @@ def user_lookup(request):
         return HttpResponse(status=404)
 
     path = request.get_full_path().replace('/user_lookup', '')
-    remote_url = f'{"https" if not settings.DEBUG else "http"}://{settings.ELASTIC_SEARCH_URL}:9200{path}'
+    remote_url = f'{settings.HAYSTACK_ELASTIC_SEARCH_URL}{path}'
```
nicc938, line 880: nicc938

Issue #7310, opened 2020-09-01
bug: tribe sorting 


thelostone-mc:
#### Description

https://github.com/gitcoinco/web/pull/7171 introduced sorting of participants by their rank but this doesn't work as it attempts to sort within a paginated result.

Ideally, the sort has to be done on the backend and all the frontend changes added needs to be removed.

There is a revert PR if this cannot be addressed 
https://github.com/gitcoinco/web/pull/7295

PixelantDesign:
@sanchaymittal could you take a look at this?

PR #7350, opened 2020-09-10
Feature/tribe v2

sanchaymittal:
##### Description

1. Gitcoin staff can edit header view of tribes
- With this tribe header can be edited by gitcoin staff now.

2. Ordered User profile & Townsquare side-bar UI
- Allows to fetch correct ordering of user profile based on rank_coder
- sidebar UI elements will now show name only.
- before it was showing followers count only.

##### Refers/Fixes

#7310
#7295

##### Testing
https://www.loom.com/share/56038d60ecf34d3aadf6e3980c2f5fd3


app/assets/v2/js/vue-components.js
```diff
@@ -701,28 +700,23 @@ Vue.component('suggested-profile', {
...
+<b-media tag="li" class="row mx-auto mx-md-n1 mb-1">
   <template v-slot:aside>
     <a :href="profile_url" class="d-flex nav-link nav-line pr-0 mr-0">
       <b-img :src="avatar_url" class="nav_avatar"></b-img>
     </a>
   </template>
-  <div class="row">
-    <span class="col-6 col-md-12 col-xl-7 font-caption">
+  <div class="col">
+    <span class="row-6 row-md-12 row-xl-7 font-caption">
```
thelostone-mc, line 710: Does this work ?

app/assets/v2/js/users.js
```diff
@@ -9,7 +9,7 @@ let hackathonId = document.hasOwnProperty('hackathon_id') ? document.hackathon_i
 Vue.mixin({
   computed: {
     orderedUsers: function() {
-      return _.orderBy(this.users, 'position_contributor', 'asc');
+      return _.orderBy(this.users, 'rank_coder', 'desc');
```
thelostone-mc, line 12: The ordering should happen on the backend !
Do remove the `_` library and other changes as well which were introduced

thelostone-mc (changes requested):
left a few comments

app/assets/v2/js/vue-components.js
sanchaymittal: Yesss

sanchaymittal: change

cc: @PixelantDesign

octavioamu: what exactly are those row- classes?

octavioamu: '.row' is a bootstrap grid classe to hold cols, cols are responsive but doesn't make sense to me have row responsive as is not what control sizes.

octavioamu: you are using 4.5 booastrap doc, if you notice those row- doesn't exist on the bootstrap version we are using 
https://getbootstrap.com/docs/4.3/layout/grid/ 
If you use the inspector will notice those doesn't do nothing

app/assets/v2/js/users.js
sanchaymittal: removing it.

app/assets/v2/js/vue-components.js
sanchaymittal: Ohh right! I will make the changes based on v4.3 and push it.

thelostone-mc (approved):

octavioamu (approved):

PR #7356, opened 2020-09-11
zkSync updates part 3

mds1:
Final set of zkSync changes. This has been rebased against the integration branch, though I have not yet tested it here

### Changes

1. zkSync fees are now additive, instead of subtracting fees from the contribution amount. Subtractive approach caused errors when fee was greater than transfer amount (this was a common situation, where user donates $1, so $0.05 cents go to Gitcoin, and fee for that $0.05 cent transfer is $0.10). Note — we are very conservative on fee estimates here, as explained [here](https://github.com/ScopeLift/gitcoin-web/blob/52db97def188a0b654885142617d4ccb1313128c/app/assets/v2/js/cart.js#L1660-L1663)

2. Gas estimates and user recommendations about which checkout approach to use are implemented. We can improve estimates as the round goes on.

3. Transaction validator now covers all cases. 

### Testing

Here are the main flows that should be tested. For each case, we also need to make sure the transaction validator works as expected.

1. Ordinary L1 checkout
2. User has insufficient zkSync funds so deposits into zkSync. User stays on page during checkout, and only uses 1 token
3. User has insufficient zkSync funds so deposits into zkSync. User stays on page during checkout, and uses 2+ tokens
3. Repeat item 2, but refresh cart page after final web3 prompt (i.e. user closes tab because they don't want to wait 10 confirmations, so they come back later)
4. User has enough funds on zkSync so never sends L1 tx, and only uses 1 token
5. User has enough funds on zkSync so never sends L1 tx, and uses 2+ tokens


docs/GRANTS.md
```diff
@@ -1,6 +1,74 @@
...
+logging in to your zkSync account, such as at https://wallet.zksync.io/, you are asked to sign a
+message. That signature is used to generate a private key, and each transfer then requires two
+signatures&mdash;one from this private key, and as an additional security measure, one more from
+your regular web3 wallet. This is great from a security perspective, but not so great from a UX
+perspective. Requiring one signature for each transfer can be a pain if you want to send a lot of
+transfers, so you'd be doing a lot of clicking when checking out with large carts. 
+
+To remedy this, we ask you to login to a "Gitcoin zkSync" account. This account is created by
+asking you to sign a different message from the regular zkSync message, and this gives us an L1 account
+that the Gitcoin app has the private key to. We then login to zkSync with *this account*, and now
```
apbendi, line 21: Great write up! I would perhaps clarify this a bit, something like:

"this generates an L1 account that the Gitcoin frontend can use to sign transactions directly. The private key to this account comes from the signature, and is never stored or transmitted anywhere."

mds1: That sounds good! Updated in 7ff3ef4

thelostone-mc (approved):

PR #7417, opened 2020-09-16
Add 1.4x Multiplier Verification Case

frankchen07:
##### Description

If someone is Bright ID and SMS verified, they achieve a even higher multiplier. We want to account for this in the calculation. I've added 4 additional cases:

1.  x and y are both Bright and SMS verified (1.4x)
2. x is both and y is Bright (1.3x)
3. x is both and y is SMS (1.225x)
4. x is both and y is none (1x) 

There were also some weird merge conflicts. Parts of the CLR code looked a little outdated between the two versions, but I resolved them

##### Refers/Fixes

Doesn't fix anything, just an add-on.

##### Testing

`if` and `elif` logic follows previous iterations currently in the code.

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -205,8 +205,20 @@ def calculate_clr(aggregated_contributions, pair_totals, sms_verified_list, brig
 
             # pairwise matches to current round
             for k2, v2 in contribz.items():
+                # both (sms & bright)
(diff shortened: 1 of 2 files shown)


thelostone-mc:
Closing this out as we had discussed offline

PR #7301, opened 2020-08-30
Add Github Link to repo on Grant Detail info

molecula451:
##### Description

This PR will optionally request github link repo on the grants new form and append it to the grant details info for the looks

Screencast-2020-08-30-124937

##### Refers/Fixes

fix #7270

PR #7423, opened 2020-09-16
register field on admin

molecula451:
admin for #7301


thelostone-mc (approved):

thelostone-mc:
nice one @molecula451

PR #7456, opened 2020-09-17
Enable SMS Verification directly from Trust Bonus tab

apbendi:
Instead of kicking users over to the cart to verify their SMS, they can now do it directly from the Trust Bonus tab.

Screen Shot 2020-09-17 at 2 55 31 PM

Note: This PR extracts and modifies code from the cart to create a vue component used on the Trust Bonus tab. The component is NOT currently shared with the cart, but should be, to DRY things up. This needed refactoring is noted in TODO comments, and should be tackled post Round 7.

### DIFF
--- a/app/assets/v2/js/pages/profile-trust.js
+++ b/app/assets/v2/js/pages/profile-trust.js
@@ -139,3 +139,231 @@ let show_brightid_verify_modal = function(brightid_uuid) {
   $(content).appendTo('body');
   $('#verify_brightid_modal').bootstrapModal('show');
 };
+
+Vue.component('sms-verify-modal', {
+  delimiters: [ '[[', ']]' ],
+  data: function() {
+    return {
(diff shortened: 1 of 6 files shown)


app/assets/v2/js/pages/profile-trust.js
```diff
@@ -139,3 +139,234 @@ let show_brightid_verify_modal = function(brightid_uuid) {
...
+
+  $(document).on('click', '#verify_offline', function(e) {
+    $(this).remove();
+    $('#verify_offline_target').css('display', 'block');
+  });
+});
+
+if (document.getElementById('gc-sms-modal')) {
+
+  const app = new Vue({
```
octavioamu, line 367: can we do something like const appTrust  here? I want to avoid coalitions with others apps loaded on the same page.

octavioamu (changes requested):
just a comment, rest look good to me

octavioamu (approved):

PR #7457, opened 2020-09-17
User Directory Pagination Size change from 10 -> 10000

androolloyd:
##### Description
Temporary fix for increased usability while a proper fix is finished.

##### Refers/Fixes

Allows the csv export to export more than 10 users(the default page size)


owocki:
let me cherry-pick to stable so i can test; i dont have the UD setup on my local

owocki:
per user directory channel convo; this wont work.  new approach was discussed in the user directory channel

PR #7464, opened 2020-09-18
Add hackathons events

zoek1:
##### Description
image

### DIFF
--- a/app/assets/onepager/css/main.css
+++ b/app/assets/onepager/css/main.css
@@ -544,6 +544,23 @@ body #main a:not(.button):hover {
   color: #ec5597;
 }
 
+
+body #main a.no-pink {
+  color: var(--gc-blue);
+}
+
+body #main a.no-pink:hover {
+  color: var(--gc-blue-hover);
+}
+
+body #main a.no-pink.btn-gc-blue {
+  color: #ffffff;
+}
+
+body #main a.text-black.no-pink {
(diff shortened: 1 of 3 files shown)


jdorfman:
Looking good @zoek1! Is there a way to add UTM's to the "Join" buttons?

jdorfman:
@PixelantDesign @octavioamu this will replace the Disqus widget that has 2+ years of stale content on the [gas history page](https://gitcoin.co/gas/history). This page gets a lot of organic search traffic (see GAnalytics) and advertising our current hackathons seems like a better use of space.

thelostone-mc:
@zoek1 could we make the code reusable ! I believe we have the exact same format on the hackathon list page 
So ideally make that into it's own file and import it in both places

thelostone-mc (changes requested):
DRY code needed

jdorfman:
@thelostone-mc will this allow @zoek1 to add custom UTM's so we know where the sign-ups are coming from?

thelostone-mc:
@jdorfman So this would be mapped to the buttons right ?  yeah that's possible even with DRY code

thelostone-mc:
@jdorfman closing this out as this code is also present is open in PR https://github.com/gitcoinco/web/pull/7502/commits

jdorfman:
Thanks @thelostone-mc !

PR #7473, opened 2020-09-19
Make computed properties reactive to document.web3network changes

mds1:
### Problem

The list of supported zkSync tokens is stored as a computed property called `zkSyncSupportedTokens`. It's dependent on `document.web3network`. Vue cannot react to changes in `document`, so when the network changed this was not updated.

This means if a user visits the cart, then connects their wallet, the network change would not be detected, and `zkSyncSupportedTokens` would return an empty list. As a result, they would not be able to checkout with zkSync.

### Solution

We fix this by adding a new property to the component. Vue can watch and react to changes in this new property, and we update this new property when the `dataWalletReady` event is detected. We also make the mainnet token list the default when nothing is detected

cc @octavioamu @thelostone-mc

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -329,11 +329,13 @@ Vue.component('grants-cart', {
 
     // Array of supported tokens
     zkSyncSupportedTokens() {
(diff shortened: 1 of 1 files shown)


mds1:
I also added a commit to add SNT support here since that was recently added by zkSync

PR #7506, opened 2020-09-23
fix(dashboard/views): invalid attribute call

chibie:
##### Refers/Fixes

https://sentry.io/organizations/gitcoin/issues/1839085529/?project=1398424


thelostone-mc:
@chibie I see commits from your previous PR as well! Could we fix that up?

thelostone-mc (approved):

PR #7510, opened 2020-09-23
Start tx validator v2, and fix bug in zkSync validations

mds1:
Creates a new validator that will be updated as necessary going forward. Also fixes a bug in the zkSync validator.

cc @owocki


app/economy/tx.py
```diff
@@ -294,19 +482,26 @@ def get_token_originators(to_address, token, from_address='', return_what='trans
     endpoint = 'token-transfers' if token != '0x0' else 'ether-transfers'
     url = f'https://api.aleth.io/v1/{endpoint}?filter[to]=' + address + '&filter[token]=' + token + '&page%5Blimit%5D=100'
     if token == '0x0':
+        print('qqqq')
         url = f'https://api.aleth.io/v1/{endpoint}?filter[account]=' + address + '&page%5Blimit%5D=100'
     if from_address:
+        print('wwww')
```
thelostone-mc, line 488: ^ do we wanna remove all the print stmt :P

octavioamu, line 488: do we need the prints?

octavioamu (changes requested):

app/economy/tx.py
mds1: Oops, yea those were leftover from my expert debugging techniques. Removed in f61690e24012169ba8d24a85fb99dc21edf26e24

mds1: Oops, yea those were leftover from my expert debugging techniques. Removed in f61690e24012169ba8d24a85fb99dc21edf26e24

thelostone-mc (approved):

PR #7489, opened 2020-09-21
zkSync UX Improvements

mds1:
Resolves #7397, #7469

### Changes

1. To fix https://github.com/gitcoinco/web/issues/7397: When the cart loads we check the DB for a deposit tx hash from the user to see if their checkout was interrupted. If it was, we now have a check to see if the transaction failed. If it did fail, we mark the user as "not interrupted" so they can try checking out again

2. To fix https://github.com/gitcoinco/web/issues/7469: During zkSync checkout Flow B (user makes L1 deposit), the zkSync fees were not shown to the user anywhere, so we now show them at the bottom of the modal along with a link to the docs/GRANTS.md file to learn more about fees and how the checkout flow works. The modal is shown in the screenshot below. (Note that we don't need to show these fees in Flow A since the user signs a message containing the fee amount and that wallet should show that message to the user)

image

cc @willsputra

PR #7511, opened 2020-09-23
More zkSync UX updates

mds1:
## Changes

1. `cart.js` now has an `isZkSyncDown` flag. This PR sets it to true. When true the modal looks like the screenshot below when loading. **We must remember to set this back to false when things are ok!**

2. Fixes a bug in https://github.com/gitcoinco/web/pull/7489 where the full fee string was not always displayed at checkout

image

@owocki


owocki:
@thelostone-mc look ok to u?

PR #7519, opened 2020-09-24
Implement Trust Bonus Twitter Verification

apbendi:
Add's Twitter verification to the Trust Bonus tab.

Screen Shot 2020-09-23 at 10 04 12 PM

Adds two new fields to the Profile model: `is_twitter_verified`, which can be used for the CLR Matching functionality, and `twitter_handle`, which is the twitter name they verified as theres.

This PR does NOT add any interface for adding Twitter to your profile, or updating your Twitter. These features can be added at a later date.

On the Trust Bonus screen, there is a Vue based modal flow for sending the required  tweet, entering your twitter name, and having the presence of the Tweet validated.

Screen Shot 2020-09-23 at 10 04 22 PM

Screen Shot 2020-09-23 at 10 04 44 PM

Screen Shot 2020-09-23 at 10 05 00 PM


app/dashboard/models.py
```diff
@@ -2871,6 +2871,8 @@ class Profile(SuperModel):
     objects_full = ProfileQuerySet.as_manager()
     brightid_uuid=models.UUIDField(default=uuid.uuid4, unique=True)
     is_brightid_verified=models.BooleanField(default=False)
+    is_twitter_verified=models.BooleanField(default=False)
+    twitter_handle=models.CharField(blank=True, max_length=15)
```
thelostone-mc, line 2875: could we set null=True as well ? else we won' be able to edit Profiles via admin if they don't have a value for twitter_handle

apbendi: Done!

thelostone-mc (approved):

PR #7522, opened 2020-09-24
n:n matching rounds <> grants

owocki:
##### Description

Hello, this PR introduces the ability to add a grant to multiple GrantCLR Matching Rounds.

It does this by abstracting the clr_prediction_curve into the `GrantCLRCalculation` object, which is fk'd to a grant and a GrantCLR object.

when the frontend accesses clr_prediction_curve array, it has no idea that is actual interacting with multiple grant clr rounds. ha! stupid frontend

anyway, this PR allows grants to be in multiple clr rounds in at once. hoorayyyyy for foriegn keys!

We manage the frontend to this interface by changing the clr_prediction_curve on the Grant object to compute the clr_preidction_curve across all grants as a property.  this manages the complexity of this information from user/frontendn perspective.

##### Refers/Fixes

the vision

##### Testing

`./manage.py estimate_clr` then poke around the frontend

Screen Shot 2020-09-23 at 10 57 13 PM
Screen Shot 2020-09-23 at 10 56 29 PM


app/grants/models.py
```diff
@@ -1087,8 +1119,18 @@ def successful_contribution(self, tx_id):
 
 @receiver(pre_save, sender=Grant, dispatch_uid="psave_grant")
 def psave_grant(sender, instance, **kwargs):
+    # TODO: move to grant task
+    instance.clr_prediction_curve = instance.calc_clr_prediction_curve
+    instance.clr_round_num = instance.calc_clr_round_nums
+
+    # todo: move to grant task
+    active_clr_rounds = GrantCLR.objects.filter(is_active=True)
```
owocki, line 1127: @thelostone-mc theres probably a smarter way to manage the `instance.active_clr_rounds` object i have not found it yet. kind of feels like it should be a computed property & not a fk to me, but open to feedback of course.

app/grants/models.py
```diff
@@ -316,18 +328,6 @@ class Meta:
...
         help_text=_('The fundingamount across all rounds with phantom funding'),
     )
-    # TODO-CROSS-GRANT: [{round: fk1, value: time}]
-    clr_prediction_curve = ArrayField(
-        ArrayField(
-            models.FloatField(),
-            size=2,
-        ), blank=True, default=list, help_text=_('5 point curve to predict CLR donations.'))
-    # TODO: REMOVE
-    backup_clr_prediction_curve = ArrayField(
```
thelostone-mc, line 326: ohh nice nice i like

app/grants/models.py
```diff
@@ -375,6 +375,12 @@ class Meta:
 
     funding_info = models.CharField(default='', blank=True, null=True, max_length=255, help_text=_('Is this grant VC funded?'))
 
+    clr_prediction_curve = ArrayField(
+        ArrayField(
+            models.FloatField(),
+            size=2,
+        ), blank=True, default=list, help_text=_('5 point curve to predict CLR donations.'))
```
thelostone-mc, line 382: ^ This works for now -> but curious if you've given any thought on what the this would be the total accumulated match irrespective of chains / would we wanna move this to a table by itself which has 
`grant_fk` | `clr_round_fk` | `clr_prediction_curve` which would store finer details ?

Grant A is part of Zcash Tech & Eth Tech 
`clr_prediction_curve` -> would be a property which show the total prediction match across all active rounds/chains

and the new table would would have 2 entries in fk table 
`grant_fk` | `clr_round_fk` | `clr_prediction_curve` with clr_prediction_curve for both Zcash & EthTech

app/grants/models.py
thelostone-mc: @owocki I think this can be removed 
This function does the same 
https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L1312

owocki: per our slack convo; we are keeping it right? so that if a grant is updated its fk's are updated

app/grants/models.py
owocki: > the this would be the total accumulated match irrespective of chains / would we wanna move this to a table by itself which 

not sure i totally understand the question (typo in the middle?) but the GrantCLRCalculation basically hadnles this i think

i think if we want to extend the match estimates to only look at one (or a subset of) of curves, the existing FK structure handles that.  i dont see a strong reason to store it in the grant-level one

owocki:
responded to both queries

thelostone-mc:
@danlipert what we spoke about the other day !
it looks like this gets us there

frankchen07:
if a grant an be in multiple CLR rounds, does that mean we're calculating the CLR prediction over all CLR rounds for a particular grant, or does the grant have a different prediction in each CLR round?

owocki:
@frankchen07 both!  checkout the models.py changes in this PR; there is now a `GrantCLRCalculation` object that contains the prediction per round, and the `clr_prediction_curve` for a round is the sum of all `GrantCLRCalculation.clr_prediction_curve` for each round it is in

danlipert (approved):

app/grants/models.py
```diff
@@ -1824,3 +1866,21 @@ class GrantStat(SuperModel):
...
     def __str__(self):
         return f'{self.snapshot_type} {self.created_on} for {self.grant.title}'
+
+
+class GrantCLRCalculation(SuperModel):
+
+    latest = models.BooleanField(default=False, db_index=True, help_text="Is this calc the latest?")
+    grant = models.ForeignKey(Grant, on_delete=models.CASCADE, related_name='clr_calculations',
+                              help_text=_('The grant'))
+    grantclr = models.ForeignKey(GrantCLR, on_delete=models.CASCADE, related_name='clr_calculations',
```
thelostone-mc, line 1876: @owocki why do both of these have the same related name ?

thelostone-mc:
i'll be pulling this in and fixing up conflicts and testing out !

app/grants/models.py
owocki: bc they are both `clr_calculations`, just reference from a diff spot.

owocki:
thx; onto master?

thelostone-mc:
@owocki nope! it goes into https://github.com/gitcoinco/web/pull/7870 cause this touches DB and would need regression from team as well

PR #7544, opened 2020-09-28
User Directory UI Fixes and Filter updates

androolloyd:
##### Description
Fixes for the UI as well as some filter data being unavailable, some performance issue with the production set being loaded in terms of varying options, will work on making it an autocomplete system with a better typeahead setup.


thelostone-mc (approved):

app/dashboard/tasks.py
```diff
@@ -4,6 +4,7 @@
 import os
 from datetime import datetime
 
+from appdirs import unicode
```
thelostone-mc, line 7: @androolloyd CI fails cause of this

thelostone-mc:
@androolloyd looks like travis is erroring out cause of the migration rename -> mind just creating a new migration?

Issue #7512, opened 2020-09-23
make grant collections unfurl image into the grant collectoin on-sit eimage


owocki:
http://localhost:8000/grants/collections?collection_id=1

the twitter unfurl image for this page should be an collage of the first 4 grants' images in the cart.

<img width="584" alt="Screen Shot 2020-09-23 at 10 46 18 AM" src="https://user-images.githubusercontent.com/513929/94043388-17830c80-fd8a-11ea-806f-f9f169f12739.png">

on-site it already looks good: 

<img width="307" alt="Screen Shot 2020-09-23 at 10 46 12 AM" src="https://user-images.githubusercontent.com/513929/94043398-1a7dfd00-fd8a-11ea-9425-a5468feef880.png">

please use PIL to make it so!

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.5 ETH (169.12 USD @ $338.24/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/7512/100023762).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/7512/100023762).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://chat.gitcoin.co/'>Gitcoin Chat</a>
* $737,277.06 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 266 years, 1 month from now.
Please review their action plans below:


**1) [zoek1](https://gitcoin.co/zoek1) has been approved to start work.**

I implemented the collections, currently i'm creating the collage. I didn't know exists a bounty for this
**2) [hhio618](https://gitcoin.co/hhio618) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/web/7512/100023762?mutate_worker_action=approve&worker=hhio618) | [reject worker](https://gitcoin.co/issue/gitcoinco/web/7512/100023762?mutate_worker_action=reject&worker=hhio618))_.**

[...]

PR #7587, opened 2020-10-02
Generate collection thumbnail

zoek1:
##### Description
https://www.loom.com/share/df2bc6a33c38487d956e05a9eb552353
DeepinScreenshot_select-area_20201002012410

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/7512


app/grants/views.py
```diff
@@ -2247,3 +2252,18 @@ def add_grant_from_collection(request, collection_id):
     return JsonResponse({
         'grants': grants,
     })
+
+
+@login_required
+@staff_member_required
```
thelostone-mc, line 2258: ^ why does need a login + staff decorator?

zoek1: This Is for development and debug purposes like the email views.

thelostone-mc:
@zoek1 could you fix up the conflicts as well

zoek1:
conflicts fixed @thelostone-mc

app/grants/views.py
```diff
@@ -2277,3 +2282,18 @@ def add_grant_from_collection(request, collection_id):
...
+def collection_thumbnail(request, collection_id):
+    width = int(request.GET.get('w', 600))
+    height = int(request.GET.get('h', 400))
+    collection = GrantCollection.objects.get(pk=collection_id)
```
thelostone-mc, line 2292: ^ try + catch ?

app/grants/templates/grants/components/collection.html
```diff
@@ -33,9 +33,10 @@
               <img :src="getGrantLogo(3)" alt="" v-if="collection.grants.length >= 4">
             </div>
           </div>
-          <div class="collection-item__owner-image--cover">
+          <div class="collection-item__owner-image--cover" v-if="!collection.cover">
             <img :src="collection.owner.avatar_url || 'https://c.gitcoin.co/avatars/57e79c0ae763bb27095f6b265a1a8bf3/thelostone-mc.svg'">
```
octavioamu, line 37: wait this is from the old code but if not user is default to thelostone avatar? users always have an avatar_url, if not the fallback should be github one, the same way we do in other places

octavioamu (changes requested):
looking good just left a comment about a default user avatar

PixelantDesign (changes requested):
Could we make the dark border thinner and put a circle behind the avatar, and make it 30% larger? 

See example here

Screen Shot 2020-10-13 at 1 23 21 AM

zoek1:
what do you think @PixelantDesign? 
image

app/grants/views.py
zoek1: IMHO since this method is for development purposes is better to show the exceptions instead of caching them.

chibie (approved):

PR #7590, opened 2020-10-02
Email users with pending contributions

zoek1:
##### Description
image

Check email local
`http://localhost:8000/_administration/email/pending_contributions`

Send emails
`docker-compose exec -e DJANGO_SETTINGS_MODULE="app.settings" web python3 app/manage.py  pending_contributions`

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/7497


thelostone-mc (approved):

thelostone-mc:
@mds1 / @owocki how often should we trigger the mgmt command ?
Does it need be in the crontab / do we invoke it when we need to ?

mds1:
@thelostone-mc Right now, given that the round is just about over, I think we only need to send these emails one time. For next round we can revisit to make it a little smarter if needed

owocki:
concur with matt

app/grants/management/commands/pending_contributions.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Command(BaseCommand):
+
+    help = 'Send reminder for pending grant contributions'
+
+    def handle(self, *args, **options):
+        contributions = Contribution.objects.filter(
+            validator_comment__contains="User may not be aware so send them email reminders")
+        for contribution in contributions:
```
thelostone-mc, line 32: @mds1 adding a print statement here would do the trick

app/marketing/mails.py
```diff
@@ -407,6 +408,21 @@ def successful_contribution(grant, subscription, contribution):
...
+    from_email = settings.CONTACT_EMAIL
+    to_email = contribution.subscription.contributor_profile.email
+    cur_language = translation.get_language()
+
+    try:
+        setup_lang(to_email)
+        html, text, subject = render_pending_contribution_email(contribution)
+
+        if not should_suppress_notification_email(to_email, 'pending_contribution'):
+            send_mail(from_email, to_email, subject, text, html, categories=['transactional', func_name()])
```
mds1, line 421: Print statement here

mds1:
Hey @zoek1, it seems this doesn't filter out users and would send a user multiple emails. I added two items to my cart, did a checkout with zkSync, updated the validator comments as required, then ran this script with those print statements in the two lines of code above. And the print statements ran twice which shows that it was sending an email twice to the same user

PR #7652, opened 2020-10-11
decreaseses the login from from 7s to 2.3s by making the sync_profile tasks happen on a delay

owocki:
##### Description

decreaseses the login from from 7s to 2.3s by making the sync tasks happen on a delay

##### Refers/Fixes

self evident

##### Testing

tested locally

### DIFF
--- a/app/app/utils.py
+++ b/app/app/utils.py
@@ -185,7 +185,20 @@ def get_upload_filename(instance, filename):
     return f"docs/{getattr(instance, '_path', '')}/{salt}/{file_path}"
 
 
-def sync_profile(handle, user=None, hide_profile=True):
+def sync_profile(handle, user=None, hide_profile=True, delay_okay=False):
+    from dashboard.models import Profile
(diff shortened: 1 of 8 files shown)


app/app/utils.py
```diff
@@ -185,7 +185,20 @@ def get_upload_filename(instance, filename):
...
+    delay = delay_okay and profile
+    if delay:
+        from dashboard.tasks import sync_profile as sync_profile_task
+        user_pk = user.pk if user else None
+        sync_profile_task.delay(handle, user_pk, hide_profile)
+    else:
+        actually_sync_profile(handle, user=user, hide_profile=hide_profile)
+
+
+def actually_sync_profile(handle, user=None, hide_profile=True):
```
owocki, line 201: many of the chat related tasks here could be done on a delay as well; a further PR  would possibly do that so that upon first login were not blocking on things like chat tasks to create the profile

thelostone-mc:
getting this into master and then into stable for this weeks deploy

PR #7685, opened 2020-10-15
refactor of grantclr to make the round_num an actual int + add a subround/customer field

owocki:
##### Description

 what do you all think about changing the round num to be 7 for these grants https://bits.owocki.com/z8uY64nB
and then making a ‘subround_slug’ field that contains the info about which category/detail it is for.  i worry that were not using the fields correctly here, and its going to byte us ( get it!!!) when we try to do an infra-tech round for round 8.

##### Refers/Fixes

slack https://gitcoincore.slack.com/archives/G01AL7XRCF2/p1602783199004000

##### Testing

not tested fully; this needs to be finished up & tested


owocki:
@apbendi @thelostone-mc do you all think this is a worthwhile effort?

the migrations run on this, and i edited some code paths.. but we'd need to fully test it + clean it up before moving forward fully

thelostone-mc:
@owocki i do see a few todo here! is that something where you still need to throw in code ?

owocki:
yes there is still work to be done.

it'd be great if people would respond to my questions about whether i'm on the right track before i put in the effort to continue rewriting this.

> @apbendi @thelostone-mc do you all think this is a worthwhile effort?

i was gating on the response to this, but from slack it seems like at least you aditya think its a good idea.

app/grants/models.py
```diff
@@ -118,7 +118,10 @@ def active_clrs_sum(self):
 
 
 class GrantCLR(SuperModel):
-    round_num = models.CharField(max_length=15, help_text="CLR Round Number")
+    
+    customer_name = models.PositiveIntegerField(help_text="CLR Round Number")
```
thelostone-mc, line 122: @owocki what is the need for customer name?

app/grants/urls.py
```diff
@@ -31,8 +31,8 @@
 app_name = 'grants'
 urlpatterns = [
     path('', grants, name='grants'),
-    path('clr/<slug:round_num>', clr_grants, name='clr_grants'),
-    path('clr/<slug:round_num>/', clr_grants, name='clr_grants'),
+    path('clr/<int:round_num>', clr_grants, name='clr_grants'),
+    path('clr/<int:round_num>/<str:sub_round_slug>', clr_grants, name='clr_grants'),
```
thelostone-mc, line 35: why would you have round num and slug ?
Is this to group all the clr for round 8 under clr/8/...?

For custom CLR like matic / zcash 
I'm not sure if this makes sense 

I'd lean towards 
- keeping how it was 
- keeping both

app/grants/views.py
```diff
@@ -332,11 +332,11 @@ def bulk_grants_for_cart(request):
     return JsonResponse({'grants': grants})
 
 
-def clr_grants(request, round_num):
+def clr_grants(request, round_num, sub_round_slug=''):
     """CLR grants explorer."""
 
     try:
-        clr_round = GrantCLR.objects.get(round_num__icontains=round_num)
+        clr_round = GrantCLR.objects.get(round_num=round_num, sub_round_slug=sub_round_slug)
```
thelostone-mc, line 339: if sub_round_slug is empty -> you would get more than one result!

app/grants/models.py
owocki: this is for if we need to do a custom round for zcash, matic, filecoin etc.

app/grants/urls.py
owocki: round num = 7
slug = community, infra tech, or dapp tech

there can be multiple slugs in the same round.

chibie:
@thelostone-mc @owocki

how does this affect filecoin, matic, zcash clr?

if,

```
round num = 7
slug = community, infra tech, or dapp tech
```

will filecoin/matic/zcash be slug?

app/grants/views.py
chibie: fixed.

owocki:
pls make sure this goes into the `grantsround8` branch that @thelostone-mc is creating; not master. probably shouldnt go out ntil grants round8 does

thelostone-mc:
Okay i synced up with kevin on slack

we have 3 identifiers for constructing URL
- round_num : int (compulsory)
- customer_name: string (optional)
- sub_round_slug: string (optional)



If `customer_name` **is set** , `sub_round_slug` **is not set**
URL :`/clr/<customer_name>/<round_num>`

If `customer_name` **is not set**,  `sub_round_slug` **is set**
URL: `/clr/<round_num>/<sub_round_slug>`

If `customer_name`  **is not set**,  `sub_round_slug`  **is not set**
URL: `/clr/<round_num>`

If `customer_name` **is set**,  `sub_round_slug` **is set**
URL : `/clr/<customer_name>/<round_num>/<sub_round_slug>`

app/grants/views.py
```diff
@@ -386,11 +386,18 @@ def get_grants(request):
...
     clr_round = None
     try:
         if round_num:
-            clr_round = GrantCLR.objects.get(round_num=round_num)
+            params = {
+                'round_num': round_num,
+                'sub_round_slug': sub_round_slug,
+                'customer_name': customer_name
+            }
+            clr_round = GrantCLR.objects.get(**params)
```
danlipert, line 400: @chibie love me some **splat!

thelostone-mc:
@chibie looks there are a few conflicts! could you fix them up ?

chibie:
> @chibie looks there are a few conflicts! could you fix them up ?
> 
> migrations would need to be recreated
> Also with round_num being created. this will break
> https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L2278
> 
> just update round_num to be 0 !
> All CLR created by pledges have to be set to the same round_num and the admin will have to change it before launches

@thelostone-mc are you saying `round_num = 'plegde'` should be `round_num = 0` ?

maybe `sub_round_slug` can be 'pledge' ?

thelostone-mc:
@chibie that works for me

thelostone-mc:
@chibie that works for me

chibie:
@thelostone-mc all set here!

PR #7686, opened 2020-10-15
Contribution ingest script updates

mds1:
Update contribution ingestion script:
- Only use checksum addresses
- Add comments to clarify input arguments


thelostone-mc (approved):

chibie (approved):

danlipert (approved):
Looks good - will have to have a v2 or update this one once we move to batch txns right? so hopefully round 9

mds1:
Yep, we will need to update once we implement the zksync batch txs, since there no longer will be a Gitcoin zkSync address. It should only be a small change though

PR #7704, opened 2020-10-19
Add project submission

### LINKED ISSUE #7695: Subit work error for non integer hours worked. ".5"
**Describe the bug**
On submitting work I get a very non-intuitive error: "Unable to fulfill bounty. Please try again later" as a popup when all fields are completed. 

I stumbled around trying to figure the issue out, and only **after looking in the console did I find a hint**:
`error: bounty fulfilment failed with status: 400 and message: error: missing hoursWorked`

I tried again with "1" for hours, and it processed fine. 

**To Reproduce**
1. submit work for bounty with all fields complete, hours worked as non integer (with no leading 0 -- ".5"

**Expected behavior**
non-integer values for time worked actually work, or user MUST bet an error message in GUI that makes sense. 
(could be I needed a leading 0?)

octavioamu:
##### Description

if start work project post failed the user will not able to submit work
Gives the user the ability to create a project on submission if there is not projects.
working hours not decimals 

image

##### Refers/Fixes

fix #7695

### DIFF
--- a/app/assets/v2/js/hackathon-projects.js
+++ b/app/assets/v2/js/hackathon-projects.js
@@ -29,7 +29,11 @@ const submitProject = (logo, data, callback) => {
     if (callback) {
       callback(response);
     }
+    if (document.location.pathname.includes('fulfill')) {
+      document.location.reload();
+    }
     return _alert({message: response.msg}, 'info');
+
 
   }).fail(function(data) {
     _alert(data.responseJSON['error'], 'error');
--- a/app/dashboard/templates/bounty/fulfill.html
+++ b/app/dashboard/templates/bounty/fulfill.html
@@ -132,7 +132,7 @@
                   <div class="mt-2 row">
                     <div class="col-12">
                       <label class="form__label" for="hoursWorked">{% trans "Hours Worked" %}</label>
-                      <input name='hoursWorked' id='hoursWorked' class="form__input font-body" type="number" placeholder="1 (feel free to approximate)" value="" min="0" step="1" required/>
+                      <input name='hoursWorked' id='hoursWorked' class="form__input font-body" type="number" placeholder="1 (feel free to approximate)" value="" min="0" step="0.1" required/>
                     </div>
 
                     <div class="col-12">
--- a/app/dashboard/templates/bounty/fulfill.html
+++ b/app/dashboard/templates/bounty/fulfill.html
@@ -93,6 +93,12 @@
                         {% endfor %}
                       </select>
                     </div>
+
+                  {% elif bounty.event %}
+                    <div class="mt-2">
+                      <label class="form__label" for="projectId">{% trans "Please create your project:" %}</label>
+                      <button type="button" onclick="projectModal({{bounty.pk}})" class="btn btn-sm btn-gc-green d-block">Add project</button>
+                    </div>
                   {% endif %}
 
                   {% if bounty.web3_type == 'fiat' %}
@@ -188,7 +194,7 @@
     {% include 'shared/footer_scripts.html' %}
     {% include 'shared/footer.html' %}
 
-  </body>
+  {% include 'shared/current_profile.html' %}
 
   <script>
     $('[data-toggle="popover"]').popover()
@@ -210,6 +216,8 @@
       });
     })
   </script>
+    <script src="{% static "v2/js/hackathon-projects.js" %}"></script>
+    <script src="{% static "v2/js/user-search.js" %}"></script>
 
   {% if is_bounties_network %}
     <script src="{% static "v2/js/abi.js" %}"></script>
@@ -224,5 +232,6 @@
   {% endif %}
   <script src="{% static "v2/js/rating.js" %}"></script>
   <script src="{% static "v2/js/pages/fulfill_bounty/index.js" %}"></script>
+</body>
 
 </html>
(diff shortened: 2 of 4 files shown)


app/dashboard/templates/bounty/fulfill.html
```diff
@@ -126,7 +132,7 @@ <h4 class="text-center mb-4">{% trans "Submit Work" %}</h4>
                   <div class="mt-2 row">
                     <div class="col-12">
                       <label class="form__label" for="hoursWorked">{% trans "Hours Worked" %}</label>
-                      <input name='hoursWorked' id='hoursWorked' class="form__input font-body" type="number" placeholder="1 (feel free to approximate)" value="" min="0" step="0.1" required/>
+                      <input name='hoursWorked' id='hoursWorked' class="form__input font-body" type="number" placeholder="1 (feel free to approximate)" value="" min="0" step="1" required/>
```
thelostone-mc, line 135: why do we do this ?

octavioamu: because https://github.com/gitcoinco/web/issues/7695

octavioamu: we are encouraging decimals on the ui when actually the backend doesnt accept it

app/assets/v2/js/hackathon-projects.js
```diff
@@ -29,8 +29,12 @@ const submitProject = (logo, data, callback) => {
     if (callback) {
       callback(response);
     }
+    if (document.location.pathname.includes('fulfill')) {
+      document.location.reload();
```
thelostone-mc, line 33: why do we reload ? i didn't get this 😅

app/assets/v2/js/hackathon-projects.js
```diff
@@ -29,8 +29,12 @@ const submitProject = (logo, data, callback) => {
...
+    }
```
aamustapha, line 34: Wouldn't this lead to recursion with no breakout?

app/dashboard/templates/bounty/fulfill.html
aamustapha: I think we could watch and do a Math.ceil.

app/assets/v2/js/hackathon-projects.js
octavioamu: is not the most elegant way I know but the form need the project pk and that hackathon projects is used in many places, so in some point we can just add that to the data form, but that will need a lot of changes on the form approach, and as we are planning to move that form to vue doesn't make sense to invest a lot of time. fair?

aamustapha (changes requested):
I'm curious

chibie:
@octavioamu ux review: instead of row for creating a project with "Add Project" button, may be better to include a text like "You can create a project here if you haven't" under the project URL input as a helptext with hyperlinked "create a project" to trigger the modal.

app/dashboard/templates/bounty/fulfill.html
```diff
@@ -93,6 +93,12 @@ <h4 class="text-center mb-4">{% trans "Submit Work" %}</h4>
                         {% endfor %}
                       </select>
                     </div>
+
+                  {% elif bounty.event %}
+                    <div class="mt-2">
+                      <label class="form__label" for="projectId">{% trans "Please create your project:" %}</label>
```
chibie, line 99: @octavioamu ux review: instead of row for creating a project with "Add Project" button, may be better to include a text like "You can create a project here if you haven't" under the project URL input as a helptext with hyperlinked "create a project" to trigger the modal.

chibie (changes requested):

app/assets/v2/js/hackathon-projects.js
octavioamu: no because this post action is just to create a project not to submit the submission form once the project exist this post will not be call anymore.

app/dashboard/templates/bounty/fulfill.html
octavioamu: do you want to add the backend validation Abdul?

octavioamu: right now is `if not hours_worked or not hours_worked.isdigit():`

thelostone-mc: the model does support decimals 
`fulfiller_hours_worked = models.DecimalField(null=True, blank=True, decimal_places=2, max_digits=50)`

thelostone-mc: i know this was a specific requirement from scott a while back

app/dashboard/templates/bounty/fulfill.html
octavioamu: image
check the image, the project link (in some point need to be named diff)  is just the link or where the code is.
the project select is for hackathons, contributors create a project based on their submission but is not related to the project link (but probably will have the same link)

octavioamu: image
this is the project form

app/dashboard/templates/bounty/fulfill.html
octavioamu: mm probably is because the comma vs dot?

app/dashboard/templates/bounty/fulfill.html
chibie: @octavioamu okay this makes sense... 💯

chibie (approved):

app/dashboard/templates/bounty/fulfill.html
octavioamu: image

octavioamu: seems the problem is with hours_worked.isdigit()

thelostone-mc: yup that's the issue we have to update that to check if has numbers or . and not characters

PR #7712, opened 2020-10-20
Elastic User Directory / sharable searches v0.

androolloyd:
##### Description
Updates for Sharing Searches. 
- [x] Url Params are now fed into the internal store and the UI is updated to reflected the loaded searches. 
- [x] Url Rewrites for sharing, as the search is updated the URL is adjusted as well to enable easy sharing.


app/assets/v2/js/shared.js
```diff
@@ -1010,6 +1010,75 @@ function shuffleArray(array) {
   return array;
 }
 
+
+const getAllUrlParams = () => {
```
octavioamu, line 1014: maybe this function can be a lot more simple with https://developer.mozilla.org/pt-BR/docs/Web/API/URLSearchParams

danlipert (approved):

thelostone-mc (approved):

PR #7713, opened 2020-10-20
fixes/convert2grant

sanchaymittal:
##### Description 

fixes for convert to grant feature.
- condition check if grant exists for the project
- condition check if the grant is created by project team members.
- Add support of the project to grant for l2 matic grants.

video demo: https://drive.google.com/file/d/1xON7Y-FUxwG4eqIghvSAdkX6JmfnXQb6/view?usp=sharing

Note: @octavioamu instructions to create this pr to stable.
https://gitcoincore.slack.com/archives/CDML7FS7M/p1602701818068000?thread_ts=1602698951.056900&cid=CDML7FS7M


PixelantDesign (approved):

danlipert:
@sanchaymittal I don't think that slack link is working, at least it isnt for me. Can you share the contents here?

app/grants/views.py
```diff
@@ -1400,6 +1400,30 @@ def grant_new_whitelabel(request):
 
     profile = get_profile(request)
 
+    project = None
+    project_id = request.GET.get('project_id', None)
+    if project_id is not None:
+        hackathon_project = HackathonProject.objects.filter(pk=project_id).nocache().first()
+        if hackathon_project is not None:
+            if hackathon_project.grant_obj is None:
+                if request.user.profile in hackathon_project.profiles.all():
```
danlipert, line 1409: A better (faster) way to do this is `if hackathon_project.profiles.filter(id=request.user.profile.id).exists():` Since we don't actual need the profile object from the queryset.

app/grants/views.py
```diff
@@ -1487,21 +1520,35 @@ def grant_new(request):
...
-    # if project_id is not None:
-    #     hackathon_project = HackathonProject.objects.filter(pk=project_id).nocache().first()
-    #     if request.user.profile in hackathon_project.profiles.all():
-    #         project = hackathon_project
+    project_id = request.GET.get('project_id', None)
+    if project_id is not None:
+        hackathon_project = HackathonProject.objects.filter(pk=project_id).nocache().first()
+        if hackathon_project is not None:
+            if hackathon_project.grant_obj is None:
+                if request.user.profile in hackathon_project.profiles.all():
```
danlipert, line 1534: same optimization here

danlipert (changes requested):
One small performance trick otherwise looking good

app/grants/views.py
PixelantDesign: @sanchaymittal could you take a look at @danlipert request ?

danlipert:
Seems like this PR is not working due to incompatibility with changes made for the ZCash grants round, closing for now until we decide what we want to do with it.

PR #7725, opened 2020-10-21
chore: backend fixes

thelostone-mc:
##### Description

- add check to ensure txn is not older than 5 min
- ensure contirbution mail is sent only after txn is validated
- removed address as a mandatory param on the backend
- set deploy_tx_id default to 0x0

### DIFF
--- a/app/grants/models.py
+++ b/app/grants/models.py
@@ -1137,7 +1137,7 @@ next_valid_timestamp: {next_valid_timestamp}
         return contribution
 
 
-    def create_contribution(self, tx_id):
+    def create_contribution(self, tx_id, is_successful_contribution=True):
         from marketing.mails import successful_contribution
         from grants.tasks import update_grant_metadata
 
@@ -1173,7 +1173,8 @@ next_valid_timestamp: {next_valid_timestamp}
         self.save()
         grant.updateActiveSubscriptions()
         grant.save()
-        successful_contribution(self.grant, self, contribution)
+        if is_successful_contribution:
+            successful_contribution(self.grant, self, contribution)
 
         update_grant_metadata.delay(self.pk)
         return contribution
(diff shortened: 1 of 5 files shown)


app/grants/models.py
```diff
@@ -1173,7 +1173,8 @@ def create_contribution(self, tx_id):
         self.save()
         grant.updateActiveSubscriptions()
         grant.save()
-        successful_contribution(self.grant, self, contribution)
```
thelostone-mc, line 1176: @mds1 / @octavioamu this is what always causes the mail to be sent asap !
For cross chain flows it will happen only after txn is validated 
For eth -> we will have to shift this to the celery task which does the validation 

`process_grant_contribution`  can do this in follow up PR after running it with team

app/grants/sync/zcash.py
```diff
@@ -73,6 +77,19 @@ def get_zcash_txn_status(txnid):
     return None
 
 
+def is_txn_done_recently(time_of_txn):
```
thelostone-mc, line 80: this check ensures that the txn was signed within 5 hours of submission on the Contribution record being created

app/grants/views.py
```diff
@@ -2613,13 +2612,13 @@ def contribute_to_grants_v1(request):
...
             continue
 
-        tx_id = contribution.get('tx_id', None)
-        if not tx_id:
-            invalid_contributions.append({
-                'grant_id': grant_id,
-                'message': 'error: tx_id is mandatory param'
-            })
-            continue
+        # tx_id = contribution.get('tx_id', None)
```
thelostone-mc, line 2615: made this not mandatory as other cross chains might not be able to provide tx_id , commented out in case if want to keep this for some reason

danlipert: Good call - do you want to add a quick comment in the code explaining it?

danlipert (approved):

octavioamu (approved):

app/grants/views.py
thelostone-mc: will add it in a follow up commit 🙌

PR #7727, opened 2020-10-21
remove bad tests

danlipert:
##### Description

This PR removes some tests that a) don't make sense, testing a that a library loads in the case of IPFS or b) I don't think ever worked, as maybe this was missed because of the test failures in a).

##### Refers/Fixes

Fixes travis

##### Testing

tested locally


octavioamu:
they use to work but there is a conflict between ipfs-go version and ipfshttpclient, but now as we don't use bounty network I think make sense. Maybe also remove it from the tests packages installation.

danlipert:
> they use to work but there is a conflict between ipfs-go version and ipfshttpclient, but now as we don't use bounty network I think make sense. Maybe also remove it from the tests packages installation.

yeah i saw your comment on the ipfshttpclient repo, thanks for doing that. I took a look but not sure why we are getting the conflict now. either way I think its a silly test

octavioamu (approved):

PR #7536, opened 2020-09-25
add the ingest_community_events_calender cronjob

hhio618:
##### Description

- Add a cron job to ingest Gitcoin Community events calendar into the marketing.models.upcomingdate model.
- Add the ingest_community_events_calender management command with a url argument set default to [this iCal link](https://calendar.google.com/calendar/ical/7rq7ga2oubv3tk93hk67agdv88%40group.calendar.google.com/public/basic.ics)).

##### Refers/Fixes
Fixes: https://github.com/gitcoinco/web/issues/7499

PR #7758, opened 2020-10-25
remove auto_add=now from last-modified field

hhio618:
##### Description

Fix bad test: app/marketing/tests/management/commands/test_ingest_community_events_calender.py

##### Refers/Fixes

Refs #7536, #7727


app/marketing/models.py
```diff
@@ -420,7 +420,7 @@ class UpcomingDate(SuperModel):
     """Define the upcoming date model"""
     # These fields are meant to use for update UpcomingDate based on the icalendar updates
     uid = models.CharField(max_length=255, null=True, blank=True)
-    last_modified = models.DateTimeField(db_index=True, auto_now=True)
+    last_modified = models.DateTimeField(db_index=True)
```
aamustapha, line 423: is there a reason for this? to the best of my knowledge, removing this means this field would no longer be updated when a record is changed.

If you're doing this because of failing tests, what you want to do is a `unittest.mock`

aamustapha (changes requested):

app/marketing/models.py
hhio618: This field needs to be filled manually from an input iCalendar file. I had added a ```auto_now=True``` to it accidentally. So yes definitely this attribute need to be deleted in my opinion.

hhio618: This field is different from ```modified_on``` that is defined in SuperModel and tracks last modification time of the object.

danlipert: since we already have modified_on then we dont need those one as well

danlipert (approved):

app/marketing/models.py
hhio618: As may you already knew, this field is related to the last_modified field of an actual event modification date in an iCal file @danlipert. I used this field to check if some event needs to get updated. 

If I remove this field and use modified_on then there will be no modification date of actual db record, Is that ok?

thelostone-mc (approved):

PR #7773, opened 2020-10-27
Fix NoReverseMatch when project_name is empty

aamustapha:
##### Description
Currently, when a project name is empty the reverse url method raises a NoReverseMatch error. This can prevent customers from temporarily being able to access the system. A quickfix could be to modify defaulting instances but a more longterm solution would be to allow null project_name. This is done by 
1. Creating a new url path that doesn't accept project_name
2. Check if a project has a name, if it doesn't do not include empty string in the kwargs


app/dashboard/templates/dashboard/hackathon/projects.html
```diff
@@ -169,7 +169,12 @@ <h1>{{hackathon.name}} Projects</h1>
                 </div>
                 <div class="card-body">
                   <h5 class="card-title text-uppercase">{{ project.name }}</h5>
-                  <a href="{% url 'hackathon_project_page' hackathon=project.hackathon.slug project_id=project.id project_name=project.name|slugify %}" class="font-weight-bold card-subtitle">Project Home</a>
+                  <a href="
+                   {% if project.name%}
```
danlipert, line 173: needs extra space here

danlipert (approved):
Nice! Just one little typo, but looks great

thelostone-mc (approved):

PR #7787, opened 2020-10-29
fix: properly display project info in activity

chibie:
##### Description

distortion in html if there's no project name, app/retail/templates/shared/activity.html - line 82

```
title="Go to {{ row.project.name }}">via {{ row.project.name }}</a>
```
will be

```
via
```

instead, display `via project <project id>`

##### Refers/Fixes

#7773


danlipert (approved):

thelostone-mc (approved):
much neater

Issue #7796, opened 2020-10-30
zcash grant contributions dont show activity feed items, or on the transaction view


owocki:
reproduction steps 
1. add grant to cart https://gitcoin.co/grants/1486/zcashzeal?tab=transactions
2. do the checkout + get the 'grant contributions recorded' message https://bits.owocki.com/p9uwBlB9

what i observe vs what i should observe:
1. transaction says it failed on trasnsactions tab https://bits.owocki.com/Blu1qJvp
2. grant card still says 0 contributors worth 0 USD https://bits.owocki.com/kpuweQvB , vs it should say 1 contributor worth 0.01 ZEC * conv rate
3. no activity feed item for the contribution ; vs it should show
4. i see the contribution in the db https://gitcoin.co/_administrationgrants/contribution/49933/change/ but it is never marked as successful.  (in the ETH flow we assume success=True when contribution is created + only flip it back to failed if the validator fails)

owocki:
looks like the txn did succeed on the block explorer https://explorer.zcha.in/transactions/85bb9c41f77cbea06091606fd99c7a8d9cf7db69e3599a7cc38dd4a7b923014b

thelostone-mc:
Looks like it updated! 🙌  
The transaction was probably in pending state when you created the issue and got picked up cron a short while later

PR #7797, opened 2020-10-30
showing txn in pending state

thelostone-mc:
##### Description

- email gets sent to user on clicking payout (doesn't wait for confirmation)
- entry in transaction tab is marked as pending state 

image

### DIFF
--- a/app/grants/sync/helpers.py
+++ b/app/grants/sync/helpers.py
@@ -55,7 +55,8 @@ def record_contribution_activity(contribution):
                 comment= subscription.comments
             )
 
-        successful_contribution(grant, subscription, contribution)
+        # note: commenting out for optimistic UI
+        # successful_contribution(grant, subscription, contribution)
         new_supporter(grant, subscription)
         thank_you_for_supporting(grant, subscription)
         update_grant_metadata.delay(grant.pk)
--- a/app/grants/sync/helpers.py
+++ b/app/grants/sync/helpers.py
@@ -57,9 +57,9 @@ def record_contribution_activity(contribution):
 
         # note: commenting out for optimistic UI
         # successful_contribution(grant, subscription, contribution)
+        # update_grant_metadata.delay(grant.pk)
         new_supporter(grant, subscription)
         thank_you_for_supporting(grant, subscription)
-        update_grant_metadata.delay(grant.pk)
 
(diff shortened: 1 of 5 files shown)


thelostone-mc:
@owocki quick check -> the amount should get updated as well! I'm not sure what that happens

app/grants/sync/helpers.py
```diff
@@ -55,7 +55,8 @@ def record_contribution_activity(contribution):
                 comment= subscription.comments
             )
 
-        successful_contribution(grant, subscription, contribution)
+        # note: commenting out for optimistic UI
+        # successful_contribution(grant, subscription, contribution)
```
owocki, line 59: where is this moved to?  seems like it's not re-added anywhere

app/grants/views.py
```diff
@@ -2698,7 +2698,9 @@ def contribute_to_grants_v1(request):
             subscription.save()
 
             # step 3: create contribution + fire celery
-            contribution = subscription.create_contribution(tx_id, False)
+            contribution = subscription.create_contribution(tx_id, True)
+            contribution.success = True
```
owocki, line 2702: is it marked as failed if the contribution fails elsewhere?

owocki:
which of the feedback items in my ticket ( https://github.com/gitcoinco/web/issues/7796 ) does this fix?  just the tx history tab, or the other items too?

owocki:
which of the feedback items in my ticket ( https://github.com/gitcoinco/web/issues/7796 ) does this fix?  just the tx history tab, or the other items too?

app/grants/views.py
thelostone-mc: nope! there is no way for us to fail -> hence it'll be in forever pending 
The fallback we could have is to auto fail it after 5 hours

app/grants/sync/helpers.py
thelostone-mc: this is taken care  https://github.com/gitcoinco/web/pull/7797/files#diff-f630b64d112cbde98b1992af074c13f88ce29c0f6a81dffdc2fdfa967fb8a2c6R2701

changing it to true adds it

`contribution = subscription.create_contribution(tx_id, True)`

danlipert (approved):

PR #7819, opened 2020-11-03
Autofill Grants

### LINKED ISSUE #7746: feat: auto-fill grant/new form
#### Description

- prefill grant type and grant subtype via URL as shown below http://gitcoin.co/grants/new?category=matic&subcategory=test&clr=matic
 - - category -> what populates the category field
- - subcategory -> what populates the subcategory field (this can be an array)
- - chain (selects which chain to prefill -> ETH / Zcash)

---
For showing a custom logo on grant creation form
- Add an image field to Grant Type model
- The logo on the grant creation form should change based on the category set (If logo is not available -> use default logo)

---

- [ ] Design
- [ ] Development

aamustapha:
##### Description
To make creating grants more user friendly, a user can prefill some fields from the URL parameters. e.g `gitcoin.com/grants/new/?title=Some%20Sample%20Grant`. Some of the applications of this is to make forking grants easeir from the frontend or allow a user to easily create a grant based on previous information known about that user.

- [x] Allow autofill input fields
- [x] Allow autofill radio
- [x] Allow autofill select 
- [x] Allow autofill select multiple
- [x] Load appropriate marketing logo at the top
`

##### Refers/Fixes
fixes #7746 

##### Testing

- [x] Test Autofill Input field: `grants/new?title=New%20Test%20Grant`
- [x] Test Autofill chain: `grants/new?title=New%20Test%20Grant&chainId=eth`
- [x] Test Autofill grant_type (select field):: `grants/new?title=New%20Test%20Grant&chainId=eth&grant_type=health`
- [x] Test Autofill category (multiple select): `grants/new?title=New%20Test%20Grant&chainId=eth&grant_type=media&grant_categories=[3,4]`
- [x] Test Banner changes: `grants/new?title=New%20Test%20Grant&chainId=eth&grant_type=change`

Video: https://www.loom.com/share/02881364a6f04570b048cd6339a3f8fb


thelostone-mc:
Is this ready for review ? @aamustapha 
The adding of a logo wiring etc is pending as well (or are you planning that as a separate PR ?)

aamustapha:
No, I'm trying to figure how the logo would work.

app/grants/views.py
```diff
@@ -1585,12 +1585,18 @@ def grant_new(request):
...
             'id': g_type.pk,
             'name': g_type.name,
             'label': g_type.label,
             'categories': grant_categories
-        })
+        }
+        if g_type.logo:
+            grant_type_temp['image_url'] = request.build_absolute_uri(g_type.logo.url)
+
+        grant_types.append(grant_type_temp
```
thelostone-mc, line 1598: This seems broken

thelostone-mc: @octavioamu did pitch the idea of sending all the banners to the frontend and having the frontend decide which type to use! 

Reason being: if i the matic logo is present and I switch from to filecoin -> shouldn't the image change as well ?

aamustapha: Yes, if the logo for filecoin is available, it would change. Otherwise, it uses the default icon.

thelostone-mc (approved):

chibie (approved):

danlipert (approved):
This looks great - a few comments and a question though:

Let's squash the migrations before merging this into master. Easiest way to do this @aamustapha is to just delete the migration files from your branch and recreate the migrations. This might mess up your local db though. 

Are there any security implications to think about with this branch? For example, can I create a URL that has query parameters such that I'm able to escape the input fields and add arbitrary JS/HTML via the query params?

Lastly, I noticed the template has a lot of style changes in the form of indentations. Since every contributor tends to use different settings in terms of auto-styling, lets make sure to limit the changes to whats described in the PR. If we want to do style changes, lets make that a separate PR. That way its easy to review! 👍

thelostone-mc:
Oh yeah could we undo the indent changes ? Usually we stick to 1 tab == 2 spaces for frontend code

aamustapha:
- [x] Let's squash the migrations before merging this into master. Easiest way to do this @aamustapha is to just delete the migration files from your branch and recreate the migrations. This might mess up your local db though.

> Are there any security implications to think about with this branch? For example, can I create a URL that has query parameters such that I'm able to escape the input fields and add arbitrary JS/HTML via the query params?
This is a good catch, i thought about this and it's the reason I manually converted arrays back via splitting rather than `eval`, but I think still it'd accept `<script>...</script>`. I'd use regex to see if any tags are in what is passed, if they exist tags, we wouldn't autofill the field.


> Lastly, I noticed the template has a lot of style changes in the form of indentations. Since every contributor tends to use different settings in terms of auto-styling, lets make sure to limit the changes to whats described in the PR. If we want to do style changes, lets make that a separate PR. That way its easy to review! +1

The indent changes was because I added an additional wrapper on LN: 44-48. I noticed that as per the design specs, that already existed default logo renders within the form card while custom logo are displayed onto of it. as shown in the screenshots below

Default Image
image

Custom Logo
image

app/assets/v2/js/grants/_new.js
```diff
@@ -296,6 +296,55 @@ if (document.getElementById('gc-new-grant')) {
           placeholder: 'Give a detailed desciription about your Grant'
         }
       };
+    },
+    computed: {
+      queryParams () {
+        let query = window.location.search.substr(1)
```
octavioamu, line 302: a recommendation for future url params logic https://developer.mozilla.org/pt-BR/docs/Web/API/URLSearchParams

octavioamu (approved):

PR #7824, opened 2020-11-05
fix: update QR code for CELO + cUSD

thelostone-mc:
##### Description

Updating the QR string based on https://github.com/celo-org/celo-monorepo/pull/5515
This will be available in version 1.4 of Valora

### DIFF
--- a/app/assets/v2/js/pages/bounty_details2.js
+++ b/app/assets/v2/js/pages/bounty_details2.js
@@ -140,11 +140,11 @@ Vue.mixin({
             `ethereum:${address}`;
           break;
 
-        case 'CELO': // waiting : pending
+        case 'CELO':
         case 'cUSD':
           qr_string = value ?
-            `celo://wallet/pay?address=${address}&amount=${value}` :
-            `celo://wallet/pay?address=${address}`;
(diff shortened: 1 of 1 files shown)


chibie (approved):

app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -140,11 +140,11 @@ Vue.mixin({
...
           break;
 
-        case 'CELO': // waiting : pending
+        case 'CELO':
         case 'cUSD':
           qr_string = value ?
-            `celo://wallet/pay?address=${address}&amount=${value}` :
-            `celo://wallet/pay?address=${address}`;
+            `celo://wallet/pay?address=${address}&amount=${value}&token=${token_name}` :
+            `celo://wallet/pay?address=${address}&token=${token_name}`;
```
medhak1, line 147: to clarify, if the `token` param is not given, the deeplink will default to cUSD, so you can choose whether or not you want to provide it or not. For CELO though, the param is necessary.

thelostone-mc: Yup! thanks for clarifying that @medhak1

PR #7825, opened 2020-11-05
chore: remove un-used code from grants DB + fix cancel grant code

thelostone-mc:


thelostone-mc:
Closing this out as it's not outdated

Issue #7817, opened 2020-11-03
As a new user, I find the new onboarding model extremely annoying, and I would like for it to show up every visit, not every pageload.


owocki:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)
As a new user, I find the new onboarding model extremely annoying, and I would like for it to show up every visit, not every pageload.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)
*Summary*:
Customer feedback

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*:
Feature

### Current Behavior
[comment]: # (Describe what actually happened.)
Modal shows/ can be dismissed once per pageload

### Expected Behavior
[comment]: # (Describe what you expected to happen.)
Modal shows/ can be dismissed once per VISIT; or when user clicks it in the nav/profile completeness tab.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

### Data Requirements
[comment]: # (How will we measure the success of this feature? What kind of tracking is needed for this feature (clicks, impressions, flag)?)

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

nglglhtr:
Wondering if work has already been started on this? Considering taking this as a test to go through a developer's experience of contributing to Gitcoin

owocki:
@octavioamu made it sound like it was  apretty small change; idk if hes started on it tho

octavioamu:
I didn't yet. I'm ok if you want to take it.

nglglhtr:
@octavioamu my idea is to use the `window.sessionStorage` in onboarding modal display

octavioamu:
@nglglhtr was fixed on friday here is the commit 393984c

owocki:
tyty @octavioamu

PR #7847, opened 2020-11-09
Displays onboarding modal to a new user on each new session

nglglhtr:
##### Description
Displays onboarding modal to a new user on each new session, not on every reload 

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/7817

##### Testing

Opening a page in a new tab or window creates a new session with the value of the top-level browsing context, which differs from how session cookies work. ([source](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage))

### DIFF
--- a/app/assets/v2/js/onboard.js
+++ b/app/assets/v2/js/onboard.js
@@ -243,10 +243,12 @@ if (document.getElementById('gc-onboard')) {
       if (
         document.contxt.github_handle &&
         !document.contxt.persona_is_funder &&
-        !document.contxt.persona_is_hunter
+        !document.contxt.persona_is_hunter &&
+        !window.sessionStorage.getItem('alreadyViewed')
       ) {
         // show_persona_modal();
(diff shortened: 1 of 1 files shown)


thelostone-mc:
@nglglhtr why did we end up using sessionstorage as opposed to localstorage?

octavioamu:
This one was already fixed as informed on slack, here is the commit  https://github.com/gitcoinco/web/commit/393984c816f26b674de918e8060941a08397544c

nglglhtr:
@thelostone-mc [this](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) gives a good description. basically, data localStorage doesn't expire - an expiry needs to be set here while data in sessionStorage is cleared when the page session ends -- giving the display on 'every visit'

nglglhtr:
@octavioamu Oh, just saw. Alright!

PR #7870, opened 2020-11-13
Kevin fix multigrant

thelostone-mc:
#### Description

This resolves conflicts in https://github.com/gitcoinco/web/pull/7522/files
and ensure a grant can be present in multiple ETH rounds at the same time 

Flow/Steps Tests
- [x] Create multiple Grant CLR active at the same time and ensure few grant are present in all the rounds 
- [x] running estimate_clr -> populates the newly Created Table `GrantCLRCalcualtion` for individual curves with latest = True
- [x] making a contribution / running clr calculation calculated the total prediction curve is a sum of the individual prediction curve . 
- [x] Check that total prediction curve is reflected on the UI
- [x] On running estimate_clr for the same round, ensure older prediction curves are set to False  

cc @frankchen07 from the CLR standpoint does this math make sense.
If a grant is part of CLR A and CLR B 
we calculation curves for CLR A + CLR B and sum the results and that is the effective pot match 

@danlipert 
this table will bloat up quick after a few rounds.
Should we work on moving records with latest=False into it's own table when we reach this problem and clean up this table?

@owocki


danlipert:
No need to move data to diff tables - we'll just clean this table if it gets too big. Lets add an index on current so that the lookup is fast @thelostone-mc

app/grants/migrations/0093_auto_20201113_0559.py
```diff
@@ -0,0 +1,35 @@
...
+            model_name='grant',
+            name='backup_clr_prediction_curve',
+        ),
+        migrations.CreateModel(
+            name='GrantCLRCalculation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_on', models.DateTimeField(db_index=True, default=economy.models.get_time)),
+                ('modified_on', models.DateTimeField(default=economy.models.get_time)),
+                ('latest', models.BooleanField(db_index=True, default=False, help_text='Is this calc the latest?')),
```
danlipert, line 26: oh its already there! nice!

danlipert (approved):

chibie (approved):

PR #7880, opened 2020-11-16
Send a single email for all grant  items in cart

### LINKED ISSUE #7755: Consolidate emails so that user who checks out gets only 1 email not n emails for n contributions.
@apbendi will tackle this one

corydickson:
##### Description

This allows it so that a single email is sent when a user funds a list of grants

Still working on improvements: 

- [x] Not sending an email containing the grant with negative contributions
- [x] Removing dead code
- [x] Performance in the double retrieval of the subscription obj from the database (not sure how to do this when only submitting grant_ids that have successfully completed the checkout task

##### Refers/Fixes

Closes #7755 

##### Testing

No tests currently, still investigating where to add since haven't seen any integrations with SendGrid. 

I also had trouble viewing the email (it won't show it up in my inbox including spam) on my end but I think it has to do with protonmail policy and my SendGrid accounts trust level so would appreciate if someone could test this branch with their creds and post the screenshot here. There definitely is some work to be done for the styling because right now it's reusing existing HTML. 

Screen Shot 2020-11-17 at 5 58 21 PM

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -1716,7 +1716,8 @@ def bulk_fund(request):
     successes = []
     failures = []
 
-    grant_tasks = []
+    batch_grants_mail = []
+    profile = get_profile(request)
     for (index, grant_id) in enumerate(grant_ids_list):
         try:
             grant = Grant.objects.get(pk=grant_id)
@@ -1730,8 +1731,6 @@ def bulk_fund(request):
                 'success': False
             })
             continue
-
-        profile = get_profile(request)
 
         if not grant.active:
             failures.append({
@@ -1825,7 +1824,7 @@ def bulk_fund(request):
                 'token_address': request.POST.get('token_address').split(',')[index],
                 'token_symbol': request.POST.get('token_symbol').split(',')[index],
             }
-            grant_tasks.append(process_grant_contribution(grant_id, grant.slug, profile.pk, payload))
+            process_grant_contribution.delay(grant_id, grant.slug, profile.pk, payload)
         except Exception as e:
             failures.append({
                 'active': 'grant_error',
@@ -1843,11 +1842,9 @@ def bulk_fund(request):
             'text': _('Funding for this grant was successfully processed and saved.'),
             'success': True
         })
+        batch_grants_mail.append((grant_id, payload))
 
-    try:
-        chord(grant_tasks)(thank_you_for_supporting([success.get('grant') for success in successes], profile))
-    except Exception as e:
-        print(e)
+    thank_you_for_supporting(batch_grants_mail, profile)
 
     return JsonResponse({
         'success': True,
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -1841,7 +1841,7 @@ def bulk_fund(request):
             'text': _('Funding for this grant was successfully processed and saved.'),
             'success': True
         })
-        batch_grants_mail.append((grant_id, payload))
+        batch_grants_mail.append(grant_id)
 
     thank_you_for_supporting(batch_grants_mail, profile)
 
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -25,7 +25,6 @@ import logging
 import random
 import re
 import time
-from celery import chord
 from decimal import Decimal
 
 from django.conf import settings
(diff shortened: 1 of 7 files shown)


corydickson:
Screen Shot 2020-11-20 at 1 20 10 AM

I'm having trouble verifying this email in my inbox due to phishing/spam filters. Is there anyway someone with more trust on SendGrid can test the code above and make sure everything looks like what's inside the admin dashboard

thelostone-mc:
@corydickson code looks sane ! I'll get it into the integration branch and we'll get it tested over there

app/marketing/mails.py
```diff
@@ -300,18 +301,24 @@ def new_supporter(grant, subscription):
         translation.activate(cur_language)
 
 
-def thank_you_for_supporting(grant, subscription):
-    if subscription and subscription.negative:
-        return
+def thank_you_for_supporting(grants, contributor):
```
thelostone-mc, line 304: Ah this would break things over  at https://github.com/gitcoinco/web/blob/master/app/grants/sync/helpers.py#L62

which is an alternate API we use for non-Ethereum Chains

app/marketing/mails.py
```diff
@@ -300,18 +301,24 @@ def new_supporter(grant, subscription):
...
+    subscriptions = []
+    for (grant_id, payload) in grants:
+        subscription = Subscription.objects.filter(grant_id=grant_id, contributor_profile=contributor).first()
```
thelostone-mc, line 307: If i made a contribution to grant A, grant B -> I'd get a mail for it
Now a while later if i make another contribution for grant B + grant C

Wouldn't my grant A contribution also turn up in the mail?

thelostone-mc (changes requested):
left a few comments

app/marketing/mails.py
corydickson: Gotcha I'll update that as well

app/marketing/mails.py
corydickson: My main motivation to use this approach was because it received a prior approval; but I don't think so, it would only grab the subscriptions from the grants in the most recent bulk cart payload, they would however receive another email with grant B in the later contribution.

Maybe this addresses your concern: https://github.com/gitcoinco/web/pull/7880/commits/04c423222b8d9cf3cf9b605ac83364099fab6974 

If we went in this direction, it would only send emails after the grant processing task successfully completes.

thelostone-mc:
@corydickson tested locally and it seemed to work fine

aamustapha (changes requested):
This looks solid to me, would just appreciate being dryier.

app/retail/templates/emails/grants/thank_you_for_supporting.html
```diff
@@ -46,6 +49,10 @@
     line-height: 1.5em;
   }
 
+  p {
+    margin-top: 0;
+  }
+
```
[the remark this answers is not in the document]
thelostone-mc: Ah for email -> we don't include the boostrap lib AFAIK! (unless we've defined .mt-0 is already defined in email styles)

thelostone-mc (approved):

app/grants/views.py
```diff
@@ -1840,6 +1841,9 @@ def bulk_fund(request):
             'text': _('Funding for this grant was successfully processed and saved.'),
             'success': True
         })
+        batch_grants_mail.append(grant_id)
+
+    thank_you_for_supporting(batch_grants_mail, profile)
```
thelostone-mc, line 1846: this won't work @corydickson as subscriptions are created by a celery task
`process_grant_contribution.delay(grant_id, grant.slug, profile.pk, payload)`

This means when you try to contribute to 2-3 grants
It causes the email to be sent before the subscription objects have been created!

Hence we don't have the subscription object while we send the mail causing it to error out

Screenshot 2020-11-30 at 12 56 34 PM
Screenshot 2020-11-30 at 12 56 51 PM



Screenshot 2020-11-30 at 12 33 51 PM


cc @apbendi / @mds1

corydickson: Make sense, using a chord see: https://github.com/gitcoinco/web/commit/e622584e31080c49110107c06df6e42d830914ae

Should resolve this issue

PR #7537, opened 2020-09-25
Add Twitter Verified Combinations & Trust Bonus

frankchen07:
##### Description

Added Twitter Verified handling (https://github.com/gitcoinco/web/pull/7537/commits/f6e336511eb5762bbe820dfe399f3f96cc22fdf5).
Simplified combination loops for different verification statuses (https://github.com/gitcoinco/web/pull/7537/commits/3c3e18f9788e91359c708483199b37d473c8c7c6).
Accounting for Trust Bonus data model, which overwrites and nullifies both previous commits (https://github.com/gitcoinco/web/pull/7537/commits/4d20cb4731ca4b2017fc9818c9b0e886cc75d550).

##### Refers/Fixes

1. No reference, added twitter verification, and we need to be able to handle it.
2. Adding the Trust Bonus data model to massively simplify handling of combinations.

##### Testing

`any` and `for` logic tested with 3 test cases:

```
x = [1, 2, 3]

In [10]: 9 > 1 and any(i in x for i in [9, 1])
Out[10]: True

In [11]: 9 > 6 and any(i in x for i in [9, 6])
Out[11]: False

In [12]: 3 > 1 and any(i in x for i in [3, 1])
Out[12]: True

In [13]: 4 > 1 and any(i in x for i in [4, 1])
Out[13]: True

In [14]: 9 > 7 and any(i in x for i in [9, 7])
Out[14]: False
```

With https://github.com/gitcoinco/web/pull/7537/commits/4d20cb4731ca4b2017fc9818c9b0e886cc75d550, the trust bonus data model massively simplifies the logic above (and eliminates the need for it). 

closes https://github.com/gitcoinco/web/issues/7821

PR #7887, opened 2020-11-16
Add Saturation Documentation

frankchen07:
##### Description

Added some documentation on saturation modifications in `grants/models.py` 

##### Refers/Fixes

Adds documentation on what to do to increase saturation. Once #7537 gets in though, the threshold will be modified to just one, rather than an unverified and verified threshold. Same should apply though. 

1. increase contribution multiplier
2. increase threshold, but not to infinity.

##### Testing

Just documentation.


thelostone-mc:
I think this would involve creating a migration file as well @frankchen07

frankchen07:
ah because we've changed the method and added help text. Updated

PR #7890, opened 2020-11-17
fix(grant): validate twitter handle for /grants/new

### LINKED ISSUE #7867: grant creation: ensure twitter input starts with @ and is not an actual link
#### Description

- In grants/new, add client JS validation to ensure twitter handles start with `@` and is not the user link 
- Round 7 lot of folks using the adding twitter link and weren't able to verify their grant.  This causes a lot of support requests being created which can be avoided

We wanna allow: 
`@gitcoin` and disallow
`https://twitter.com/gitcoin`

If `@` is not present -> frontend should append `@` and set it to backend

**End State**

- [x] Ensure Client validation works for twitter handles
- [x] Ensure Twitter verification works for a grant

chibie:
##### Description

Validate Twitter handles to accept only '@username' pattern during grant creation.

##### Refers/Fixes

fixes #7867

### DIFF
--- a/app/grants/tasks.py
+++ b/app/grants/tasks.py
@@ -11,7 +11,7 @@ from celery import app, group
 from celery.utils.log import get_task_logger
 from dashboard.models import Profile
 from grants.models import Grant, Subscription
-from marketing.mails import new_supporter, thank_you_for_supporting, new_grant_admin, new_grant
+from marketing.mails import new_grant, new_grant_admin, new_supporter, thank_you_for_supporting
 from townsquare.models import Comment
 
 logger = get_task_logger(__name__)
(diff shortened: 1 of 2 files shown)


thelostone-mc:
@chibie What are your thoughts on doing a frontend validation as well ?
That would ensure , the user getting immediate feedback -> less frustrated

chibie:
> @chibie What are your thoughts on doing a frontend validation as well ?
> That would ensure , the user getting immediate feedback -> less frustrated

@thelostone-mc i did it initially but reverted when i saw other field validations in BE.

i'll make another commit with FE validation soon

chibie:
@thelostone-mc all set now

thelostone-mc (approved):

PR #7896, opened 2020-11-18
Update deploy address, fix dashboard bug

mds1:
DRAFT: Waiting for final deployment of contracts before merging

Changes:
- Updates pToken factory address to the final factory address deployed on mainnet and Rinkeby
- Bug fix: A token creator was able to click "complete" on a request, but this should not be possible. Only requestors should mark it as complete


mds1:
Replaced by https://github.com/gitcoinco/web/pull/8275

PR #7898, opened 2020-11-18
Check if query params exists before attempting to access elements / Grants auto fill

aamustapha:
##### Description
Grants autofill was failing because there wasn't an explicit check as to whether there are query parameters. Hence I fixed by ensuring query parameters exist before attempting to access index. Major change on L346


chibie (commented):
@aamustapha also observe commented L301 - L332

thelostone-mc:
@aamustapha octo had left a comment on using getURLParams instead !

https://gitcoincore.slack.com/archives/G01AN94AQP9/p1605713759035300?thread_ts=1605698345.014400&cid=G01AN94AQP9

PR #7899, opened 2020-11-18
zkSync refactor

mds1:
Changes:
- Uses new zkSync `CheckoutManager` class to complete checkout atomically on zkSync's site
- Should have all required UX helpers (e.g. gas estimation to recommend a checkout method, error alerts, "token not supported on zkSync" messages)
- UI and functionality matches the [Figma mockups](https://www.figma.com/file/GUnBaAsN9k6XQi5RNHnxyK/Grants-R8?node-id=267%3A204)

cc @apbendi


app/assets/v2/js/cart-ethereum-zksync.js
```diff
@@ -0,0 +1,273 @@
...
+     *   1. Make sure the token is supported by zkSync
+     *        Mainnet list: https://api.zksync.io/api/v0.1/tokens
+     *        Rinkeby list: https://rinkeby-api.zksync.io/api/v0.1/tokens
+     *   2. Add the token symbol to the appropriate list below
+     * @dev We hardcode the list here instead of fetching from the API to improve performance and
+     * reduce problems that arise if zkSync's server is slow
+     */
+    supportedTokens() {
+      const mainnetTokens = [ 'ETH', 'DAI', 'USDC', 'TUSD', 'USDT', 'SUSD', 'BUSD', 'LEND', 'BAT', 'KNC', 'LINK', 'MANA', 'MKR', 'REP', 'SNX', 'WBTC', 'ZRX', 'MLTT', 'LRC', 'HEX', 'PAN', 'SNT', 'YFI', 'UNI', 'STORJ', 'TBTC', 'EURS', 'GUSD', 'RENBTC', 'RNDR', 'DARK', 'CEL', 'AUSDC', 'CVP', 'BZRX', 'REN' ];
+      const rinkebyTokens = [ 'ETH', 'USDT', 'USDC', 'LINK', 'TUSD', 'HT', 'OMG', 'TRB', 'ZRX', 'BAT', 'REP', 'STORJ', 'NEXO', 'MCO', 'KNC', 'LAMB', 'GNT', 'MLTT', 'XEM', 'DAI', 'PHNX' ];
```
apbendi, line 68: I wonder if it might be worth stuffing these in a JSON blob in the database. The reason I say this is, as it stands, we need to do a PR + redeploy to support a new token. With something in the DB, it could be edited in the admin w/o updating code. What do you think?

apbendi (commented):
Love deleting code 😍 nice work @mds1. Left a couple of comments!

app/assets/v2/js/cart-ethereum-zksync.js
```diff
@@ -0,0 +1,273 @@
...
+
+      // If we're here, user needs at least one L1 deposit, so let's calculate the total cost
+      this.cart.tokenList.forEach((tokenSymbol) => {
+        const zksyncBalance = toBigNumber(this.user.zksyncState.committed.balances[tokenSymbol]);
+        
+        if (requiredAmounts[tokenSymbol].gt(zksyncBalance)) {
+          if (tokenSymbol === 'ETH') {
+            totalCost = totalCost.add('200000'); // add 200k gas for ETH deposits
+          } else {
+            totalCost = totalCost.add('250000'); // add 250k gas for token deposits
```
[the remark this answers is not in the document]
mds1: I believe zkSync uses 200k and 300k as the gas limits respectively, but since we aren't sending the transaction ourselves anymore we probably can be a bit less conservative here. I'll make a note of this and perhaps we circle back later once the rest of the core changes are made?

app/assets/v2/js/cart-ethereum-zksync.js
mds1: That's a good point, and you're right it's probably the better approach. @thelostone-mc I know this is how regular tokes are configured, so what's involved with doing that for these lists as well? 

On the other hand, considering over 99.9% of donations are made in tokens already on this list, it may not be worth the effort right now

thelostone-mc: hmm having a new token table just for zksync seems like an overkill given that it won't change much 🤔
if we support other L2 (not sure about that) -> would it mean having a new table for every L2 ?

apbendi: You know, I think you're right @thelostone-mc, especially per @mds1's point that 99% of donations come from just a few tokens. I withdraw my suggestion!

app/assets/v2/js/cart-ethereum-zksync.js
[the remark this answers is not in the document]
apbendi: Fine with me!

app/assets/v2/js/cart-ethereum-zksync.js
thelostone-mc: i'd lean towards keeping it as a list in code for now and switch it over later is needed

mds1: Sounds good, I'll mark this as resolved then 👍

thelostone-mc (approved):

PR #7900, opened 2020-11-18
Grants round 8

octavioamu:
##### Description

Integration branch for grants round 8 
https://github.com/gitcoinco/web/projects/52


app/grants/templates/grants/components/card.html
```diff
@@ -98,7 +98,7 @@ <h2 class="grant-item__title font-subheader">
                   </span>
                 </div>
               </div>
-              <template v-if="cred.is_authenticated">
+              <template v-if="isUserLogged">
```
octavioamu, line 101: @thelostone-mc this will not work this componet doesn't have `isUserLogged` (this only exist on grant details component) the prop for this component is named `cred`

thelostone-mc: There wasn't a prop called cred causing this check to fail! 
I did see credentials but I wasn't able to access that as well

thelostone-mc (approved):

app/grants/management/commands/estimate_clr.py
```diff
@@ -49,24 +49,11 @@ def handle(self, *args, **options):
 
         if active_clr_rounds:
             for clr_round in active_clr_rounds:
-                print(f"CALCULATING CLR estimates for ROUND: {clr_round.round_num}")
-                predict_clr(
+                process_predict_clr(
```
owocki, line 52: fwiw; this doesnt actually delegate the tasks to the queues.  it just runs it inline.

PR #7913, opened 2020-11-20
chore(grants): estimate clr via celery

### LINKED ISSUE #7854: move estimate clr to celery task
#### Description 

speed up grant clr calcs - if we want to speed up the calcs, i could see offloading the `estimate_clr`jobs to the celery queues tho.   that’d be an elegant way to paralellize them

- update mgmt command estimate.clr to invoke predict_clr as a celery task 
- when you haven multiple CLR rounds -> all-round calculations should be offloaded to celery task

note:
In prod -> we can create more celery workers to make the task run in parallel

chibie:
##### Description

This moves invocation of `predict_clr` by `estimate clr` management command to a celery task

##### Refers/Fixes

closes #7854

##### Testing

image


chibie:
@thelostone-mc all set here?

app/grants/tasks.py
```diff
@@ -206,6 +212,19 @@ def recalc_clr(self, grant_id, retry: bool = True) -> None:
...
+@app.shared_task(bind=True, max_retries=1)
+def process_predict_clr(save_to_db=False, from_date=None, clr_round=None, network='mainnet', only_grant_pk=None):
+    from grants.clr import predict_clr
+    return predict_clr(
+        save_to_db=False,
+        from_date=None,
+        clr_round=None,
+        network='mainnet',
+        only_grant_pk=None
+    )
```
thelostone-mc, line 225: ^ I don't get why we are doing this?

thelostone-mc: Did you test this out @chibie ? 
Could we pass only the needed arguments as is

thelostone-mc: Also do we need a return ?

app/grants/management/commands/estimate_clr.py
```diff
@@ -49,14 +49,14 @@ def handle(self, *args, **options):
...
                     save_to_db=True,
                     from_date=timezone.now(),
                     clr_round=clr_round,
                     network=network
                 )
-                print(f"finished CLR estimates for {clr_round.round_num}")
+                print(f"finished CLR estimates for {clr_round.round_num} {clr_round.sub_round_slug}")
 
                 # TOTAL GRANT
                 # grants = Grant.objects.filter(network=network, hidden=False, active=True, link_to_new_grant=None)
```
thelostone-mc, line 62: ^ I believe we'd have to move this commented out code as well into process_predict_clr
this is used for debugging

thelostone-mc:
@chibie could you through in a screenshot as well just to make sure this works as expected and the logs are printed in order and stuff

app/grants/tasks.py
chibie: `predict_clr` function in clr.py returns `debug_output`

app/grants/management/commands/estimate_clr.py
chibie: oh right!

app/grants/tasks.py
thelostone-mc: Ah i don't know what we use that for ! 
Cause if you look at other places -> we don't do anything with it

thelostone-mc: could we switch it to 
```
process_predict_clr(save_to_db, from_date, clr_round, network):

    predict_clr(
        save_to_db,
        from_date,
        clr_round,
        network
    )
```

app/grants/management/commands/estimate_clr.py
```diff
@@ -49,24 +49,13 @@ def handle(self, *args, **options):
...
-                # grants = Grant.objects.filter(network=network, hidden=False, active=True, link_to_new_grant=None)
-                # grants = grants.filter(**clr_round.grant_filters)
-
-                # total_clr_distributed = 0
-                # for grant in grants:
-                #     total_clr_distributed += grant.clr_prediction_curve[0][1]
-
-                # print(f'Total CLR allocated for {clr_round.round_num} - {total_clr_distributed}')
-
+                print(f"finished CLR estimates for {clr_round.round_num} {clr_round.sub_round_slug}")
```
thelostone-mc, line 59: This print statement also would have to be pushed into it right ?

PR #7914, opened 2020-11-20
 purge old data

thelostone-mc:


castroed47 (commented):
https://m.facebook.com/story.php?story_fbid=247161453438087&id=100044327024735

PR #7830, opened 2020-11-06
Feature: Show Recent People I've Tipped

sebastiantf:
##### Description
This PR adds a new feature that shows the list of users (max of 7) that you've recently tipped, when you visit http://gitcoin.co/tip

##### Refers/Fixes
Fixes #7681 

##### Screenshots
1. If user has not tipped anyone yet:

**GIF**: 

**Video**: https://www.loom.com/share/bdd437d9ea5245df9005173b90f0e9ab

2. If user has tipped before:

**Video**: https://www.loom.com/share/cffb44d8cb8345fbb051b6edb50011bc

**Screenshot**: 
FireShot Capture 314 - Send Tip - Gitcoin - Gitcoin - localhost

Done:
- [x] Display icons of the last 7 people tipped below the search (max of 7).
- [x] On hover, show the persons gitcoin username
- [x] Clicking on the name takes user to profile
- [x] If only one person has been tipped, left align the entire row

PR #7938, opened 2020-11-25
Fix error in /tip for non-logged in user

sebastiantf:
##### Description
Follow-up for #7830 

Fixes showing this error for non-logged-in user:
> 'AnonymousUser' object has no attribute 'profile'

##### Refers/Fixes
https://github.com/gitcoinco/web/pull/7830#issuecomment-733157241

##### Screenshot
This user is not logged in:

fix:recent-tips

### DIFF
--- a/app/dashboard/tip_views.py
+++ b/app/dashboard/tip_views.py
@@ -420,6 +420,7 @@ def send_tip_3(request):
 
 
 @ratelimit(key='ip', rate='5/m', method=ratelimit.UNSAFE, block=True)
+@login_required
 def send_tip_2(request):
     """Handle the second stage of sending a tip.
 


PixelantDesign (approved):

thelostone-mc:
@sebastiantf a neater way to go about this would be to add the @login_required decorator in views.py file

sebastiantf:
@thelostone-mc When `@login_required` is used, it automatically redirects to Github login and then to Townsquare without being able to see the /tip page. Is that fine?

Also, `@login_required` was not used for the /tip page

Should I revert this commit and use the decorator?

aamustapha (approved):
LGTM. Just a syntatic sugar enhancement. could do good.

thelostone-mc:
> Should I revert this commit and use the decorator?

@sebastiantf yup would be cleaner ! plus there is no reason for folks to use /tiop without being logged in

sebastiantf:
> > Should I revert this commit and use the decorator?
> 
> @sebastiantf yup would be cleaner ! plus there is no reason for folks to use /tiop without being logged in

@thelostone-mc Done 👍

thelostone-mc (approved):
thanks :D

Issue #7792, opened 2020-10-29
a/b test of the sorting method on the main  page.  even pk users = weighted shuffle (which maxes the weight at $5000), odd pk users; = random shuffle (not weighted).  we'll then report the data to blockscience to see how thisi changes the rich get richer.


owocki:
MZ - suggestoin - a/b test of the weighting mechanism; weighted_shuffle vs inverse weighted_shuffle

i can do this - @owocki

thelostone-mc:
looks like this was the same as https://github.com/gitcoinco/web/issues/7822

PR #7941, opened 2020-11-25
a/b test of sort order

owocki:
##### Description

splits users into two groups, one that sees weighted shuffle grants + another that sees random grants (based upon even/odd-ness of their PK).

this will be evaluated by blockscience down the line, so they can see what effect gitcoin's default sort order has on outcomes

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/7792

##### Testing

tested locally


app/grants/views.py
```diff
@@ -538,6 +538,9 @@ def get_grants(request):
     page = request.GET.get('page', 1)
     collections_page = request.GET.get('collections_page', 1)
     sort = request.GET.get('sort_option', 'weighted_shuffle')
+    if request.user.is_authenticated and request.user.profile.pk % 2 == 1:
+            if sort == 'weighted_shuffle':
+                sort = 'random_shuffle'
```
aamustapha, line 543: The indentation here looks off.

aamustapha (changes requested):
LGTM. Just one

chibie (approved):
only note @aamustapha 's comment

app/grants/views.py
thelostone-mc: Yup it def is off ! @owocki you sure this worked locally ?  I'm fixing it up for now cause i need everything in round 8 branch 

nice catch @aamustapha

thelostone-mc:
fixed up indent + redid migrations + refactord code

PR #7942, opened 2020-11-26
Grants checkout POST request + tx validator updates

### LINKED ISSUE #7744: chore: Fix (or majorly triage) the issue where the system doesnt know about a contribution + we have to manually ingest them.
#### Description 

**Part1**
As the title states, figure out root cause and fix up missing contributions so as to avoid manual adding them in db

@mds1 

**Part2**

See if UI/UX improves needed now that for bulks checkout we move user to zksync site 

sample app which does it:
https://zksync.curve.fi/

### LINKED ISSUE #7751: chore: update payout_type in Contribution
#### Description

- ~prefill contributions with payout_type field added to Contribution model~
- ensure new contributions have the payout_type set

mds1:
Closes #7744 
Closes #7751

This PR updates the full Ethereum grant checkout flow to work as follows:
1. User adds items to their cart
2. User starts checkout with the L1 `BulkCheckout` contract or with zkSync Checkout
3. User's cart is backed up in a JSON Store with `key=user_address` and `view='ethereum_cart_data'`
4. Once a transaction hash is received, we use the `bulk-fund` endpoint to create all contribution objects
    1. Contributions for L1 checkouts have `split_tx_id` equal to the L1 transaction hash. This transaction executed all transfers for all items in the cart. They also have checkout_type=eth_std`
    2. Contributions for zkSync checkouts have `split_tx_id` equal to `sync-tx:uniqueZkSyncTxHash`. Each contribution has its own unique transaction hash, `uniqueZkSyncTxHash` corresponding to a single transfer on zkSync. All transaction hashes come are provided to us as the promise resolution from zkSync checkout, and they are already prefixed with `sync-tx:` when we receive them. They also have checkout_type=eth_zksync`
5. Once the `bulk-fund` endpoint finishes, the response it sends to the frontend returns success/failure data for all contributions. Successes indicate contributions were created successfully, failures indicate the contribution could not be created
6. If there are no failures, we delete the JSON Store created in step 3, and we are done!
7. If there are failures, we replace the JSON Store created in Step 3 with a new one where `key=user_address-timestamp` and `view='ethereum_cart_data'`. This way we have their cart data backed up and future checkouts do not overwrite it. Now we are done!

The transaction validator now works as follows:
1. If `checkout_type == 'eth_zksync'`
    1. Parse the transaction hash from the `split_tx_id` field
    2. Using the zkSync [REST API](https://zksync.io/api/v0.1.html#rest-api), fetch transaction data
    3. Update the `originated_address` and `success` status based on the transaction data (and update other fields to indicate validator was successful)
2. If `checkout_type == 'eth_std`
   1. Continue to use the standard L1 validation process from last round that parses event logs to make sure the transfer was sent
3. If `checkout_type` is anything else, return and do nothing. (In a separate PR it may be worth moving the zCash validation here, that way all contribution validation is from the same script/cron)

Other changes and notes:
- Removed the `zksync-recovery` page and functions that are used to recover funds from a user's Gitcoin zkSync wallet and send it to their regular zkSync wallet. Since we no longer use Gitcoin zkSync wallets, this code can be removed. It can be added back later if it becomes necessary for users to recover funds.
- In `app/grants/views.py`, we remove 3 outdated grant contribution checks that now will never occur
- We never actually figured out the root cause of #7744 or how to reproduce it. However, because our architecture and validator are now significantly simpler, and because we backup cart data in a JSON Store, I don't think it should be a problem. If it becomes a problem, we should update the `bulk-fund` endpoint to return a list of contribution IDs, then add a new endpoint that the frontend can use to validate that each one exists before redirecting the user to the success modal


thelostone-mc (approved):
LGTM

PR #7944, opened 2020-11-26
Create groups for ptokens

corydickson:
##### Description

Returns a 403 on the create ptoken endpoint if a user does not have the necessary permissions and hides the model the dashboard page. 

##### Testing

Run the `ptoken` migration to add a user to the group and create the necessary permissions. Then, login and navigate to the dashboard page. If a user has permission should see the modal, else it will not appear

### DIFF
--- a/app/ptokens/management/commands/setup_ptoken_launch.py
+++ b/app/ptokens/management/commands/setup_ptoken_launch.py
@@ -1,0 +1,35 @@
+'''
+    Copyright (C) 2020 Gitcoin Core
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU Affero General Public License for more details.
+    You should have received a copy of the GNU Affero General Public License
+    along with this program. If not, see <http://www.gnu.org/licenses/>.
+'''
+from django.contrib.auth.models import Group, Permission, User
+from django.core.management.base import BaseCommand
+from django.db.models import Q
+
+
+class Command(BaseCommand):
+    help = "Setup pToken Permissions"
+
+    def handle(self, *args, **options):
+        pToken_group_name = "pToken-Seed-Round"
+        pToken_group = Group.objects.get_or_create(name=pToken_group_name)[0]
+        add_pToken_auth = Permission.objects.get(codename="add_pToken_auth")
+        pToken_group.permissions.add(add_pToken_auth)
+
+        print('Adding seeded users to pToken Group')
+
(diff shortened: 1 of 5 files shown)


app/ptokens/management/commands/setup_ptoken_launch.py
```diff
@@ -0,0 +1,37 @@
...
+        pToken_group_name = "pToken-Seed-Round"
+        pToken_group = Group.objects.get_or_create(name=pToken_group_name)[0]
+        ct = ContentType.objects.get_for_model(model=User)
+        add_pToken_auth = Permission.objects.get_or_create(name="Add pToken", codename="add_pToken_auth", content_type=ct)[0]
+        pToken_group.permissions.add(add_pToken_auth)
+
+        print('Adding seeded users to pToken Group')
+
+        # TODO: How else will we determine which users should be added to the group
+        valid_users = User.objects.filter(~Q(groups__name__in=[pToken_group_name]), profile__trust_profile=True)
```
corydickson, line 33: Just need clarity on the identifier to seed the initial user base. Email / ids ?

app/ptokens/views.py
```diff
@@ -204,6 +204,12 @@ def ptoken(request, token_id='me'):
             return JsonResponse(
                 {'error': _('You must be authenticated via github to use this feature!')},
                 status=401)
+
+        if not user.has_perm('auth.user.add_pToken_auth'):
```
corydickson, line 208: @mds1 Brought up the good point that a user would be able to abuse Vue here and possibly open the modal; this ensures that they cannot make the request given their creds

aamustapha (approved):
LGTM

chibie (approved):

HelloIPM:
OK NOTED

HelloIPM (approved):
DONE

app/ptokens/management/commands/setup_ptoken_launch.py
```diff
@@ -0,0 +1,37 @@
...
+        for user in valid_users:
+            user.groups.add(pToken_group)
+
+        print('Valid users added to pToken group')
```
thelostone-mc, line 37: Do we run this just once and be done with it ?

corydickson: yup, unless you had another way in mind. Let me know how seeding is determined and I can open another PR and add it to this command

thelostone-mc (approved):

thelostone-mc:
Just tested this ! I ran `setup_ptoken_launch` and `has_ptoken_auth` is true but I don't see a creation modal popping up anywhere

Untitled

mds1:
I'll defer to @corydickson on this one. From a quick glance [it looks like](https://github.com/corydickson/web/blob/96a5903de81cf3a919df7f443174b14a4bfd08ed/app/dashboard/templates/board/ptokens.html#L60-L73) it's set up correctly, where the "Create" button is behind `v-if="has_ptoken_auth"`.

Can we also add instructions/comments somewhere (even if just as a reply here) that explain what `setup_ptoken_launch` and `has_ptoken_auth` are both used for + how to configure them in dev mode? I've never used Django groups so a brief explanation would help!

Issue #7947, opened 2020-11-26
round 8 bug / improve list


thelostone-mc:
**MAJOR**
- [x] checkIsCurator function is broken as the logic doesn't fit the data (map collections array and check curators on each one). 
- [x] removeCollection doesn't work not id is found.
- [x] fix to 2 decimals  [in prediction curve UI](https://user-images.githubusercontent.com/5358146/100424743-07521b00-30b4-11eb-9500-70394c730490.png) **FIX: 913db656f**
- [x] make background image in RoutingPolicy as optional **FIX: fc9b24a9c**
- [x] add new field name in RoutingPolicy to make it easier to identify **FIX: fc0c8bd54**
- [x] in UI customer-name is not appended to URL [image](https://user-images.githubusercontent.com/5358146/100439252-4fc80380-30c9-11eb-9003-58cf8e75886d.png) **FIX: 76ff4383d**
- [x] CLR banner not being updating in UI despite setting image in GrantCLR.logo **FIX: 97370ea7f**
- [x] Appending `/` URL to endpoint should not return 404 **FIX: 5e56dff43**
- [x] If grant is part of multiple rounds -> the round names are not appended
- [x] made CLR banners into accordion so that it is collapsible and store that boolean in local-storage
- [x] edit grant broken. **FIX: 7547b5fbf**

**MINOR**
- [x] ensure error are cleared after submitting a match pledge  **FIX: c7767fc63**
- [x] fix broken collections UI  **FIX: 821de8c9c** 
- [ ] fix alignment in prediction curve UI [image](https://user-images.githubusercontent.com/5358146/100424503-b3dfcd00-30b3-11eb-95d2-8c709af765a6.png)
- [x] Show My Items on landing only if user is logged in **FIX: 1b33dcdfe**
- [x] update text in [Did you know section](https://user-images.githubusercontent.com/5358146/100425472-4765cd80-30b5-11eb-87bf-853f4df621f9.png) **FIX: fe0b6f11c**
- [x] Show `Add to collection` in card only if user is logged in and has a collection of their own [image](https://user-images.githubusercontent.com/5358146/100426793-7715d500-30b7-11eb-9a64-a6aba5676ace.png)
- [x] Sidebar should auto-close after canceling grant
- [x] Remove Current & Past Sponsors in UI **FIX: 42c4d73ab**
- [x] Hide prediction curve when a grant is not part of a CLR **FIX: ad530e813**
- [x] related grants styles

owocki:
[...]

PR #7961, opened 2020-11-30
fix(grants): backend twitter validation

chibie:
##### Description

Fixes for backend twitter validation

* validate twitter handle in grant edit flow
* include '@' if it wasn't included in frontend

PR #7962, opened 2020-11-30
Fixes grants 8

zoek1:
##### Description

- Fix checkIsCurator, now is space and '@' aware
- Fix removeCollection due to some browser doesn't construct the object properly
- Make optional the '@' in twitter handles
- Extract the username from twitter urls
- Validate twitter accounts using twitter urls and handles, also manage the '@' scenario
https://www.loom.com/share/2f6cabf8424b4a2b8e4fae3e841a4804

##### Refers/Fixes
https://github.com/gitcoinco/web/issues/7947

### DIFF
--- a/app/assets/v2/js/grants/_detail-component.js
+++ b/app/assets/v2/js/grants/_detail-component.js
@@ -270,6 +270,28 @@ Vue.mixin({
       }
 
     },
+    isTwitterUrl(input) {
+      const twitter_re = /^https?:\/\/(www\.)?twitter\.com\/(#!\/)?([^/]+)(\/\w+)*$/;
+      const twitter_match = input.match(twitter_re);
+
+      if (twitter_match) {
+        return twitter_match[3];
+      }
+    },
+    transformTwitterHandle1() {
+      const match_twitter = this.isTwitterUrl(this.grant.twitter_handle_1);
+
+      if (match_twitter) {
+        this.grant.twitter_handle_1 = match_twitter;
+      }
+    },
+    transformTwitterHandle2() {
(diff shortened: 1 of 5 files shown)


chibie (changes requested):
@zoek1 i see you validate against twitter url (https?://www?.twitter.com)

we intend for users to only input twitter username (`@johndoe` or johndoe), if '@' is missing, then backend includes it.

i'm handling this here #7961 , initial was #7890 

can you take out twitter validation changes from this PR?

zoek1:
Thanks @chibie for your quick review, here some comments: 
1. Check the vídeo, if the user pastes the Twitter URL, the username is extracted and replaced with the URL. 
2. The validator supports now Twitter URLs and usernames, so it's more strong. It just adds support for grants that already have twitter URLs.
3. I think is a bad idea to store the usernames with the @ character, because that requires being aware and manipulate the username to build just the Twitter URL. The @ doesn't help on anything, just introduces more validations and transformations in the code IMHO.

what do you think @thelostone-mc @octavioamu

chibie:
@zoek1 really good points. i'll prefer your solution then

i guess i'll update backend validations strictly for twitter URL.

cc: @octavioamu @thelostone-mc

thelostone-mc:
synced up with @chibie offline ! makes sense

apbendi:
Hey @zoek1, was just regression testing the Grants Round 8 branch, and Twitter validation fails with the error message "Request must include a Twitter handle", which seems to be related to the "Match Twitter URL" check you added. It's unclear to me why this is failing or what it was trying to achieve. Can you please test on your end and see if you can reproduce?

zoek1:
I see the problem @apbendi, let me make the patch for this

PR #7979, opened 2020-12-01
Prevent past expiration date in redis for Idena validation ceremony time

apbendi:
During the Idena network's validation window itself, the endpoint used for the "next" validation window returns the one currently going on. This was causing a 500 because we were attempting to set an expiration date in the past for redis.

With this change, the Trust Bonus may show a window in the past until the network itself gives us an updated value.


apbendi:
Superceded by https://github.com/gitcoinco/web/pull/7987

PR #7981, opened 2020-12-01
Make active CLR rounds togglable

aamustapha:
##### Description
Displaying active CLR rounds at the top of the Grant's page helps users to easily navigate to a spceicific CLR round they're interested in. However, in a case where there are several active CLR rounds, probably above 5, the interface easily becomes cluttered up.

This PR adds a toggle functionality to the Active CLR rounds banner at the top of the page.

##### Testing
https://www.loom.com/share/8bab707e0d80440e9fe3505abda5ca20


thelostone-mc:
@aamustapha could you cherry-pick your changes and raise a pR

chibie (approved):

aamustapha:
@thelostone-mc @danlipert @octavioamu

danlipert (approved):

thelostone-mc (approved):

app/assets/v2/js/grants/index.js
```diff
@@ -162,6 +162,7 @@ if (document.getElementById('grants-showcase')) {
       page: 1,
       collectionsPage: 1,
       limit: 6,
+      show_active_clrs: window.localStorage.getItem('show_active_clrs') != 'false',
```
octavioamu, line 165: false string? seems to me this will be === true

octavioamu: maybe you should use JSON.parse(window.localStorage.idx_status = 'false')
 and also handle when there is not localstorage show_active_clrs to avoid errors

octavioamu (changes requested):

app/assets/v2/js/grants/index.js
octavioamu: Anyway works 
default null  >  true
'false' > false
anything > true 
but in general I feel is better when we deal with booleans to keep it booleans

octavioamu (approved):

PR #8001, opened 2020-12-02
tx to json

octavioamu:
##### Description

move transaction tab to rest json

### DIFF
--- a/app/assets/v2/js/grants/_detail.js
+++ b/app/assets/v2/js/grants/_detail.js
@@ -60,6 +60,8 @@ Vue.mixin({
     fetchTransactions: function() {
       let vm = this;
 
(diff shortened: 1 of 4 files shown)


owocki (approved):
just tested this. looks good, is much more performant, would love to deploy tomororw.

nice work @octavioamu 

a few very very small things.
1. ADMIN urls seem to go to subscriptoin/ID instead of contribution/ID like they should
2. the contribution count in the top left of transaction tab is wrong + changes as you press ’load more” .. maybe it should say “50+” instead of “50"? (edited)

thelostone-mc:
@octavioamu / @owocki I'm getting this in now ! we can do a follow up PR with the minor changes as they seem trivial

PR #8018, opened 2020-12-03
Bulk cart email single celery task

corydickson:
Looks like we want to ignore the error handling, thanks for checking this out @thelostone-mc

### DIFF
--- a/app/grants/sync/helpers.py
+++ b/app/grants/sync/helpers.py
@@ -59,7 +59,7 @@ def record_contribution_activity(contribution):
         # successful_contribution(grant, subscription, contribution)
         # update_grant_metadata.delay(grant.pk)
         new_supporter(grant, subscription)
-        thank_you_for_supporting(grant, subscription)
+        thank_you_for_supporting([subscription], subscription.contributor_profile)
 
     except Exception as e:
         logger.error(f"error in record_contribution_activity: {e} - {contribution}")
--- a/app/grants/tasks.py
+++ b/app/grants/tasks.py
@@ -4,6 +4,7 @@ from decimal import Decimal
 
 from django.conf import settings
 from django.utils.text import slugify
+from django.utils.translation import gettext_lazy as _
 
 import pytz
 from app.services import RedisService
@@ -11,6 +12,7 @@ from celery import app, group
 from celery.utils.log import get_task_logger
 from dashboard.models import Profile
 from grants.models import Grant, Subscription
+from grants.utils import is_grant_team_member
 from marketing.mails import new_grant, new_grant_admin, new_supporter, thank_you_for_supporting
 from townsquare.models import Comment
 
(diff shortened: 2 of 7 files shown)


app/grants/tasks.py
```diff
@@ -118,78 +120,202 @@ def process_grant_contribution(self, grant_id, grant_slug, profile_id, package,
...
+        if grant.link_to_new_grant:
+            failures.append({
+                'active': 'grant_error',
+                'title': _('Fund - Grant Migrated'),
+                'grant':grant_id,
+                'text': _('This Grant has ended'),
+                'subtext': _('Contributions can no longer be made to this grant. <br> Visit the new grant to contribute.'),
+                'success': False
+            })
+            continue
```
thelostone-mc, line 172: ^ changes from L130 to L172 can be done in sync ! cause they contain useful checks.
So we'd do all these basic checks and pass the data here

app/grants/tasks.py
```diff
@@ -108,7 +110,7 @@ def update_grant_metadata(self, grant_id, retry: bool = True) -> None:
 
 
 @app.shared_task(bind=True, max_retries=1)
-def process_grant_contribution(self, grant_id, grant_slug, profile_id, package, retry: bool = True) -> None:
+def process_grant_contribution(self, grant_ids_list, profile_id, post_request, retry: bool = True):
```
thelostone-mc, line 113: rename this to `process_grant_contributions` and could you add a comment saying what this function does

app/grants/tasks.py
```diff
@@ -118,78 +120,202 @@ def process_grant_contribution(self, grant_id, grant_slug, profile_id, package,
...
+                'title': _('Fund - Grant Not Configured'),
+                'grant':grant_id,
+                'text': _('This Grant is not configured to accept funding at this time.'),
+                'subtext': _('Grant is not properly configured for funding.  Please set grant.contract_address on this grant, or contact founders@gitcoin.co if you believe this message is in error!'),
+                'success': False
+            })
+            continue
+
+        try:
+            package = {
```
thelostone-mc, line 200: it seems pointless to create this object here! wouls it make sense to create this outside while we do this 
https://github.com/gitcoinco/web/pull/8018/files#r537294299

app/grants/sync/helpers.py
```diff
@@ -59,7 +59,7 @@ def record_contribution_activity(contribution):
         # successful_contribution(grant, subscription, contribution)
         # update_grant_metadata.delay(grant.pk)
         new_supporter(grant, subscription)
-        thank_you_for_supporting(grant, subscription)
+        thank_you_for_supporting([grant.pk], subscription.contributor_profile)
```
thelostone-mc, line 62: Is this right?  seems like you're passing grant.pk instead of subscription

PR #8021, opened 2020-12-03
ensure amount_per_period_minus_gas_price is not 0 

thelostone-mc:
#### Description

Ensure amount_per_period_minus_gas_price is not 0 if contribution is made Gitcoin Grant

### DIFF
--- a/app/grants/models.py
+++ b/app/grants/models.py
@@ -1009,6 +1009,10 @@ class Subscription(SuperModel):
 
     @property
     def amount_per_period_minus_gas_price(self):
+        if self.grant.pk == 86:
+            # If contribution is to Gitcoin Grant
+            return float(self.amount_per_period)
+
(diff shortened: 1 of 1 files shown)


app/grants/models.py
```diff
@@ -1009,6 +1009,10 @@ def status(self):
 
     @property
     def amount_per_period_minus_gas_price(self):
+        if self.grant.pk == 86:
```
owocki, line 1012: wouldnt it be better to jsut check if `amount_per_period_to_gitcoin` is falsy?

thelostone-mc: Oh you mean something like ?
```
if self.amount_per_period ==  self.amount_per_period_to_gitcoin:
  return self.amount_per_period
```

thelostone-mc: that would fix it up

PR #8024, opened 2020-12-03
Added Idena Node Support

iRhonin:
##### Description

Added support for using Idena Node as a data source instead of `api.idena.io`.
`api.idena.io` still exists as a fallback option.

##### Refers/Fixes
#7646


apbendi:
Thanks @iRhonin! Interesting idea to fetch the data from a local node instead of the API, but I have a question: doesn't this open a security hole? A user could run a local service that responded with the expected results for a validated user, and convince the Gitcoin App that s/he was. Am I missing something here?

iRhonin:
> Thanks @iRhonin! Interesting idea to fetch the data from a local node instead of the API, but I have a question: doesn't this open a security hole? A user could run a local service that responded with the expected results for a validated user, and convince the Gitcoin App that s/he was. Am I missing something here?

Hi @apbendi, thanks for your feedback.
The user can not do that because this node is hosted on the Gitcoin side. actually, this is Gitcoin's Idena node.
Also, the user has no control over where the data coming from and everything is transparent here.

apbendi:
Ahh yes, thanks @iRhonin. I see now the localhost request is made from the backend, not the frontend, so that should have been obvious. Forgive my hasty glance at the first pass.

Sorry if I'm missing some context, but will/is Gitcoin going to be running an Idena node full time? Has that been configured and setup independent of this PR? cc @thelostone-mc

iRhonin:
> Ahh yes, thanks @iRhonin. I see now the localhost request is made from the backend, not the frontend, so that should have been obvious. Forgive my hasty glance at the first pass.
> 
> Sorry if I'm missing some context, but will/is Gitcoin going to be running an Idena node full time? Has that been configured and setup independent of this PR? cc @thelostone-mc

Sure, don't hesitate to ask anything.
Yes, for using this feature Gitcoin should run a node full time.
I have linked a document for the node setup, it is in third-party integration docs.

But this is not mandatory, without the node everything works like before.
Running a node can increase security and remove the dependency on a centralized API.

app/dashboard/idena_utils.py
```diff
@@ -100,12 +110,21 @@ def next_validation_time():
...
+            logging.warn('Idena node error: {ex}')
+
+    if status is None: 
+        url = f'https://api.idena.io/api/Identity/{address}'
+        r = requests.get(url).json()
+
+        if 'error' in r:
+            status = 'Not validated'
+        else:
+            status = r['result']['state']
```
aamustapha, line 128: Why not use `r.get('result').get('state')` should `result` fail to be resolved, so it can fail silently, since we are already handling the worse case scenario on L130.

aamustapha (changes requested):
Great work, just a little improvement suggestion

app/dashboard/idena_utils.py
iRhonin: Definitely, that is better, thank you for mentioning.

thelostone-mc:
I'm not sure if we are looking an running a full node at this point in time ! I'd say we stick to using the public API and if the feature gets traction , we evaluate running our own node

iRhonin:
@thelostone-mc should I close this PR? 
Although this feature is backward compatible.

apbendi:
Thanks @iRhonin, really appreciate you taking the time to pull this together, but I agree with @thelostone-mc that we probably won't be looking to run our own full node in the short to medium. While your changes are backwards compatible, my recommendation would be that we hold off on including the code to support full nodes until we actually make the decision to run one. Given all that, I think closing this PR for now is probably correct, but legitimate thanks again for taking the time to write it.

iRhonin:
@apbendi ok, thank you for the detailed answer.

owocki:
it works! https://bits.owocki.com/Z4u6XkD2

@iRhonin do we owe you any ETH for the build out still? let us know

iRhonin:
> it works! https://bits.owocki.com/Z4u6XkD2
> 
> @iRhonin do we owe you any ETH for the build out still? let us know

Hi @owocki, Hooray! Let me know in case of any problem.
I have received my dna coins from Idena team months ago, and Idena integration was not listed for a prize from gitcoin team, it was a seperate bounty.

PR #8049, opened 2020-12-07
Bump zksync version

mds1:
Only change is that zkSync checkout URL is updated from `https://checkout.zksync.dev` to `https://checkout.zksync.io`. The `*.dev` URL they use for testnets, and wanted to make it more clear that checkout is production code, not dev code


octavioamu (approved):

aamustapha (approved):
LGTM

PR #8067, opened 2020-12-09
Create a new migration to converge leafs.

aamustapha:
##### Description
Migration files arent' fully converged with multiple nodes. Runing migrations return
```CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0167_hackathonevent_metadata, 0157_auto_20201021_2319 in dashboard).```

I created a new migration files to depend on both leaf nodes, hence creating a single leaf node.


thelostone-mc:
Closing cause this seems to be already be fixed in master

PR #8077, opened 2020-12-10
p1: cross chain grants integration

thelostone-mc:
#### Description

- splits code into multiple files within cart
- fixes up bugs in bounties sync logic
- updated backend + frontend to accommodate newer chains. starting with CELO + ZIL for starters


app/grants/management/commands/sync_pending_contributions.py
```diff
@@ -54,3 +56,65 @@ def handle(self, *args, **options):
 
             for contribution in zcash_pending_contributions.all():
                 sync_payout(contribution)
+
+
+        # PENDING CELO TXN
+        celo_pending_contributions = pending_contribution.filter(subscription__tenant='CELO')
+        if celo_pending_contributions:
```
danlipert, line 63: I believe `if Queryset` always returns true since it hasn't been evaluated. Normally you would use `.exists()` to see if a queryset returns any records without having to evaluate the whole thing. But in this case, I don't even think the `if` statements here are needed. Within the if statement you call update on a queryset and then iterate over it, both of which will just be no-ops if the queryset is empty.

app/grants/management/commands/sync_pending_contributions.py
```diff
@@ -54,3 +56,65 @@ def handle(self, *args, **options):
...
+                tx_cleared=False
+            )
+            for contribution in contrib_to_be_expired:
+                update_grant_metadata.delay(contribution.subscription.grant.pk)
+
+            for contribution in celo_pending_contributions.all():
+                sync_payout(contribution)
+
+
+        # PENDING ZIL TXN
```
danlipert, line 77:
```
tenants = ['ZIL', 'CELO', 'POLKADOT', 'HARMONY']
for t in tenants:
    pending_contributions = pending_contribution.filter(subscription__tenant=t)
    ...
```

app/grants/utils.py
```diff
@@ -290,3 +294,11 @@ def sync_payout(contribution):
 
     if subscription.tenant == 'ZCASH':
```
danlipert, line 295: check this out:
```
tenant_payout_mapper = {
    'CELO': sync_celo_payout,
    'ZIL': sync_zil_payout,
    'POLKADOT': sync_polkadot_payout,
    'HARMONY': sync_harmony_payout}

tenant_payout_mapper[subscription.tenant](contribution)
```

put the tenant payout_mapper as a "global" in this submodule so it is loaded into memory when the server boots for even more performance.

danlipert (changes requested):
A couple of fun performance improvements, probably more lurking here

aamustapha (commented):
A suggestion, why not merge the migrations 107/108

thelostone-mc:
@aamustapha forgot to push that in ! nice catch

PR #8082, opened 2020-12-10
zksync version update

mds1:
We bump the zksync version to a version which requires us to provide API keys for node providers. Without API keys, ethers falls back to shared API keys which can be throttled. For Infura, `document.contxt.INFURA_V3_PROJECT_ID`. Do we have API keys that we can use for Etherscan, Alcehmy, or Pocket?


mds1:
@octavioamu @thelostone-mc This is now ready for review/merge, pending any additional API keys we want to include

thelostone-mc (approved):

thelostone-mc:
let's roll with INFURA for now !

danlipert (approved):
We have a paid infura account so we should be good in terms of throttling

aamustapha (approved):

thelostone-mc:
@mds1 could you resolve the conflict and we can get this in todayn

mds1:
@octavioamu Looks like the conflict is because you changed the CDN for the zkSync package. Do you have any context for why this change was made? I don't think it should matter which CDN we use but want to confirm first before resolving the conflict

thelostone-mc:
@mds1 had issues with unpkg and it wasn't loading in a few situations so @octavioamu switched it over to jsdeliver which fixed that issue. 
Here is the slack link https://gitcoincore.slack.com/archives/CAXFMVDGD/p1608218741052000

mds1:
Ok, just updated the URL 👍

PR #8083, opened 2020-12-10
Travis tests

octavioamu:


app/quests/migrations/0028_auto_20201109_2213.py
```diff
@@ -4,7 +4,8 @@
 
 
 def forwards(apps, schema_editor):
-    from quests.models import Quest
+    Quest = apps.get_model('quests', 'Quest')
```
danlipert, line 7: niiiiice

danlipert (approved):

app/quests/migrations/0028_auto_20201109_2213.py
octavioamu: that one was a hard find and the reason why people wasn't able to build locally.

Issue #8143, opened 2020-12-17
Checkout failing, funds stuck in zkSync, "API code 300; Error communicating core server"


neil-lowry:
**Describe the bug**
Trying to check out a large Gitcoin grant cart. zkSync got to the "Step 2. Transfer" stage but then fails repeatedly (5+ times now) with this error: "TxBatch failed. Reason: zkSync API response error: code 300; message: Error communicating core server". My funds are still locked in zkSync so I seem to be unable to fall back to standard checkout.

**To Reproduce**
I'm not sure you can reproduce it. I have 77 items in the cart.

**Expected behavior**
I expected to make a successful payment.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Chrome
- Browser Version: 87.0.4280.88

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

mds1:
Hey @neil-lowry, can you try doing two separate zkSync checkouts of about 39 items each? zkSync only supports up to 50 transactions in one batch (we will make this more clear from the Gitcoin cart)

neil-lowry:
I'm happy to do that, but right now my funds seem to be trapped in zkSync.

StanislavBreadless:
> I'm happy to do that, but right now my funds seem to be trapped in zkSync.

If the funds have been successfully deposited to zkSync you can simply open the checkout again. (You won't have to re-deposit them)

neil-lowry:
So remove 39 items from my cart and then check out, and the funds should still be spendable from zkSync? That seems fine if safe, I'll try that.

StanislavBreadless:
> So remove 39 items from my cart and then check out, and the funds should still be spendable from zkSync? That seems fine if safe, I'll try that.

Yep. If want to check the funds that you have on zkSync, you can log in to zkSync [primary wallet](https://wallet.zksync.io) and you'll see the assets that you have in zkSync.

neil-lowry:
Now, with 39 items in my cart, "Step 1: Setting things up" pauses for about 20 seconds, then fails with this message: "Failed to estimate batch transaction fee. Reason: Error: Network Error. The error above makes it either impossible to proceed to checkout or prevents the opener page from recording that you completed the checkout. You should leave the page."

I have tried this a few times, plus closing and restarting from the cart page.

neil-lowry:
[...]

PR #8145, opened 2020-12-17
Limit zkSync checkout to 45 items in cart

### LINKED ISSUE #8114: Trying to donate to 98 grants with zksync doesnt work
I put the deets into their github cause it's their issue... but wanted to keep you in the loop

https://github.com/matter-labs/zksync/issues/266

mds1:
Closes #8114 and #8143

To test:
1. Either add 45 items to your cart, OR change `maxItems` in `cart-ethereum-zksync.js` to a number like 2. The number set there is the max allowed
2. If you change it to 2, add 2 items to your cart and things should look normal. Add a third and the zkSync button should be disabled with a message

The actual zkSync limit is 50, but we use 45 cart items to leave room for automatic tip contributions which are added as separate transfers


thelostone-mc (approved):
locally tested ! LGTM

Issue #8215, opened 2021-01-08
Quests - Kudos issuance fails when a user already holds the same Kudos issued from a different Quest


gdixon:
**Describe the bug**

When a previously uncompleted Quest is successfully completed and the user goes on to claim the associated Kudos, if the user has already received that Kudos from a different Quest, then the issuance fails with the following error: 

> "Sorry but the coupon for a free kudos has been used already."

**To Reproduce**

- Find a Kudos that has multiple Quests associated
- Complete more than one of those Quests

**Expected behavior**
- The user should receive multiples of the same Kudos so long as the Kudos is still mintable.

Mamu9:
Any solution? I'm getting the same issue.

PR #8216, opened 2021-01-08
Quests - Fixes failing Kudos issuance when a user already holds the same Kudos issued from a different Quest

gdixon:
##### Description

Adds quest_pk to the BulkTransferCoupon model and adds associations inside get_or_create_price_url so that a Kudos may be issued multiple times if rewarded by multiple Quests.

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/8215

##### Testing

No tests were added as I couldn't find any that addressed this usecase, please point me in the right direction if this is something we need a test case for.


app/kudos/models.py
```diff
@@ -646,6 +646,7 @@ class BulkTransferCoupon(SuperModel):
 
     sender_address = models.CharField(max_length=255, blank=True)
     sender_pk = models.CharField(max_length=255, blank=True)
+    quest_pk = models.CharField(max_length=255, blank=True)
```
owocki, line 649: `models.ForeignKey` is probalby what u want to use not a charfield

app/quests/helpers.py
```diff
@@ -188,13 +188,15 @@ def get_or_create_prize_url(quest, profile):
         btcs = BulkTransferCoupon.objects.filter(
             token=quest.kudos_reward,
             tag='quest',
+            quest_pk=quest.pk,
```
owocki, line 191: hmmm so the only side effect of this we should be aware of is that, as of w2 jan when this is released, then all of the 100k beaten quests in the platform, will all generate new kudos reward links if the user beats them a second time.

maybe thats ok tho now that kudos are on xdai + easy/cheap to issue.  if ppl want a second reward then nbd.

app/kudos/models.py
gdixon: I did consider a foreignKey - but didn't think we would need to follow that association? I suppose it would future proof us though?

app/quests/helpers.py
gdixon: Ahh, that could be an issue, but without retroactively assigning the pk from previous quests into the db, I can't see a way to solve this in the code? 

From what I can see, it's either; allow them all to be re-rewarded, or block any that have already been rewarded from this check (ie put an or into the filter, so that we're checking for a match OR an isNull on the quest_pk)?

What would your preference be?

owocki: > Ahh, that could be an issue, but without retroactively assigning the pk from previous quests into the db, I can't see a way to solve this in the code?

yeah, thats probably the most elegant solution.   we'd have to guess which quest it was if a user had gotten multiple bulk checkouts with the same kudos, but otherwise itd work fine

app/kudos/models.py
owocki: no idea if we plan to follow the foreign key in the future, im just a fan of using the right type for the job.

thelostone-mc: could we add in help_text as well while we add new fields ? 
I'm getting that done for existing fields

gdixon: @owocki - That's great, thanks for confirming.

@thelostone-mc - No worries, I'll add help_text with this next round of changes.

app/quests/helpers.py
gdixon: Brilliant, I think I have a fairly concise data migration written up - I'll push it now.

thelostone-mc:
@gdixon is this something we still wanna get in ?

gdixon:
I believe it is still a bug, but the migration is going to take a long time to run, it would be better in a management command! 

I'll refactor when I get chance!

thelostone-mc:
Closing this out as we intend to retire kudos

PR #8220, opened 2021-01-08
grants: add harmony support

thelostone-mc:
##### Description

- [x] Grant Creation
- [x] Grant Detail Page (View + Update)
- [x] Add to Cart
- [x] Contribute Flow 
   - when one wallet extension is not installed
   - when not enough balance
   - when wallet is locked
   - when sign is rejected
   - when signature is approved
- [x] Chain validation + sending email on successful validation 

https://share.vidyard.com/watch/E47CWxua89CM8xM5PtcCLA?


app/grants/views.py
```diff
@@ -3048,12 +3048,12 @@ def contribute_to_grants_v1(request):
             })
             continue
 
-        if is_grant_team_member(grant, profile):
-            invalid_contributions.append({
-                'grant_id': grant_id,
-                'message': 'error: team members cannot contribute to own grant'
-            })
-            continue
+        # if is_grant_team_member(grant, profile):
```
chibie, line 3051: why is this commented?

chibie (changes requested):

app/grants/views.py
thelostone-mc: cause i don't think should be present at this point ! commenting it out and will take it out later if folks are good with it

app/assets/v2/js/cart.js
```diff
@@ -371,6 +374,9 @@ Vue.component('grants-cart', {
         case 'ZIL':
           vm.chainId = '102';
           break;
+        case 'HAROMONY':
```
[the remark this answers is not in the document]
thelostone-mc: oh thanks for that ^_^

app/grants/views.py
thelostone-mc: this is most likely being removed in my follow up PR with the polkadot stuff

chibie (approved):

app/grants/views.py
octavioamu: but we are not longer validate if the contributor is also part of the grant?

octavioamu (approved):
looking good the only thing I noticed is the harmony tab is trying to connect with web3modal

thelostone-mc:
> looking good the only thing I noticed is the harmony tab is trying to connect with web3modal

Yup this is something we'll have to fix for all non web3 related flows

PR #8242, opened 2021-01-13
feat(grants): enable grant owner opt-out of clr matching

### LINKED ISSUE #7272: allow a grant to opt out of matching, so they dont partciipate in QF
allow a grant to opt out of matching, so they dont partciipate in QF

this is a requirement i got from the YAM finance community; apparently they dont want any QF money since they hav ea business model for their project.. and dont want to be accused of trying to shill us a non public good on a public good platform.

doing htis would enable them to still use grants

chibie:
##### Description

Allow grant owner opt-out of CLR matching

image

##### Refers/Fixes

closes #7272 

##### Testing

https://www.loom.com/share/b81e5090a61b48cf922907c5ffe7c2f6


app/assets/v2/js/grants/_new.js
```diff
@@ -313,6 +314,7 @@ if (document.getElementById('gc-new-grant')) {
           description_rich: '',
           team_members: [],
           region: null,
+          opt_out_clr: null,
```
thelostone-mc, line 317: why null ?  shouldn't default be false ?

app/assets/v2/js/grants/_new.js
```diff
@@ -165,7 +165,8 @@ Vue.mixin({
         'grant_type': form.grant_type,
         'categories[]': form.grant_categories,
         'network': form.network,
-        'region': form.region
+        'region': form.region,
+        'opt_out_clr': form.opt_out_clr ? '1' : '0'
```
thelostone-mc, line 169: could we stick to true / false as opposed to 1/0 ?

app/grants/templates/grants/_new.html
```diff
@@ -192,6 +190,16 @@ <h1 class="text-center font-title-xl">Create a Grant</h1>
                 </v-select>
               </div>
 
+              <!-- CLR Matching -->
+              <div class="col-12 mb-3">
+                <label class="font-caption letter-spacing text-black-60 text-uppercase" for="grant_clr">CLR Matching</label>
+
+                <div>
+                  <input name="grant_opt_out_clr" id="grant_opt_out_clr" type="checkbox" value="1" v-model="form.opt_out_clr">
+                  <label for="grant_opt_out_clr">Opt Out of CLR Matching</label>
```
thelostone-mc, line 199: might need some UX explaning what CLR is.
@willsputra / @PixelantDesign

app/grants/templates/grants/cart/eth.html
```diff
@@ -49,7 +49,7 @@
           <div v-else-if="grant.is_on_team" class="clr-match-box font-caption font-weight-semibold d-flex flex-nowrap align-items-center">
             Cannot match own grant
           </div>
-          <div v-else-if="grant.is_clr_eligible == true" class="clr-match-box d-flex flex-nowrap align-items-center">
+          <div v-else-if="grant.is_clr_eligible == true && grant.opt_out_clr == false" class="clr-match-box d-flex flex-nowrap align-items-center">
```
thelostone-mc, line 52: can we create a new property is_grant_eligible_for_clr which combines `grant.is_clr_eligible == true && grant.opt_out_clr == false` and we just use that ?

app/grants/views.py
```diff
@@ -1634,7 +1634,10 @@ def grant_edit(request, grant_id):
         if region:
             grant.region = region
 
+        grant.opt_out_clr = bool(int(request.POST.get('opt_out_clr')))
```
thelostone-mc, line 1637: have a default as False ?

app/grants/views.py
```diff
@@ -1822,6 +1825,7 @@ def grant_new(request):
             'region': request.POST.get('region', None),
             'clr_prediction_curve': [[0.0, 0.0, 0.0] for x in range(0, 6)],
             'grant_type': GrantType.objects.get(name=grant_type),
+            'opt_out_clr': bool(int(request.POST.get('opt_out_clr')))
```
thelostone-mc, line 1828: default as False?

thelostone-mc (changes requested):
Left a few comments

app/assets/v2/js/grants/_new.js
chibie: `true` / ` false` / `null` / `undefined` are sent as string values to backend, so it will always be evaluated as python `True`

that's why i did `bool(int(request.POST.get('opt_out_clr')))` as a way to convert '1' / '0' to python boolean

what do you think?

app/grants/views.py
chibie: yes in db, but also here `'opt_out_clr': form.opt_out_clr ? '1' : '0'`

string '0' evaluates as python `False` i.e `bool(int(<'0'>))`

PixelantDesign (changes requested):
Lookin' good!

Could we **change** the word CLR to Quadratic Funding?
For the explanation...let's also[ link the word Quadratic Funding to this page](https://wtfisqf.com/?grant=&grant=&grant=&grant=&match=1000) and have it open in a new window.

chibie:
> Lookin' good!
> 
> Could we **change** the word CLR to Quadratic Funding?
> For the explanation...let's also[ link the word Quadratic Funding to this page](https://wtfisqf.com/?grant=&grant=&grant=&grant=&match=1000) and have it open in a new window.
> 
> If we are moving away from the term CLR, we should do this everywhere @Kweiss

image

frankchen07:
LGTM, maybe we can try the staging environment Wednesday for this!

app/assets/v2/js/grants/_new.js
thelostone-mc: I'd still say stick to true/false! I know we had this discussion a few months back cause I had the same question -> and we agreed to use true / false -> which is I ended up sticking to that 

https://gitcoincore.slack.com/archives/G01LRE6F39A/p1606144083009000

app/grants/templates/grants/cart/eth.html
aamustapha: Totally agree with this since there's so much repetition. Also, I'd recommend sticking to use either 
`grant.opt_out_clr == false` or `!grant.opt_out_clr` but not both.

aamustapha (commented):
Amazing work. kudos, just a few fixes

app/grants/templates/grants/cart/eth.html
chibie: here

```
    @property
    def is_grant_eligible_for_clr(self):
        return self.is_clr_eligible == True and self.opt_out_clr == False
 ```

is there a need to keep both `is_clr_eligible` and `is_grant_eligible_for_clr` ?

thelostone-mc (approved):

app/assets/v2/js/grants/_new.js
chibie: done

thelostone-mc (approved):

thelostone-mc:
looks like we have travis failures

thelostone-mc:
closing this out for https://github.com/gitcoinco/web/pull/8607 which uses exisiting field

Issue #8238, opened 2021-01-13
chore: send 1 mail instead of n mails


thelostone-mc:
#### Description 

Project: Grants

What is it: 
Grants is a product where open source maintainers (web3+non web3) can start a grant for their project and configure it to receive funding on a ETH address (other chains are also supported)

Scope:
For the sake of simplicity / keeping the scope cleaner, this task would require focussing only on grants which have been created with an ETH address (and not other chains)

Flow:
- Once a grant has been created. Users of Gitcoin can discover the created grants on gitcoin.co/grants and add them to the side cart by clicking on 'Add to Cart' button. The user is allowed to add multiple grants to the cart and click on checkout in order to fund the multiple grants.
- On the checkout page -> the user is presented with an option to payout (standard checkout / zksync checkout). Currently when a user has at least 2 grants in the cart and checkout with one of the options , they get a mail sent to them for each grant they contribute to 

This feels spammy as we have scenarios where users add like 5 grants -> checkout once but they get 5 emails for it
Ideally, we'd want them to receive only 1 email.
This would also involve tweaking the email template to allow it to show multiple contributions within the same email

Things to check once this task is completed:
- User funds 1..n grants -> always gets one mail
- Txn is created successfully in Subscription and Contribution Table and you can see the txn on the grant transaction (displayed on the UI)

starting point to see the code: https://github.com/gitcoinco/web/blob/master/app/grants/views.py#L2048
trigger for email (which happens within a celery task : https://github.com/gitcoinco/web/blob/master/app/grants/tasks.py#L255)

Note: 
- Pleasure ensure your env variables are configured to ensure stuff like github login, sendgrid etc have been configured to ensure you are able to login to the app and also send an email

Related PR where we attempted this: https://github.com/gitcoinco/web/pull/8018
Related Issue: https://github.com/gitcoinco/web/issues/7755

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.005 BNB (0.19 USD @ $37.54/BNB)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/8238/100024591).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/8238/100024591).
[...]

PR #8258, opened 2021-01-18
Send one mail instead of n mail to supporter on checkout of n grants

tarunbhm:
##### Description

- Made process_grant_contribution task return grant and created subscription objects
- Added batch_process_grant_contributions task to process multiple grants in one task
- Added batch_thank_you_for_supporting method to render and send email for multiple grant subscriptions
- Added batch_thank_you_for_supporting.html and batch_thank_you_for_supporting.txt templates
- Updated grants/views.py to use new batch_process_grant_contributions task

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/web/issues/8238


thelostone-mc:
could you add a video @tarunbhm ?

thelostone-mc (commented):
@tarunbhm code wise this make sense !
Although as opposed to adding new email template, could we replace the old ones ?
`thank_you_for_supporting` is used in one more place  in `helper.py`

This is when grants are created using CELO / ZIL address

tarunbhm:
@thelostone-mc Do you want to see mail in the video? I had tested by adding logs for generated mails and had verified that only one mail was generated. I wan to avoid creating an account on sendgrid.

thelostone-mc:
@tarunbhm the email being sent out is new code added ->. they does qualify as having frontend changes
Plus having a video attached -> make it easier for the non-engineers to review the flow as well

> I had tested by adding logs for generated mails and had verified that only one mail was generated.

I would be a lot more comfortable if we could have this tested out as well cause having to rollback in cause is something was wrong in the email is painful

tarunbhm:
> I would be a lot more comfortable if we could have this tested out as well cause having to rollback in cause is something was wrong in the email is painful

I had tested by logging rendered email body both txt and html so we are covered there.

As you mentioned video is required for non tech people, I will create a video with mail in inbox.

aamustapha (approved):

thelostone-mc (approved):
image


Tested multiple ETH grants + 1 cross chain grants
Flow seems to work without any issues

PR #8260, opened 2021-01-18
manual fallback to binance address autofill for new bounty

### LINKED ISSUE #8249: Address autofill fails with no Binance wallet extension installed
**Describe the bug**
Address autofill fails when a user doesn't have binance wallet extension installed.

**To Reproduce**
* Disable binance wallet extension (if already installed)
* Navigate to gitcoin.co/new
* Select Binance as bounty chain

**Expected behavior**
there's two cases:

**case 1:** user doesn't have binance wallet extension or extension is disabled => show user a modal with instructions to either download extension or enable installed extension.

**case 2**: user doesn't have binance wallet extension or extension is disabled => allow user manually enter binance address, disable autofill.

**N.B**: binance addresses are exactly same as eth addresses (0x...xxx..xxx) so it's not possible to validate the binance address.

As discussed in Slack, case 2 is simpler and cleaner, a disclaimer will be added so people make sure they double-check they are using the right address.

**Screenshots**
image

cc: @PixelantDesign @Kweiss

chibie:
##### Description

Allow manual filling of binance address as a fallback to autofilling when user doesn't have Binance Wallet Extension installed.

##### Refers/Fixes

#8249 

##### Testing

image

### DIFF
--- a/app/assets/v2/js/pages/hackathon_new_bounty.js
+++ b/app/assets/v2/js/pages/hackathon_new_bounty.js
@@ -62,6 +62,15 @@ Vue.mixin({
       });
 
     },
+    getBinanceSelectedAccount: async function() {
+      let vm = this;
+
+      try {
+        vm.form.funderAddress = await binance_utils.getSelectedAccount();
+      } catch (error) {
+        vm.funderAddressFallback = true;
+      }
+    },
     getAmount: function(token) {
       let vm = this;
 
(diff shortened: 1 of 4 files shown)


chibie:
i removed autofill but i'm working on a better flow

```
if user doesn't have wallet extension
        display(Looks like you don't have the Binance Wallet Extension or the extension is disabled.
                     Please enter your Binance Wallet address. We are unable to validate a Binance address. Please confirm that you 
                     are using the correct Binance address.)
else:
        autofill address from extension
```

@PixelantDesign

aamustapha (approved):

thelostone-mc (approved):

PR #8274, opened 2021-01-20
Let users ingest missing contributions

### LINKED ISSUE #8250: Create a way for users to ingest Tx that were missed by our system:
We can have a page with a simple form, and have that form POST to an endpoint which runs the ingestion script. If users try to put in fake contributions this way, the backend transaction validator will catch it. The obvious benefit of this approach is we can just point users to a special page so they can ingest it themselves.

Use this form as a template for the UI: https://gitcoin.co/grants/matching-partners/new

mds1:
Closes #8250 

- The new page is located at `/grants/add-missing-contributions`
- The logic used in `ingest_grant_txns.py` was copied in the new endpoint that lives in the `ingest_contributions` method of `views.py`
- Video shown below

https://user-images.githubusercontent.com/17163988/105187216-65651280-5ae7-11eb-871b-4096b5f1122e.mp4


PixelantDesign (commented):
Looks good. How do users get here?

mds1:
@PixelantDesign For now we will just send the link to users when they report missing contributions

app/assets/v2/js/grants-ingest-missing-contributions.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+});
+
+if (document.getElementById('gc-grants-ingest-contributions')) {
+
+  appIngestContributions = new Vue({
+    delimiters: [ '[[', ']]' ],
+    el: '#gc-grants-ingest-contributions'
+  });
+}
```
thelostone-mc, line 150: can we throw this into `/grants/ingest-missing-contributions.js`

app/grants/views.py
```diff
@@ -3173,3 +3174,249 @@ def contribute_to_grants_v1(request):
...
     return JsonResponse(response)
+
+@login_required
+def ingest_contributions_view(request):
+    context = {
+        'title': 'Add missing contributions',
+        'EMAIL_ACCOUNT_VALIDATION': EMAIL_ACCOUNT_VALIDATION
+    }
+    if not request.user.is_authenticated:
+        return redirect('/login/github?next=' + request.get_full_path())
```
thelostone-mc, line 3185: this wouldn't be needed as `@login_required` is used

app/assets/v2/js/grants-ingest-missing-contributions.js
mds1: Do you mean move this whole JS file from `app/assets/v2/js/grants-ingest-missing-contributions.js` into `app/grants/ingest-missing-contributions.js`?

If so, sure. One question though—is there some sort of guidelines we're using to decide where to put JS/Vue files? It seems most JS files (including grants related ones) are located in `app/assets/v2/js/` which is why I put it there

app/grants/views.py
mds1: Good point. Remove this check in def1e4a

app/assets/v2/js/grants-ingest-missing-contributions.js
thelostone-mc: well it's always easier to group files together unless it's to be reused !
Folks do end up creating multiple files and I end up having to port it when I get the time 😅

thelostone-mc: @mds1 ^

mds1: Moved the JS file into `app/assets/v2/js/grants/ingest-missing-contributions.js` in 191d3f2

thelostone-mc (approved):

thelostone-mc (approved):

app/grants/views.py
```diff
@@ -3173,3 +3174,247 @@ def contribute_to_grants_v1(request):
...
+                "grant": subscription.grant,
+                "activity_type": "new_grant_contribution",
+                "metadata": metadata,
+            }
+
+            Activity.objects.create(**kwargs)
+            print("Saved!\n")
+
+        except Exception as e:
+            print(e)
```
owocki, line 3303: u probably want to `logger.exception` this

mds1: Since this was merged, I created a new PR with this change here: https://github.com/gitcoinco/web/pull/8320

cc @thelostone-mc

PR #8276, opened 2021-01-20
Update Dockerfile

Kweiss:
Removing version form PIP to grab latest so that Docker builds on M1 Mac.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN mkdir -p /usr/share/GeoIP/ && \
     mv GeoLite2-Country_20200128/*.mmdb /usr/share/GeoIP/
 
 # Upgrade package essentials.
-RUN pip3 install --upgrade pip==20.0.2 setuptools wheel dumb-init pipenv
+RUN pip3 install --upgrade setuptools wheel dumb-init pipenv
 
 COPY requirements/ /code/
 RUN apt-get update


thelostone-mc:
Does this boot up on your machine @Kweiss ?
Also will need to test on other machines -> to ensure it works as expected
I remember the version was pinned there for a reason

Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN mkdir -p /usr/share/GeoIP/ && \
     mv GeoLite2-Country_20200128/*.mmdb /usr/share/GeoIP/
 
 # Upgrade package essentials.
-RUN pip3 install --upgrade pip==20.0.2 setuptools wheel dumb-init pipenv
+RUN pip3 install --upgrade setuptools wheel dumb-init pipenv
```
aamustapha, line 38: The problem I see with removing specific version is compatibility. As new packages are released almost weekly, it's hard for things not to break.

aamustapha (commented):
If there's a problem with this version we could look for a more recent but stable version and fix to that instead.

octavioamu:
We added the version to avoid the travis failing, the build will not work with this change

octavioamu:
PIP versions above 20.0 changed a lot of stuff making our setup to break for some reason and installing all the versions of the packages and missing dependencies. 
If you did that locally is probably the reason of the errors you are having.

Issue #8310, opened 2021-01-27
Text in a grant is not looking good


Solexplorer:
**Describe the bug**

This grant: https://gitcoin.co/grants/1923/doruk-ismen is not looking good as the text is overlaying the card in the right:
![immagine](https://user-images.githubusercontent.com/50027175/106024154-c54f5080-60c7-11eb-87ff-ede8ccb18bb7.png)

**To Reproduce**

Check the grant I linked above

**Expected behavior**

Text should not be overlaying

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Browser:
- Browser Version:

**Smartphone (please complete the following information):**

- Device: 
- OS: 
- Browser:
- Version:

**Additional context**

gdixon:
On Firefox? It looks like we need to add `width: 100%;` to `.grant-data`.

Solexplorer:
Exactly, it's on firefox

PR #8311, opened 2021-01-27
Fix overflowing grant-data element

gdixon:
##### Description

This will fix the .grant-data elements sizing issue on Firefox, without a width it was defaulting to auto which was filling the whole container.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/8310

##### Testing

Before:

Screenshot 2021-01-27 at 17 11 10

After:

Screenshot 2021-01-27 at 17 04 43

### DIFF
--- a/app/grants/templates/grants/detail/template-grant-details.html
+++ b/app/grants/templates/grants/detail/template-grant-details.html
@@ -3,7 +3,7 @@
   <div>
     <slot></slot>
     <div class="container grid grid-template-sidebar mb-5">
-      <div class="grant-data">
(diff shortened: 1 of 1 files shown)


thelostone-mc (changes requested):
oh nice ! but could we use `w-100` instead from bootstrap ? that way we avoid introducing duplicate rules

gdixon:
@thelostone-mc Brilliant - I'll have to have a read over bootstrap's docs - I've not used it in a while, looks like a lot has changed!

thelostone-mc (approved):

PR #8322, opened 2021-01-28
Rename user-facing instances of 'personal token' to 'time token'

mds1:
Rename "Personal Tokens" to "Time Tokens". To simplify the renaming process, variables, classes, and contracts continue to use the old name, but user-facing text uses the new name. Personal tokens and Time tokens are the same thing, so you will likely see those two phrases used interchangably throughout the codebase.

@thelostone-mc One of the changes is to a migration file `0150_auto_20200904_1505.py`—should this file be changed, or should it be left alone?


aamustapha (approved):

chibie (commented):
why not replace occurrences of ```ptoken / ptokens``` with ```ttoken / ttokens```  ?  👀

thelostone-mc (changes requested):
@mds1 I'd say let's not update the 0150_auto_20200904_1505! 
Instead create a new migration file 
It makes it cleaner and less messier on stable as 0150 has already been executed in live

mds1:
> why not replace occurrences of ptoken / ptokens with ttoken / ttokens ? 👀

@chibie Just because it's a lot simpler not to! If we rename the variables we have to re-test all the functionality to make sure there's nothing we missed. Additionally, the contracts would have to be redeployed if we also wanted to rename them to be consistent, but at current gas prices that costs over $250 so doesn't seem worth it

>  I'd say let's not update the 0150_auto_20200904_1505!
> Instead create a new migration file
> It makes it cleaner and less messier on stable as 0150 has already been executed in live

@thelostone-mc Just force-pushed with a new commit that doesn't update the migration file. Do we need a new migration though? We only updated user-facing text, not variable names/DB fields, so I don't think we need one!

thelostone-mc:
@mds1 nah it should be fine!

PR #8329, opened 2021-01-29
add sync email pref from settings

octavioamu:
##### Description

Sync email pref with mautic when user submit changes.
- create fields on mautic contact

### DIFF
--- a/app/retail/templates/settings/email.html
+++ b/app/retail/templates/settings/email.html
@@ -41,6 +41,33 @@
 
 {% block scripts %}
 <script>
+
+  function changePreferences() {
+    let mauticObj = {
+      email:  document.contxt.email,
+      github_handle: document.contxt.github_handle
+    };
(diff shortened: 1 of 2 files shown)


app/retail/templates/settings/email.html
```diff
@@ -41,6 +41,32 @@ <h5>{% trans "Email Preferences" %}</h5>
...
 {% block scripts %}
 <script>
+
+  function changePreferences() {
+    let mauticObj = {
+      email:  document.contxt.email,
+      github_handle: document.contxt.github_handle
+    };
+    let newValues = $('#email-settings input.priv_checkbox').serializeArray()
+    let emailPref = {
```
thelostone-mc, line 51: ^ const vs let

octavioamu: Done

octavioamu: emailPref will change values and variables so is the only one now with let

thelostone-mc (approved):

PR #8337, opened 2021-02-02
Add a refresh button to update BrightID status on Trust tab

corydickson:
##### Description

We may want to add more sophisticated styles to this link

##### Refers/Fixes

Addresses #8161 

This no longer creates a blocking loading situation with BrightID and instead gives the user an option to click a button to manually refresh the page that would update the backend.

##### Testing

Create a BrightID a new verified account while signed in and navigate to the trust tab. Before one hour has passed click the `Refresh Status` button and see that your status has updated to verified.

### DIFF
--- a/app/dashboard/templates/profiles/trust-vue.html
+++ b/app/dashboard/templates/profiles/trust-vue.html
@@ -114,7 +114,10 @@
 
                 <br><br>Connect
                 with other users and join groups to meet the threshold.
-                
+                <a href="https://brightid.org">Learn more.</a>
+                <br>
+                <br>
+                <a href={{ get_brightid_status }} class="text-nowrap">Refresh Status</a>
               {% elif brightid_status == 'verified' %}
                 You're verified with <a href="https://brightid.org" target="new">BrightID</a>, the unique, decentralized digital identity solution.
                 <br><br>
--- a/app/dashboard/templates/profiles/trust-vue.html
+++ b/app/dashboard/templates/profiles/trust-vue.html
@@ -112,8 +112,8 @@
               {% elif brightid_status == 'not_verified' %}
                 Verify yourself on <a href="https://brightid.org" target="new">BrightID</a> by proving you're human to other humans. 
 
-                <br><br>Connect
-                with other users and join groups to meet the threshold.
+                <br><br>
+                <p class="mt-3">Connect with other users and join groups to meet the threshold.</p>
                 <a href="https://brightid.org">Learn more.</a>
                 <a href={{ get_brightid_status }} class="text-justify d-block mt-2">Refresh Status</a>
               {% elif brightid_status == 'verified' %}
--- a/app/dashboard/templates/profiles/trust-vue.html
+++ b/app/dashboard/templates/profiles/trust-vue.html
@@ -115,9 +115,7 @@
                 <br><br>Connect
                 with other users and join groups to meet the threshold.
                 <a href="https://brightid.org">Learn more.</a>
-                <br>
-                <br>
-                <a href={{ get_brightid_status }} class="text-nowrap">Refresh Status</a>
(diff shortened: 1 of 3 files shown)


app/dashboard/views.py
```diff
@@ -3074,6 +3080,23 @@ def authenticate_idena(request, handle):
         }
     })
 
+@login_required
```
thelostone-mc, line 3083: ^ if you use the `login_required` decorator -> you don't need the `is_logged_in_user` checked

app/dashboard/templates/profiles/trust-vue.html
```diff
@@ -55,6 +55,9 @@ <h5>Active Now</h5>
                 Verify yourself on BrightID by proving you're human to other humans. Connect
                 with other users and join groups to meet the threshold.
                 <a href="https://brightid.org">Learn more.</a>
+                <br>
+                <br>
```
thelostone-mc, line 59: ^ could we use css instead of adding 2 <br>

app/dashboard/templates/profiles/trust-vue.html
```diff
@@ -55,6 +55,9 @@ <h5>Active Now</h5>
...
+                <a href={{ get_brightid_status }} class="text-nowrap">Refresh Status</a>
```
thelostone-mc, line 60: add `"` to `href`

app/dashboard/views.py
```diff
@@ -3074,6 +3080,23 @@ def authenticate_idena(request, handle):
...
+def recheck_brightid_status(request, handle):
+    is_logged_in_user = request.user.is_authenticated and request.user.username.lower() == handle.lower()
+    if not is_logged_in_user:
+        return JsonResponse({
+            'ok': False,
+            'msg': f'Request must be for the logged in user'
+        })
+    
+    profile = profile_helper(handle, True)
```
aamustapha, line 3092: Do we want to allow any user to be able to run this? or every user only runs this for his profile + admin | is_staff can run for other users??

aamustapha (commented):
Looking good, quick note.

app/dashboard/templates/profiles/trust-vue.html
corydickson: the quotes escapes the html, without them the link renders correctly. [
Screen Shot 2021-02-12 at 12 16 47 AM
](url)

app/dashboard/views.py
corydickson: I think this should only be available for the logged in user, there is a job that kicks off every hour that performs this action to update bright_id_status to the db so not sure if we need the staff role to do this as well.

octavioamu (approved):

app/dashboard/templates/profiles/trust-vue.html
```diff
@@ -55,9 +55,9 @@ <h5>Active Now</h5>
               {% elif brightid_status == 'not_verified' %}
                 Verify yourself on <a href="https://brightid.org" target="new">BrightID</a> by proving you're human to other humans. 
 
-                <br><br>Connect
-                with other users and join groups to meet the threshold.
-                
+                <p class=""mt-3>Connect with other users and join groups to meet the threshold.</p>
```
thelostone-mc, line 58: Error: Only support send transaction with ipc

corydickson: 🤦

corydickson:
@octavioamu @thelostone-mc should be ready for another pass

thelostone-mc:
@frankchen07 is this something we'd still want to get in ?

frankchen07:
is there a video where I can see the functionality?

corydickson:
@frankchen07 I'll rebase and record something

frankchen07:
bumping to see if we have a video?

corydickson:
Seems to be redundant...closing

Issue #8309, opened 2021-01-27
Update Sentry Python SDK


frankchen07:
### Who
As an internal team member

### What 
let's make sure to update our Sentry Python SDK

### Why
so we can take advantage of the new SDK's stability, performant capabilities, and the ability to send transactions.

### Description
Luckily, this only takes a few lines of code:

```
import sentry_sdk
sentry_sdk.init("your DSN",
    # Set traces_sample_rate to 1.0 to capture 100%
    # of transactions for performance monitoring.
    # We recommend adjusting this value in production.
    traces_sample_rate=1.0,
)
```

but here's the [documentation](https://learn.sentry.io/e2t/tc/DVttwpCmZqW1ZJd1h69KWn6W58c1_74mhzPfN8JCsD73lGmwV1-WJV7Cg-JCVWXJKz3lSCZ8W4qJG-b1hX4WJW8Ggy4y2wt9GVW1zW4gk7pCgpwW3FG0Cf2G0Y8DN2fsDpV4sly7N1lLnT0W1MDTW7wC5gk5gQ8t_W17WlYn2NdwFFMjGRNHj06c5W5m6pQH2mqyChW1L095z1MX8s0W1RCmLT3YFV1QW32L5fw6D-31fW35v5G18QwrX5V5D8M_4_SC6BW67rJ1V6mHdtqW8VDHCR4YSXFJW24qwT37cWwbtW83ZLjh42Skp934Nv1) too.

gdixon:
@frankchen07 What would be a sensible traces_sample_rate for the expected load?

PR #8343, opened 2021-02-02
Updates sentry-sdk (0.7.7->0.19.5) and introduces traces_sample_rate

gdixon:
##### Description

Updates the sentry-sdk version and introduces traces_sample_rate (set to sample 35% of requests).

##### Refers/Fixes

Refers: https://github.com/gitcoinco/web/issues/8309

##### Testing

@octavioamu provided figures on reqs/min to get us to the 0.35 figure. If the load changes, we might need to readdress how much sampling we do.

### DIFF
--- a/app/app/settings.py
+++ b/app/app/settings.py
@@ -301,7 +301,8 @@ RAVEN_JS_VERSION = env.str('RAVEN_JS_VERSION', default='3.26.4')
 if SENTRY_DSN:
     sentry_sdk.init(
         SENTRY_DSN,
-        integrations=[DjangoIntegration(), CeleryIntegration()]
+        integrations=[DjangoIntegration(), CeleryIntegration()],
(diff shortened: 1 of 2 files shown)


thelostone-mc:
yo @octavioamu mind dropping the reasons why we set value as 0.35 -> just so folks can refer this PR later if needed

thelostone-mc (approved):

gdixon:
> yo @octavioamu mind dropping the reasons why we set value as 0.35 -> just so folks can refer this PR later if needed

@thelostone-mc  - theres ~700 reqs/min and the example given in the docs (https://docs.sentry.io/product/performance/distributed-tracing/) is aiming for around ~250/min, (700 * 0.35 will give us ~245). 

If this looks like too few or too many samples in practice we can ratchet up or down once we see the data coming into sentry.

PR #8362, opened 2021-02-05
Fix None reference in JS + conditional check if style is not undefined

aamustapha:
##### Description
There's a None reference from Python to JS where JS doesn't recognize None.
Also, if there's no style, we shouldn't access it's properties.


aamustapha:
Fixed in #8361

PR #8373, opened 2021-02-08
remove crossorigin

octavioamu:


gdixon (approved):
Everything looks good to me (apart from `weekly_founder_recap.html` - but we can drop the \<link\> as part of the redesign/tidy up)

PR #8375, opened 2021-02-09
Start cart refactor

mds1:
**This PR is still a work in progress and is not yet complete**

# Questions

1. What's the "right" way to stretch content so the footer is at the bottom here? I'm assuming this will be handled at the site-wide layout level, not the page level, so I can leave this as is
image

------------------------
2. Do we have mockups for the "loading" and "empty cart" screens? Above you see loading, below is empty cart
image

------------------------
3. What are the bulleted list and heart icons in the figma mockup, shown below:
image

------------------------
4. What class should we use to create the light purple background of the "Save 40%" text shown in the screenshot below? Right now I'm using Bootstrap's `bg-light`, but I don't think that's quite what we're looking for
image

------------------------
5. Why do row elements spill outside the row? Am I not using Bootstrap classes the right way? For example, see the image below where the grant images hang over the left image, and the standard checkout button hangs over the right edge

image

------------------------

6. Is there any way to support hot module reloading so we don’t need to keep refreshing the page to see updates? This significantly slows me down, so even if it takes a few hours to implement I think it’s definitely worth it. My workflow is typically: make small changes > view them > repeat until done > move to next part of page. So you can see how this results in a lot of time waiting around for refreshes. I could try to batch multiple tweaks of different sections of the page to reduce refreshes, but it’s harder to keep track of what you are doing in each section

------------------------

7. I don't see mobile cart mockups, so the mobile formatting currently does not look good. More generally, should mobile be given the same priority as desktop views? Or are most users desktop users and we should focus on that first, then circle back to fix mobile at the end?


app/grants/templates/grants/cart-vue.html
```diff
@@ -43,76 +43,25 @@
...
-                    </font>
-                  </div>
-                  <div class="col col-12 col-md-4 col-lg-3 mt-3 mt-md-0 mx-auto text-center">
-                    <a href="/{{ username }}/trust" class="button button--primary text-nowrap" role="button" target="_blank">Improve Trust Bonus</a>
-                  </div>
-                </div>
-              </div>
-              {% comment %} Cart header {% endcomment %}
-
+            <div v-else style="margin: 0 auto;">
```
octavioamu, line 61: m-auto

thelostone-mc:
Is this good to review @mds1  ?

mds1:
Oops, yes forgot to take WIP out of the title, but it's ready!

app/grants/templates/grants/cart/eth.html
```diff
@@ -3,408 +3,260 @@
...
+    Please remove [[ grantsByTenant.length - maxCartItems ]] item<span v-if="grantsByTenant.length - maxCartItems > 1">s</span>
+    from your cart to continue checkout.
+    </div>
+    <div v-else-if="isZkSyncDown">
+      zkSync is down at the moment, so if you'd like to checkout now you may need to use standard checkout
+    </div>
+  </small>
+
+  <!-- Add some empty space before the footer -->
+  <div class="m-5"></div>
```
thelostone-mc, line 261: I'm assuming these changes are restricted to ETH in checkout and will not affect the other cross chain tabs

mds1: Correct

thelostone-mc (approved):

app/grants/templates/grants/cart/eth.html
```diff
@@ -3,408 +3,260 @@
...
 
-  {% comment %} Summary {% endcomment %}
-  <div class="flex-container font-bigger-1 text-left grant-header-row black"
-    style="font-weight: bold; margin-top: 1rem;">
-    <div style="width: 100%">
-      <div class="flex-container" style="justify-content: start">
-        SUMMARY
+  <!-- Checkout recommendation -->
+  <div class="row justify-content-end mb-3">
+    <small class="col-auto bg-light p-2">
```
gdixon, line 124: Thank you for checking in on this @mds1: `.bg-light` is used throughout the codebase, and I'm not sure what colour we will need to settle on for the rest of the design to make sense with minimal changes, I think the safest option for now would be `.bg-lightpurple`.

chibie:
can you throw in before/after screenshot(s)? 👀

chibie (commented):
can you throw in before/after screenshot(s)? 👀

app/grants/templates/grants/cart/eth.html
mds1: Thanks, updated to use `.bg-lightpurple` in 146d26a25, and will use that going forward

mds1:
@chibie Before is the same as the current site on https://gitcoin.co/grants/cart, and after is below

image

thelostone-mc:
@mds1 would be worth cherry-picking your changes and throwing just those in ?

mds1:
@thelostone-mc Let me try that. I suspect I'll still get conflicts but maybe it will be easier to resolve. I'll squash all commits on this branch, then try to cherry pick it over to a new one. If it works I'll close this PR and open a new one

mds1:
@thelostone-mc Good idea, that worked pretty well and seemed fairly easy to resolve. This PR has been replaced by https://github.com/gitcoinco/web/pull/8445

PR #8395, opened 2021-02-15
sets datetime of server on deploy to prevent https://gitcoincore.slack.com/archives/CAXQ7PT60/p1610986391011200

owocki:
##### Description

sets datetime of server to prevent https://gitcoincore.slack.com/archives/CAXQ7PT60/p1610986391011200

happened today on the celery server

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1610986391011200
##### Testing

tested locallly

### DIFF
--- a/scripts/deploy.bash
+++ b/scripts/deploy.bash
@@ -105,6 +105,9 @@ fi
 cd ~/gitcoin/coin || echo "Cannot find coin directory!"
 bash scripts/run_management_command.bash ping_google https://gitcoin.co/sitemap.xml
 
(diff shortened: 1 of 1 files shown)


thelostone-mc (approved):
Oh did we run this only on the crobox last time ?

owocki:
> Oh did we run this only on the crobox last time ?

yeah, but the celery box was never updated.  might as well do it every deploy to make sure the boxes always stay in sync (unless we want to fully setup NTP)

PR #8407, opened 2021-02-17
Setup home page, QF active

mds1:
Here's the bulk of the refactor for the home page with QF active. Some notes/questions:

1. Collections are not yet shown. This is a good example of where I think @octavioamu could implement this much faster than I can
2. Similarly there's no logic to about when QF is active vs. inactive—this page will always show
2. Hero image is an old one, guessing this is just the localhost default one
3. @gdixon Seems something is off with the footer as you can see. The page is already encapsulated with `content__main min-vh-100 d-md-flex flex-column` (which you added) but isn't pushed totally to the bottom. There's also some whitespace on the sidess
4. Header font sizes are too big on mobile, but I'm guessing that's something GPG will fix
5. This is merging into `gpg/new-components` since I rebased against that to get the live reloading
6. What are [these inline styles](https://github.com/ScopeLift/gitcoin-web/blob/91f69cbc63a6fd57f67f135fb6cfbd2fcae78a34/app/grants/templates/grants/index.html#L31-L56) used for? I think we can delete them but want to confirm

https://user-images.githubusercontent.com/17163988/108268219-124fa100-7121-11eb-9c27-290490e1c4bf.mp4


app/grants/templates/grants/index.html
```diff
@@ -146,6 +147,7 @@
           {% include 'grants/detail/side-cart.html' %}
         </div>
       </div>
+      {% endcomment %}
```
gdixon, line 152: There's a missing closing tag here (`</div>`) - That's what has knocked the footer out :)

app/grants/templates/grants/index.html
```diff
@@ -146,6 +147,7 @@
...
```
gdixon, line 151: There's a missing closing tag here (ln151: </div>) - That's what has knocked the footer out :)

gdixon, line 151: There's a missing closing tag here (ln151: `</div>`) - That's what has knocked the footer out :)

app/grants/templates/grants/shared/landing_qf_active.html
```diff
@@ -0,0 +1,97 @@
...
+      </div>
+      <div class="col-auto text-muted d-block d-md-none">
+        Selections of grants curated by the community
+      </div>
+      <div class="col-auto">
+        <a href="/grants/collections">{% trans "View all collections" %}</a>
+        <i class="ml-2 text-primary fas fa-long-arrow-alt-right"></i>
+      </div>
+    </div>
+    <!-- TODO show collections-->
```
octavioamu, line 90: this mean this pr is ready or still need work?

octavioamu:
There is an small design problem here as seems the original desings doesn't have the notifications 

so seem now the structure will be with twoo announces together 
today:
HEADER
NOTIFICATION
CONTENT
NOTIFICATION FOOTER
FOOTER

This pr:
HEADER
CONTENT
NOTIFICATION
NOTIFICATION FOOTER
FOOTER

app/grants/templates/grants/index.html
mds1: Ha, yep that fixed it! Good catch 

Dev feature request: If possible I'd love to have a formatter/linter on the HTML files, like we do with JS files. I've noticed missing or extra closing tags multiple times so far

app/grants/templates/grants/shared/landing_qf_active.html
mds1: This TODO was one of the things that became a time suck so I postponed it for now with that TODO comment

So to answer your question—up to you. Since we're merging things into `grants-9` now, my preference would to "merge fast" even if incomplete. That way the `grants-9` branch always has the latest completed state of the grants redesign, and it  minimizes merge/rebase conflicts

mds1:
> There is an small design problem here as seems the original desings doesn't have the notifications
> 
> so seem now the structure will be with twoo announces together
> today:
> HEADER
> NOTIFICATION
> CONTENT
> NOTIFICATION FOOTER
> FOOTER
> 
> This pr:
> HEADER
> CONTENT
> NOTIFICATION
> NOTIFICATION FOOTER
> FOOTER

Sorry @octavioamu, I'm not sure I understand the issue you are referencing? Is this something I need to fix, or something on `gpg/new-components` that will affect this branch? I also just rebased against the latest `gpg/new-components` if that helps

octavioamu:
I noticed on the video the footer is maybe on the wrong wrapper as have some white padding/margin.

> Sorry @octavioamu, I'm not sure I understand the issue you are referencing? Is this something I need to fix, or something on gpg/new-components that will affect this branch? I also just rebased against the latest gpg/new-components if that helps

What I posted is the structure of the page compared with the new one, notice that purple/blue line before footer is a notification area(was behind the top banner on top of the content) but as the changes here get next to the other footer notification

mds1:
> I noticed on the video the footer is maybe on the wrong wrapper as have some white padding/margin.

Yep! That was fixed here: https://github.com/gitcoinco/web/pull/8407#discussion_r582912064

> What I posted is the structure of the page compared with the new one, notice that purple/blue line before footer is a notification area(was behind the top banner on top of the content) but as the changes here get next to the other footer notification

Hmm, I didn't change the location of the notification/footer components though, and I see that purple/blue line on other pages on the site too, including on gitcoin.co. Is there a spot in the diff this was introduced? Perhaps I'm misunderstanding, but I basically just swapped out `landing_hero.html` with `landing_qf_active.html`, so I don't think that would have changed the structure

app/grants/templates/grants/shared/landing_qf_active.html
```diff
@@ -0,0 +1,97 @@
...
+  </div>
+
+  <!-- Summary of matching pools this round -->
+  <div class="row justify-content-between align-items-center text-secondary mt-3">
+    <div class="col-auto">$125K Community</div>
+    <div class="col-auto">$125K Infra Tech</div>
+    <div class="col-auto">$125K dApp Tech</div>
+    <div class="col-auto">$125K Eask Asia</div>
+    <div class="col-auto">$125K NFTs</div>
+  </div>
```
thelostone-mc, line 52: ^ This shouldn't be hardcoded right ?

thelostone-mc (changes requested):
left a comment

app/grants/templates/grants/shared/landing_qf_active.html
octavioamu: ohh true this need to came from the filters, the info already exist on the api endpoint, good catch

thelostone-mc: @mds1 I am getting this in for now => but we def want this from the API endpoint

PR #8368, opened 2021-02-05
New brand styles integration

octavioamu:
##### Description

**EPIC** - New brand styles integration

This PR:

- Builds on top of #8360
- Copies all stylesheets from ./app/assets/v2/css/\*\*/\*\*/\*.css (and other locations) -> ./app/assets/v2/scss/\*\*/\*\*/\*.scss
- Internalises vue-innersearch.css and quill.\*.css (as .scss)
- Changes all stylesheet paths to sass equivalents (with django-libsass block's) in templates (* ~100 files)
- Removes all stylesheets from ./app/assets/v2/css/\*\*/\*\*/\*.css
- Drops preloads from shared/head.html
- Fixes stylelint path and introduces stylelint-scss (temporarily adding `no-descending-specificity` until we've sorted out nesting in the sass files)
- Fixes stylesheets manually and with `stylelint --fix`
- Drops typography.css from email templates (ignored by premailer) and tests
- Adds collectstatic to travis build to enable tests to pass (with precompilation over sass assets)
- Fixes minor styling bug on quests (global .btn being overridden by page styles)
- Drops stylesheets that don't exist (having checked against live site)
- Adds offline precompilation steps for production
- Adds JS bundling to further reduce requests
- Removes duplicated script and stylesheet inclusions
- Removes jquery-old where appropriate (moving select2 out of the if slim block)
- Adds jsrender to the if slim block

##### Refers/Fixes

Refers: #8363
Refers: #8308
Refers: #8358

PR #8409, opened 2021-02-18
Adds local static cache for .scss and .js files

gdixon:
##### Description

- Saves a local copy of .scss and .js assets (with and without hash) during collectstatic so that compressor can instrument cache over STATIC_ROOT

##### Refers/Fixes

Refers: #8368

##### Testing

Tested in dev with production ENV's

Screenshot 2021-02-18 at 09 48 49

### DIFF
--- a/app/app/static_storage.py
+++ b/app/app/static_storage.py
@@ -5,7 +5,8 @@ from os.path import basename
 from secrets import token_hex
 
 from django.conf import settings
-from django.contrib.staticfiles.storage import ManifestFilesMixin
+from django.core.files.storage import get_storage_class
+from django.contrib.staticfiles.storage import ManifestFilesMixin, HashedFilesMixin
 
 from storages.backends.s3boto3 import S3Boto3Storage, SpooledTemporaryFile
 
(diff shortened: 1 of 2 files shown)


chibie (approved):
LGTM codewise... let's see how it plays out in prod experiment.

thelostone-mc (approved):

app/app/static_storage.py
```diff
@@ -47,6 +53,21 @@ def _save_content(self, obj, content, parameters):
...
 
+    def save(self, name, content):
+        # record the clean file content (pre gzip)
+        file_content = content.file
+        # Save remote copy to S3
+        super(SilentFileStorage, self).save(name, content)
+        # Only save .scss and .js files locally
+        if ".scss" in name or ".js" in name:
+            # restore the clean file_content
+            content.file = file_content
```
thelostone-mc, line 64: why do we do this ? would content.file change after we invoke L60 ?

gdixon: We have gzip enabled and from what I can tell, it's applied locally before the file is uploaded. We might not need it, but because we're saving multiple copies, I didn't want to risk losing the content.

octavioamu (approved):

PR #8428, opened 2021-02-22
Implement post-checkout Grants Modal Redesign and Refactor

apbendi:
Implements the redesigned "Thank You" modal after grants checkout, with one major limitation:

Currently the image on the modal is an old-school Gitcoin robot, left as a placeholder. The updated design calls for a collections-like image with four quandrants for grant images, plus the user's photo in the middle. However, since the user's cart is not yet a collection at this screen, I don't believe the endpoint for generating these images is set up for this use case. I was wondering if the GPG folks would have some thoughts on how to address this. Alternatively, we could use an updated static image for a version 1.

Screen Shot 2021-02-22 at 4 49 51 PM

Screen Shot 2021-02-22 at 4 50 10 PM

Screen Shot 2021-02-22 at 4 58 53 PM

Screen Shot 2021-02-22 at 4 59 08 PM

### DIFF
--- a/app/assets/v2/js/grants/funding.js
+++ b/app/assets/v2/js/grants/funding.js
@@ -1,22 +1,19 @@
-Vue.component("contribution-thanks-modal", {
-  delimiters: ["[[", "]]"],
-  data: function () {
+Vue.component('contribution-thanks-modal', {
+  delimiters: [ '[[', ']]' ],
+  data: function() {
     return {
       modalId: 'contribution-thanks',
       numberOfContributions: 0,
       donations: [],
-      tweetUrl: '',
+      tweetUrl: ''
     };
   },
-  mounted: function () {
-    const shouldShow = Boolean(
-      localStorage.getItem("contributions_were_successful")
-    );
-    this.numberOfContributions = Number(
-      localStorage.getItem("contributions_count")
-    );
+  mounted: function() {
+    const shouldShow = Boolean(localStorage.getItem('contributions_were_successful'));
 
-    this.tweetUrl = `https://twitter.com/intent/tweet?text=I just funded ${this.numberOfContributions} grants on @gitcoin ${CartData.share_url()}`;;
+    this.numberOfContributions = Number(localStorage.getItem('contributions_count'));
+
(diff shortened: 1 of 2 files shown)


apbendi:
I've rebased the PR to the update `gpg/new-components` branch, resolved linting errors, and updated one of the buttons which was not yet using a proper bootstrap variant. All changes are pushed, however, I'm not able to test locally, because after rebasing, clicking the "Standard Checkout" button fails immediately with the following error:

Screen Shot 2021-02-23 at 9 00 25 PM

app/assets/v2/js/grants/funding.js
```diff
@@ -157,46 +231,49 @@ function sideCartRowForGrant(grant, index) {
...
-    tokenDataList = tokenDataList.filter(token => token.chainId === 58);
+    tokenDataList = tokenDataList.filter((token) => token.chainId === 58);
     tokenDefault = 'DOT';
   } else if (grant.tenants && grant.tenants.includes('KUSAMA')) {
-    tokenDataList = tokenDataList.filter(token => token.chainId === 59);
+    tokenDataList = tokenDataList.filter((token) => token.chainId === 59);
     tokenDefault = 'KSM';
   } else {
-    tokenDataList = tokenDataList.filter(token => token.chainId === 1);
+    tokenDataList = tokenDataList.filter((token) => token.chainId === 1);
```
thelostone-mc, line 270: ^ could we undo these @apbendi

app/assets/v2/js/grants/funding.js
```diff
@@ -205,8 +282,8 @@ function tokenOptionsForGrant(grant) {
             <option value="${tokenDefault}">${tokenDefault}</option>
         `;
 
-    tokenDataList = tokenDataList.filter(tokenData => {
-      return (tokenData.address === grant.grant_token_address);
+    tokenDataList = tokenDataList.filter((tokenData) => {
```
thelostone-mc, line 285: this change is not needed

app/assets/v2/js/grants/funding.js
apbendi: hey @thelostone-mc, I believe I made this change due to linting errors, but happy to undo if that's not an issue.

thelostone-mc (approved):

PR #8450, opened 2021-02-26
bug: ensure bulk add works on CLR page

thelostone-mc:
##### Description

- bulk add doesn't work as expected when no type is selected on dedicated CLR page
- This PR populates clr_pk on the clr page -> which is passed as a query param to the bulk add param

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1614276119230100

##### Testing

https://share.vidyard.com/watch/yTBSJ4xB2bQ9ejjzKSLvDQ?

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -494,8 +494,8 @@ def bulk_grants_for_cart(request):
     only_contributions = request.GET.get('only_contributions', '') == 'true'
 
     try:
-        round_num = request.GET.get('clr_round')
-        clr_round = GrantCLR.objects.get(pk=round_num)
+        clr_round_pk = request.GET.get('clr_round')
+        clr_round = GrantCLR.objects.get(pk=clr_round_pk)
     except GrantCLR.DoesNotExist:
         clr_round = None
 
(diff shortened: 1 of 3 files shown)


app/grants/views.py
```diff
@@ -493,6 +493,12 @@ def bulk_grants_for_cart(request):
     following = request.GET.get('following', '') != ''
     only_contributions = request.GET.get('only_contributions', '') == 'true'
 
+    try:
+        round_num = request.GET.get('clr_round')
+        clr_round = GrantCLR.objects.get(pk=round_num)
```
chibie, line 498: there's an actual `round_num` field in `GrantCLR` model. so it looks confusing using `round_num` as a variable name here since it's actually the `clr_round_pk` that's sent as `clr_round` param.

maybe use `clr_round_pk` ?

consider using `

chibie (changes requested):

app/grants/views.py
thelostone-mc: addressed

PR #8480, opened 2021-03-03
chore: setup pgadmin

chibie:
##### Description

pgadmin allows you a GUI view into the postgres db container.

if migration issues ever made you cry, this could unblock you quicker.

##### Testing

https://www.loom.com/share/9d0a7b9a8ee149f6be343624f64dfe26

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,17 @@ services:
         reservations:
           memory: 128M
 
+  pgadmin:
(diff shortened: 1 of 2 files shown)


octavioamu:
IMO just using a UI tool like tableplus is better than introducing another image to run on our docker just for some cases when you need to fix something. + this will run on others envs like stage travis etc and better avoid.

PR #8522, opened 2021-03-10
ensure in_active_clr is cleared before recalculating

thelostone-mc:
##### Description

- comment out `create_grant_active_clr_mapping`
- invoke ` grant.in_active_clrs.clear()` in `create_grant_clr_cache` to ensure old CLR are removed 

https://docs.djangoproject.com/en/3.0/topics/db/examples/many_to_many/

##### Test Flow

- Ensure a grant is part of 2 CLRS
- Set one CLR as `is_active` : `False`
- run `create_grant_clr_cache` -> Grant should be part of only 1 CLR // this means clear works as expected
-  Set the same CLR as `is_active` : `Trur`
- run `create_grant_clr_cache` -> Grant should be part of 2 CLR


app/perftools/management/commands/create_page_cache.py
```diff
@@ -70,6 +70,7 @@ def create_grant_clr_cache():
     pks = Grant.objects.filter(active=True, hidden=False).values_list('pk', flat=True)
     for pk in pks:
         grant = Grant.objects.get(pk=pk)
+        grant.in_active_clrs.clear()
```
owocki, line 73: i think this should be a part of the `calc_clr_round` function so it can be called from anywhere with same results

PR #8546, opened 2021-03-11
Start improving missing contribution page

mds1:
- Frontend verifies that connected wallet address matches form inputs
- Frontend asks for a signature and passes that data to backend
- Backend verifies address recovered from signature matches user address

I *think* this should work, but before merging needs more testing first against a mainnet token list. The reason is because web3py does not let you recover addresses from PoA networks like Rinkeby and it throws with:

`web3.exceptions.ValidationError: The field extraData is 97 bytes, but should be 32. It is quite likely that you are connected to a POA chain. Refer http://web3py.readthedocs.io/en/stable/middleware.html#geth-style-proof-of-authority for more details. The full extraData is: HexBytes('0xd883010918846765746888676f312e31352e35856c696e757800000000000000c4f6ccfbbfd37487321a523acfb0d11475512cb299470d2991199c781e5594e80349e828106f55eb2f2a7734b377050624ea517aba9f55178d9fcd1165d9f31000')`

So to verify this works:
- Visit https://localhost:8000/grants/add-missing-contributions
- Connect wallet to mainnet and make sure your DB has the mainnet token list
- Try both options and make sure ingestion succeeds

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -68,6 +68,7 @@ from dashboard.tasks import increment_view_count
 from dashboard.utils import get_web3, has_tx_mined
 from economy.models import Token as FTokens
 from economy.utils import convert_amount, convert_token_to_usdt
+from eth_account.messages import defunct_hash_message
 from gas.utils import conf_time_spread, eth_usd_conv_rate, gas_advisories, recommend_min_gas_price_to_confirm_in_time
 from grants.models import (
     CartActivity, Contribution, Flag, Grant, GrantAPIKey, GrantBrandingRoutingPolicy, GrantCategory, GrantCLR,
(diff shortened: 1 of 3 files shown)


app/grants/views.py
```diff
@@ -3397,9 +3398,28 @@ def ingest_contributions(request):
...
     txHash = request.POST.get('txHash')
     userAddress = request.POST.get('userAddress')
+    signature = request.POST.get('signature')
+    message = request.POST.get('message')
     network = request.POST.get('network')
     ingestion_types = [] # after each series of ingestion, we append the ingestion_method to this array
 
+    # Setup web3
+    PROVIDER = f"wss://{network}.infura.io/ws/v3/{settings.INFURA_V3_PROJECT_ID}"
+    w3 = Web3(Web3.WebsocketProvider(PROVIDER))
```
owocki, line 3408: `w3 = get_web3(network)`

owocki:
just tested on my local, works gr8.

owocki (approved):

owocki:
@thelostone-mc any comments on this one?

thelostone-mc (approved):

app/grants/views.py
mds1: Just pushed 9847312 to update this if you want to retest > merge > deploy

PR #8563, opened 2021-03-12
Checkout falls back to manual ingestion if regular ingestion fails

mds1:
Checkout falls back to manual ingestion if regular ingestion fails. For simplicity, we notify the user of this in the message they are signing. Rinkeby tests do not work due as explained in #8546, so testing process is similar to that PR:

1. Add item(s) to your cart
2. Connect wallet to mainnet and make sure your DB has the mainnet token list
3. Open `cart.js`, and in `postToDatabase()` on line 932, intentionally throw an error at the beginning of the `try` statement
4. Checkout with standard checkout, and repeat for zkSync. Both should ask you to sign a message which is the fallback to manual ingestion. If that fails, an error message is shown telling the user to visit https://gitcoin.co/grants/add-missing-contributions

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -928,133 +928,207 @@ Vue.component('grants-cart', {
         });
     },
 
-    // POSTs donation data to database
+    // POSTs donation data to database. Wrapped in a try/catch, and if it fails, we fallback to the manual ingestion script
     async postToDatabase(txHash, contractAddress, userAddress) {
-      // this.grantsByTenant is the array used for donations
-      // We loop through each donation to configure the payload then POST the required data
-      const donations = this.donationInputs;
-      const csrfmiddlewaretoken = document.querySelector('[name=csrfmiddlewaretoken]').value;
+      try {
+        // this.grantsByTenant is the array used for donations
+        // We loop through each donation to configure the payload then POST the required data
+        const donations = this.donationInputs;
+        const csrfmiddlewaretoken = document.querySelector('[name=csrfmiddlewaretoken]').value;
 
-      // If txHash has a length of one, stretch it so there's one hash for each donation
-      let txHashes = txHash;
+        // If txHash has a length of one, stretch it so there's one hash for each donation
+        let txHashes = txHash;
 
-      if (txHash.length === 1) {
-        txHashes = new Array(donations.length).fill(txHash[0]);
+        if (txHash.length === 1) {
+          txHashes = new Array(donations.length).fill(txHash[0]);
+        }
+
+        // TODO update celery task to manage this so we can remove these two server requests
+        // Update the JSON store with the transaction hashes. We append a timestamp to ensure it
+        // doesn't get overwritten by a subsequent checkout
+        await this.manageEthereumCartJSONStore(`${userAddress} - ${new Date().getTime()}`, 'save', txHashes);
+        // Once that's done, we can delete the old JSON store
+        await this.manageEthereumCartJSONStore(userAddress, 'delete');
+
+        // All transactions are the same type, so if any hash begins with `sync-tx:` we know it's
+        // a zkSync checkout
+        const checkout_type = txHashes[0].startsWith('sync') ? 'eth_zksync' : 'eth_std';
+
+        // Configure template payload
+        const saveSubscriptionPayload = {
+          // Values that are constant for all donations
+          checkout_type,
+          contributor_address: userAddress,
+          csrfmiddlewaretoken,
+          frequency_count: 1,
+          frequency_unit: 'rounds',
+          gas_price: 0,
+          gitcoin_donation_address: gitcoinAddress,
+          hide_wallet_address: this.hideWalletAddress,
+          anonymize_gitcoin_grants_contributions: false,
+          include_for_clr: this.include_for_clr,
+          match_direction: '+',
+          network: document.web3network,
+          num_periods: 1,
+          real_period_seconds: 0,
+          recurring_or_not: 'once',
+          signature: 'onetime',
+          splitter_contract_address: contractAddress,
+          subscription_hash: 'onetime',
+          // Values that vary by donation
+          'gitcoin-grant-input-amount': [],
+          admin_address: [],
+          amount_per_period: [],
+          comment: [],
+          confirmed: [],
+          contract_address: [],
+          contract_version: [],
+          denomination: [],
+          grant_id: [],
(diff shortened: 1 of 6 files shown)


app/assets/v2/js/cart.js
```diff
@@ -928,133 +928,207 @@ Vue.component('grants-cart', {
...
-        body: new URLSearchParams(saveSubscriptionPayload)
-      };
+        // if (json.failures.length > 0) {
+        //   // Something went wrong, so we create a backup of the users cart
+        //   await this.manageEthereumCartJSONStore(`${userAddress} - ${new Date().getTime()}`, 'save');
+        // }
+
+        // // Clear JSON Store
+        // await this.manageEthereumCartJSONStore(userAddress, 'delete');
+      } catch (err) {
```
owocki, line 1059: is everything above here just identation changes for the try/catch? or are there logical changes?

app/assets/v2/js/cart.js
```diff
@@ -928,133 +928,207 @@ Vue.component('grants-cart', {
...
+        // Something went wrong, so we use the manual ingestion process instead
+        console.error(err);
+        console.log('Standard contribution ingestion failed, falling back to manual ingestion');
+        await this.postToDatabaseManualIngestion(txHash, userAddress);
+      }
+    },
+
+    // Alternative to postToDatabase that uses the manual ingestion process
+    async postToDatabaseManualIngestion(txHash, userAddress) {
```
owocki, line 1068: anyway to DRY this with the one on the manual ingestion page?

app/assets/v2/js/cart.js
mds1: No logic changes in this section, just indentation

app/assets/v2/js/cart.js
```diff
@@ -928,133 +928,207 @@ Vue.component('grants-cart', {
...
+      const network = document.web3network || 'mainnet';
+      const payload = { csrfmiddlewaretoken, txHash, userAddress, signature, message, network };
+      const postParams = { method: 'POST', headers, body: new URLSearchParams(payload) };
+      let json; // response
 
       // Send saveSubscription request
-      const res = await fetch(url, saveSubscriptionParams);
-      const json = await res.json();
+      try {
+        _alert('Please be patient, as manual ingestion may take 1-2 minutes', 'info', '3000');
```
owocki, line 1088: do we want to tell the user we are even doing manual ingestion?  seems more fluid to me to just abstract it from the suer

app/assets/v2/js/cart.js
```diff
@@ -928,133 +928,207 @@ Vue.component('grants-cart', {
...
+        throw new Error(message);
+      }
+    },
 
-      // if (json.failures.length > 0) {
-      //   // Something went wrong, so we create a backup of the users cart
-      //   await this.manageEthereumCartJSONStore(`${userAddress} - ${new Date().getTime()}`, 'save');
-      // }
+    // Asks user to sign a message as verification they own the provided address
+    async signMessage(userAddress) {
```
owocki, line 1107: can we DRY this with the func in the other file?

owocki (commented):
just 👁️ 'd ; curious what @thelostone-mc thinks tho

app/assets/v2/js/cart.js
mds1: TL;DR yes, but wanted to get this out ASAP and not have to worry about potentially breaking manual ingestion page and re-testing

I did it this way because it keeps things simpler for now:
1. We ask the user to sign a different message, which is the way we inform them why we are signing a message
2. In this method, we'll only ever be ingesting Standard OR zkSync, but manual ingestion page supports both
3. Error handing is a bit different here, since it will tell users to use manual ingestion page if that fails

app/assets/v2/js/cart.js
mds1: Up to you. I think we need to indicate it somehow, even if just say "Please be patient, this may take 1-2 minutes". But due to the extra signature required you can't really abstract it away entirely

app/assets/v2/js/cart.js
mds1: Yes, but didn't for similar reason to here: https://github.com/gitcoinco/web/pull/8563#discussion_r593255675

thelostone-mc:
@mds1 code looks alright apart from needing to DRY it

mds1:
Ok, pushed a few new changes to this PR:
- Pagination for the zkSync API is supported (closes #8581)
- The `/grants/add-missing-contributions` page uses the `is_staff` flag to let you enter a username, and the backend will skip signature verification if the caller is an admin. This lets staff ingest contributions on behalf of other users
- Makes the code more DRY by pulling signature logic + ingest logic into `shared.js`
- Contributor address now saves the `from` address of the transaction instead of `"N/A"`, cc @owocki 

In addition to the testing steps in the PR description:
- Try toggling your `is_staff` status and make sure the username input field on `/grants/add-missing-contributions` shows/hides, and the signature verification is executed/skipped
- Make sure the contributor address of an ingested contribution is set

owocki:
hmmm im a bit nervous about making changes to how the cart / checkout works during a live round. how well tested is that portion of the code? donno if thats needed now that we haven't seen the transactionhashes error in several days, zksync seems to have fixed it

mds1:
Yea that's a fair point. One one hand, I tested it locally and it works. On the other, testing on my local Rinkeby env is not representative of prod, and that's causes issues in the past. 

Maybe we get a bit more specific about the specific features bunched into this PR, and if you only want to merge a subset I'll open a new PR with that subset. Below is the list of changes + my risk assessment:

1. Removes our backups to JSON Store during checkout flow—I'd say this is safe to remove, and we're not benefiting from these backups anymore anyway

2. Checkout flow falls back to manual ingestion + DRY manual ingestion code—This is the part I think is arguably risky since we're changing checkout flow mid-round. On the other hand, we only really added/changed the checkout logic for the case where `postToDatabase` (the method that saves contributions) fails, by wrapping it in a try/catch, so it *should* be safe/not risky

3. Pagination of zkSync API on the backend, used for manual ingestion—Ok to keep

4. The `is_staff` functionality + signature verification bypass for manual ingestion—Ok to keep

5. Manual ingestion saves `from` address as the contributor address—Ok to keep

mds1:
Closing, as this is now replaced by https://github.com/gitcoinco/web/pull/8621, which adds items 3, 4, and 5 of the above comment. Items 1 and 2 should be done in a separate PR after the round, so we can avoid changing the checkout flow mid-round

PR #8583, opened 2021-03-13
Tune tx validator to be more conservative in how it treats pending txns.

owocki:
##### Description

Tune tx validator to be more conservative in how it treats pending txns.
##### Refers/Fixes

https://gitcoincore.slack.com/archives/CBDTKB59A/p1615655279157300

##### Testing

tested locally on a backup from today


thelostone-mc (approved):
codewise lgtm

Issue #8571, opened 2021-03-13
Help fix this signed message that will not validate.


owocki:
The scope of this bounty is to help me understand why a signed message from web3py will not validate in [this PR](https://github.com/gitcoinco/web/pull/8478).

the signed message is created [here](https://github.com/gitcoinco/web/pull/8478/files#r587697299) with `web3==4.5.0` `w3.eth.account.signHash` , and validated [here](https://github.com/gitcoinco/NFTPOC/blob/master/contracts/GitcoinPassport.sol#L19-L22) in solidity.   i can get the message to validate in the [js hardhat tests](https://github.com/gitcoinco/NFTPOC/blob/master/test/GitcoinPassport.js#L37-L38), so i know its working, i just can't figure out how to get web3py signed messages to work.

i will offer 0.4 ETH to anyone who can help me solve it.  

rules
1. first come first serve.  
2. a valid submission is a PR to https://github.com/gitcoinco/NFTPOC/ and/or to https://github.com/gitcoinco/web/pull/8478/
3. a valid submission must use `web3==4.5.0` and must sign a message in python backend that is validated by the solidity contract

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.4 ETH (706.58 USD @ $1766.44/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/web/8571/100025071).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/web/8571/100025071).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $1,463,893.61 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

developerfred:
@owocki do you need use **V5** https://web3py.readthedocs.io/en/latest/web3.eth.account.html#sign-a-message

```python
from eth_account import Account, messages

msg_hash_hex = "058c3b4c8e5dc4632b5c6b861b2c1861d53e426dc673c907ddf2651942b0f230"
private_key_hex = "b25c7db31feed9122727bf0939dc769a96564b2de4c4726d035b36ecf1e5b364"

[...]

PR #8584, opened 2021-03-13
Fixed This Web3py Signed Message That Will Not Validate In Solidity

iRhonin:
##### Description
I have found [defunct_hash_message](https://ethaccount.readthedocs.io/en/stable/eth_account.html#eth_account.messages.defunct_hash_message) from web3py that should use to hash a msg before signHash method.
I did not change any requirement version as discussed in #8571.

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/8571

##### Testing
Check this tx from my contract:
https://rinkeby.etherscan.io/tx/0xce4c7763b9d5aae69ccd9f0038fd6bf561c8021a9fedc1babb68425ddcd3158a

and here is my Gitcoin Passport
https://rinkeby.etherscan.io/token/0x7b02739251d06dc0669465db00c7986badae50b0?a=0x18ae9fc06bed0637b1d46063d6b7af1a4f97b02c

### DIFF
--- a/app/passport/views.py
+++ b/app/passport/views.py
@@ -2,6 +2,7 @@ from django.shortcuts import render
 from django.http import Http404, JsonResponse
 from dashboard.utils import get_web3
 import web3
+from eth_account.messages import defunct_hash_message
 import json
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


app/passport/views.py
```diff
@@ -2,6 +2,7 @@
 from django.http import Http404, JsonResponse
 from dashboard.utils import get_web3
 import web3
+from eth_account.messages import defunct_hash_message
```
owocki, line 5: what version of `eth_account` u using?

owocki: seems to work on my local at `eth-account==0.2.2`

PR #8594, opened 2021-03-15
prevent quickstart from redirecting to /new page

thelostone-mc:
##### Description

https://gitcoincore.slack.com/archives/CAXQ7PT60/p1615596513109800

Untitled


octavioamu:
Instead of this changes can we just add `btn-closeguide` class to the button? is what controls if a modal is present or not on the js logic  https://github.com/gitcoinco/web/pull/8381/files#diff-66f8019a317f08929e6e6724999990dcc23295c6beaa3c6bada45274e7b56bcdL68

octavioamu (approved):

PR #8599, opened 2021-03-15
only load chains lib scripts when needed

octavioamu:
##### Description

This pr move the chains js dependencies to an object and add a computed prop to with the tenants presents on the cart. For each tenant present on the cartData (grants)  this will inject the proper chain js lib needed avoiding load unused big js libs if not going to be used.

For future chains additions we will need to just add the tenant name and the object with all the js endpoints (external or internal) following the same pattern we do for the tabs.

There is a special rule for kuzama as this one have diff chain id / name but is actually part of polkadot and use the same files.

With this approach cart seems to be loading normally again and even if you pick a polkadot grant then the load time doesn't fill tedious as the other js are not loaded (unless you pick one grant of each of our supported chains, then will be the same time as today online)

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -65,7 +65,31 @@ Vue.component('grants-cart', {
       // Checkout, zkSync
       zkSyncUnsupportedTokens: [], // Used to inform user which tokens in their cart are not on zkSync
       zkSyncEstimatedGasCost: undefined, // Used to tell user which checkout method is cheaper
-      isZkSyncDown: false // disable zkSync when true
+      isZkSyncDown: false, // disable zkSync when true
+      chainScripts: {
+        'POLKADOT': [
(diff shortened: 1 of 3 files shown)


owocki:
YESSSS!

have you tested eth/zksync checkout with this.  everything still works?

octavioamu:
> YESSSS!
> 
> have you tested eth/zksync checkout with this. everything still works?

Tested only eth for now, shouldn't be a problem for zksync as didn't changed nothing there but still want to test other chains on stage to be 100% sure there are no regressions.

gdixon (approved):
Looks good to me 👍  I've tested on staging with ETH, DOT, KSM and RSK - everything connects as it should and I've successfully made a contribution with DAI 🙌

chibie (approved):
awesome! much needed!

thelostone-mc (approved):

PR #8607, opened 2021-03-16
allow admin to opt out of CLR

thelostone-mc:
##### Description

Untitled

This is duplicate of https://github.com/gitcoinco/web/pulls and uses `is_clr_elgible` instead of adding a new field cause we had a few support requests asking us to remove their grant from CLR

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -1593,6 +1593,9 @@ def grant_edit(request, grant_id):
         if not description_rich:
             description_rich = description
 
+        is_clr_eligible = json.loads(request.POST.get('is_clr_eligible', 'true'))
+        grant.is_clr_eligible = description_rich = is_clr_eligible
+
         eth_payout_address = request.POST.get('eth_payout_address', '0x0')
         zcash_payout_address = request.POST.get('zcash_payout_address', '0x0')
         celo_payout_address = request.POST.get('celo_payout_address', '0x0')
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -1593,8 +1593,9 @@ def grant_edit(request, grant_id):
         if not description_rich:
             description_rich = description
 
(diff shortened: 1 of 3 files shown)


chibie (approved):

app/grants/views.py
```diff
@@ -1593,6 +1593,9 @@ def grant_edit(request, grant_id):
         if not description_rich:
             description_rich = description
 
+        is_clr_eligible = json.loads(request.POST.get('is_clr_eligible', 'true'))
+        grant.is_clr_eligible = description_rich = is_clr_eligible
```
gdixon, line 1597: Why are we assigning `description_rich = is_clr_eligible`, this will overwrite  `description_rich`?

thelostone-mc: Ugh issues with copy pasting

gdixon (changes requested):
Should we be only be setting this state when the grant is `active`? Otherwise users will be able to mark their grants as eligible before we've approved them?

app/grants/views.py
gdixon: Awesome! 😅

app/grants/views.py
```diff
@@ -1593,6 +1593,9 @@ def grant_edit(request, grant_id):
         if not description_rich:
             description_rich = description
 
+        is_clr_eligible = json.loads(request.POST.get('is_clr_eligible', 'true'))
+        grant.is_clr_eligible = is_clr_eligible
```
gdixon, line 1597: Can we wrap this with `if grant.active:`?

thelostone-mc: hmm well if a grant is not active -> this endpoint shouldn't be accessible via UI.
So ideally they would not be able to edit their grants right ?

thelostone-mc: changed it anyways to be safer

PR #8621, opened 2021-03-17
zkSync and ingestion improvements

### LINKED ISSUE #8581: zkSync reconciliation and adjust to "Add Contribution" pages
### What 
Update the "Add contributions" page to paginate to 500 Txs instead of the current 100, to confirm we have them all.

After the round closes, we may need to a larger sweep to ensure we dont miss any transactions. Let's brainstorm how we do this too :/

mds1:
Similar to #8563, but removes items 1 and 2 from [this comment](https://github.com/gitcoinco/web/pull/8563#issuecomment-800257576), since those affect the checkout logic, and we don't want to change that mid-round

Therefore the changes in this PR are:
- Pagination for the zkSync API is supported (closes #8581)
- The `/grants/add-missing-contributions` page uses the `is_staff` flag to let you enter a username, and the backend will skip signature verification if the caller is staff. This lets staff ingest contributions on behalf of other users
- Contributor address now saves the from address of the transaction instead of "N/A"

To test, toggle your staff status in make sure both cases ingest contributions as expected. When you are a staff, make sure you can ingest on behalf of someone else


app/assets/v2/js/grants/ingest-missing-contributions.js
```diff
@@ -98,52 +106,45 @@ Vue.component('grants-ingest-contributions', {
...
         const formParams = this.checkForm();
 
         if (!formParams) {
-          return;
+          return; // return if form is not valid
         }
 
         // Make sure wallet is connected
-        let walletAddress;
+        const walletAddress = web3 ? (await web3.eth.getAccounts())[0] : undefined;
```
octavioamu, line 121: Should we be triggering  web3modal to be sure people is using it and connected to the site instead?

octavioamu (approved):

app/assets/v2/js/grants/ingest-missing-contributions.js
```diff
@@ -26,30 +27,37 @@ Vue.component('grants-ingest-contributions', {
...
+      const isFormComplete = Boolean(txHash) != Boolean(userAddress);
 
       if (!isFormComplete) {
         // Form was not filled out
-        this.$set(this.errors, 'invalidForm', 'Please enter a valid transaction hash or a valid wallet address');
+        this.$set(this.errors, 'invalidForm', 'Please enter a valid transaction hash OR a valid wallet address, but not both');
       } else {
         // Form was filled out, so validate the inputs
-        isValidTxHash = txHash && txHash.length === 66 && txHash.startsWith('0x');
+        isValidTxHash = txHash && txHash.length === 66 && ethers.utils.isHexString('0x');
```
gdixon, line 42: This will be true for any 66 char string - `... && ethers.utils.isHexString(txHash)`?

app/assets/v2/js/grants/ingest-missing-contributions.js
mds1: If they don't have a wallet, this will throw an error and they can just connect by clicking in the header. But if you prefer to manually trigger web3modal that's fine too. Let me know.

If you do want web3modal, is this the correct way to do it?

```javascript
initWallet();
await onConnect();
const walletAddress = (await web3.eth.getAccounts())[0];
```

octavioamu: yes that should work, is just a detail to be sure the UX helps the users to understand they are connected with the right wallet and help them understand if not working. 
to get the web3modal address 
`const walletAddress = (await selectedAccount); `

gdixon (approved):
@mds1 - I've left a comment, but it shouldn't block deploy - can you get this sorted when you get chance please Matt?

app/assets/v2/js/grants/ingest-missing-contributions.js
mds1: Oops this is a typo, good catch. It should be `ethers.utils.isHexString(txHash)`

I'll fix this + @octavioamu's comment in the next commit, coming shortly...

Issue #8575, opened 2021-03-13
Funding not credited with Argent smart contract wallet without zkSync


mmatous:
**Describe the bug**
After successful transaction grants remain in card, funding doesn't appear in grant list of funding. I don't know whether the sent ETH got credited to grant total. My profile says "No Grants Funded."

**To Reproduce**
 Try to fund grants with Argent over WalletConnect WITHOUT using zkSync checkout. 

**Expected behavior**
Cart is emptied and funding appears in the grant funding list.

**Desktop (please complete the following information):**

- OS: Gentoo Linux
- Browser: Mozilla Firefox
- Browser Version: 86.0.1

**Smartphone (please complete the following information):**

- Device: Moto Z Play (addison)
- OS: Android 8.0.0. (stock)

**Additional context**
TX [0xc571168e05fa0b488b80d009f7f1c6524476288aa1ac4c0329936587bfd44b0c](https://etherscan.io/tx/0xc571168e05fa0b488b80d009f7f1c6524476288aa1ac4c0329936587bfd44b0c)

mmatous:
I just found out that https://gitcoin.co/grants/add-missing-contributions exists. Tried it out with TX hash but kinda unsurprisingly it doesn't work because relayers. "Sender of the provided transaction does not match connected wallet address".

mds1:
Ahh, yes that makes sense. Someone from Gitcoin will need to ingest on your behalf. I'm hoping to add that functionality in the next day or two, and will let you know + close this issue once your contributions are in the database!

mmatous:
Hi, is there something I should do or someone I should contact now? With GR9 closing soon I'd like to make sure that my contribution goes through with appropriate CLR matching.

mds1:
@thelostone-mc @Kweiss Can we get PR https://github.com/gitcoinco/web/pull/8621 merged today? It's been ready for a while now, and we need it to ingest @mmatous's contributions + any other users with Argent wallet having this issue

mds1:
Hey @mmatous, I've ingested these contributions for you, which you can see by checking the "All Activity" section of https://gitcoin.co/mmatous

mmatous:
Great, thank you!

PR #8632, opened 2021-03-18
Add management command to update contributor addresses of N/A

mds1:
To run the script, use network of `rinkeby` or `mainnet` in the command below.

```
docker-compose exec -e DJANGO_SETTINGS_MODULE="app.settings" web python3 app/manage.py update_contributor_address "rinkeby" 
```

This script will search for all subscriptions that have a `contributor_address` of `N/A`, then find the address of the user who sent the transaction and set that as the `contributor_address`. This requires one network request per subscription, so it may take some time to run if there's a lot of subscriptions.

Note: This approach does not guarantee the correct contributor address. Because we are using the sender of the transaction as the contributor, we get the wrong address for users with wallet's that use a relayer or meta-transactions, such as Argent. In those cases, the relayer address is incorrectly listed as the sender. A more robust approach would take a non-trivial amount of work since it requires recognizing relayed transaction and parsing them to find the wallet address, and there's no universal standard for relayed transaction format

@owocki

### DIFF
--- a/app/grants/management/commands/update_contributor_address.py
+++ b/app/grants/management/commands/update_contributor_address.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""
+
+For any Subscription objects that have a contributor address of "N/A", put in the transaction's from address
+
+Copyright (C) 2020 Gitcoin Core
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""
+
+from django.core.management.base import BaseCommand
+import requests
+from dashboard.utils import get_web3
+from grants.models import Subscription
+from web3 import Web3
+
+
+class Command(BaseCommand):
+
+    help = "Inserts missing subscriptions and contributions into the database"
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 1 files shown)


octavioamu (approved):

octavioamu:
@mds1 the job need to be added to the crontab?

mds1:
Nope, @owocki will just run this as a one-time thing to backfill previous entries. Going forward we shouldn't see any more N/A entries show up, but if they do we'll have this script to fix them

gdixon (approved):
Looks good to me 🙌  Thanks Matt!! 

With the relayer issue, what sort of approach would be appropriate? How do we spot a relayed transaction and how do we find the true sender? Does each relayer follow a similar procedure or are we going to have to figure out each on a case-by-case basis? 

Is it worth writing up some pseudo code so that we can gauge the effort involved in covering these edge cases?

mds1:
@gdixon I think it's a case by case basis unfortunately. AFAIK Argent is the most popular wallet that uses a relayer, and there are probably others but I'm not sure of them offhand. We probably could do something like this to special case Argent, which actually isn't too much work now that I think about it:

For reference we'll use [the transaction](https://etherscan.io/tx/0xc571168e05fa0b488b80d009f7f1c6524476288aa1ac4c0329936587bfd44b0c) linked in #8575, and below is some JS pseudocode of how to do this:

```javascript
// Function selector for the bulk checkout `donate` method
const bulkCheckoutFunctionSelector = '0x9120491c';

// Function selector for Argent's `execute` method used to relay transactions
const argentExecuteFunctionSelector = '0xc1169548';

// Determine transaction type
const tx = ethersProvider.getTransaction('0xc571168e05fa0b488b80d009f7f1c6524476288aa1ac4c0329936587bfd44b0c');
const data = tx.data;

if (data.startsWith(bulkCheckoutFunctionSelector)) {
  // Regular wallet
  contributorAddress = tx.from;
} else if (data.startsWith(argentExecuteFunctionSelector)) {
  // Argent relayer
  contributor_address = `0x${tx.data.slice(34,74)}`;
} else {
  // However we want to handle other cases, probably fallback to tx.from
} 
```

PR #7682, opened 2020-10-15
add poap badges verification

hhio618:
##### Description

Added POAP badges verification.

###### Done
- If the ethereum account logged in on Gitcoin holds a POAP NFT that has been sitting in that wallet for more than 15 days, it can be assumed that it's a unique human.
###### WIP
- Add test
- Add a mechanism when  "Someone has prepared beforehand the round (more than fifteen days earlier) and shuffled the POAPs to many different accounts." (@poapxyz)

##### Refers/Fixes

[gitcoinco/skunkworks#187](https://github.com/gitcoinco/skunkworks/issues/187)

##### Testing

output

PR #8640, opened 2021-03-19
add poap owner account

hhio618:
##### Description
- Save POAP account owner to the user profile
- Make sure that a POAP owner account used by just one profile
##### Refers/Fixes
refs #7682


gdixon (approved):
LGTM! Thank you @hhio618! 

One thing I have noticed, but not as part of this PR - the RPC intermittently fails to return matching data when we filter the events. I tried to find an alternative, but couldn't find any way to get the block_number from the erc721 without using events, does anybody know how we could make this more reliable?

hhio618:
@gdixon You're welcome!
Can I help you with anything else here?

thelostone-mc (approved):

PR #8675, opened 2021-03-24
adds a read replica for production queue processors

owocki:
##### Description

adds a read replica for production queue processors, per my notes in #proj-performance
we will also move metabase over to the read replica
app servers will continue to hit the default DB

##### Refers/Fixes

https://docs.djangoproject.com/en/3.1/topics/db/multi-db/
https://aws.amazon.com/premiumsupport/knowledge-center/create-read-replica-rds/

##### Testing

tested locally, working a-okay!

test it yourself 
```
from grants.models import Grant
Grant.objects.using('read_replica_1').all().count()
Grant.objects.using('default').all().count()
```

important note about deployments; make sure the ENV variable is set on prod before deploying


thelostone-mc (approved):
code wise lgtm

PR #8682, opened 2021-03-24
Proof of Humanity Integration

iRhonin:
##### Description
Implemented Proof of Humanity (POH) sybil resistance.

Inspired (and copy/pasted) from `POAP` integration
 

- Enabled `Proof of Humanity` row in `trust tab`
- Check registration on POH using `isRegistered` method of [contract](https://etherscan.io/address/0xc5e9ddebb09cd64dfacab4011a0d5cedaf7c9bdb#readContract)
- Added `is_poh_verified` and `poh_handle` fields to the profile model and related migration
- Added POH to trust bounce calculation

##### Refers/Fixes

https://github.com/gitcoinco/skunkworks/issues/208

##### Testing

Grow-Open-Source-Gitcoin (2)
Grow-Open-Source-Gitcoin-_1_


app/dashboard/migrations/0174_profile_google_user_id.py
```diff
@@ -7,7 +7,7 @@
 def populate_google_id(apps, schema_editor):
     from dashboard.models import Profile
 
-    profiles_with_google_connection = Profile.objects.filter(is_google_verified=True)
+    profiles_with_google_connection = Profile.objects.filter(is_google_verified=True).only('identity_data_google', 'google_user_id')
```
iRhonin, line 10: I have to change this, accessing `is_poh_verified` and `poh_handle` before the migration (app/dashboard/migrations/0176_poh_fields.py) causes this to fail.

owocki:
amazing!

@octavioamu or @thelostone-mc can you and i tag team the review on this?

app/dashboard/migrations/0174_profile_google_user_id.py
thelostone-mc: updating an older migration might have side effects on the live server cause this would have already been executed. Is there a way we could do without this ?

app/dashboard/views.py
```diff
@@ -6494,6 +6496,62 @@ def verify_user_poap(request, handle):
...
 
+@login_required
+@require_POST
+def verify_user_poh(request, handle):
+    is_logged_in_user = request.user.is_authenticated and request.user.username.lower() == handle.lower()
+    if not is_logged_in_user:
+        return JsonResponse({
+            'ok': False,
+            'msg': f'Request must be for the logged in user'
+        })
```
thelostone-mc, line 6507: ^ this is not needed as `login_required` is already used

thelostone-mc:
Let's try and get this in this week !

app/dashboard/views.py
iRhonin: So is it ok to remove `handle` URL parameter and use `request.user.username`?
Otherwise, the check is required to make sure the user entered her/his handle only.

app/dashboard/migrations/0174_profile_google_user_id.py
iRhonin: I think this change is safe because does not change any field on DB side.
Also without this change, adding any further filled to `Profile` causes migrations to fail in this line.

iRhonin:
Hey @thelostone-mc,
I have fixed the conflict and left a comment on your review,
Is there anything else I can do for this PR?

octavioamu (approved):
LGTM

app/assets/v2/js/pages/profile-trust.js
```diff
@@ -518,6 +518,176 @@ Vue.component('poap-verify-modal', {
...
+          this.validationStep = 'validate-address';
+          this.hideTemporarily = false;
+        });
+    },
+    generateSignature() {
+      web3.eth.personal.sign('verify_poh_registration', this.ethAddress).then((signature) => {
+        this.signature = signature;
+        this.verifyPOH();
+      });
+    },
```
octavioamu, line 627: maybe missing some error handling here if something fails like wallet disconnected or user deny the verification on metamask?

iRhonin: Thank you for mentioning, I will try different scenarios and add the required error handlings.

thelostone-mc (approved):

PR #8693, opened 2021-03-24
gpg/jtbd-backend

### LINKED ISSUE #8659: gpg - jobs landing pages / backend
### Description
Handle query of data in JTBD landing pages in a performant and DRY way

Part of a larger epic: #8365 > #8658 

### Workflow

* Create the queries for each data query on the designed pages in jsonStores, queries need to be limited on used values and amounts of each based on designs (ex just query 3 grants and fields title, description, logo, etc)
* Add the queries to create_page_cache and add it to the command so will be run on the crontab together with others queries and generate a jsonstore entry for each.
* Create the views def for each of the pages querying the jsonStores by view/key and use cache for those (see context['last_posts'] = cache.get_or_set('last_posts', fetchPost, 5000) for reference)

### Queries
- [x] Top earners from leaderboard (earn)
- [x] Latest Hackathons (earn/learn/connect)
- [x] Latest Bounties (earn)
- [x] Successful grants (earn/fund)
- [x] Successful projects (connect)

### Definition of Done
- query the required information from the cache

chibie:
##### Description

backend views for jtbd landing pages

- [x] Top earners from leaderboard (earn)
- [x] Latest Hackathons (earn/learn/connect)
- [x] Latest Bounties (earn)
- [x] Successful grants (earn/fund)
- [x] Successful projects (connect)
- [x] Passionate Builders

##### Refers/Fixes

closes #8659 

##### Testing

https://share.getcloudapp.com/mXupRj6P

to test locally, first run `docker-compose exec -e DJANGO_SETTINGS_MODULE="app.settings" web python3 app/manage.py create_page_cache`


app/perftools/management/commands/create_page_cache.py
```diff
@@ -49,6 +49,285 @@
...
+
+    top_earners = bounties = []
+
+    for earner in leaderboard:
+        top_earners.append({
+            'rank': earner.rank,
+            'amount': earner.amount,
+            'avatar_url': earner.avatar_url,
+            'username': earner.github_username,
+        })
```
octavioamu, line 86: instead of looping maybe is better to do the query using values or values_list:

`LeaderboardRank.objects.filter(
        active=True, product='bounties', leaderboard='monthly_earners',
    ).values('rank', 'amount', 'avatar_url', 'username').order_by('-amount')[0:4].cache()`
    
check the docs https://docs.djangoproject.com/en/3.1/ref/models/querysets/#values
> Returns a QuerySet that returns dictionaries, rather than model instances, when used as an iterable.

  https://docs.djangoproject.com/en/3.1/ref/models/querysets/#values-list

app/retail/views.py
```diff
@@ -1574,3 +1575,39 @@ def clean_url(url):
...
 
     return TemplateResponse(request, 'admin_index.html', context)
+
+
+@cached_as(JSONStore.objects.filter(view='posts', key='posts'), timeout=1200)
+@cached_view
+def fetchPost(qt='2'):
+    jsonstores = JSONStore.objects.filter(view='posts', key='posts')
+    if jsonstores.exists():
+        return jsonstores.first().data
```
octavioamu, line 1585: this is needed here?

octavioamu (changes requested):
left a few comments code is looking good but still be more dry using values()  instead of itinerating each element in that way I believe also  the query will be faster as instead of looking for all the fields it just fetch the columns you will use.

app/retail/views.py
chibie: oops nope... i put it there for reference... taking it out now

app/perftools/management/commands/create_page_cache.py
chibie: this will throw and error cause `avatar_url` is a property not a field. @octavioamu

app/retail/views.py
```diff
@@ -1574,3 +1575,39 @@ def clean_url(url):
...
+
+
+@cached_view(timeout=86400) # cached for 24 hours
```
octavioamu, line 1588: maybe is better to name the cached object so if we need to invalidate it later will be easier

app/perftools/management/commands/create_page_cache.py
chibie: for reference

@octavioamu said

```
well you can always go with github_username and call the avatar from the html

check marketing/utils.py
def get_platform_wide_stats(since_last_n_days=90):
L291
```

app/perftools/management/commands/create_page_cache.py
```diff
@@ -49,6 +49,231 @@
...
+        },
+    }
+    view = 'jtbd'
+    keyword = 'fund'
+    JSONStore.objects.filter(view=view, key=keyword).all().delete()
+    data = json.loads(json.dumps(data, cls=EncodeAnything))
+    JSONStore.objects.create(
+        view=view,
+        key=keyword,
+        data=data,
```
thelostone-mc, line 273: would it be worth wrapping everything in try catch ?

octavioamu (approved):

octavioamu:
merging to the integration branch to facilitate work and as seems is in a good spot , we can do changes there if needed

app/perftools/management/commands/create_page_cache.py
```diff
@@ -49,6 +49,231 @@
...
+    ).values('rank', 'amount', 'github_username').order_by('-amount')[0:4].cache())
+
+    thirty_days_ago = timezone.now() - datetime.timedelta(days=30)
+    
+    bounties = list(Bounty.objects.filter(
+        network='mainnet', event=None, idx_status='open', created_on__gt=thirty_days_ago
+    ).order_by('-_val_usd_db').values('_val_usd_db', 'title', 'issue_description')[0:2])
+
+    # WalletConnect
+    featured_grant = Grant.objects.filter(pk=275).first()
```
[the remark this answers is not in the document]
octavioamu: yes they decide which grant they want by id and we put it on the query

PR #8697, opened 2021-03-25
remove override and allow close button clickable by admin

thelostone-mc:
##### Description

From connor
>Problem: “Override status” button no longer works for admins after moving from standard bounties, admins have to “view in admin” and then type in “done” in override status field to close
Request: Fix “Override status” button on bounty frontend

The fix instead :
- to expose close bounty button to admin
- remove override status  (this is done cause we don't want bounty jumping the states)

##### Testing

Screenshot 2021-03-25 at 4 20 34 PM

### DIFF
--- a/app/dashboard/templates/bounty/details2.html
+++ b/app/dashboard/templates/bounty/details2.html
@@ -41,7 +41,6 @@
             </span>
             <a v-for="link in quickLinks" class="btn btn-outline-primary px-2 mx-2 btn-sm" :href="[[link.href]]" :title="[[ link.title ]]">[[ link.label ]]</a>
             <button class="btn btn-outline-primary px-2 mx-2 btn-sm" @click="contactFunder()">Contact Funder</button>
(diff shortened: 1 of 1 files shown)


chibie (approved):

PixelantDesign (approved):

connoroday:
Looks great! This will definitely help. A further improvement (which may be another open issue) would be prompting the funder to close the bounty every time they pay out for a cooperative or contest bounty:

A modal pops up saying:
"Thanks for paying out! Would you like to close the bounty, or are you paying more users?"
With two buttons:
"Close bounty" or "I'm paying more prizes"

^This makes an admin having to close issues on behalf of funders much more rare

frankchen07:
I would say exposing the close bounty button for the funder is a great start to solving this problem. With contest or cooperative, let's see how they use the exposed closed button first before we go and add menu options for them.

PR #8698, opened 2021-03-25
update bounty expiry date to hackthon end date

thelostone-mc:
#### Issue

@connoroday 
> Problem: When extending hackathon end date, bounties expiration doesn’t change
Request: Bounty deadlines get updated when hackathon is extended

##### How to use

- Head over to a hackathon event in admin
- Update the end date and click on SAVE
- Once the hackathon end date has been updated, scroll to the bottom and hit the `Update Related Bounties Expiry Date` button and all the bounties for the hackathon event should have the same expiry date. 

Check the video for more info 

Untitled

### DIFF
--- a/app/dashboard/admin.py
+++ b/app/dashboard/admin.py
@@ -488,7 +488,7 @@ class HackathonEventAdmin(admin.ModelAdmin):
 
 
     def response_change(self, request, obj):
-        if "_auto_update_expiry" in request.POST:
+        if "_bulk_update_expiry" in request.POST:
             try:
                 bounties_to_extend = Bounty.objects.filter(event=obj)
                 for bounty in bounties_to_extend:
--- a/app/dashboard/admin.py
+++ b/app/dashboard/admin.py
@@ -19,6 +19,7 @@
 from __future__ import unicode_literals
 
 from django.contrib import admin
+from django.shortcuts import redirect
 from django.utils import timezone
 from django.utils.html import format_html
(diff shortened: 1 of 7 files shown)


PixelantDesign (approved):

app/dashboard/admin.py
```diff
@@ -486,6 +487,19 @@ def explorer_link(self, instance):
         return mark_safe(f'<a href="{url}">Explorer Link</a>')
 
 
+    def response_change(self, request, obj):
+        if "_auto_update_expiry" in request.POST:
+            try:
+                bounties_to_extend = Bounty.objects.filter(event=obj)
+                for bounty in bounties_to_extend:
+                    bounty.expires_date=obj.end_date
+                    bounty.save()
```
thelostone-mc, line 496: ^ tried using bulk save here but that didn't seem to work in admin as expected hence defaulting to saving one record at a time

connoroday:
This looks perfect! Thank you

app/dashboard/admin.py
gdixon: There's never going to be an unreasonable amount of entries anyway, and this is an admin function so should be fine

gdixon (approved):
Codewise LGTM! The only niggle I have is with the `_auto_` part of the name, this would imply that it's not something that can be ran manually -- would `_bulk_update_expiry` make more sense?

thelostone-mc:
@gdixon updated

PR #8723, opened 2021-03-29
add client side validation

thelostone-mc:
##### Description

- update placeholder to let folks know to enter chain address
- added client side validation for chains when user submits work 

##### Refers/Fixes

Screenshot 2021-03-29 at 7 10 34 PM

Screenshot 2021-03-29 at 7 11 25 PM

### DIFF
--- a/app/assets/v2/js/pages/fulfill_bounty/token.js
+++ b/app/assets/v2/js/pages/fulfill_bounty/token.js
@@ -17,6 +17,9 @@ fulfillBounty = data => {
     _alert({ message: gettext('Add valid address you would want the bounty to be sent to') }, 'error');
     unloading_button($('.js-submit'));
     return;
+  } else if (!is_valid_address(data.payoutAddress)) {
+    unloading_button($('.js-submit'));
+    return;
   }
 
   const url = '/api/v1/bounty/fulfill';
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/pages/fulfill_bounty/token.js
```diff
@@ -68,3 +71,73 @@ fulfillBounty = data => {
...
+
+const is_valid_address = (address) => {
+  switch (web3_type) {
+
+    // etc
+    // celo
+    // rsk
+
+    case 'binance_ext':
+      if (!address.toLowerCase().startsWith('bnb')) {
```
chibie, line 84: @thelostone-mc binance starts with `0x`

app/assets/v2/js/pages/fulfill_bounty/token.js
```diff
@@ -68,3 +71,73 @@ fulfillBounty = data => {
...
+      if (!address.toLowerCase().startsWith('xdc')) {
+        _alert('Enter a valid xinfin address', 'error');
+        return false;
+      }
+      return true;
+
+    case 'qr':
+
+      if (token_name == 'BTC') {
+        if (address.toLowerCase().startsWith('0x')) {
```
chibie, line 116: does BTC start with `0x` ?

chibie (changes requested):

app/assets/v2/js/pages/fulfill_bounty/token.js
gdixon: This is displaying the error when a BTC address starts with 0x 👍

gdixon (commented):
Would it be worth moving this error state reporting to the method we use in the style guide?
http://localhost:8000/styleguide/components 

It will mean we need to write up data-patterns for each of the chains so that we can properly toggle the ::invalid state, but it would add consistency. Bounty creation currently follows this pattern if you want to have a look at how its working atm.

app/assets/v2/js/pages/fulfill_bounty/token.js
thelostone-mc: forgot to remove that ! thanks for catching that

thelostone-mc:
@gdixon done
image

gdixon (approved):
Awesome!! LGTM! 😃

PR #8761, opened 2021-04-08
workaround for data migrations for 04082021 command

owocki:
workaround for data migrations for 04082021 command

1. removes the google_user_id migration
2. TODO for next release cycle, fold in the google_user_id migration to the rest of the anti sybil migrations (maybe even move all this info to a new table so we dont have to ALTER the profile table anymore
3. deploy + bite the bullet on migrating dashboard_profile then (along with all the other dashboard_profile changes that are coming in the next cycle)
4. in the future, be more careful about ALTER's during deploy (possibly stopping manage.py jobs on cronbox + celery queues before running migrations, and running any --empty migrations as async scripts instead of migraitons


octavioamu (approved):

gdixon (approved):
Looks good to me!

PR #8771, opened 2021-04-09
fix: 🐛 TypeError: Cannot read property 'eth' of null

heldrida:
##### Description

Fixes Uncaught TypeError: Cannot read property 'eth' of null

Screenshot 2021-04-09 at 23 57 35


thelostone-mc:
Oh this doesn't fix it !  If you head over to a hackathon page without web3 installed (incognito) -> the error still persists
Instead commit 265db9eee89ddaff6be33782391e9518b1075eb4 should do that and be deployed in a bit

PR #8783, opened 2021-04-15
bug: invalid literal for int() with base 10: 'found'

thelostone-mc:
##### Description

- Issue: https://sentry.io/share/issue/a9f6f0a8fef040489899661635e45e7e/
- report: 14k since deployed
- flow affected: slim clr estimate  which is triggered by cronjob

### DIFF
--- a/app/grants/management/commands/estimate_clr.py
+++ b/app/grants/management/commands/estimate_clr.py
@@ -47,10 +47,10 @@ class Command(BaseCommand):
         what = options['what']
         print (network, clr_pk, what)
 
-        if clr_pk == "all":
-            active_clr_rounds = GrantCLR.objects.filter(is_active=True)
-        else:
+        if clr_pk.isdigit():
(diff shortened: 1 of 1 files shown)


app/grants/management/commands/estimate_clr.py
```diff
@@ -47,10 +47,10 @@ def handle(self, *args, **options):
         what = options['what']
         print (network, clr_pk, what)
 
-        if clr_pk == "all":
-            active_clr_rounds = GrantCLR.objects.filter(is_active=True)
-        else:
+        if clr_pk.isdigit():
             active_clr_rounds = GrantCLR.objects.filter(pk=clr_pk)
```
chibie, line 51: why not `GrantCLR.objects.filter(pk=clr_pk, is_active=True)` ?

thelostone-mc: cause we'd want to have the ability to run clr on a round even if it's not active manually to calculate the curve after round ends (aka before payout)

chibie (approved):

PR #8791, opened 2021-04-16
feat: have ability to set latest flag to 0 in GrantCLRCalculation

thelostone-mc:
##### Description

Introduce 2 new actions on Grant CLR admin
  - `Reset CLR calculations for this round` -> set the `GrantCLRCalculations` for that round to `latest` = `False` 
  - `Reset all CLR calculations for all rounds` -> -> set ALL `GrantCLRCalulations` to `latest` = `False`  
  
###### Why do need this ?

-  Any time the estimate_clr runs for a round -> the earlier records for that CLR`GrantCLRCalculations`  has the `latest` column set to `False` and the once all the CLR are calculated , the effective CLR for each grant is nothing but the same of the latest `GrantCLRCalculations` with `latest=True` and the that grant.pk
- When the round runs
   - the older ones are invalidated (aka latest is set to False) 
   - new records are created
- When a round ends, the records with latest=True are created. Let's call these `Record X`.
- When a new round is started  -> the `record X` would be taken into account cause the older ones cause they were never invalidated 

This PR lets admin reset those for an individual round / bulk reset for all GrantCLRCalculations

##### How to use it

- Head over to GrantCLR object once it's ended and payouts are done 
- Scroll Down to bottom of the page 
- Click on `Reset CLR calculations for this round` OR `Reset all CLR calculations for all rounds`

##### Testing

https://share.vidyard.com/watch/j16VkR3kC5y9tKrjUsCRHb?

### DIFF
--- a/app/grants/admin.py
+++ b/app/grants/admin.py
@@ -479,7 +479,7 @@ class GrantCLRAdmin(admin.ModelAdmin):
                 recalc_clr.delay(grant.pk)
             self.message_user(request, "submitted recaclulation to queue")
 
-        if "_set_current_grant_clr_calulations_to_false" in request.POST:
+        if "_set_current_grant_clr_calculations_to_false" in request.POST:
             latest_calculations = GrantCLRCalculation.objects.filter(grantclr=obj, latest=True)
 
             if latest_calculations.count() == 0:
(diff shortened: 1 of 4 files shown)


octavioamu (approved):

chibie (changes requested):
lgtm!

typo: `_set_{current|all}_grant_clr_calulations_to_false` should be `_set_{current|all}_grant_clr_calculations_to_false`

calculations > calulations 😉

thelostone-mc:
Lol thanks @chibie

PR #8823, opened 2021-04-26
Fixes content cards so that they properly link to respective pages

Lerog1724:


gdixon:
This has already been merged

Issue #8741, opened 2021-04-05
We need better tooling to provision our production environment


gdixon:
### User Story

[comment]: # (As a <user type>, I want to <task> so that <goal>.)

As a devops engineer, I want an automated means of provisioning new instances and updating existing instances of our application/celery servers.

### Why Is this Needed

[comment]: # (Describe the problem and why this task is needed. Provide description of the current state, what you would like to happen, and what actually happen)

At the minute, if we want to provision a new instance we must manually install and configure each of the required packages, additionally, if code changes bring around [new requirements](https://github.com/gitcoinco/web/pull/7844#issuecomment-802912824), or if we need to make upgrades, we must roll out these changes manually.

We have documentation [here](https://docs.gitcoin.co/mk_alternative_setup/) which details our basic requirements of a host machine as well as this [Dockerfile](https://github.com/gitcoinco/web/blob/master/Dockerfile) which is used to stand-up our local environments.

It would be great if we could combine these two set-ups so that we only had to maintain our requirements in one place (a script), and use this solution to provision our application in any environment.

A possible solution might be to write an [Ansible playbook](https://docs.ansible.com/ansible/latest/user_guide/playbooks_intro.html), which could replace the majority of our Dockerfile, and allow us to provision our production environment in an automated fashion.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

### Current Behavior
[comment]: # (Describe what actually happened.)

Currently, we follow our documentation and stand-up each server manually.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

We would like to automate this process in the future, and the first step is to automate the provisioning stage.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

When we can stand up a new instance with a single command.

### Data Requirements

The solution should be a part of this repository, and should be appropriately commented/documented so that we are able to make upgrades as required.

### Additional Information
[comment]: # (Any other information that would be useful, content, screenshots, etc.)

We have included links to the following in the body of this issue:

- [Gitcoin alternative setup documentation](https://docs.gitcoin.co/mk_alternative_setup/)
- [Our Dockerfile](https://github.com/gitcoinco/web/blob/master/Dockerfile)
- [An example of our requirements changing](https://github.com/gitcoinco/web/pull/7844#issuecomment-802912824)
- [Ansible playbook documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_intro.html)

PR #7844, opened 2020-11-07
Feat: add duniter trust bonus 

developerfred:
##### Description

 [skunkworks/issues/187](https://github.com/gitcoinco/skunkworks/issues/187)

So it was the validation flow

1. The user must have the same userid on Duniter and Gitcoin
2. The user needs to be verified as a member
3. User needs to add their Gitcoin URL on Duniter

with these 3 validations OK my code returns the user as validated in Duniter

Env endpoints
``` 
BMAS_ENDPOINT = "BMAS g1.duniter.org 443"
ES_CORE_ENDPOINT = "ES_CORE_API g1.data.duniter.fr 443"
ES_USER_ENDPOINT = "ES_USER_API g1.data.duniter.fr 443"
```
When a Duniter user has 5 certificates he is automatically verified for 5 years in my implementation I check if he is a member and if the account has the same username as the gitcoin account

the Gitcoin user only needs to post using Duniter's public key to do the validations.

## Trust Duniter
[screen movie](https://youtu.be/CrKTciCpO9g)

Screen Shot 2020-11-13 at 08 44 27
Screen Shot 2020-11-13 at 08 44 38

**I don't have the checks on my duniter profile, but any gitcoin user who has these checks can validate his account.**
Screen Shot 2020-11-11 at 12 45 13
Screen Shot 2020-11-11 at 12 45 06

# User Verify on Duniter
Screen Shot 2020-11-11 at 12 44 48

PR #8825, opened 2021-04-26
[WIP] Adding a local vagrant+ansible setup option for testing deployments and package lists

zacheryschiller:
##### Description

Testing an alternative approach to running Gitcoin web locally by provisioning vms with vagrant to more closely match the production environment, test configuring with ansible, and find a common way to list and install packages across deployments types. 

Once testing is complete and configuration is working smoothly, the playbook calls in `Vagrantfile` can be updated to call the `config-<host>.yml` playbooks achieving the full install from the single command `vagrant up`.
/cc @gdixon @thelostone-mc

##### Refers/Fixes

Refs: #8741 

##### Testing

Requires additional testing to clear up package conflicts.

##### Notes and TODOs                                                             
* Many paths and variables are hardcoded for quick testing and should be pulled out and set separately
* There is a python package that is causing a dependency issue. This is [discussed in more detail here](https://github.com/gitcoinco/web/pull/7844#issuecomment-802912824). I am temporarily removing `duniterpy` but I'm not sure what package versions should be set to deconflict correctly.


octavioamu:
this look very interesting

thelostone-mc:
Closing this out but will keep branch alive if we need to revisit this

PR #8834, opened 2021-04-28
fix: use only json renderer in prod

chibie:
##### Description

drf browsable api should only be shown in local

##### Refers/Fixes

https://gitcoin.co/api/v0.1/


thelostone-mc:
Could you add a bit more on what this PR solves / does ?

PR #8835, opened 2021-04-28
makes the /api/v0.1/grants endpoint responsive again

owocki:
##### Description

makes the /api/v0.1/grants endpoint responsive again
adds field that roy stoof needs for grandois (grants L2)

##### Refers/Fixes

"Grandiose - A gitcoin grant discovery tool" email

##### Testing

tested locally

### DIFF
--- a/app/grants/router.py
+++ b/app/grants/router.py
@@ -75,11 +75,16 @@ class GrantViewSet(viewsets.ModelViewSet):
         if 'grant_type' in param_keys:
             queryset = queryset.filter(grant_type__name=self.request.query_params.get('grant_type'))
 
-        # adds a limit
-        if 'limit' in param_keys:
-            limit = int(self.request.query_params.get('limit'))
-            queryset = queryset[0:limit]
-
+        max_limit = 100
+        limit = str(self.request.GET.get('limit', max_limit))
+        offset = str(self.request.GET.get('offset', '0'))
+        if not limit.isnumeric() or int(limit) > max_limit:
+            limit = max_limit
+        if not offset.isnumeric() or int(offset) < 0:
+            offset = 0
+        offset = int(offset)
+        limit = int(limit)
+        queryset = queryset[offset:limit]
 
         return queryset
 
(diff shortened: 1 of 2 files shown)


Shammah:
tl;dr - Looks good to me

**These changes add the following four API properties:**
1. `amount_received_in_round` - The total amount received so far in the current round.
2. `positive_round_contributor_count` - The total number of contributors so far in the current round.
3. `clr_round_num` & `is_clr_active` - These allow filtering grants such that you only get the grants that participate in a certain funding round.

With these four properties I should be able to track the progress over time of each grant participating in the current active funding round. I'm assuming the numbers returned are just for the current active round and do not contain historical values of previous rounds. I personally don't need these for now.

**These changes also allow paging**
Together with `offset` & `limit` I should be able to page through all the grants, collecting the all the data of all grants I need.

`limit` is capped to `100` which should be fine.
`offset` is limited in the range of `[0, inf)`, which should be fine as well when used with Python slice notation.

thelostone-mc (approved):

thelostone-mc (approved):

Shammah:
@owocki  I've given the new features a test and it seems there's a bug in the offset implementation

If you go to the following URL you can see 5 grants being returned.
https://gitcoin.co/api/v0.1/grants/?grant_type=tech&offset=0&limit=5

If I wanted to just have the 3rd, I should be able to get just that with the following URL:
https://gitcoin.co/api/v0.1/grants/?grant_type=tech&offset=2&limit=1

However you get an empty result. To get our proper result, you have to enter the URL:
https://gitcoin.co/api/v0.1/grants/?grant_type=tech&offset=2&limit=3

This means 'limit' means the end of 'offset'. This wouldn't be a problem is limit wasn't capped to `100` 😉 
The fix is probably to change line 87 in `router.py` from:
`queryset = queryset[offset:limit]`
to
`queryset = queryset[offset:offset+limit]`

I don't have the code checkout out so I'm unable to quickly make a PR. Should be an easy fix I think?

owocki:
3030f823e6dfe0190a56280bc8ec08019e4be823 should handle it

PR #8467, opened 2021-03-01
Adds commit hash to the 'manifest.json' making manifest generation deploy specific

gdixon:
##### Description

Adds the HEAD commit hash to the manifest filename so that for each deployment we generate a new manifest.*.json file, allowing the old version to carry on serving files whilst the deployment finishes (avoiding OfflineGenerationError's).

Depends on updates to -> https://github.com/gdixon/django-compressor/commit/97cf285af96110da88810f5e19b3809ef748a322

##### Refers/Fixes

Refers: #8368

##### Testing

Tested locally (with production .env variables)

PR #8360, opened 2021-02-04
 GPG/redesign

gdixon:
##### Description

**EPIC** - new brand styles integration

This PR:

- Introduces django-libsass
- ~~Contains a test stylesheet _(./app/assets/v3/scss/main.scss)_ to check that libsass will work in our environment~~
- Replaces bootstrap.min.css with bootstrap.scss
- Replaces bootstrap-vue.min.css with bootstrap-vue/index.scss
- Replaces vue-select.css with vue-select.scss
- Adds minification rules for production env
- ~~Moves libs to v3/libs~~
- Updates bootstrap v4.3.1 -> v4.6.0
- Updates bootstrap.min.js v4.3.1 -> v4.6.0
- Updates popper.min.js  v1.16.0 -> v1.16.1
- Moves v3/scss assets into v2/scss - fixing problems with collectstatic
- Updates and internalises jQuery (v3.3.1 -> v3.5.1)
- Fixes google tagmanager on landing
- Implements new scss libs on landing
- Moves libs back in to "lib" dir _(./app/assets/v2/scss/lib/*)_

##### Refers/Fixes

Fixes: #8363
Refers: #8308
Refers: #8358

PR #8860, opened 2021-05-04
Feature: Webpack bundling of JS/SCSS

gdixon:
##### Description

This PR replaces django-compressor with an in-house alternative to avoid problems on production:

- All templates now use `bundle` as opposed to `compress` template tags
- All bundling/compressing works against local files (./assets) and doesn't make use of manifest.json
- Management command is initiated with `./manage.py bundle`
- Adds a second step to compress the bundles (`yarn run build`)
- GIves us access to babel and webpack (we can pull dependencies from NPM and avoid storing them in this repo)
- Includes all variables in all SCSS scripts without requiring us to manually import them
- Adds `merge_js` and `merge_css` options for when we want to avoid webpack - we can instead merge the files and compress using python
- Webpack is a module loader, it doesn't play nicely with globals, some JS files have been modified so that they assign to window/this explicitly
  - we can revert these changes if we have no intention of importing libs from NPM
  - at the minute, we are bundling using `{% bundle merge_js ... %}`,  but I have left the assignment changes in place so that we can easily switch back to `{% bundle js ... %}` (this would allow us to import from NPM and remove the libs from our repo)
  
- Build steps are integrated into `docker-command.bash` (for dev) and `deploy.bash` (for prod)

--

- Syntax:

        {% bundle [js|css|merge_js|merge_css] file [block_name] %}
        <script src="..."></script>
        <script>
            ...
        </script>
        --or--
        <link href="..."/>
        <style>
            ...
        </style>
        {% endbundle %}

    (dev) to compress:

        yarn run webpack

    (prod) to compress:

        ./manage.py bundle && yarn run build

--

- ##### Note: 
    For dev - you will need to rebuild your docker image (or manually install yarn on the web image)
    For prod - we will need to manually install yarn (on the cron box):
               
             
        curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
        echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
        apt-get update
        apt-get install -y yarn
        yarn global add n
        n stable
             

##### Refers/Fixes

Fixes: #8467
Fixes: 43d44398582819377ed87a3c966269d5470ca915
Fixes: #8360 

##### Testing

Tested locally with dev and prod envs

--

It would be a good idea to create a second PR with a reduced scope to test this on production before merging


octavioamu:
closed as #8865 was merged

PR #8883, opened 2021-05-11
chore: rewrite report endpoint

thelostone-mc:
### Description

This PR retries the report_real which takes a lot of resources and times out on heavy grants
Instead, Gitcoin will now offer 2 endpoints for public consumption 

**Note:** 
- The results are paginated and return upto 30  
- These endpoints are rate limited by IP to 5/s
- The API will return a valid response only for a max of 30 days 
- These endpoints aren't restricted to just eth address but instead provide info on all chains supported 
- A new metadata object has been added to the response which provides details like total count, current page, how many pages, next_page, etc
- The original API is being retired but the data structure stays the safe and the info which was provided by the older endpoint, continues to be provided in the 2 new endpoints 

---------

#### Get all grant contributions made by an address

**Endpoint**:  `api/v0.1/grants/contributions_sent_report/?address=0x997D35b300bA1775fdB175dF045252e57D6EA5B0&format=json`

**Parameters**: 

- `address` : `String` 
- `format`: `json`
- `page`: `number`
- `from_timestamp`: `YYYY-MM-DD`
- `to_timestamp`: `YYYY-MM-DD`

sample response:
```json
{
  "metadata": {
    "from": "2021-04-11T06:22:23.082709",
    "to": "2021-05-11T06:22:23.082709",
    "count": 2,
    "current_page": "1",
    "num_pages": 1,
    "has_next": false
  },
  "data": [
    {
      "grant_name": "Gitcoin Grants Round 8 + Dev Fund",
      "asset": "ETH",
      "timestamp": "2020-08-14T22:13:01.132221Z",
      "grant_amount": "1000000000000000",
      "gitcoin_maintenance_amount": "0",
      "grant_usd_value": 0.43796,
      "gitcoin_usd_value": 0.0
    },
    {
      "grant_name": "Gitcoin Grants Round 8 + Dev Fund",
      "asset": "ETH",
      "timestamp": "2020-08-14T22:13:01.392117Z",
      "grant_amount": "0",
      "gitcoin_maintenance_amount": "0",
      "grant_usd_value": 0.0,
      "gitcoin_usd_value": 0.0
    }
  ]
}
```

---------

#### Get all grant contributions received by an grant

**Endpoint**:  `api/v0.1/grants/contributions_rec_report/?id=23&format=json`

**Parameters**: 

- `id` : `number` (grant id) 
- `format`: `json`
- `page`: `number`
- `from_timestamp`: `YYYY-MM-DD`
- `to_timestamp`: `YYYY-MM-DD`

sample response: 

```json
{
  "transactions": [
    {
        "asset": "DAI",
        "timestamp": "2021-04-05T18:09:13.664911Z",
        "amount": "2958149082487300",
        "clr_round": null,
        "usd_value": 0.0029581490824873
    }
  ],
  "clr_payouts": [
    {
      "amount": 530,
      "asset": "DAI",
      "usd_value": 530,
      "timestamp": "2019-06-04T16:06:59.354432Z",
      "round": 2
    }
  ],
  "metadata": {
    "grant_name": "The Gitcoin Open Source Support Fund",
    "from_timestamp": "2021-04-11T06:17:15.472045",
    "to_timestamp": "2021-05-11T06:17:15.472045",
    "count": 1,
    "current_page": "1",
    "num_pages": 1,
    "has_next": false
  }
}
```

### DIFF
--- a/app/grants/router.py
+++ b/app/grants/router.py
@@ -3,6 +3,7 @@ from datetime import datetime, timedelta
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 
 import django_filters.rest_framework
+from ratelimit.decorators import ratelimit
 from rest_framework import routers, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound
@@ -94,6 +95,7 @@ class GrantViewSet(viewsets.ModelViewSet):
 
 
     @action(detail=False)
+    @ratelimit(key='ip', rate='5/s')
     def contributions_rec_report(self, request):
         """Genrate Grantee Report for an Grant"""
 
@@ -180,6 +182,7 @@ class GrantViewSet(viewsets.ModelViewSet):
 
 
     @action(detail=False)
+    @ratelimit(key='ip', rate='5/s')
     def contributions_sent_report(self, request):
         """Generate report for grant contributions made by an address"""
 
--- a/app/grants/router.py
+++ b/app/grants/router.py
@@ -97,7 +97,10 @@ class GrantViewSet(viewsets.ModelViewSet):
     @action(detail=False)
     @ratelimit(key='ip', rate='5/s')
     def contributions_rec_report(self, request):
-        """Genrate Grantee Report for an Grant"""
+        """
+            Genrate Grantee Report for an Grant
+            URL: api/v0.1/grants/contributions_rec_report/?id=<grant-id>&format=json
+        """
 
         results_limit = 30
 
@@ -136,9 +139,9 @@ class GrantViewSet(viewsets.ModelViewSet):
                     'error': 'to_timestamp is is not in the format YYYY-MM-DD'
                 })
 
-        if (to_timestamp - from_timestamp).days > 32:
+        if (to_timestamp - from_timestamp).days > 31:
             return Response({
-                'error': 'timeperiod should be less than 32 days'
+                'error': 'timeperiod should be less than 31 days'
             })
 
         try:
@@ -184,7 +187,10 @@ class GrantViewSet(viewsets.ModelViewSet):
     @action(detail=False)
     @ratelimit(key='ip', rate='5/s')
     def contributions_sent_report(self, request):
-        """Generate report for grant contributions made by an address"""
+        """
+            Generate report for grant contributions made by an address
+            URL: api/v0.1/grants/contributions_sent_report/?address=<address>&format=json
+        """
 
         donor_serializer = DonorSerializer
         results_limit = 30
@@ -225,9 +231,9 @@ class GrantViewSet(viewsets.ModelViewSet):
                     'error': 'to_timestamp is is not in the format YYYY-MM-DD'
                 })
 
-        if (to_timestamp - from_timestamp).days > 32:
+        if (to_timestamp - from_timestamp).days > 31:
             return Response({
-                'error': 'timeperiod should be less than 32 days'
+                'error': 'timeperiod should be less than 31 days'
             })
 
         # Filter Contributions made by given address
--- a/app/grants/router.py
+++ b/app/grants/router.py
@@ -1,4 +1,6 @@
-from datetime import datetime
+from datetime import datetime, timedelta
+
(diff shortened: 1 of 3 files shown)


app/grants/router.py
```diff
@@ -88,59 +93,161 @@ def get_queryset(self):
...
+        else:
             try:
                 to_timestamp = datetime.strptime(to_timestamp, format)
             except ValueError:
-                raise NotFound(detail="Please provide to_timestamp in the format: "+format)
+                return Response({
+                    'error': 'to_timestamp is is not in the format YYYY-MM-DD'
+                })
+
+        if (to_timestamp - from_timestamp).days > 32:
```
zacheryschiller, line 139: This wouldn't catch the value `== 32`. Should be `>= 32` or `> 31`.

app/grants/router.py
```diff
@@ -88,59 +93,161 @@ def get_queryset(self):
...
-        grants_queryset = grants_queryset.filter(subscriptions__subscription_contribution__created_on__lte=to_timestamp)
-        grants_queryset = grants_queryset.distinct()
+            try:
+                to_timestamp = datetime.strptime(to_timestamp, format)
+            except ValueError:
+                return Response({
+                    'error': 'to_timestamp is is not in the format YYYY-MM-DD'
+                })
+
+        if (to_timestamp - from_timestamp).days > 32:
```
zacheryschiller, line 228: Same as above. This wouldn't catch the value `== 32`. Should be `>= 32` or `> 31`.

app/grants/router.py
gdixon: Why are we using 32 days as the marker here? Wouldn't 31 (max days in a month) make more sense?

app/grants/serializers.py
```diff
@@ -96,28 +96,6 @@ class Meta:
 
         fields = ('amount', 'asset', 'usd_value', 'timestamp', 'round')
```
gdixon, line 97: Consider: Now that we're including all chains, would it make sense to include the chain that the asset is associated with in the output?

app/grants/router.py
```diff
@@ -88,59 +93,161 @@ def get_queryset(self):
...
-    @action(detail=False)
-    def report(self, request):
-        return Response({'error': 'reports temporarily offline'})
 
     @action(detail=False)
-    def report_real(self, request):
-        """Generate Grants report for an ethereum address"""
+    @ratelimit(key='ip', rate='5/s')
+    def contributions_rec_report(self, request):
+        """Genrate Grantee Report for an Grant"""
```
gdixon, line 100: Consider: Would it be worth including details of the expected request format here?

app/grants/router.py
```diff
@@ -88,59 +93,161 @@ def get_queryset(self):
...
+                'has_next': all_contributions.page(page).has_next()
+            }
+
+        })
+
+
+    @action(detail=False)
+    @ratelimit(key='ip', rate='5/s')
+    def contributions_sent_report(self, request):
+        """Generate report for grant contributions made by an address"""
```
gdixon, line 187: Consider: Same again here (detail what a request should look like)?

gdixon (approved):
LGTM! I've left a few comments, but nothing that is blocking!

app/grants/router.py
thelostone-mc: I can tweak it to `>= 31` but aren't they the same ?

app/grants/serializers.py
thelostone-mc: This is for pubic consumption! 
Do you see value adding it on? I figured we'd provide it if folks need it.
The data provided was what folks initially wanted

app/grants/router.py
thelostone-mc: I can throw it in a readme file ! Would you rather we add it here ?

gdixon: We can assume (or can we?) that users who will be accessing our end-points will likely check the code, I think that it would be useful to include these details inline, and later we can look to create documentation from the code itself? If theres a better place we can leave these details though, we can explore that, but I still feel inline would be a safe approach?

PR #8910, opened 2021-05-18
Loads in wallpapers from the assets directory and fixes bool check

gdixon:
##### Description

We were reading the static directory for wallpaper entries, but in production we upload our assets to S3 so this directory would have been empty/not present.

The bool check was also failing because `bool('0')` is True

##### Refers/Fixes

Fixes: #8908 

##### Testing

Tested locally


octavioamu:
we need to add ${media_url}  on the profile-edit.js 
also can we throw all the new pngs here https://tinypng.com/ ? Just to be sure we don't add more loading times to the profiles.

octavioamu:
looking deep with Graham the code we probably need a refactor of this as is not correctly coded to use S3
form prod :  
image

octavioamu (approved):

PR #8936, opened 2021-05-24
Remove staff

thelostone-mc:
##### Description

- remove decorator + staff
- uncomment banner


gdixon (approved):
LGTM!!! 🚀

zacheryschiller (commented):
We should remove the `staff_member_required` import in `app/quadraticlands/views.py` since it will now be unused.

PR #8999, opened 2021-06-06
Removing code where normalize_url() function is used when the input i…

0xrishabh:
Removing code where normalize_url() function is used when the input is already Normalized.

get_bounty_id_from_db function is getting normalized URL but was again normalizing it for no good reason.


thelostone-mc (approved):
Ah nice catch 🙌

Issue #8852, opened 2021-05-01
missing contributions in the gitcoin grants UI


frankchen07:
### Circumstance

when gitcoin grants r9 was in progress, contributors were seeing diffs between the grants UI and on-chain transactions, which prompted a stopfix solution (missing contributions link).

### Description

- [ ] what was the exact problem why contributions weren't showing up? (gas prices? zksync issue?)
- [ ] zksync failures 
- [ ] transaction always pending, 
- [ ] transaction dropped or replaced 

### Discussion

@frankchen07 @thelostone-mc @mds1 to sync up and clarify the problem

possible solution suggestion is to not show contributions as complete until they post on the etherscan and are successful

mds1:
For reference, the latest known details / potential causes / things to investigate are listed here: https://github.com/gitcoinco/web/issues/7744#issuecomment-754823192

thelostone-mc:
@mds1 I'm not clear on the root cause -> any thoughts on how we can go about 
- reproducing this
- how to fix this

PR #9009, opened 2021-06-08
remove manageEthereumCartJSONStore

thelostone-mc:
##### Description

removes the EthereumCartJSONStore as this is not being used right now and this could possibly be causing contributions to be missed 

##### Refers/Fixes

https://github.com/gitcoinco/web/pull/8563/files
fixes https://github.com/gitcoinco/web/issues/8852

##### Testing

**ZKSYNC**

Untitled

--------------
**BULK CHECKOUT**

Untitled

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -982,8 +982,8 @@ Vue.component('grants-cart', {
       const bulkTransaction = new web3.eth.Contract(bulkCheckoutAbi, bulkCheckoutAddress);
       const donationInputsFiltered = this.getDonationInputs();
 
-      // Save off cart data
-      await this.manageEthereumCartJSONStore(userAddress, 'save');
+      // // Save off cart data
+      // await this.manageEthereumCartJSONStore(userAddress, 'save');
 
       // Send transaction
       indicateMetamaskPopup();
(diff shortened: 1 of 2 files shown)


app/assets/v2/js/cart.js
```diff
@@ -1003,148 +1000,204 @@ Vue.component('grants-cart', {
...
-      const csrfmiddlewaretoken = document.querySelector('[name=csrfmiddlewaretoken]').value;
-
-      // If txHash has a length of one, stretch it so there's one hash for each donation
-      let txHashes = txHash;
-
-      if (txHash.length === 1) {
-        txHashes = new Array(donations.length).fill(txHash[0]);
-      }
+      try {
+        // this.grantsByTenant is the array used for donations
```
mds1, line 1006: Nit, just to clarify this comment
```suggestion
        // this.grantsByTenant is the array used for donations, and this.donationInputs is computed from it
```

app/assets/v2/js/cart.js
```diff
@@ -1003,148 +1000,204 @@ Vue.component('grants-cart', {
...
-        grant_id: [],
-        split_tx_id: [], // Bulk donation hash for L1, or specific hash for L2
-        sub_new_approve_tx_id: [],
-        token_address: [],
-        token_symbol: []
-      };
+        for (let i = 0; i < donations.length; i += 1) {
+          // Get URL to POST to
+          const donation = donations[i];
+          const grantId = donation.grant.grant_id;
```
mds1, line 1058:
```suggestion
          // Get donation information
          const donation = donations[i];
          const grantId = donation.grant.grant_id;
```

app/assets/v2/js/cart.js
```diff
@@ -1003,148 +1000,204 @@ Vue.component('grants-cart', {
...
+            {
+              'name': 'product',
+              'attributes': {
+                'product': 'grants',
+                'persona': 'grants-contributor',
+                'action': 'contribute'
+              }
+            }
+          ]
+        });
```
mds1, line 1125: Is this an analytics thing? It occurs after the `fetch` so I don't think it should be an issue, but just wanted to confirm there's no way this could cause issues similar to how saving JSON stores caused issues

mds1 (approved):
Looks good! There's some comments that could be improved to help readers understand what's going on, so I made some rephrasing suggestions (those were probably my fault!) 

But otherwise, marked as approved

app/assets/v2/js/cart.js
thelostone-mc: Yup this is an analytics thing ! @octavioamu would it be worth doing this after the POST call is complete
@mds1 has a valid point here

octavioamu: yes is a post to our internal endpoint and from there to mautic.

octavioamu (approved):
This look nice just not sure the relation with the issue linked.
Also are we not using that data for nothing? 
anyway this seem a great cleanup

gdixon (approved):

PR #9023, opened 2021-06-09
sentry: remove ddtrace (105k)

thelostone-mc:
##### Description

Could possible fix https://sentry.io/organizations/gitcoin/issues/1892482690/?project=1398424&query=is%3Aunresolved
which is now the highest error counts on sentry


octavioamu:
mm not sure if stop sending data to datadog will be an actual fix there are some docs here maybe we need to configure properly to go around the request limit https://ddtrace.readthedocs.io/en/v0.46.0/troubleshooting.html

PR #9041, opened 2021-06-11
feat(grants): email rollup for new contributions

### LINKED ISSUE #8917: email rollup for standard transactions on gitcoin grants
### Circumstance

Using bulk checkout allows for an email rollup, we should do the same for standard checkout so folks who donate to more than 40+ grants don't get bombarded with hundreds of grant emails.

### Description

- [x] implement email rollup according to cart size
- [x] how do we rollup for standard checkout (where I believe the problem exists)

for grant owners:
- [ ] have an email sent out every cron job interval
- [ ] on the email itself, there will be no actual contributor or individual contribution details, instead, we will add aggregate metrics and link them to their grant activity feed
- [ ] "in the last 6 hours your grant has earned approximately $$$ from XXX users,  please click here to see more"

chibie:
##### Description

 - [x] have an email sent out every cron job interval
 - [x] on the email itself, there will be no actual contributor or individual contribution details, instead, we will add aggregate metrics and link them to their grant activity feed
 - [x] "in the last XX hours your grant has earned approximately $$$ from XXX users, please click here to see more"

##### Refers/Fixes

#8917

##### Testing

https://www.loom.com/share/d1b1caabdc6c406d8642c2fbcc542f2c


app/app/urls.py
```diff
@@ -652,6 +652,7 @@
     path('_administration/email/new_grant', retail.emails.new_grant, name='admin_new_grant'),
     path('_administration/email/new_grant_approved', retail.emails.new_grant_approved, name='admin_new_grant_approved'),
     path('_administration/email/new_supporter', retail.emails.new_supporter, name='admin_new_supporter'),
+    path('_administration/email/new_contributions', retail.emails.new_contributions, name='admin_new_contributions'),
```
thelostone-mc, line 655: Isn't this supposed to replace `new_supporter` ? Can we reuse that or remove that ?

app/grants/tasks.py
```diff
@@ -390,3 +384,12 @@ def save_contribution(self, contrib_id):
...
     contrib.save()
+
+
+@app.shared_task(bind=True, max_retries=3)
+def process_new_contributions_email(self, grant_id):
+    try:
+        grant = Grant.objects.get(pk=grant_id)
+        new_contributions(grant)
+    except Exception as e:
+        print(e)
```
thelostone-mc, line 395: might be worth to log exeception so that we can catch it on sentry

app/marketing/mails.py
```diff
@@ -326,6 +326,26 @@ def new_supporter(grant, subscription):
         translation.activate(cur_language)
 
 
+def new_contributions(grant):
```
thelostone-mc, line 329: Let's remove the older one (new supporter)

thelostone-mc (changes requested):
Left a few comments but genrerally looking good

app/app/urls.py
thelostone-mc: It seems like you did this in https://github.com/gitcoinco/web/pull/9043 ?
I'm not sure though cause it's 2 diff PRs

thelostone-mc: Could we get it as 1 cause that way it's wrapped into 1 -> where we replace new_supporter with new_contributions ?

thelostone-mc:
Changes over at https://github.com/gitcoinco/web/pull/9064

PR #9042, opened 2021-06-11
add round active boolean to turn off the counter

octavioamu:
##### Description

- add "active" field on json data 
- add condition to disable the countdown display based on "active"
- set default dates to 14 days round

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -973,7 +973,7 @@ def grants_by_grant_type(request, grant_type):
         if _type.get("keyword") == grant_type:
             grant_label = _type.get("label")
 
-    clr_round, round_start_date, round_end_date, round_active = get_clr_rounds_metadata()
+    _, round_start_date, round_end_date, _ = get_clr_rounds_metadata()
 
     params = {
         'active': 'grants_explorer',
@@ -1150,7 +1150,7 @@ def grants_by_grant_clr(request, clr_round):
         if _type.get("keyword") == grant_type:
             grant_label = _type.get("label")
 
-    clr_round, round_start_date, round_end_date, round_active = get_clr_rounds_metadata()
+    clr_round, round_start_date, round_end_date, _ = get_clr_rounds_metadata()
 
     params = {
         'active': 'grants_landing',
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -102,6 +102,7 @@ def get_clr_rounds_metadata():
         clr_round = CLR_ROUND_DATA['round_num']
         start_date = CLR_ROUND_DATA['round_start']
         end_date = CLR_ROUND_DATA['round_end']
+        round_active = CLR_ROUND_DATA['active']
 
         # timezones are in UTC (format example: 2021-06-16:15.00.00)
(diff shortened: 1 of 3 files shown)


app/grants/views.py
```diff
@@ -1147,7 +1150,7 @@ def grants_by_grant_clr(request, clr_round):
         if _type.get("keyword") == grant_type:
             grant_label = _type.get("label")
 
-    clr_round, round_start_date, round_end_date = get_clr_rounds_metadata()
+    clr_round, round_start_date, round_end_date, round_active = get_clr_rounds_metadata()
```
thelostone-mc, line 1153: `clr_round, round_start_date, round_end_date, _ = get_clr_rounds_metadata()`

app/grants/views.py
```diff
@@ -970,7 +973,7 @@ def grants_by_grant_type(request, grant_type):
...
```
thelostone-mc, line 976: `clr_round, round_start_date, round_end_date, _ = get_clr_rounds_metadata()`

thelostone-mc:
Can we rename `active` to `show_banner` or a more meaningful name ?

app/grants/views.py
thelostone-mc: fixed this

app/grants/views.py
thelostone-mc: fixed this

octavioamu:
> Can we rename `active` to `show_banner` or a more meaningful name ?

I don't think so, this doesn't hide the banner just the counter, we don't want to hide the banner. 
Also the idea is just follow the state of the round /active or not  without worry where will be used.

app/grants/views.py
```diff
@@ -973,7 +973,7 @@ def grants_by_grant_type(request, grant_type):
         if _type.get("keyword") == grant_type:
             grant_label = _type.get("label")
 
-    clr_round, round_start_date, round_end_date, round_active = get_clr_rounds_metadata()
+    _, round_start_date, round_end_date, _ = get_clr_rounds_metadata()
```
octavioamu, line 976: this conflict with `from django.utils.translation import gettext_lazy as _` why need to be _ ?

gdixon: Yeah good point, we can import that as anything 👍 Theres 124 uses in `app/dashboard/views.py` - should we go for something like `__` - Or do we want to change it to something more meaningful?

octavioamu: My opinion is keep the name of what that function return, is more easy to understand and learn about the code.

app/grants/views.py
```diff
@@ -102,6 +102,7 @@ def get_clr_rounds_metadata():
         clr_round = CLR_ROUND_DATA['round_num']
         start_date = CLR_ROUND_DATA['round_start']
         end_date = CLR_ROUND_DATA['round_end']
+        show_bannner = CLR_ROUND_DATA['show_bannner']
```
gdixon, line 105: I think `clr_active`/`active` makes more sense - we can use this to toggle the round dates in the trust requirements banner too 👍

Issue #9074, opened 2021-06-13
bug: activity tab is not loading


gdixon:
**Describe the bug**

The activity tab is not being populated with activity data

**To Reproduce**

Visit staging and check a profile which should have activity (./thelostone-mc)

**Expected behavior**

Activity should be loaded

**Screenshots**

<img width="1206" alt="Screenshot 2021-06-13 at 22 34 29" src="https://user-images.githubusercontent.com/5897836/121822549-8fd28a80-cc97-11eb-9b64-5368ff2ce0d7.png">

**Desktop (please complete the following information):**

_N/A_

**Smartphone (please complete the following information):**

_N/A_

**Additional context**

This issue seems to be isolated to staging and local (the issue is present even on `stable` on my local (but everything is fine on production))

PR #9075, opened 2021-06-13
Ensures activity is supplied in the profile view

gdixon:
##### Description

This PR fixes missing profile data from the context in `app/dashboard/views.py` which causes the activity to never load

##### Refers/Fixes

Fixes: #9074

##### Testing

Tested locally

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -2770,10 +2770,10 @@ def get_profile_tab(request, profile, tab, prev_context):
 
                 return TemplateResponse(request, 'profiles/profile_activities.html', context, status=status)
 
-
-        all_activities = context.get('activities')
+        as_dict = profile.to_dict()
+        all_activities = as_dict.get('activities')
         tabs = []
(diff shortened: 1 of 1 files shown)


thelostone-mc (approved):

app/dashboard/views.py
```diff
@@ -2770,10 +2770,10 @@ def get_profile_tab(request, profile, tab, prev_context):
 
                 return TemplateResponse(request, 'profiles/profile_activities.html', context, status=status)
 
-
-        all_activities = context.get('activities')
+        as_dict = profile.to_dict()
```
owocki, line 2773: @gdixon heads up this method takes 3-5s to load in prod, which caused https://github.com/gitcoinco/web/issues/9119

owocki: use `profile.as_dict` which is already cached/pre-computed in the db

owocki: https://github.com/gitcoinco/web/pull/9124

Issue #9081, opened 2021-06-14
GR10 Trust Bonus and Min Contribution QA


PixelantDesign:
@gdixon @octavioamu the build is looking great! Noticed just a few things during QA, please see below

## **Trust Bonus**

- [x] **Profile Page**-  When user navigates to the profile page from Grants Home, there was a note in the design spec to jump the user down to the Trust bonus section. Is this possible?

- [ ] **Profile Page** - Did not get a chance to review Idena as Graham noted it's not working on staging. 

- [x] **Profile Page** - When disconnecting from a service, there should be a confirmation at the top of the page. Please see designs.

- [x] **Profile Page**  - On ENS modal, please update the word validate in the last line to verify.

- [x] **Profile Page** - Disconnect Modals - The header should reflect that he user is disconnecting, the build currently says ' Connect to <Service Name>. Please see designs https://www.figma.com/file/CP57XgGQwC9ChPCMkTuQ7i/GR10?node-id=851%3A12660

- [x] **Save to Collection**  - After user clicks save to a collection, we redirect the user to see the collection they've just created. The Grant confirmation banner doesn’t display long enough for the user to read the confirmation message. It flashes very quickly, can we lengthen time to 3 seconds and then fade?

- [ ] **Saving Collections** - Also when saving to existing collection, is it possible to speed up the refresh not show the no data state collections while the grant page is refreshing?  

## **Min Contribution**

- [x] Minimum contribution amount is not showing in the cart immediately, there is a short lag after the amount has been changed to less than 1 Dai

- [x] Message over checkout is tiny, please use design system small, 14px

Thanks guys! Lookin' great!

octavioamu:
>  Minimum contribution amount is not showing in the cart immediately, there is a short lag after the amount has been changed to less than 1 Dai

each time you put a value the system need to fetch the data to convert the value to USD, that was something already discussed that will happen.

octavioamu:
>  Message over checkout is tiny, please use design system small, 14px

we are using <small>  that need to be reviewed on the design system small == 80% of base font size. As that is a convention on the html tag and used on other part of the system already.

willsputra:
> we are using that need to be reviewed on the design system small == 80% of base font size. As that is a convention on the html tag and used on other part of the system already.
> 
> I will change it to use a 14px font

thanks! yea for now just making sure it's 14px and not too small would be great.

we can sync sometime to make sure the font styles we're using on the figma design system are the same as the tags on the live site.

[...]

PR #9098, opened 2021-06-15
Implements feedback from @PixelantDesign

gdixon:
##### Description

This PR:

- [x] **Profile Page**-  When user navigates to the profile page from Grants Home, there was a note in the design spec to jump the user down to the Trust bonus section. Is this possible?

- [ ] ~**Profile Page** - Did not get a chance to review Idena as Graham noted it's not working on staging.~

- [x] **Profile Page** - When disconnecting from a service, there should be a confirmation at the top of the page. Please see designs.

- [x] **Profile Page**  - On ENS modal, please update the word validate in the last line to verify.

- [x] **Profile Page** - Disconnect Modals - The header should reflect that he user is disconnecting, the build currently says ' Connect to <Service Name>. Please see designs https://www.figma.com/file/CP57XgGQwC9ChPCMkTuQ7i/GR10?node-id=851%3A12660

- [x] **Save to Collection**  - After user clicks save to a collection, we redirect the user to see the collection they've just created. The Grant confirmation banner doesn’t display long enough for the user to read the confirmation message. It flashes very quickly, can we lengthen time to 3 seconds and then fade?

- [ ] ~**Saving Collections** - Also when saving to existing collection, is it possible to speed up the refresh not show the no data state collections while the grant page is refreshing?~

    _Could we open a separate ticket for this @PixelantDesign  - this sounds more like a bug than a regression?_

##### Refers/Fixes

Fixes: #9081

##### Testing

Tested locally


app/assets/v2/js/pages/profile-trust.js
```diff
@@ -278,6 +278,7 @@ Vue.component('sms-verify-modal', {
       } finally {
         this.awaitingResponse = false;
         this.dismissVerification();
+        _alert("You have sucessfully disconnected your SMS verification.", "success", 3000);
```
zacheryschiller, line 281: Typo `successfully`

app/assets/v2/js/pages/profile-trust.js
```diff
@@ -494,6 +495,7 @@ Vue.component('twitter-verify-modal', {
       } finally {
         this.awaitingResponse = false;
         this.dismissVerification();
+        _alert("You have sucessfully disconnected from Twitter.", "success", 3000);
```
zacheryschiller, line 498: Typo `successfully`

gdixon: Resolved in c086cab

app/assets/v2/js/pages/profile-trust.js
gdixon: Resolved in c086cab

thelostone-mc (approved):

PR #9099, opened 2021-06-15
Adds some sanity to the kudos minter

owocki:
##### Description

Adds some sanity to the kudos minter by decoupling the 'sync' actions from the mint action.  The reason its worth doing this is that the site 504s sometimes bc syncing takes so long.  This allows us to mint them 1 by 1 then alllows a sync job thereafter.

##### Refers/Fixes

my own empirical experience.

##### Testing

 tested locally


gdixon (approved):
LGTM!

thelostone-mc (approved):

Issue #9119, opened 2021-06-16
profiles are slow to load


owocki:
profiles are slow to load

```
owocki@local /Users/owocki~ % time curl https://gitcoin.co/owocki > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  293k  100  293k    0     0  33962      0  0:00:08  0:00:08 --:--:-- 84538
curl https://gitcoin.co/owocki > /dev/null  0.03s user 0.02s system 0% cpu 8.872 total
owocki@local /Users/owocki~ % time curl https://gitcoin.co/vs77bb > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  295k  100  295k    0     0  64155      0  0:00:04  0:00:04 --:--:-- 64155
curl https://gitcoin.co/vs77bb > /dev/null  0.03s user 0.01s system 0% cpu 4.725 total
```

probably worth profiling these to see whats up + remediating

owocki:
ahhhh!!! much faster now

PR #9124, opened 2021-06-16
fix slow profiles by shaving off 5s of load time to gitcoin.co/owocki

owocki:
##### Description

fix slow profiles by shaving off 5s of load time to gitcoin.co/owocki

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/9119

##### Testing

tested locally

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -2770,7 +2770,7 @@ def get_profile_tab(request, profile, tab, prev_context):
 
                 return TemplateResponse(request, 'profiles/profile_activities.html', context, status=status)
 
-        as_dict = profile.to_dict()
+        as_dict = profile.as_dict
         all_activities = as_dict.get('activities')
         tabs = []
         counts = as_dict.get('activities_counts', {'joined': 1})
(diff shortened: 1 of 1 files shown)


app/dashboard/views.py
```diff
@@ -3936,7 +3936,7 @@ def profile(request, handle, tab=None):
         return redirect(profile.url)
 
     # setup context for visit
-    if not len(profile.tribe_members) and tab == 'tribe':
+    if not profile.tribe_members.exists() and tab == 'tribe':
```
owocki, line 3939: much more db efficient bc it just pulls a boolean back from the DB; whereas the prvious call pulled all the members back. shaves off 1s

app/dashboard/views.py
```diff
@@ -2770,7 +2770,7 @@ def get_profile_tab(request, profile, tab, prev_context):
 
                 return TemplateResponse(request, 'profiles/profile_activities.html', context, status=status)
 
-        as_dict = profile.to_dict()
+        as_dict = profile.as_dict
```
owocki, line 2773: https://github.com/gitcoinco/web/pull/9075/files#diff-9d08214144399c16f205c599ce310ef6a3edb83fe9b0c47ab7df5c6bbf6b0c3cL2774

instead of computing the dict on the fly, use the saved one in the db.  the saved one in the db is updated by the queues on the backend (not on load time)

cc @gdixon

gdixon (approved):

app/dashboard/views.py
gdixon: Awesome 🙌 Would it be worth making this same optimisation to kudos?

owocki: which line is that?

PR #9169, opened 2021-06-21
fix: Ensures the submenu els are present before attempting to display them

gdixon:
##### Description

This PR solves the navbar.js issue --- the mobile submenu implementation has not been rolled out over the products menu, so the errors are coming from users hitting one of the product links and navigating away, the href works, but the errors get thrown before the browser navigates.

This will place the checks in the right place to make sure the sub-submenu elements are present before attempting to place the menu in the correct position.

##### Refers/Fixes

Refers: #9144

Fixes: [Cannot read property 'classList' of null - Events: 1.1k ~ Users: 594](https://sentry.io/organizations/gitcoin/issues/2464485024/?project=1398424&query=is%3Aunresolved+assigned_to%3Ame&statsPeriod=14d)
Fixes: [computedRootStyles is not defined - Events: 742 ~ Users: 544](https://sentry.io/organizations/gitcoin/issues/2445050840/?project=1398424&query=is%3Aunresolved+assigned_to%3Ame&statsPeriod=14d)

##### Testing

Tested locally

### DIFF
--- a/app/assets/v2/js/navbar.js
+++ b/app/assets/v2/js/navbar.js
@@ -31,6 +31,12 @@ const makeMenu = (navbarEl) => {
     }
   };
 
+  // pull computedRootStyles from shared.js or compute here
+  const computedRootStyles = computedRootStyles || getComputedStyle(document.documentElement);
+
+  // pull breakpoint_md from shared.js or from root styles if not present
+  const breakpoint_md = breakpoint_md || parseFloat(computedRootStyles.getPropertyValue('--breakpoint-md'));
+
   // read the transition duration from navbar.scss (computedRootStyles is defined in shared.js)
   const transitionDuration = parseFloat(computedRootStyles.getPropertyValue('--gc-menu-transition-duration'));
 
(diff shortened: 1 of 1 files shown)


app/assets/v2/js/navbar.js
```diff
@@ -186,18 +192,13 @@ const makeMenu = (navbarEl) => {
   // remove isVisible transitions to reduce jank
   const resetVisibility = () => {
     // removing isVisible will reset transition to just opacity
-    if (backgroundEl)
-      backgroundEl.classList.remove('isVisible');
-    if (contentEl)
-      contentEl.classList.remove('isVisible');
-    if (caretEl)
-      caretEl.classList.remove('isVisible');
+    backgroundEl.classList.remove('isVisible');
```
zacheryschiller, line 195: @thelostone-mc added these `if`s a few days ago to fix some of the Sentry bugs. Do we need to remove them? Or does this branch just need to be updated with those changes?

gdixon: This negates the need to make those checks -- they never fully resolved the issue, that function shouldn't be getting called so we don't need to check for the els there.

thelostone-mc: Makes sense ! if we can prevent it from being called at the start, that means my changes would be redundant

thelostone-mc (approved):

app/assets/v2/js/navbar.js
zacheryschiller: Sounds good. Thanks for the extra info!

PR #9176, opened 2021-06-22
GITC-49: Adds notion integration to support sybil-hunting

gdixon:
##### Description

This PR integrates notion's api and posts newly created grants metadata (title/url) to the grants-review page.

Before we deploy these changes we have 2 new env's to add: `NOTION_API_KEY` & `NOTION_SYBIL_DB`

##### Refers/Fixes

Refers: GTC-49

##### Testing

Tested locally


zacheryschiller (approved):

app/grants/views.py
```diff
@@ -1864,6 +1864,15 @@ def grant_new(request):
 
         grant = Grant.objects.create(**grant_kwargs)
 
+        try:
+            # record to notion for sybil-hunters
+            save_grant_to_notion(title, grant.url)
+        except:
+            # delete the newly created grant and error if we can't save to notion
+            grant.delete()
```
octavioamu, line 1872: so if notion integration field fails this will delete the grant?? if that the case I don't feel super comfortable with the approach.

gdixon: I've altered the process to send an email on failures instead of blocking the save 👍 

Resolved in 351055f

octavioamu (approved):

app/grants/views.py
bambers936: 13135340

PR #9184, opened 2021-06-23
GITC-64: Purges claim and mission pages

gdixon:
##### Description

This PR:

- Purges the `claim` & `mission` pages

- Removes the _claim value_ from the QL `nav` & `QL-banner`

- Adds "& Workstreams" to the `QL-banner` copy

- Changes QL _Token Distribution_ copy from: 

    "There are no GTC associated with your account. If you think this is in error, please reach out on Discord. Otherwise join a workstream to participate in Gitcoin governance!"
    
    _To_ 
    
    "Token Distribution has now concluded, but we welcome you to join a workstream to help shape the future of digital public goods!"

##### Refers/Fixes

Fixes: GITC-64

##### Testing

Tested locally


thelostone-mc:
Any reason we are holding onto all the models ? Is that cause we need to preserve data/ just cause this a quick fix?

PixelantDesign (approved):

octavioamu (approved):

zacheryschiller:
> Any reason we are holding onto all the models ? Is that cause we need to preserve data/ just cause this a quick fix?

I agree that we should remove content we aren't using. If we need to pull it back we can checkout an old commit/tag. However, with the current timeline maybe we can do that in another pr? What do you think @thelostone-mc?

zacheryschiller (approved):

gdixon:
> Any reason we are holding onto all the models ? Is that cause we need to preserve data/ just cause this a quick fix?

We're still using the mission details on `./quadraticlands/dashboard` - so the main model we could drop is the dist table, but I would prefer it if we got a nod from product that we're safe to get rid of it? @frankchen07 @Kweiss

octavioamu:
I will say for now to don't remove models as this wasn't actually planned on the QL conception, just to be precautious if there is something we let go on our thinking.

PR #9199, opened 2021-06-24
GITC-73:  remove unused imports

chibie:
##### Description

Brief reason why we should remove unused imports

https://medium.datadriveninvestor.com/java-why-you-remove-unused-imports-7a3327a0b877

This was done using a combination of reports from Vulture and Pylance (VSCode extension)

This is the vulture report → https://rentry.co/uumip

#### some extra care taken

1. hacks for circular imports aren’t regarded as unused
2. modular imports in apps.py or other places that aren’t auto imported aren’t regarded as unused
3. i discovered other shit codes but intentionally limited this pr to the specific purpose

example use of pylance -> https://www.loom.com/share/35bc02a1c6b04e249cb17fc3517551d1


app/dashboard/tests/test_binance_sync.py
```diff
@@ -67,7 +64,7 @@ def setUp(self):
         )
         BountyFulfillment.objects.create(
             fulfiller_address='0x0000000000000000000000000000000000000000',
-            bounty=bounty,
+            bounty=self.bounty,
```
thelostone-mc, line 67: was this intentional ?

app/marketing/apps.py
```diff
@@ -26,4 +26,4 @@ class MarketingConfig(AppConfig):
     name = 'marketing'
 
     def ready(self):
-        from .signals import create_email_subscriber
+        import .signals # noqa
```
thelostone-mc, line 29: was this intentional ?

app/kudos/test_models.py
```diff
@@ -1,22 +0,0 @@
-# -*- coding: utf-8 -*-
-"""Test the Kudos models.
```
zls2, line 2: @chibie Good catch that this test isn't implemented. Should we make a ticket to implement this test?

zacheryschiller, line 2: @chibie Good catch that this test isn't implemented. Should we make a ticket to implement this test?

app/dashboard/tests/test_binance_sync.py
chibie: yes

app/kudos/test_models.py
chibie: @zacheryschiller i don't think it's worth the effort to test just models... more focus should be on testing request/response cycle and helper functions

thelostone-mc (approved):

zacheryschiller (approved):

PR #9204, opened 2021-06-25
GTC-84: could not convert string to float: '1\xa0000'

thelostone-mc:
##### Description

https://sentry.io/organizations/gitcoin/issues/1098741849/?project=1398424&query=is%3Aunresolved
Ensures POST call fails gracefully when values are not valid

##### Testing

Tested locally

https://gitcoin.atlassian.net/browse/GITC-84

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -2524,6 +2524,11 @@ def profile_job_opportunity(request, handle):
         profile.save()
     except (ProfileNotFoundException, ProfileHiddenException):
         raise Http404
+    except ValueError:
+        return JsonResponse(
+            {'error': 'Bad request'},
+            status=401
+        )
 
     response = {
         'status': 200,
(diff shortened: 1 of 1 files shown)


app/dashboard/views.py
```diff
@@ -2524,6 +2524,11 @@ def profile_job_opportunity(request, handle):
         profile.save()
     except (ProfileNotFoundException, ProfileHiddenException):
         raise Http404
+    except ValueError:
+        return JsonResponse(
+            {'error': 'Bad request'},
+            status=401
```
chibie, line 2530: @thelostone-mc bad request should have status=400

chibie (changes requested):

app/dashboard/views.py
thelostone-mc: udpated

PR #9207, opened 2021-06-25
GITC-80: remove unused templates

chibie:
##### Description

Remove unused templates based on output from https://github.com/AMeng/django-unused-templates/blob/master/django_unused_templates.py

#### some extra care taken

1. dynamically rendered templates were exempted, example  

```return TemplateResponse(request, f'quadraticlands/mission/{mission_name}/question_{question_num}.html', context)```


zacheryschiller:
Looks good cleaning up all this! I didn't see anything under "Testing" above. Did you test all the products locally to make sure we aren't accidentally removing something we need?

Lerog1724 (approved):
Thank you..

thelostone-mc (approved):
manually checked if files were used in codebase. couldn't find reference. LGTM

thelostone-mc:
> Looks good cleaning up all this! I didn't see anything under "Testing" above. Did you test all the products locally to make sure we aren't accidentally removing something we need?


@zacheryschiller I don't think you'd actually be able to test this. Doing a manual check while reviewing is probably the safest way to review this PT

PR #9237, opened 2021-06-30
hotfix - twilio ddos fix

owocki:
##### Description
some butthead is going in and ddosing our twilio account with twilio verify requests, which cost like 1c each... and in aggregate costing us hundreds of $$$$ per round.  

this PR adds a rate limit to the twilio endpoint call so that it cant be done anymore.

ideally we'd extend this even further and block ppl who abuse the system as such.. but i dont have cycles to code that today

##### Refers/Fixes

https://app.datadoghq.com/logs?query=request_user_sms&index=
https://www.twilio.com/console/verify/services

##### Testing

did not test, but its a simple addition of a rate limit so hope its ok

### DIFF
--- a/app/assets/v2/js/navbar.js
+++ b/app/assets/v2/js/navbar.js
@@ -32,10 +32,10 @@ const makeMenu = (navbarEl) => {
   };
 
   // pull computedRootStyles from shared.js or compute here
-  const computedRootStyles = (window.hasOwnProperty("computedRootStyles") ? window.computedRootStyles : getComputedStyle(document.documentElement));
+  const computedRootStyles = (Object.hasOwnProperty.call(window, 'computedRootStyles') ? window.computedRootStyles : getComputedStyle(document.documentElement));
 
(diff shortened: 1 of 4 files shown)


owocki:
note engineering team: - i did not get a chance to test as im on a laptoip without docker privisioned and in between calls. can you quickly test before shipping?

gdixon (approved):
LGTM!! 

Screenshot 2021-06-30 at 18 11 08

Do we want to add any error messages to the frontend?

owocki:
a graceful  failure eg "Youre ratelimited -Please try again in 1 minute" would prob be nice

zacheryschiller (approved):

gdixon:
It looks like the reason we were getting so many requests to the whitepaper was because it was set as the response to ratelimited requests - I've changed that and added the error response to the front-end for sms-verification 👍

octavioamu (approved):

PR #9260, opened 2021-07-05
GITC-165: Increases celery task_time_limit to account for longer running tasks

gdixon:
##### Description

This PR will double the time given to a celery task before it will fail (if we still get errors we will need to profile the longest running tasks on production/staging and set these values accordingly)

##### Refers/Fixes

Fixes: GITC-165

##### Testing

Tested configuration locally

### DIFF
--- a/app/taskapp/celery.py
+++ b/app/taskapp/celery.py
@@ -27,3 +27,6 @@ class CeleryConfig(AppConfig):
         app.autodiscover_tasks(lambda: installed_apps, force=True)
         app.conf.task_routes = {ele[0]: ele[1] for ele in settings.CELERY_ROUTES}
         app.conf.task_default_queue = 'default'
+        # Fix Hard time limit (300.0s) exceeded for task
+        app.conf.task_time_limit = 50 * 60 # 50mins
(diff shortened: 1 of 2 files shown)


app/taskapp/celery.py
```diff
@@ -27,3 +27,6 @@ def ready(self):
         app.autodiscover_tasks(lambda: installed_apps, force=True)
         app.conf.task_routes = {ele[0]: ele[1] for ele in settings.CELERY_ROUTES}
         app.conf.task_default_queue = 'default'
+        # Fix Hard time limit (300.0s) exceeded for task
+        app.conf.task_time_limit = 50 * 60 # 50mins
+        app.conf.task_soft_time_limit = 10 * 60 # 10mins
```
thelostone-mc, line 32: question -> do we wanna increase it universally or just for the clr ones ?

gdixon: Yeah it looks like this is only hitting the limit for new_bounty_daily - I've moved the settings there 👍

thelostone-mc (approved):

PR #9261, opened 2021-07-05
Fix AnonymousUser post on status

octavioamu:
##### Description

- Fix messages feedback
- Fix errors
- Fix blocking POST action on JS
- Fix validation on server to not raise 500. 

##### Refers/Fixes

fix https://gitcoin.atlassian.net/browse/GITC-166

### DIFF
--- a/app/retail/views.py
+++ b/app/retail/views.py
@@ -923,7 +923,12 @@ def create_status_update(request):
     issue_re = re.compile(r'^(?:https?://)?(?:github\.com)/(?:[\w,\-,\_]+)/(?:[\w,\-,\_]+)/issues/(?:[\d]+)')
     response = {}
 
-    if request.POST:
+    if request.POST and request.user.is_authenticated:
+        if (request.user.profile.is_blocked or request.user.profile.shadowbanned):
+            response['status'] = 400
+            response['message'] = 'Status updated!'
+            return JsonResponse(response, status=400)
+
         profile = request.user.profile
         title = request.POST.get('data')
         resource = request.POST.get('resource', '')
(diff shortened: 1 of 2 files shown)


app/retail/views.py
```diff
@@ -1030,6 +1029,12 @@ def create_status_update(request):
             response['message'] = 'Bad Request'
             logger.error('Status Update error - Error: (%s) - Handle: (%s)', e, profile.handle if profile else '')
             return JsonResponse(response, status=400)
+    else:
+        response['status'] = 401
+        response['message'] = 'Not logged in!'
+        print('here', response)
```
chibie, line 1035: we can take out the print statement

chibie (changes requested):

app/retail/views.py
```diff
@@ -1030,6 +1029,12 @@ def create_status_update(request):
...
+        return JsonResponse(status=response['status'], data={'status': 401,'message':response['message']})
```
thelostone-mc, line 1036: can't we just sent return JsonResponse(response, status=401) ?

octavioamu: how it works is status 401 http status data.status json pattern

app/retail/views.py
zacheryschiller: @octavioamu can you remove this? Other than that I think this is good to go.

zacheryschiller (changes requested):
Remove print statement :-)

app/retail/views.py
thelostone-mc: Yes that cleaner

chibie (approved):

gdixon (approved):

PR #9281, opened 2021-07-11
/administration/mesh performance fix, so i can ship GR10 data viz

owocki:
##### Description

https://gitcoin.co/_administration/mesh wasnt loading the most recent grants rounds data anymore, because its just TOO much data (400k rows).  so i wrote a trimmer that thins out the data so the network can still be representatively viewed without the system suffering in-browser-memory issues or 504 gateway timeouts on the server side.

this PR trims nodes from the mesh network for browser perf reasons, so we can view GR10 mesh data and update this thread https://twitter.com/owocki/status/1375278049420144645

##### Refers/Fixes

https://gitcoin.co/_administration/mesh if you try to view most recent grants round

##### Testing

tested locally


thelostone-mc (approved):
LGTM

Issue #9282, opened 2021-07-11
When I return to http://gitcoin.co/ after being away for an hour or so, the site is slow to load.


owocki:
When I return to http://gitcoin.co/ after being away for an hour or so, the site is slow to load.

Repro steps:
0. login.  
1. avoid site for a few hours
2. go back
3. first pageload back takes 10s

Ref https://gitcoincore.slack.com/archives/CDXGQRZB3/p1625157745005400

PR #9295, opened 2021-07-15
GITC-210: Record visit/join actions inside a celery task instead of context.py

gdixon:
##### Description

This PR moves `record_join` and `record_visit` logic into celery tasks to avoid saving data on page load

##### Refers/Fixes

Fixes: #9282 
Fixes: GITC-210

##### Testing

Tested locally

### DIFF
--- a/app/dashboard/tasks.py
+++ b/app/dashboard/tasks.py
@@ -357,10 +357,10 @@ def record_visit(self, user_pk, profile_pk, ip_address, visitorId, useragent, re
             profile.save()
         except Exception as e:
             logger.exception(e)
-        try:
-            profile_dict.delay(profile.pk)
-        except Exception as e:
-            logger.exception(e)
+
+        # enqueue profile_dict recalc
+        profile_dict.delay(profile.pk)
+
         try:
             metadata = {
                 'visitorId': visitorId,
--- a/app/dashboard/tasks.py
+++ b/app/dashboard/tasks.py
@@ -8,11 +8,13 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.http import HttpRequest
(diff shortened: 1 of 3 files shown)


app/dashboard/tasks.py
```diff
@@ -330,3 +332,69 @@ def recalculate_earning(self, pk, retry: bool = True) -> None:
...
+    :param visitorId: request.COOKIES.get("visitorId", None)
+    :param useragent: request.META['HTTP_USER_AGENT']
+    :param referrer: request.META.get('HTTP_REFERER', None)
+    :param path: request.META.get('PATH_INFO', None)
+    :param session_key: request.session._session_key
+    :param utm: _get_utm_from_cookie(request)
+    :return: None
+    """
+
+    from django.contrib.auth.models import User
```
thelostone-mc, line 353: shall we move this to the top ?

app/dashboard/tasks.py
```diff
@@ -330,3 +332,69 @@ def recalculate_earning(self, pk, retry: bool = True) -> None:
...
+        try:
+            profile.last_visit = timezone.now()
+            profile.save()
+        except Exception as e:
+            logger.exception(e)
+        try:
+            profile_dict.delay(profile.pk)
+        except Exception as e:
+            logger.exception(e)
+        try:
```
thelostone-mc, line 366: Se can remove L363-L366

thelostone-mc (approved):

app/dashboard/tasks.py
gdixon: We can likely remove the `try...except` from `profile_dict.delay(profile.pk)` - this was copied over from `context.py` as is, but lets keep the recalc here so that it's triggered regularly 👍

chibie (approved):

PR #9312, opened 2021-07-22
GITC-223: ignore squelched profiles in clr

thelostone-mc:
##### Description

We realized in Gr10 that the team was using shadow ban users to ignore contributions to be considered for matching.
Upon digging into it -> it came to light that our CLR calculations doesn't really ignore banned contributions. We would need to fix this

This PR tweaks that to ensure squelch profiles are taken into account while fetching the data

##### Refers/Fixes

GITC-223

##### Testing

Screenshot 2021-07-22 at 9 22 23 AM


gdixon (approved):

zacheryschiller (approved):
👍

PR #9333, opened 2021-08-04
GITC-276 purge: gitcoin faucet

thelostone-mc:
##### Description

- retires the faucet app
- removes cron + dead code + emails related to faucet 

##### Refers/Fixes

GITC-276

##### Testing

Locally tested


app/dashboard/templates/shared/zero_balance_error.html
```diff
@@ -22,13 +22,10 @@ <h3>{% trans "You're almost there!" %}</h3>
       <ol class="pl-3 mb-0 font-smaller-1">
         <li>{% trans "In order to continue, you'll need some Ether." %}</li>
         <li>{% trans "'Ether' is a digital asset that can be used to pay for the computational resources needed to run an application or program. In this case, Gitcoin." %}</li>
-        <li>{% trans "The easiest way to get Ether is to request it via the Gitcoin Faucet:" %}</li>
```
chibie, line 25: wouldn't it be nice to point to a trusted faucet website? e.g https://faucet.metamask.io/

app/retail/templates/gas_faq.html
```diff
@@ -12,8 +12,6 @@
 
 
 <h1>{% trans "FAQ" %}</h1>
-<a href="{% url 'gas_faucet_list' %}" target="_blank" rel="noopener noreferrer">{% trans "I need gas money, where can I get it?" %}</a>
```
chibie, line 15: do we need to take this out? it's a valid faq... maybe point to https://faucet.metamask.io/

octavioamu (approved):
look good and scary

chibie (approved):

app/marketing/mails.py
```diff
@@ -866,7 +844,7 @@ def notify_deadbeat_quest(quest):
         setup_lang(to_email)
         subject = _("Dead Quest Alert")
         body = f"This quest is dead ({quest.title}): https://gitcoin.co/{quest.admin_url} "
-        if not should_suppress_notification_email(to_email, 'faucet'):
```
zacheryschiller, line 869: Should we just remove these entirely rather than changing to the quest or kudos emails?

thelostone-mc: hmm but this would mean that the quest email never gets fired ! is that what we want ?
It was mapped to the wrong setting from what i can tell

zacheryschiller (approved):

PR #9341, opened 2021-08-05
Extract grants models into individual files

Jer-Sch:
##### Description

In order to unit test each model in `app/grants/models.py`, it will be helpful to have each model in its own file rather than one big file. This will make the code more readable, and reduce the amount of coupling and potential breakable changes that might occur.

Once this work is complete, unit tests can be added to each model in smaller PRs.
##### Testing

This work is preparatory to writing unit tests for the grants models


app/grants/models/grant.py
```diff
@@ -16,13 +16,141 @@
...
-from .grant_clr import GrantCLR
+from .grant_clr_calculation import GrantCLRCalculation
+from .grant_collection import GrantCollection
 
 from grants.utils import get_upload_filename, is_grant_team_member
 
 from townsquare.models import Favorite
 from web3 import Web3
 
+class GrantCLR(SuperModel):
```
thelostone-mc, line 27: Why are we not keeping this in it's own file?

Jer-Sch: @thelostone-mc Grant and GrantCLR are coupled and having GrantCLR in its own file was causing a circular dependency. Grant directly references GrantCLR in its `calc_clr_round` method (line 509), and vice versa GrantCLR references Grant in its `grants` method (line 128). I felt like keeping GrantCLR with Grant was the quickest way to move forward, but I'm happy to work on separating them.

octavioamu (approved):

app/grants/models/grant.py
thelostone-mc: Ah gotcha !

chibie (approved):
lgtm!

app/grants/migrations/0017_matchpledge_end_date.py
```diff
@@ -14,6 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='matchpledge',
             name='end_date',
-            field=models.DateTimeField(default=grants.models.next_month),
+            field=models.DateTimeField(default=grants.models.match_pledge.next_month),
```
thelostone-mc, line 17: Hmm I'd say let's not touch older migration files cause it's already been run on production

I'm good adding it as a try-catch

zacheryschiller (approved):

gdixon (approved):

gdixon:
Lets hold off getting this merged until all the GR11 changes are in 🙌 

Nice work here @Jer-Sch!!

PR #9354, opened 2021-08-16
chore: set max-width to 100 on videos in mission

thelostone-mc:
##### Description

 set max-width to 100 on videos in mission
 
##### Refers/Fixes

GITC-180

##### Testing

Untitled


octavioamu (changes requested):
I believe will be better to use responsive videos https://getbootstrap.com/docs/4.6/utilities/embed/

octavioamu (approved):

zacheryschiller (approved):

chibie (approved):

PR #9356, opened 2021-08-16
Update docker files and scripts to set up cypress

ksolo:
##### Description

Updating docker configurations to support cypress testing.

dockerfile:
* add dependencies for cypress
* add dependencies for chrome

docker-compose:
* add ipc to share memory with host
* add shm_size increase /dev/shm size
* update accounts in ganache-cli to 10

docker-command.bash
* add cypress install command after yarn install

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,6 +45,8 @@ services:
 
   web:
     restart: unless-stopped
+    ipc: host
+    shm_size: '4gb'
     environment:
       - PYTHONUNBUFFERED=1
       - PYTHONIOENCODING=utf8
@@ -63,6 +65,7 @@ services:
       - "3030:3030"
       - "8001:8001"
       - "35729:35729"
+      - "9222:9222"
     depends_on:
       - db
     stdin_open: true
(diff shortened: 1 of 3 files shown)


zacheryschiller (approved):

chibie (approved):

docker-compose.yml
```diff
@@ -45,6 +45,8 @@ services:
 
   web:
     restart: unless-stopped
+    ipc: host
+    shm_size: '4gb'
```
thelostone-mc, line 49: Is this needed ?

ksolo: It took quite a few resources for this to run in a container. I can attempt to run them without this.

docker-compose.yml
```diff
@@ -79,7 +82,7 @@ services:
     image: trufflesuite/ganache-cli
     ports:
       - "8545:8545"
-    command: '-a 1 --seed 2 --host 0.0.0.0 --mnemonic "${TEST_MNEMONIC}"'
+    command: '-a 10 --host 0.0.0.0 --mnemonic "${TEST_MNEMONIC}"'
```
gdixon, line 85: Do we need 10 accounts? If so can we leave out `-a 10` - this is the default 🙌

PR #9368, opened 2021-08-17
fixes bounty fulfillment data (usd amounts)

owocki:
##### Description

i recently discovered that the usd amounts for the earnings table for all bountyfulfilments data.

upon further investigation, this was because the usd conversion function on bountyfulfillment assumes that `self.payment_amount` was an integer which needed to be divided by `10**decimals` , when in fact it is not.  it is a floating point number that represents the amount of tokens transfered

                # https://gitcoin.co/_administrationdashboard/earning/1108896/change
                # https://gitcoin.co/_administrationdashboard/bounty/26254/change/
                # https://gitcoin.co/_administrationdashboard/earning/1107885/change
                # https://gitcoin.co/_administrationdashboard/bountyfulfillment/982/

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CBWA6A75M/p1629222791016400

this test data

##### Testing

tested locally

##### Deploy notes

we will need to run save() on all bountyfulfillments when this is done

```
from dashboard.models import *
for bf in BountyFulfillment.objects.all():
    bf.save()
```

### DIFF
--- a/app/dashboard/models.py
+++ b/app/dashboard/models.py
@@ -1510,16 +1510,16 @@ class BountyFulfillment(SuperModel):
     def value_in_usdt_at_time(self, at_time):
         try:
             if self.token_name in ['USDT', 'USDC']:
-                return float(self.payout_amount / 10 ** 6)
+                return float(self.payout_amount)
             if self.token_name in settings.STABLE_COINS:
-                return float(self.payout_amount / 10 ** 18)
+                return float(self.payout_amount)
             if self.token_name in ['ETH']:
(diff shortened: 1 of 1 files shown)


thelostone-mc (approved):

zacheryschiller (approved):

octavioamu:
good catch Owocki, not sure how many fulfilments we have in prod will be safe to run the save or should be on a celery to be safe? I believe should be ok like this but just double checking as we had some problems with big data changes on prod.

oneemo95 (commented):
Duplicate of #

PR #9374, opened 2021-08-19
Add test env to bundle

zacheryschiller:
##### Description

- Modifies bundling so that we can push to S3 from the testing environment as well as from prod
- Resolves two pip requirements conflicts

##### Refers/Fixes

References GITC-230

##### Testing
Testing on the testing environment


app/dashboard/templatetags/bundle.py
```diff
@@ -113,8 +113,8 @@ def check_for_changes(elems, attr, kind, outputFile):
             if el.get(attr):
                 # check if we're loading an alternative source in production
                 file = el[attr]
-                if isProduction and el.get('prod'):
-                    file = el['prod']
+                if assetsHosted and el.get(settings.ENV):
+                    file = el[settings.ENV]
```
gdixon, line 117: This is a prop on the import to allow us to select a different asset in `prod` vs `dev` - check out how its being used here: https://github.com/gitcoinco/web/blob/e0fdf7704ba8deb8ee5fba9ca4161540b3d69521/app/retail/templates/shared/footer_scripts.html#L34

app/dashboard/templatetags/bundle.py
```diff
@@ -193,10 +193,10 @@ def get_bundled(elems, attr, kind, merge):
         content = sass.compile(string='%s \n %s' % (get_sass_extras(), content))
 
     # minify the content in production
-    if isProduction and 'js' in kind:
+    if assetsHosted and 'js' in kind:
         import rjsmin
         content = rjsmin.jsmin(content)
-    elif isProduction and 'css' in kind:
+    elif assetsHosted and 'css' in kind:
```
thelostone-mc, line 199: ^ wrap the common check in a parent `if`

chibie (approved):

thelostone-mc (approved):

app/dashboard/templatetags/bundle.py
```diff
@@ -38,8 +38,8 @@
 """
 
 
-# check for production env
-isProduction = settings.ENV in ['prod']
+# check if assets are hosted in the env
+assetsHosted = settings.ENV in ['prod', 'test']
```
gdixon, line 42: Can we alter this to `isDevelopment` and negate the `in` check - I feel like `assetsHosted` will have the reader asking "where are the assets hosted?"

gdixon (approved):
LGTM! 🙌 Nice work Zack!

PR #9391, opened 2021-08-26
0/Test Donation model

Jer-Sch:
##### Description

This PR adds unit tests to the Donation model


thelostone-mc (approved):

app/grants/migrations/0017_matchpledge_end_date.py
```diff
@@ -9,11 +9,19 @@ class Migration(migrations.Migration):
...
+            ),
+        ]
+    except AttributeError:
+        operations = [
+            migrations.AddField(
+                model_name='matchpledge',
+                name='end_date',
+                field=models.DateTimeField(default=grants.models.match_pledge.next_month),
+            ),
+        ]
```
thelostone-mc, line 27: How does this help? 
t seems like we are adding the same operation again

thelostone-mc (changes requested):
left a comment

app/grants/migrations/0017_matchpledge_end_date.py
Jer-Sch: @thelostone-mc You're right, this doesn't help the Donation tests at all 🙂 This is from #9483 - the integration branch for this PR that this was branched off of which is still waiting to be merged.

The issue this resolves is that with the grants models in their own files rather than one file, `next_month` needs to be accessed by calling `grants.models.match_pledge.next_month` rather than `grants.models.next_month`. I didn't want to directly edit a migration that was already run in production, so I added the change as a `try/except`.

ksolo (approved):

thelostone-mc (approved):

PR #9404, opened 2021-08-27
Remove PhantomFunding

Jer-Sch:
##### Description

This PR removes all references and code for the `PhantomFunding` model which is deprecated. It also includes a database migration to delete the model.

##### Testing

Tests for `PhantomFunding` did not exist when this work was done.


app/grants/models/grant.py
```diff
@@ -488,7 +488,7 @@ class Meta:
     )
 
     in_active_clrs = models.ManyToManyField(
-        GrantCLR,
+        "GrantCLR",
```
thelostone-mc, line 491: Should this be a string?

thelostone-mc (commented):
Left one comment but otherwise LGTM

app/grants/models/grant.py
Jer-Sch: @thelostone-mc In this case, yes. I've been doing that to prevent circular imports.

thelostone-mc (approved):

chibie (approved):

PR #9425, opened 2021-09-03
nav_auth: remove cursor-pointer

meetmangukiya:
##### Description

Cursor pointer for items where clicking has no action is really confusing. This is really more of a hover action. Not sure if this is welcomed here. But just in case. Feel free to close if  it doesn't seem fit.


thelostone-mc:
This was intentionally designed like this @meetmangukiya ! Closing this out for now 
Thanks for opening a PR though.
Would love it if you identify such things and raise an issue and have a PR against it <3

PR #9444, opened 2021-09-07
GITC-237: Alters matching pool contribution to be an "on-top" donation

gdixon:
##### Description

This PR implements the interim checkout experience. The matching pool contribution has been altered to be an "on-top" donation.

##### Refers/Fixes

Fixes: GITC-237

##### Testing

Tested locally


app/grants/models.py
```diff
@@ -1935,7 +1935,7 @@ def psave_contrib(sender, instance, **kwargs):
                     "from_profile":instance.subscription.contributor_profile,
                     "org_profile":instance.subscription.grant.org_profile,
                     "to_profile":instance.subscription.grant.admin_profile,
-                    "value_usd":instance.subscription.amount_per_period_usdt if instance.subscription.amount_per_period_usdt else instance.subscription.get_converted_amount(False),
+                    "value_usd":instance.subscription.amount_per_period_usdt if instance.subscription.amount_per_period_usdt else instance.subscription.get_converted_amount(True),
                     "url":instance.subscription.grant.url,
                     "network":instance.subscription.network,
                     "txid":instance.subscription.split_tx_id,
```
thelostone-mc, line 1941: ^ why are making the changes here ?

octavioamu (approved):

app/grants/models.py
octavioamu: I think is to ignore the gitcoin fee there as is already on the value ?

zacheryschiller (approved):
Also tested locally 👍

app/grants/models.py
gdixon: Yeah thats right @octavioamu! Passing `ignore_gitcoin_fee=True` will make sure we get the full contribution value (without deducting the matching pool contrib)

PR #9446, opened 2021-09-07
Runs clr calcs more frequently (every 15 mins)

gdixon:
##### Description

This PR updates the crontabs to run the clr calc more frequently (every 15 minutes)

##### Testing

Tested locally

### DIFF
--- a/scripts/crontab2
+++ b/scripts/crontab2
@@ -1,4 +1,4 @@
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
 
-## CLR
-1,30 * * * * cd gitcoin/coin; bash scripts/run_clr.bash slim  2>&1
+## SLIM CLR (every 30 minutes)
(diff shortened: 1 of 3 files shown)


zacheryschiller (approved):
LGTM
When we deploy this, let's make sure to watch it for a few hours to make sure it's not taking longer than we expect after many runs.

owocki:
note that i dont think the deploy script automatically installs crontab2 - crontab4 .  youll need to install these manually on the celery boxes once merged in.

chibie (commented):
@gdixon just for clarity, shouldn't the slim clr calc be more frequent than full clr calc?

chibie (approved):

octavioamu (approved):

gdixon:
> note that i dont think the deploy script automatically installs crontab2 - crontab4 . youll need to install these manually on the celery boxes once merged in.

I've got these installed 👍 

> @gdixon just for clarity, shouldn't the slim clr calc be more frequent than full clr calc?

Yes it should -- I think we can likely deprecate the slim calc now though because its only around 30secs faster than the full calc (when we run it against the end state of one of the gr10 clrs)

PR #9451, opened 2021-09-08
GITC-382: handle negligible amounts

chibie:
##### Description

This PR prevents the sending of grant contribution email for amounts less than $1


gdixon (approved):
Interesting approach!! But lgtm!! 🚀

PR #9453, opened 2021-09-08
Add db read replica 3

zacheryschiller:
##### Description

Adding the newly deployed read replica 3

##### Testing

Must test with deployment

##### Notes

Once merged we need to update the DBs in the django admin as well.

### DIFF
--- a/app/app/db.py
+++ b/app/app/db.py
@@ -9,7 +9,7 @@ class PrimaryDBRouter:
         Reads go to a randomly-chosen replica if backend node
         Else go to default DB
         """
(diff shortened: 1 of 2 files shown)


gdixon:
Can you add the new replica to `settings.py` too please

zacheryschiller:
> Can you add the new replica to `settings.py` too please (and then we need to add the config to all the boxes .env file)

These are both done now.

thelostone-mc (approved):

gdixon (approved):

PR #9461, opened 2021-09-09
Cypress testing/add grant creation test

ksolo:


thelostone-mc (approved):

app/grants/migrations/0123_auto_20210726_0703.py
```diff
@@ -1,11 +1,14 @@
...
 # Generated by Django 2.2.24 on 2021-07-26 07:03
+import os
 
 from django.db import migrations, models
 import economy.models
 
-
 def port_data(apps, schema_editor):
+    if not os.getenv('CI'):
+        pass
```
ksolo, line 10: @zacheryschiller This function was commented out preventing the data migration from categories to tags below on ln 63. I uncommented that and put this guard to prevent from running elsewhere. If this has already executed on prod, it should be skipped automatically but wanted to call it and and see if you had a different approach in mind.

zacheryschiller (approved):

Issue #9426, opened 2021-09-03
Fix checkout recommendation based on gas cost


mds1:
There are two issues with the current gas cost recommendations:

1. The [polygon gas estimation](https://github.com/gitcoinco/web/blob/gr11/app/assets/v2/js/cart-ethereum-polygon.js#L336-L384) is missing the cost of an L1 deposit the user doesn’t have sufficient balance. There should a section [like this](https://github.com/gitcoinco/web/blob/d41c1a6cb2abee0a23f5e9990bce2ebc6c9dcd72/app/assets/v2/js/cart-ethereum-zksync.js#L283-L294) part of the zkSync estimate

2. The `checkoutRecommendation()` logic in `cart.js` is incorrect. When it was just L1 and zksync, we compared gas usage directly, and this is ok because zksync gas price is directly tied to L1. But now with Polygon, you can no longer just compare gas usage directly because Polygon and L1 will have the same gas usage, but on Polygon gas cost is lower. You can fix this by updating the checkout recommendation to compare USD cost as follows:
    1. costL1 = estimateL1 * basefee * eth/usd (basefee is a good enough estimate of current gas price)
    2. costZkSync = estimateZkSync * basefee * eth/usd (you should verify this zkSync estimation approach with the zkSync team to make sure there’s no concept of L2 gas price they have that I’m forgetting, and to ensure basefee is a sufficient estimate for what they charge when you checkout)
    3. costPolygon = estimatePolygon * 2 gwei * matic/usd (polygon transactions can almost always be less than 2 gwei AFAIK so just hardcode it for simplicity)

PR #9464, opened 2021-09-10
polygon finale

chibie:
##### Description

- [x] stable fix for standard checkout bug on rinkeby
- [x] finish/restore gas estimation
- [x] update grant detail page with link to polygon explorer
- [x] remove staff flag


app/assets/v2/js/cart-ethereum-polygon.js
```diff
@@ -138,6 +134,12 @@ Vue.component('grantsCartEthereumPolygon', {
       this.polygon.checkoutStatus = 'depositing';
     },
 
+    getBulkCheckoutAddress() {
+      return appCart.$refs.cart.network === 'mainnet'
+        ? '0xb99080b9407436eBb2b8Fe56D45fFA47E9bb8877'
+        : '0x3E2849E2A489C8fE47F52847c42aF2E8A82B9973';
```
mds1, line 140: Is `0x3E2849E2A489C8fE47F52847c42aF2E8A82B9973` supposed to be the BulkCheckout contract on Polygon? It seems like there's no contract there: https://polygonscan.com/address/0x3E2849E2A489C8fE47F52847c42aF2E8A82B9973

app/assets/v2/js/cart.js
```diff
@@ -392,24 +395,19 @@ Vue.component('grants-cart', {
       const estimateZkSync = Number(this.zkSyncEstimatedGasCost); // zkSync gas cost estimate
       const estimatePolygon = Number(this.polygonEstimatedGasCost); // polygon gas cost estimate
```
mds1, line 396: It looks like the gas estimations are still directly comparing gas used, and not the actual USD cost as described in item 2 of #9426. Though maybe I'm missing it somewhere?

app/assets/v2/js/cart-ethereum-polygon.js
```diff
@@ -338,7 +344,67 @@ Vue.component('grantsCartEthereumPolygon', {
...
+          if (value.isBalanceSufficient == false) {
+            return value.amount;
+          }
+        });
+
+        for (const tokenSymbol in requiredAmounts) {
+          if (tokenSymbol === 'ETH') {
+            gasLimit += 94659; // add ~94.66k gas for ETH deposits
+          } else {
+            gasLimit += 103000; // add 103k gas for token deposits
```
mds1, line 371: Might be useful to add links to the reference deposits you used to come up with these numbers, for future reference/future devs

app/assets/v2/js/cart-ethereum-polygon.js
```diff
@@ -338,7 +344,67 @@ Vue.component('grantsCartEthereumPolygon', {
...
+      const daiAddress = this.getTokenByName('DAI')?.addr;
+      const isAllDai = donationCurrencies.every((addr) => addr === daiAddress);
+
+      if (isAllDai) {
+        if (donationCurrencies.length === 1) {
+          // Special case since we overestimate here otherwise
+          return gasLimit + 65000;
+        }
+        // https://github.com/mds1/Gitcoin-Checkout-Gas-Analysis
+        return gasLimit + 10000 * donationCurrencies.length + 80000;
```
mds1, line 388: Similarly I think I saw a plotly dashboard link somewhere, where would handy to include those here

requirements/base.txt
```diff
@@ -49,7 +49,7 @@ django-ipware==2.0.2
 geoip2==2.8.0
 django-silk==2.0.0
 django-extensions==2.2.1
-ipdb==0.11
+ipdb==0.13.9
```
mds1, line 52: Was this bumped intentionally? Just want to make sure it won't cause any issues since it wasn't mentioned in the PR description

app/assets/v2/js/cart-ethereum-polygon.js
chibie: yea... that's for mumbai.

[0xb99080b9407436eBb2b8Fe56D45fFA47E9bb8877](https://polygonscan.com/address/0xb99080b9407436eBb2b8Fe56D45fFA47E9bb8877) is for mainnet

requirements/base.txt
chibie: this was brought in from another pr => https://github.com/gitcoinco/web/pull/9443

app/assets/v2/js/cart-ethereum-polygon.js
mds1: Ah right these are only the polygon addresses, makes sense! I was thinking the first address was L1, sorry

And just to confirm `mainnet` is what the network name will be when on mainnet polygon? I would guess it's `polygon` since `mainnet` would be misleading/confusing, but I'm not sure so figured I'd ask

chibie: yes polygon also uses `mainnet` as network name and `testnet` for mumbai network.

surpising that was pretty straightforward with our system... `testnet` was more complicated to stick with

app/assets/v2/js/cart.js
chibie: this feature hasn't be done yet... it'll be handled in a follow-up pr since it isn't polygon specific

app/assets/v2/js/cart-ethereum-polygon.js
mds1: Heh interesting, well good to know, thanks!

app/assets/v2/js/cart.js
mds1: Got it, so this PR looks good to me then

mds1 (approved):

app/grants/templates/grants/cart/eth.html
```diff
@@ -173,8 +171,7 @@ <h4 class="col gc-font-base text-dark p-0">Summary</h4>
   <!-- Checkout buttons -->
   <div class="row justify-content-end">
     <div class="btn-group flex-column flex-grow-1 flex-md-grow-0 flex-md-row" role="group" aria-label="Checkout buttons">
-      
-      {% if is_staff %}
```
thelostone-mc, line 177: ^ I'd say let's keep this on until we get the green sign from the team that this is good for rollout

thelostone-mc (approved):

zacheryschiller (approved):

PR #9478, opened 2021-09-13
Cypress Testing on CI

ksolo:
# Description

Adding Cypress tests with CI configurations to run tests on Travis.

Includes:
* ignoring artifacts from cypress run in gitignore (videos and screenshots)
* update travis yml to include the xvfb service and chrome addon
* update travis yml to call the cypress-run script
* cypress-run script handles additional configuration to setup ganache, db, static assets and also run the app and cypress tests
* update docker files to build chrome to run tests in a container
* add initial cypress tests


cypress.json
```diff
@@ -1 +1,5 @@
-{}
+{
+  "chromeWebSecurity": true,
+  "baseUrl": "http://localhost:8000",
+  "projectId": "e7rv3w"
```
chibie, line 4: is `projectId` a secret?

ksolo: No, it matches the test runs to the dashboard. There is not an ability to read data or access anything from this being exposed.

chibie (approved):

zacheryschiller (approved):

thelostone-mc:
The commit history seems to have other commit not related this PR ! 
Any idea why that's there ? 

Would we need additional documentation to explain how to use/run this locally / is it not required?

ksolo:
@thelostone-mc I can rebase this and remove the other commit. I'll add documentation for running tests locally as well.

PR #9483, opened 2021-09-14
add-grants-unit-tests

Jer-Sch:
##### Description

`add-grants-unit-tests` rebased onto `master`

This PR contains the prep work done for adding unit testing to the grants models. It doesn't contain any new tests.

The `extract-grants-models-into-individual-files` branch was merged into this branch (`extract-grants ...` was already approved previously) and this branch was then rebased onto `master`, so this PR contains work already approved.

I verified that each of the new model files were exactly like the models in the original `models.py` file, including any recent changes that occurred after the original work on this was done.

Some of the changes in this PR are irrelevant to the work I did and appear to be formatting adjustments that occurred when I ran `make fix`

For easy reference, those files are:

- app/app/context.py
- app/app/db.py
- app/app/urls.py
- app/app/utils.py
- app/avatar/models.py
- app/dataviz/d3_views.py

##### Testing

No tests yet, as this is prep work for adding more tests.


app/grants/migrations/0017_matchpledge_end_date.py
```diff
@@ -9,11 +9,19 @@ class Migration(migrations.Migration):
...
+            ),
+        ]
+    except AttributeError:
+        operations = [
+            migrations.AddField(
+                model_name='matchpledge',
+                name='end_date',
+                field=models.DateTimeField(default=grants.models.match_pledge.next_month),
+            ),
+        ]
```
ksolo, line 27: I think this should be a `pass` instead of retrying the code that failed.

Jer-Sch: @ksolo It's not exactly retrying the code that failed. The code in the except block is the correct migration code because it sets the default field to grants.models.match_pledge.next_month which is correct, versus grants.models.next_month which is no longer correct. I've tried using pass in the except block, but the test suite breaks because the end_date field for MatchPledge doesn't get created.

PR #9485, opened 2021-09-15
purge: ptokens

thelostone-mc:
##### Description

Removes personal tokens

##### Refers/Fixes

GITC-409

##### Flows Tested

- bounty flow
- hackathon flow
- grants flow
- explorer
- profile


gdixon (approved):
LGTM, but let's make sure the tests are green before merging -- it looks like we need to sort out the fixtures/create a migration to drop the `!null` constraint on `profile`->`tokens`

app/dashboard/migrations/0150_auto_20200904_1505.py
```diff
@@ -8,21 +8,21 @@
...
+        # migrations.AddField(
+        #     model_name='activity',
+        #     name='ptoken',
+        #     field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='ptoken_activities', to='ptokens.PersonalToken'),
+        # ),
+        # migrations.AddField(
+        #     model_name='activity',
+        #     name='redemption',
+        #     field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='ptokens.RedemptionToken'),
+        # ),
```
ksolo, line 25: Is there a reason to keep this around? I'd prefer to delete unused code over commenting it out and leaving it in. We can always pull this back from git if it's needed.

thelostone-mc: So this is only cause it's a migration which are auto generated ! 
Ideally we wouldn't want to change this but when we remove a django it causes travis to complain as it runs through the boot

PR #9513, opened 2021-09-22
Removing the third RR reference

zacheryschiller:
##### Description

 Remove the third reference to a read-replica so that we can dedicate RR-1 and RR-2 to CLRs and then dedicate RR-3 and RR-4 to the app and asset servers.

##### Testing

We need multiple RRs and to set the environment variables to test. Which we will do on prod.

### DIFF
--- a/app/app/db.py
+++ b/app/app/db.py
@@ -9,7 +9,7 @@ class PrimaryDBRouter:
         Reads go to a randomly-chosen replica if backend node
         Else go to default DB
         """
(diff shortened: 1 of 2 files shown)


gdixon (approved):

thelostone-mc (changes requested):
commit history isn't clean ! could we cherry pick the commit ?

gdixon:
This commit was cherry-picked to stable :+1

PR #9515, opened 2021-09-22
Fix reference for testing environment

zacheryschiller:
##### Description

Fixes bug that was preventing the /earn page from loading correctly on testing environment.

##### Testing

Testing on... testing

### DIFF
--- a/app/perftools/management/commands/create_jtbd_cache.py
+++ b/app/perftools/management/commands/create_jtbd_cache.py
@@ -61,7 +61,7 @@ def create_jtbd_earn_cache():
 
     featured_grant = None
 
-    if settings.ENV in ['prod']:
+    if settings.ENV == 'prod':
         # WalletConnect
         grant_id = 275
     else:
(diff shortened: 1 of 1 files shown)


app/perftools/management/commands/create_jtbd_cache.py
```diff
@@ -61,7 +61,7 @@ def create_jtbd_earn_cache():
 
     featured_grant = None
 
-    if not settings.DEBUG:
+    if 'prod' in settings.ENV:
```
thelostone-mc, line 64: would setting.ENV be a string and not an array ?

zacheryschiller: Yes! I swapped them by accident. Thank you for catching that!

app/perftools/management/commands/create_jtbd_cache.py
```diff
@@ -61,7 +61,7 @@ def create_jtbd_earn_cache():
 
     featured_grant = None
 
-    if not settings.DEBUG:
+    if settings.ENV in ['prod']:
```
gdixon, line 64: `WalletConnect` will only have this `grant_id` on production - I'd suggest that we make this check more explicit:

```suggestion
    if settings.ENV == 'prod':
```

gdixon (changes requested):

ksolo (approved):

gdixon (approved):

PR #9516, opened 2021-09-23
bsci: add manual label

thelostone-mc:
##### Description

Add manual label to ensure bsci script does not override them
Has been locally tested

Note: After running migration
run the following command

```python
from townsquare.models import SquelchProfile
sps = SquelchProfile.objects.filter(label='Human')

for sp in sps:
  sp.label='Manual'
  sp.save()
```


gdixon (approved):
LGTM! Lets get this out asap so that we don't lose the migration step 👍

zacheryschiller (approved):

Issue #9493, opened 2021-09-18
Logout functionality does not work


nya-elimuai:
**Describe the bug**
When pressing "My Account" --> "Log Out", the user does not get logged out. The user session remains active.

**To Reproduce**
1. Go to https://gitcoin.co/settings/account
2. Press the user drop-down. Then try to log out.

**Expected behavior**
User should be logged out.

**Screenshots**
<img width="431" alt="Screen Shot 2021-09-18 at 1 47 45 PM" src="https://user-images.githubusercontent.com/15718174/133884722-64281796-970e-4350-8dd3-9eaf4dbb5996.png">

**Desktop (please complete the following information):**

- OS: Mac.
- Browser: Chrome & Brave.
- Browser Version:

**Smartphone (please complete the following information):**

- Device:
- OS:
- Browser:
- Version:

**Additional context**

PR #9555, opened 2021-10-02
Changed log out link

julian-st:
log out link was broken before

##### Description

The redirect link is now going to / instead of the current path

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/9493

##### Testing

Go to https://gitcoin.co/settings/account
Press the user drop-down. Then try to log out.

### DIFF
--- a/app/dashboard/templates/shared/nav_auth.html
+++ b/app/dashboard/templates/shared/nav_auth.html
@@ -206,7 +206,7 @@
                   {% endif %}
                   <hr/>
                   <li>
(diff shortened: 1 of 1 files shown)


thelostone-mc:
Hmm i see what you're doing here! This makes sense for pages which don't need the user to be logged in.
Would it work for pages like gitcoin.co/grants/explore ?  
where when i sign out -> it keeps me on gitcoin.co/grants/explore

chibie (approved):

ksolo (approved):

Jer-Sch (approved):

PR #9559, opened 2021-10-04
fix grants explorer

chibie:
##### Description

grants explorer broke due get_clr_rounds_metadata update

##### Refers/Fixes

https://gitcoincore.slack.com/archives/CDML7FS7M/p1633365131046300

### DIFF
--- a/app/grants/management/commands/sent_cart_reminder.py
+++ b/app/grants/management/commands/sent_cart_reminder.py
@@ -51,7 +51,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
 
-        _, round_start_date, round_end_date, _ = get_clr_rounds_metadata()
+        _, round_start_date, round_end_date, _, _, _, _, _ = get_clr_rounds_metadata()
 
         last_activity_by_user = CartActivity.objects.filter(latest=True, created_on__gt=round_start_date).exclude(metadata=[])
         count = 0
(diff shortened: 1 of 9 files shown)


ksolo (approved):

Jer-Sch (approved):

app/grants/management/commands/sent_cart_reminder.py
```diff
@@ -51,7 +51,7 @@ def add_arguments(self, parser):
 
     def handle(self, *args, **options):
 
-        _, round_start_date, round_end_date, _ = get_clr_rounds_metadata()
+        _, round_start_date, round_end_date, _, _, _, _, _ = get_clr_rounds_metadata()
```
ksolo, line 54: @chibie This is a quick fix, but this method is extremely fragile. It feels like it should be returning an object, not a tuple that we can access data on without guessing how many and where the `_` should go. I approve this so we can fix the deploy, but let's open a new tech debt ticket in Jira to fix this.

chibie: @ksolo yea makes perfect sense

ksolo: Using a named tuple would likely provide enough of an abstraction to prevent these types of bugs - https://docs.python.org/3/library/collections.html#collections.namedtuple

chibie: ticket is GITC-477

PR #9591, opened 2021-10-12
GITC-453: Add grant to collection from explorer

ksolo:
##### Description

Non related changes:
* Improve pytest collect phase by 15 seconds with pytest.ini file
* Added missing `__init__.py` for test directories
* Begin creating a command for adding grants to cypress
* update return type of `get_grant_type_cache` - which was returning a dict that the JS used as an obj and could not call `sort`

Related Changes:
* Add modal to the grant card
* Add open/close modal to component

Update card to include "add to collection"
Screen Shot 2021-10-12 at 2 33 42 PM

Modal
Screen Shot 2021-10-12 at 2 35 00 PM

Small Screen
Screen Shot 2021-10-12 at 2 36 52 PM

##### Refers/Fixes

GITC-453

##### Testing

Create a collection from the cart page
Return to the grants explore page
Click on a "add to collection" link for grant
Select a collection
Confirm

### DIFF
--- a/app/assets/v2/js/grants/components.js
+++ b/app/assets/v2/js/grants/components.js
@@ -109,6 +109,7 @@ Vue.component('grant-card', {
     },
     addGrantToSelectedCollection: async function() {
       const collection = this.collections.find(collection => collection.id == this.selectedCollection);
+
       await this.addToCollection(collection, this.grant);
 
       this.closeModal('add-to-collection');


willsputra:
looking good @ksolo! just a couple of feedback
• on the grant card, we can remove the old "Add to Collection" button since it does the same thing as the new one
Screenshot 2021-10-13 at 8 10 08 PM
 
• on the modal, could we add more left and right padding as per the design?

PR #9631, opened 2021-10-27
Fix payout amount bug

Jer-Sch:
##### Description

I would greatly appreciate feedback on this to make sure I'm doing it right. 

I think some of the changes here are linter/formatting changes.


Jer-Sch:
@thelostone-mc I have this working in the activity feed. Next steps on the ticket are to get it working in the emails, and in the user profile.

Do you know which email I should be looking for. I looked for a payout notification email in the administration section, but didn't see one.

Is there a management command I need to run to sync the user profile data?

Issue #9460, opened 2021-09-09
Search doesn't work for certain projects


abitrolly:
**Describe the bug**

It is hard to find "Cyber Academy" (https://gitcoin.co/grants/3169/cyber-academy) from https://gitcoin.co/grants/explorer/

**To Reproduce**

1. Go to https://gitcoin.co/grants/explorer/
2. Enter "cyber a"
3. Wait 5 seconds
4. Add "c" to make "cyber ac"
5. Wait 5 seconds
6. See no results

Or

1. Go to https://gitcoin.co/grants/explorer/?page=1&limit=6&me=false&sort_option=weighted_shuffle&network=mainnet&state=active&profile=false&sub_round_slug=false&collections_page=1&grant_regions=&grant_types=&grant_tags=&tenants=&idle=true&tab=grants&keyword=cyber+academy
2. Wait for Cyber Academy to appear
3. Erase "emy" from search string
4. Wait 5 seconds
5. Add "emy"
6. Wait 5 seconds
7. See no results

Also it is not in the list if you just type "academy".

![Screenshot 2021-09-09 at 11-55-16 (💰$900k Match LIVE ) Grants](https://user-images.githubusercontent.com/8781107/132661166-33ef7128-b02e-4cf4-838e-c1b1123cc227.png)

**Expected behavior**

1. (optional) Spinner that shows which phrase if being searched for
2. Find Cyber Academy page no matter how its name is typed
3. Have it in top results if the search string matches title

**Desktop (please complete the following information):**

- OS: Fedora 34
- Browser: Firefox
- Browser Version: 91.0.2

octavioamu:
Thanks for the detailed report, there are 2 things here, one is our search index full words and the other is the search box is with a lazy property. 
For a quick fix I will change the lazy search to be just debounced, is not a full fix but at least will request the new query 100% of the times.

abitrolly:
@chibie can we leave it open until it is fixed?

PR #9637, opened 2021-10-28
GITC-406: exact title match for grants search

chibie:
##### Description
As a grants contributor using the explorer search and sorting mechanism, I want exact title matches and predictable sort values when using the search bar.

Notes:

Currently, the behaviour of weighted shuffle does not return exact title matches when using the search bar, which is confusing to users.

Update weighted shuffle to ensure that exact title matches are presented first.

##### Checklist

- [x] Ensure "Weighted Shuffle" is the default sort option
- [x] Debounce search input -> perform search after user has stopped typing
- [x] When a user enters search keyword, sort option should be set to "Most Relevant"
- [x] When a user clears the search keyword, sort option should revert to "Weighted Shuffle"
- [x] When a user changes the sort option when a search result is showing, the search result should be sorted accordingly

##### Refers/Fixes

GITC-406
#9460

##### Testing

https://share.vidyard.com/watch/9sXqiZJ4Nkr6TAhJbQaJPC?


app/assets/v2/js/grants/index.js
```diff
@@ -175,18 +204,23 @@ if (document.getElementById('grants-showcase')) {
...
         let vm = this;
 
         vm.fetchedPages = [];
         vm.$set(vm, 'params', {...vm.params, ...query});
 
         if (vm.tabSelected === 'grants') {
+          if (vm.params.keyword && vm.params.sort_option == 'weighted_shuffle') {
+            vm.params.sort_option = '';
+          } else if (vm.params.keyword == '' && vm.params.sort_option != 'weighted_shuffle') {
+            vm.params.sort_option = 'weighted_shuffle';
```
gdixon, line 217: This will behave in unexpected ways - when you first add a keyword (providing you have weighted_shuffle selected on the sort) this will correctly clear that sort, but if you were to then add a different selection to the query (say you added a tag filter) we would hit the next if statement which would restore the weighted_shuffle selection.

Consider adding this to the keyword input itself - i.e onkeyup we want to be clearing the sort, any other action should leave the sort selection alone.

app/assets/v2/js/grants/index.js
```diff
@@ -23,6 +23,36 @@ $(document).ready(() => {
...
+  return function() {
+    // reference the context and args for the setTimeout function
+    let context = this; let
+        args = arguments;
+    
+    // allow user-triggered debounce from the func's supplied argument
+    if (args[0].debounce !== undefined && args[0].debounce === true)
+      immediate = false;
+    else if (args[0].debounce === undefined || args[0].debounce === false)
+      immediate = true;
```
gdixon, line 39: Why are we applying debounce options as a param to the debounced func (this func shouldn't be aware that it is being debounced)?

app/assets/v2/js/grants/index.js
```diff
@@ -23,6 +23,36 @@ $(document).ready(() => {
...
+      if (!immediate) {
+        // call the original function with apply
+        func.apply(context, args);
+      }
+    }, wait);
+
+    if (callNow)
+      func.apply(context, args);
+  };
+}
```
gdixon, line 55: We have debounce methods in the codebase already, consider moving one of those to `shared.js` rather than implementing again 👍

app/grants/views.py
```diff
@@ -733,15 +731,22 @@ def get_grants_by_filters(
...
             region_query |= Q(region=region)
         _grants = _grants.filter(region_query)
 
+    if keyword:
+        # 13. Grant search by title & description
+        _grants = _grants.annotate(search=SearchVector('description'))
+        keyword_query = Q(title__icontains=keyword)
+        keyword_query |= Q(search=keyword)
+
+        _grants = _grants.filter(keyword_query)
```
gdixon, line 747: Is this enough to ensure that the sort order is respected? From our testing the paginated result was not being ordered by the corresponding `rank` - did you verify with enough test data?

gdixon (changes requested):
Looking good @chibie!! I've left a few comments - let me know if you have any questions!

app/grants/views.py
chibie: yes it works... i tested on staging... staging is currently on this branch... you could try it with "Save AWS" keyword

chibie: > Is this enough to ensure that the sort order is respected?

sort is applied only after the search result is returned

> did you verify with enough test data?

yes on staging with "Save AWS" keyword (created multiple grants with title variations)...staging is currently on this branch

app/assets/v2/js/grants/index.js
gdixon: Ahhh, yes you are correct! My bad, sorry! I still think it would be useful to do the clearing onkeyup, because this will only cover us if the user had weighted_shuffle applied before entering a keyword. We would want to clear the sorting method no matter what the user had previously selected when they enter/alter a keyword.

chibie: gotcha... just pushed the commit for keyup

gdixon (approved):
LGTM!! 🚀

PR #9660, opened 2021-11-03
fix: removes filter on activities to avoid a bitmap heap scan

gdixon:
##### Description

This PR removes a filter which is being applied globally to the Activity that is causing postgres to perform a bitmap heap scan

##### Refers/Fixes

Refers: https://github.com/gitcoinco/web/commit/b8b8552050269708d48030e8fa9d9104c978826a

https://gitcoincore.slack.com/archives/CDXGQRZB3/p1635704073127400?thread_ts=1635690005.123700&cid=CDXGQRZB3

##### Testing

Tested locally (reduces the query time from 40s -> 100ms)


chibie (approved):

nutrina (approved):
Look good to me

PR #9661, opened 2021-11-03
Jeremy/GITC-565/fix Twitter handle feature

Jer-Sch:
##### Description

This PR resolves an issue that occurred when full Twitter urls were pasted or entered into the Twitter handle fields on the new grant form.

Now when full Twitter urls are pasted or typed, they are formatted so that only the username is present on the form.

https://user-images.githubusercontent.com/36638840/140180312-22e8822b-bf12-4457-b3c2-a2c0db813d77.mov

##### Testing

This PR includes a Cypress integration test.


app/assets/v2/js/grants/_new.js
```diff
@@ -277,7 +277,26 @@ Vue.mixin({
           _alert(err.message, 'danger');
         }
       });
-    }
+    },
+    onBlur(event) {
+      let vm = this;
+      
+      event.preventDefault();
```
ksolo, line 284: You won't need this is you add the prevent to the @blur in the html. https://vuejs.org/v2/guide/events.html#Event-Modifiers

app/assets/v2/js/grants/_new.js
```diff
@@ -277,7 +277,26 @@ Vue.mixin({
           _alert(err.message, 'danger');
         }
       });
-    }
+    },
+    onBlur(event) {
+      let vm = this;
```
ksolo, line 282: I think we can access `this` directly in the function. I don't think we need the alias.

app/assets/v2/js/grants/_new.js
```diff
@@ -277,7 +277,26 @@ Vue.mixin({
...
+      
+      event.preventDefault();
+
+      const inputField = event.target;
+      const fullTwitterURL = /https:\/\/twitter.com\/\w{1,15}/
+      const twitterUsername = /(\w{1,15})$/
```
ksolo, line 288: are twitter handles limited to 15 characters?

app/assets/v2/js/grants/_new.js
```diff
@@ -277,7 +277,26 @@ Vue.mixin({
           _alert(err.message, 'danger');
         }
       });
-    }
+    },
+    onBlur(event) {
```
ksolo, line 281: We should rename this to reveal intent. Other elements may need to respond to blur and this could lead to confusion. `validateTwitterHandle` or something like that is a better name

ksolo (changes requested):

app/assets/v2/js/grants/_new.js
Jer-Sch: Yep https://help.twitter.com/en/managing-your-account/twitter-username-rules

ksolo: It also specifies it must be at least 4 characters, it might be worth mentioning to Frank and adding a ticket to handle that (or add it to this one).

app/assets/v2/js/grants/_new.js
```diff
@@ -277,7 +277,22 @@ Vue.mixin({
...
+    },
+    handleTwitterUsername(event) {
+      const inputField = event.target;
+      const fullTwitterURL = /https:\/\/twitter.com\/\w{1,15}/
+      const twitterUsername = /(\w{1,15})$/
+
+      let extracted
+
+      if (inputField.value.match(fullTwitterURL)) {
+        extracted = '@' + inputField.value.match(twitterUsername)[0]
```
ksolo, line 289: I think you can do all this in one go using a capture group for the username.

```
const  matchResult = inputField.value.match(/https:\/\/twitter.com\/(\w{1,15}/)/)
if (matchResult) { extracted = `@{matchResult[1]}` }
```

I don't think you need two separate regexps to do this.

Jer-Sch: Oh nice. That's really slick

chibie (approved):
looks good!

app/assets/v2/js/grants/_new.js
```diff
@@ -277,6 +277,17 @@ Vue.mixin({
...
+    handleTwitterUsername(event) {
+      const inputField = event.target;
+      const matchResult = inputField.value.match(/https:\/\/twitter.com\/(\w{4,15})/);
+      let extracted;
+
+      if (matchResult) {
+        extracted = `@${matchResult[1]}`;
+      }
+
+      this.$set(this.form, inputField.id, extracted);
```
ksolo, line 290: I don't think this covers the case where a user inputs a valid twitter handle now. If they input `@gitcoin` I think it wold be set as `undefined` with this implementation.

```suggestion
      const extracted = matchResult ? `@${matchResult[1]` : inputField.value;

      this.$set(this.form, inputField.id, extracted);
```

app/grants/templates/grants/_new.html
```diff
@@ -155,7 +155,7 @@ <h1 class="text-center font-title-xl">Create a Grant</h1>
                   <label class="font-caption letter-spacing text-black-60 text-uppercase" for="twitter_handle_1">Project Twitter Handle</label>
                   <span class="font-smaller-6 badge badge-greylight text-capitalize ml-2">Required</span>
 
-                  <input id="twitter_handle_1" v-model="form.twitter_handle_1" name="twitter_handle_1" class="form__input form__input-lg" maxlength="20" type="text" placeholder="humanfund" required/>
+                  <input id="twitter_handle_1" v-model="form.twitter_handle_1" @blur.prevent="handleTwitterUsername($event)" name="twitter_handle_1" class="form__input form__input-lg" type="text" placeholder="humanfund" required/>
```
ksolo, line 158: I need to double check but I think for the `@blur.prevent=` part you can pass it the name of the function and not invoke it. `@blur.prevent="handleTwitterUsername` This applies to the other field as well. Double check me on this one though, I could be wrong here.

Jer-Sch: I'll double check. I tried it both ways and it worked, but I noticed when I invoked it I saw less warnings in the console.

ksolo: What were the warnings?

Jer-Sch: Actually, it doesn't apply now. I was seeing errors and warnings in the console earlier when I was using `window.getSelection()` to track the `input` field the user clicked on. That wasn't working properly in Chrome, so I used `event.target` instead, and now I see no errors or warnings.

ksolo (approved):

gdixon (approved):

PR #9663, opened 2021-11-04
jeremy/GITC-574/grants explorer routing change

Jer-Sch:
##### Description

This PR changes the link to the Grants Explorer on the Products dropdown to route to the explorer rather than the grants page.

https://user-images.githubusercontent.com/36638840/140388373-d48e1e47-260b-45a7-9e82-8041235f6698.mov

##### Testing

This PR includes a Cypress test

### DIFF
--- a/cypress/integration/landing_page/test_products_menu.js
+++ b/cypress/integration/landing_page/test_products_menu.js
@@ -1,0 +1,25 @@
+describe('Products menu', () => {
+  before(() => {
+    cy.setupMetamask();
+  });
+
+  beforeEach(() => {
+    cy.impersonateUser();
+  });
+
+  afterEach(() => {
+    cy.logout();
+  });
+
+  after(() => {
+    cy.clearWindows();
+  });
+
+  it('navigates to the grants explorer when \'Explore Grants\' is selected', () => {
+    cy.get('#dropdownProducts').trigger('mouseenter');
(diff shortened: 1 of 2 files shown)


cypress/integration/landing_page/test_products_menu.js
```diff
@@ -0,0 +1,25 @@
...
+    cy.setupMetamask();
+  });
+
+  beforeEach(() => {
+    cy.impersonateUser();
+  });
+
+  afterEach(() => {
+    cy.logout();
+  });
```
ksolo, line 12: are these needed for this test?

Jer-Sch: I don't think the `afterEach` is necessary, but we definitely need the `impersonateUser`

ksolo: If you have to log in for this page, then you will need to make sure to log out after the test. If you need the impersonateUser then keep the logout. Cypress will keep that user in session and it could cause state issues for the next test. I expected the grants page to be public.

Jer-Sch: Ah that makes sense. I'll put it back, but I just tested it on the production site, and the pages are public. The test breaks without `impersonateUser()` though.

gdixon (approved):
LGTM!

frankchen07:
LGTM, thanks!

thelostone-mc (approved):

PR #9668, opened 2021-11-08
Testing workshop

ksolo:
Prior to workshop:
* updated existing factories to leverage the lazy function or lazy attribute for seed data
* moved factories in grants up to `tests/factories` instead of `tests/models/factories` to make them easier to find
* added test file for bounty fulfillment

Workshop - added bountry fulfillment test.


.gitignore
```diff
@@ -61,3 +61,6 @@ _build/
 # cypress artifacts
 cypress/videos
 cypress/screenshots
+
+#asdf vm
```
chibie, line 65: asdf vm??

chibie (approved):
lgtm! just the comment above

.gitignore
ksolo: When I set this up locally without the docker context I used [asdf version manager](https://asdf-vm.com/). It can be used for Python, JS, Ruby, Elixir... To avoid committing the `.tool-versions` file I added it to the `.gitignore`

gdixon: Would it be worth including the url you provided above? It does look like a random string atm 😄

```suggestion

# asdf-vm (https://asdf-vm.com/)
```

app/grants/tests/factories/cart_activity_factory.py
```diff
@@ -1,18 +1,18 @@
...
-from dashboard.tests.factories.profile_factory import ProfileFactory
+from dashboard.tests.factories import ProfileFactory
 from grants.models.cart_activity import CartActivity
 
 from .grant_factory import GrantFactory
 
 
 @pytest.mark.django_db
 class CartActivityFactory(factory.django.DjangoModelFactory):
-    """Create mock CartActivity for testing."""
```
gdixon, line 11: How come were removing all of these descriptions?

gdixon (commented):
Nice work @ksolo :rocket - just a couple of minor comments/questions

app/grants/tests/factories/cart_activity_factory.py
ksolo: I removed them because they were technically not correct, these are factories not mocks. I also didn't think they added context. I am happy to put them back in though with the correction that they are factories that return instances of the Django Model.

gdixon (approved):

app/grants/tests/factories/cart_activity_factory.py
gdixon: Thats cool with me, I think they're well enough named that we don't need the descriptions!

PR #9677, opened 2021-11-11
GITC-475: Decouples active from latest in GrantCLRCalculation's

gdixon:
##### Description

This PR decouples `active` from `latest` and adds a command to reinstate the previous rounds `latest` state. This will allow us to always find the `latest` calc after the round has ended.

After running migrations, we should run:

``` ./manage.py place_latest_on_past_calcs  ```

##### Refers/Fixes

Fixes: GITC-475

##### Testing

Tested locally


thelostone-mc:
Would it make sense to rename active to finalised ?

gdixon:
> Would it make sense to rename active to finalised ? Also could we tweak out notion doc on payout round -> so that it's clear on how we'd proceed during payout

The `active` field should be loosely correlated to the `is_active` field on the `GrantCLR` - I think `active` makes sense as this field is just to tell us which round is currently ongoing.

I'm not sure if any changes are required, the payouts will continue to work against the `latest` field - so the only difference is that we can now run the payout scripts without the rounds being active +1 - could you point me to those documents so that I can check if we need to make updates?

PR #9692, opened 2021-11-16
GITC-599: remove duplicates in token selection input (cart + tips)

chibie:
##### Testing

tested locally

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -191,6 +191,16 @@ Vue.component('grants-cart', {
       }
       return result;
     },
+    fetchTokens() {
+      // removes duplicates from the tokens array of objects
+      let vm = this;
+
+      return vm.filterByChainId.filter((token, index, self) =>
+        index === self.findIndex((t) => (
+          t.symbol === token.symbol
+        ))
+      );
+    },
     // Returns true if user is logged in with GitHub, false otherwise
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/cart.js
```diff
@@ -191,6 +191,15 @@ Vue.component('grants-cart', {
...
       return result;
     },
+    fetchTokens() {
+      // removes duplicates from the tokens array of objects
+      let vm = this;
+      let uniq = {};
+
+      return vm.filterByChainId.filter(
+        obj => !uniq[obj['name']] && (uniq[obj['name']] = true)
+      );
```
gdixon, line 201: This is going to leave behind tokens which are not configured for polygon, will that present an issue? Instead of removing duplicates, could we filter by `chainId` AND `networkId`?

thelostone-mc:
what do we want to do here guys ?

chibie:
> what do we want to do here guys ?

i'd say we merge this

the alternative ux flow where a user selects the checkout option before selecting a token will also mean checkout recommendation may not be possible

cause the current flow is,

`select currency -> view recommendation -> choose checkout option`

gdixon (approved):
I'd agree here, let's get this merged and then loop in design to rethink the flow to get a solution in place to only show the valid tokens for a given chain.

PR #9695, opened 2021-11-16
GITC-389 Remove duplicates from explore page

ksolo:
##### Description

This is a quick fix for resolving the bug in GITC-389. When in append mode (the mode triggered from scrolling), before adding new grants received from the server, it will check to see if it already exists. A more robust fix can be included in the server, that will be added as a new ticket.

##### Refers/Fixes

GITC-389

##### Testing

Tested locally

### DIFF
--- a/app/assets/v2/js/grants/index.js
+++ b/app/assets/v2/js/grants/index.js
@@ -126,6 +126,8 @@ if (document.getElementById('grants-showcase')) {
       editingCollection: false,
       createCollectionRedirect: false,
       activeTimeout: null,
+      scrollTriggered: false,
+      previouslyLoadedGrants: {},
       selectOptions: [
         {group: 'Discover', label: null},
         {label: 'Weighted Shuffle', value: 'weighted_shuffle'},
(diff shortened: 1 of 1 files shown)


Jer-Sch (approved):

app/assets/v2/js/grants/index.js
```diff
@@ -320,8 +320,13 @@ if (document.getElementById('grants-showcase')) {
         if (!append_mode) {
           vm.grants = [];
         }
+
+        const previouslyLoadedIds = vm.grants.map(grant => grant.id);
```
gdixon, line 324: Consider: moving this to the vue model to avoid creating the mapping every time we call `fetchGrants()`

app/assets/v2/js/grants/index.js
```diff
@@ -320,8 +320,13 @@ if (document.getElementById('grants-showcase')) {
...
+        
         getGrants.grants.forEach(function(item) {
-          vm.grants.push(item);
+          if (!previouslyLoadedIds.includes(item.id)) {
```
gdixon, line 327: If `previouslyLoadedIds` was stored as an object against the viewModel we could make this check O(1) instead of O(n) and avoid the need to map `vm.grants` into `previouslyLoadedIds` each call...

```suggestion
          if (!vm.previouslyLoadedIds[item.id]) {
              vm.previouslyLoadedIds[item.id] = item;
```

gdixon (changes requested):

gdixon (approved):
LGTM! 🚀

PR #9699, opened 2021-11-18
fix: ensure that an invalid tezos txhash will not halt the sync command

gdixon:
##### Description

This PR will ensure the sync command will continue even when presented with a malformed/missing (from chain) txhash

##### Refers/Fixes

Refers to: L2 issue (https://discord.com/channels/562828676480237578/910066042284429332/910686876329988147)

##### Testing

Tested locally

### DIFF
--- a/app/dashboard/sync/tezos.py
+++ b/app/dashboard/sync/tezos.py
@@ -46,7 +46,8 @@ def get_tezos_txn_status(fulfillment):
 
     tx_response = requests.get(f'{BASE_URL}/operations/{txnid}').json()
 
-    if tx_response:
+    # a valid response will return a list whereas an invalid response will return a dict
+    if tx_response and isinstance(tx_response, list) and len(tx_response) > 0:
         tx_response = tx_response[0]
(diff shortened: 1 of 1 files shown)


app/dashboard/sync/tezos.py
```diff
@@ -46,7 +46,8 @@ def get_tezos_txn_status(fulfillment):
 
     tx_response = requests.get(f'{BASE_URL}/operations/{txnid}').json()
 
-    if tx_response:
+    # a valid response will return a list whereas an invalid response will return a dict
+    if tx_response and isinstance(tx_response, list) and len(tx_response) > 0:
```
thelostone-mc, line 50: Could we switch this to ensure that `txn_status` is set to expired for this scenarios. ?
That way the txn isn't stuck in pending forever

gdixon: We can, are we sure that after a tx is made it will appear on this API instantly? If there is a delay then we risk setting non-propagated tx's to expired?

thelostone-mc (approved):

PR #9714, opened 2021-11-22
Fixing bug in grants explorer scrolling.

nutrina:
##### Description

In grants explorer not all pages are loaded when scrolling.
This change fixes the issue by removing an erroneous setting of the flag `vm.grantsHasNext = false`

##### Testing

Before fix: in the grants explorer only a limited subset of the grants are shown in total (normally the first 2 pages, 12 items). Scrolling down doesn't load all items.
After fix: all the grants are loaded as the user scroll down.


nutrina:
@thelostone-mc This is the fix for the bug we discussed.

gdixon (approved):
Nice catch! 🚀

thelostone-mc (approved):

OoLionardo (approved):

PR #9731, opened 2021-11-25
fix: prevent CSS injection in profile banner filename filter

TkzcM:
##### Description

This PR fix a CSS injection in profile banner filename filter. 
The old filter can be bypassed with the payload below:
```
Content-Disposition: form-data; name="banner"

/static"); background-image: url(**Image Link Here**); display: (/wallpapers/2021_795.png
``` 
Any CSS without double quote could be injected by using the payload.

An injected example: https://gitcoin.co/tkzcm

##### Testing

The new filter should be fine while the banners are only from the wallpapers folder.

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -5780,7 +5780,7 @@ def change_user_profile_banner(request):
     try:
         profile = profile_helper(handle, True)
         is_valid = request.user.profile.id == profile.id
-        if filename[0:19] != '/static/wallpapers/' or filename.split('/')[-1] not in load_files_in_directory('wallpapers'):
+        if filename[0:19] != '/static/wallpapers/' or filename[19:] not in load_files_in_directory('wallpapers'):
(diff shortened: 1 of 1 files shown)


nutrina (changes requested):
This is a nice catch. 
But there are still bad things slipping through, for example this:

```
Content-Disposition: form-data; name="banner"\r\n\r\n/static/wallpapers/"); background-image: url(**Image Link Here**); display: (/wallpapers/2021_365.png\r\n- ...
```

I wold propose changing that if condition to following:

```
if filename[0:19] != '/static/wallpapers/' or filename[19:] not in load_files_in_directory('wallpapers'):
```

TkzcM:
Sorry about the insecure fix. It should be fine now.

nutrina (approved):
Looks good to me ;)

thelostone-mc (approved):

PR #9733, opened 2021-11-25
GITC-616: fetch gas price from polygon gas tracker

chibie:
##### Description

This pr fetches the gas price from polygon gas tracker as opposed to hardcoding it.

##### Testing

tested locally


app/assets/v2/js/cart-ethereum-polygon.js
```diff
@@ -381,6 +382,14 @@ Vue.component('grantsCartEthereumPolygon', {
 
       let gasLimit = 0;
 
+      // fetch gas prices from polygon gas tracker
+      if (!this.polygon.gasPrices) {
+        const priceUrl = 'https://api.polygonscan.com/api?module=gastracker&action=gasoracle&apikey=I28K1DVQAWAISBSI146I71YQDBK6N1C9GJ';
```
gdixon, line 387: We might want to switch this out for a paid account or move this to the backend and periodically update it? This is going to get hammered if we call it every time a user hits the checkout.

gdixon, line 387: We might want to switch this out for a paid account or move this to the backend and set up a cron to store and update the `safeGasPrice` periodically.

This is going to get hammered if we call it every time a user hits the checkout.

thelostone-mc: I'd say set a high default price and wrap this in a try catch for now

nutrina: Shouldn't we load the api key from the settings?

thelostone-mc: Oh that is a nice catch ! @chibie i agree with nutrina here

gdixon: Yeah we should really be protecting that key - I would recommend moving this to the backend and returning the `safeGasPrice` as part of the context.

chibie: @gdixon pushed a commit https://github.com/gitcoinco/web/pull/9733/commits/c2cfb70b9e1bb1f0a8bd3a58ff70de63c1f30aab

we should run `./manage.py fetch_gas_prices` after deploy

gdixon (approved):
Looks great @chibie, nicely done 🙌

thelostone-mc (approved):

Issue #9739, opened 2021-11-28
Quest - User not prompted to claim rewards after the message announcing the user won the quest


Sifou13:
**Describe the bug**
When winning a quest, a number of controls get rendered so that the user can take actions such as claiming the reward for this quest (Kudos NFTs), however this seems to not be for some of the quests, so it could be intermittent but I believe there is an "obvious" and logical explanation to this (non defined javascript variable referenced).

**To Reproduce**
Win a quest - However, while I can see the bug on the source code (JS), it seems that a roll back of some sort may have happened in the past couple of hours since it seems to be back on now.

**Expected behavior**
User should be able to complete the quest by claiming their reward and be provided with a number of options upon winning the Quest..

**Desktop (please complete the following information):**

- OS: Win 10
- Browser: Brave
- Browser Version: Version 1.32.113 Chromium: 96.0.4664.45 (Official Build) (64-bit)

Upon debugging the js in the browser, there is a "span" variable, and its only intended usage seems to have a mis-typed $ in front, which breaks js since "$span" does not exist.

Sifou13:
Closed, in line with Pull Request (see above)

gdixon:
Thank you @Sifou13!!

Issue #9738, opened 2021-11-28
Individual Quests Intermittently Unable to Progress passed Intro - Browser's Console Error


Sifou13:
Describe the bug:
When landing on a Quest's landing page, at the end of the intro, or as soon as the user clicks "Skip Intro", the page stalls, and nothing more happens, therefore the button to start the quest is not rendered.

To Reproduce:
Start the quest https://gitcoin.co/quests/2/pitch-your-ethereum-killer-to-the-vc-bro, and if you are using a refreshed version of GitCoin's website (not cached) from the past 24 hours, you should be able to reproduce.

Expected behavior:
When the intro finishes, or when the user clicks "Skip Intro", the button to start the quest should appear on the page so that the user can progress to the start page of the quest (the one providing the link to prepare for the quest).

**Screenshots**

![image](https://user-images.githubusercontent.com/3868034/143724823-61219922-1daf-4b90-aba7-90668826bc69.png)

I believe that there may have been a roll back of the deployed version, since I am unable to reproduce in the last hour and while github has not had any

**Desktop (please complete the following information):**

- OS: Windows
- Browser: Brave/Chrome
- Browser Version:  Brave - Version 1.32.113 Chromium: 96.0.4664.45 (Official Build) (64-bit)

Sifou13:
Closed, in line with Pull Request's comment (see above)

PR #9740, opened 2021-11-28
Bug Fixes: Users cannot make progress during different quest stages

Sifou13:
This pull request is addressing two issues that were introduced following some refactoring and XSS fixing in the past few days.

#9738 - Individual Quests Intermittently Unable to Progress passed Intro - Browser's Console Error: Changed line 184

#9739 - Quest - User not prompted to claim rewards after the message announcing the user won the quest: Changed Line 318

the code changes are obvious since they both concerns variables declaration that does not match their intended references (the declaration has dollar sign in front of name when references do not have the dollar symbol).

### DIFF
--- a/app/assets/v2/js/pages/quests.quest.quiz_style.js
+++ b/app/assets/v2/js/pages/quests.quest.quiz_style.js
@@ -181,7 +181,7 @@ var advance_to_state = async function(new_state) {
     typeWriter();
     await wait_for_typewriter();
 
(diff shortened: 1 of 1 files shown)


gdixon:
Thanks for this @Sifou13 - We got these changes in to `stable` a couple of hours ago, but thank you for taking the initiative to solve and raise a PR, we really appreciate it!

PR #9808, opened 2021-12-02
Gitc 575 checkout qa

nutrina:
##### Description

Design improvements for cart checkout.

##### Refers/Fixes

GITC-575 (except the error topic, which will be addressed separately)

##### Testing

According to DOD from Jira task.


thelostone-mc:
Hmm i'm not super keen on having the ETH checkout component sit in cart.js instead cart/eth.js
@gdixon any thoughts here ?

app/assets/v2/js/cart.js
```diff
@@ -27,9 +27,20 @@ const bulkCheckoutAddress = '0x7d655c57f71464B6f83811C55D84009Cd9f5221C';
...
 let grantData = []; // data for grants in cart, initialized in mounted hook
 
+Vue.component('eth-checkout-button', {
+  delimiters: [ '[[', ']]' ],
+  template: '#eth-checkout-template',
+  props: [ 'maxCartItems', 'network', 'isZkSyncDown', 'donationInputs', 'onPolygonUpdate', 'currentTokens', 'grantsByTenant',
+    'grantsUnderMinimalContribution', 'isCheckoutOngoing', 'standardCheckout', 'isPolygonDown', 'onZkSyncUpdate' ],
+  data: function() {
+    return {
+    };
```
thelostone-mc, line 39: why does this return an empty object?

app/grants/templates/grants/cart-vue.html
```diff
@@ -45,169 +46,426 @@
...
+                      </div>
+                    </template>
+                  </b-tabs>
+                </div>
               </div>
+              {% include 'shared/create_collection_modal.html' %}
+            </div>
+            
+            <!-- Sticky checkout for ETH -->
+            <div v-if="grantsCountByTenant.ETH !== undefined" class="bg-white border-grey border-top-2 bottom-0 px-2 py-4 position-sticky text-center w-100 d-block d-md-none">
```
thelostone-mc: This seems very messy :/ cause now we are splitting the code in eth.html and cart-vue.html.
Can this not be kept eth.html. ?

nutrina: @thelostone-mc I was not able to solve this by keeping the code in one location. The challenge for me was to render the checkout button differently on mobile (make it sticky on mobile) and desktop, as it was designed, and I could not come up with a better solution. 
I basically needed to render it in different location, and took the approach of creating a vue component out of the button so that it is reusable, and I can render it in different places.
What is not ideal in the approach I took is that the component template was moved to cart-vue.html. Having it in a different dedicated file would have made more sense.

app/assets/v2/js/cart.js
nutrina: @thelostone-mc this component has no state. I think we might as well drop it.

nutrina:
I have fixed conflicts, and have also fixed a bug that I have introduced (vue events where not propagated)
It should be fine now.

gdixon (approved):
Tested locally, all LGTM!! 🚀

app/assets/v2/js/cart.js
thelostone-mc: ^ any update on this. ?

PR #9814, opened 2021-12-02
fix: zksync - token id mismatch should result in a failed transaction

gdixon:
##### Description

If a transaction is presenting itself to have been made in one token, but was actually made in a different token, then we should mark it as a failed tx.

##### Refers/Fixes

Refs: grant contribution miscalculation thread in l2

##### Testing

Untested - can we verify the code before testing this against a single contribution on the cron box?

### DIFF
--- a/app/grants/models/contribution.py
+++ b/app/grants/models/contribution.py
@@ -195,19 +195,41 @@ class Contribution(SuperModel):
                     raise Exception('Unsupported zkSync transaction hash format')
                 tx_hash = self.split_tx_id.replace('sync-tx:', '0x') # replace `sync-tx:` prefix with `0x`
 
+                # First we get a list of zkSync tokens (TODO: this should be fetched with a management command and be cached)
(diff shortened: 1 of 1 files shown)


nutrina (approved):
LGTM

chibie (approved):

app/grants/models/contribution.py
```diff
@@ -195,19 +195,53 @@ def update_tx_status(self):
...
                 tx_hash = self.split_tx_id.replace('sync-tx:', '0x') # replace `sync-tx:` prefix with `0x`
 
+                # First we get a list of zkSync tokens (TODO: this should be fetched with a management command and be cached)
+                tokens_url = 'https://rinkeby-api.zksync.io/api/v0.1/tokens' if network == 'rinkeby' else 'https://api.zksync.io/api/v0.1/tokens'
+                r = requests.get(tokens_url)
+                r.raise_for_status()
+                token_data = r.json() # zkSync token data
+                tokens = {}
+                for token in token_data:
+                    tokens[token.symbol] = token.id
```
thelostone-mc, line 205: This is not valid :P it has to be token['id']

nutrina: Good catch @thelostone-mc

PR #9819, opened 2021-12-03
jeremy/highest match amount

Jer-Sch:
##### Description

This PR adds a `Highest Match Amount` sort option to the grants explorer. This option used to be `Match Estimate` and functions the same. It has just been renamed.

##### Testing

Cypress tests were updated to reflect this change.


gdixon (approved):
Looks good @Jer-Sch! 👍

PR #9827, opened 2021-12-03
GITC-650: inform users cap reached

nutrina:
##### Description

Displaying tooltips for caped grants

##### Refers/Fixes

GITC-650

##### Testing

Have tested locally, according to the acceptance criteria

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -1645,8 +1645,11 @@ Vue.component('grants-cart', {
             this.$set(this.grantData[i], 'grant_donation_amount_usd', amount);
 
             const matchAmount = await this.predictCLRMatch(grant, amount);
+            const clr_prediction_curve_2d = grant.grant_clr_prediction_curve;
+            const has_reached_cap = clr_prediction_curve_2d[0][1] !== 0 && clr_prediction_curve_2d[1][2] == 0 && clr_prediction_curve_2d[2][2] == 0 && clr[3][2] == 0 && clr_prediction_curve_2d[4][2] == 0 && clr_prediction_curve_2d[5][2] == 0;
 
             this.$set(this.grantData[i], 'grant_donation_clr_match', matchAmount ? matchAmount.toFixed(2) : 0);
+            this.$set(this.grantData[i], 'has_reached_cap', has_reached_cap);
           }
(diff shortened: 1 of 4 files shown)


gdixon:
Testing:

Screenshot 2021-12-03 at 18 43 48

Screenshot 2021-12-03 at 18 57 08

app/assets/v2/js/cart.js
```diff
@@ -257,7 +257,7 @@ Vue.component('grants-cart', {
       return {
         'token': donationToken,
         'total': totalEstimatedMatch.toFixed(2),
-        'total_str': totalEstimatedMatch.toFixed(2).toString() + ' ' + 'DAI'
+        'total_str': totalEstimatedMatch.toFixed(2).toString() + ' DAI'
```
gdixon, line 260: This was causing eslint to fail

app/grants/templates/grants/cart/eth.html
```diff
@@ -117,8 +117,15 @@
                     + 0 DAI
                   </span>
                   <span class="text-muted">
-                    <template v-if="grant.grant_donation_clr_match < 1000">estimated match</template>
+                    <template v-if="grant.grant_donation_clr_match < 1000">estd. match</template>
```
gdixon, line 120: If we leave this as `estimated` then it causes an overflow

gdixon (approved):
Awesome job, thank you @nutrina! 🚀

willsputra:
LGTM!

thelostone-mc (approved):

frankchen07:
What happens if a grant is in two matching pools and one cap is hit and another is not? Will the tooltip still notify the user?

gdixon:
In the situation you describe, it wouldn't be shown. We're using the summation of all clr calcs for a given grant to feed this state.

gdixon:
In the situation you describe, it wouldn't be shown, no. 

We're using the summation of all clr calcs for a given grant to feed this state, if any matching is still available then I'm not sure it would make sense to notify the user here? Although the matching might be lower than it would have been without the cap - the users contribution would still receive matching, so if we was to show the tooltip then at the very least we would have to rethink the messaging.

I think this could be something we rework for GR13?

PR #9828, opened 2021-12-03
refactor query

thelostone-mc:
##### Description

refactors ActivityIndex DB calls by allowing to ORM to handle the join as opposed to prepping a flatlist

### DIFF
--- a/app/dashboard/models.py
+++ b/app/dashboard/models.py
@@ -479,8 +479,9 @@ class Bounty(SuperModel):
     @property
     def latest_activity(self):
         # request more activityIndex items than we need to account for hidden/rinkeby/etc activity
-        activity_indexes = ActivityIndex.objects.filter(key=f'bounty:{self.pk}').values_list('activity__pk', flat=True)[0:10]
-        activity = Activity.objects.filter(pk__in=list(activity_indexes)).order_by('-pk')
+        activity = Activity.objects.filter(
+            activities_index__key=f'bounty:{self.pk}'
+        ).order_by('-pk')[0:10]
         if activity.exists():
             from dashboard.router import ActivitySerializer
             return ActivitySerializer(activity.first()).data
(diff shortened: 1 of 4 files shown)


app/dashboard/models.py
```diff
@@ -4310,20 +4311,17 @@ def get_various_activities(self):
...
         else:
             # orgs
-            activity_indexes = ActivityIndex.objects.all().values_list('activity__pk', flat=True)
             url = self.github_url
             all_activities = Activity.objects.filter(
-                Q(pk__in=list(activity_indexes)) |
                 Q(bounty__github_url__istartswith=url) |
                 Q(tip__github_url__istartswith=url)
-            )
+            ).exclude(activities_index__key__isnull=True)
```
gdixon, line 4324: Oh nice, this will clear everything from `Activity` if its not in the `ActivityIndex`?

app/retail/views.py
```diff
@@ -742,84 +744,62 @@ def results(request, keyword=None):
     context['avatar_url'] = static('v2/images/results_preview.gif')
     return TemplateResponse(request, 'results.html', context)
 
-def get_specific_activities(what, trending_only, user, after_pk, request=None, page=None, page_size=10):
+def get_specific_activities(what, trending_only, user, after_pk, request=None, page=1, page_size=10):
```
gdixon, line 747: Are we sure setting pagination here wont limit any queries where we're expecting more results to be returned? Just wanting to sanity check it!

app/retail/views.py
```diff
@@ -742,84 +744,62 @@ def results(request, keyword=None):
...
-            end_index = page * page_size
-            activity_pks = activity_pks[start_index:end_index].values_list('activity_id', flat=True)
-        else:
-            activity_pks = activity_pks.values_list('activity_id', flat=True)
+    if not activities:
+        activities = Activity.objects.exclude(activities_index__key__isnull=True)
 
     # Cross-ref the activity_pks->activity_id with the Activity objects
-    activities = Activity.objects.filter(pk__in=list(activity_pks),hidden=False).order_by('-created_on')
+    activities = activities.filter(hidden=False).order_by('-created_on')[start_index:end_index]
```
gdixon, line 802: `activities` is a list on ln 795 - we would want to wrap this to make sure `activities` can be filtered before applying this `filter`

gdixon (commented):
Nicee!! Just a couple of questions and one thing we might need to take a look at!

app/dashboard/models.py
thelostone-mc: Yup yup I ran this on gitcoin4 to check

app/retail/views.py
thelostone-mc: well without pagination we fetch the entire results which we want to avoid.
I've done a sanity check and it looks like there isn't
I say let's get this in and then wait and see if something breaks  -> we fix to ensure we get paginated requests always

app/retail/views.py
thelostone-mc: nice catch !

app/retail/views.py
```diff
@@ -792,7 +792,7 @@ def get_specific_activities(what, trending_only, user, after_pk, request=None, p
 
     # single activity
     if 'activity:' in what:
-        activities = [what.replace('activity:', '')]
+        activities = Activity.objects.filter(activities_index__key=what.replace('activity:', ''))
```
gdixon, line 795: Ahhh nice!! 🚀

gdixon (approved):

PR #9841, opened 2021-12-06
match estimate in grant checkout emails

owocki:
##### Description

match estimate in grant checkout emails

##### Refers/Fixes

common sense

##### Testing

tested locally
screencapture-localhost-8000-administration-email-thank-you-for-supporting-2021-12-06-15_10_09

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ WORKDIR /usr/src
 RUN apt-get update && apt-get install -y wget
 RUN wget "https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb"
 RUN apt-get install -y $CHROME_DEPS
-#RUN dpkg -i google-chrome-stable_current_amd64.deb
+RUN dpkg -i google-chrome-stable_current_amd64.deb
 
 # Install cypress dependencies
 RUN apt-get install -y $CYPRESS_DEPS
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ WORKDIR /usr/src
 RUN apt-get update && apt-get install -y wget
 RUN wget "https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb"
 RUN apt-get install -y $CHROME_DEPS
-RUN dpkg -i google-chrome-stable_current_amd64.deb
+#RUN dpkg -i google-chrome-stable_current_amd64.deb
 
 # Install cypress dependencies
 RUN apt-get install -y $CYPRESS_DEPS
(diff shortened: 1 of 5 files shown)


app/grants/models/subscription.py
```diff
@@ -186,6 +186,14 @@ class Subscription(SuperModel):
     def negative(self):
         return self.is_postive_vote == False
 
+    @property
+    def match_amount(self):
+        return 4.4 # TODO: actually calc it
```
owocki, line 191: s there a code sample that shows how to get the match estimate for a given contribution?
eg calculates it based upon the grants curve?
or should i wrote some code that does linear interpolation using hte match curve?

app/grants/models/subscription.py
```diff
@@ -186,6 +186,14 @@ class Subscription(SuperModel):
...
+
+    @property
+    def match_amount_token(self):
+        return 'DAI'
```
owocki, line 195: will match amounts always be DAI?  would be nice to future proof this in case we ever do matching in diff tokens

thelostone-mc: it would always be a stable coin! for the forseable future so i'd say this is ok

app/grants/models/subscription.py
thelostone-mc: you could do a linear curve

the estimation we have in DB is 
at 0 DAI => 12 DAI 
1 DAI => 40 DAI
10 DAI => 80 DAI

So if the user sends 5 DAI, we'd simply do a linear curve and write the prediction amount

Dockerfile
```diff
@@ -28,7 +28,7 @@ WORKDIR /usr/src
 RUN apt-get update && apt-get install -y wget
 RUN wget "https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb"
 RUN apt-get install -y $CHROME_DEPS
-RUN dpkg -i google-chrome-stable_current_amd64.deb
+#RUN dpkg -i google-chrome-stable_current_amd64.deb
```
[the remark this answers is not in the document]
thelostone-mc: Oh me and @nutrina had the same issue here.

app/grants/models/subscription.py
thelostone-mc: @nutrina this might be worth picking up if you've got time

gdixon: I've pushed these changes 👍

Dockerfile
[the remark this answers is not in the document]
gdixon: Hmmm - this is for cyprus! Can we install it from anywhere else @ksolo?

gdixon (approved):
LGTM!! 🚀

PR #9846, opened 2021-12-07
fix: apply the cap after normalising the distribution

gdixon:
##### Description

This PR will apply the matching_cap post normalisation (recursively)

##### Refers/Fixes

Refs: match-cap thread

##### Testing

Tested locally


thelostone-mc (approved):

Tomek00177 (commented):
Oezelo

PR #9852, opened 2021-12-08
Fixing display of total when token chosen by user is not DAI

nutrina:
##### Description

Fixing display of total when token chosen by user is not DAI


thelostone-mc (approved):

nutrina:
@thelostone-mc there was a bug, I have pushed a fix

Issue #9854, opened 2021-12-08
Checkout with Polygon BigNumber overflow error


Epick362:
**Describe the bug**

I am trying to checkout with Polygon. My first tx failed (can provide tx hash in DM) and upon filling my cart again and clicking the "Checkout with Polygon" I get an error from ethersjs library. Error seems to be due to incorrectly parsing my MATIC balance?

**To Reproduce**

Not sure if this is replicable for every Polygon checkout or if its related to my previous tx failing.

**Expected behavior**

Metamask opens for me to confirm the tx.

**Screenshots**

![image](https://user-images.githubusercontent.com/1021662/145180883-4640301a-4ca3-4713-91e9-9bd82717df4a.png)

![image](https://user-images.githubusercontent.com/1021662/145181066-c78a3e0f-3c5a-463e-89a2-f5215d7c3c27.png)

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Brave
- Browser Version: Version 1.31.88 Chromium: 95.0.4638.69 (Official Build) (x86_64)

sebala62:
Same issue with Metamask on iOS.

PR #9862, opened 2021-12-09
fix: prevents bigNumber overflow

gdixon:
##### Description

This PR will ensure that wallets which contain a large number of tokens will not throw a bigNumber overflow error

##### Refers/Fixes

Fixes: #9854

##### Testing

Tested locally


nutrina (approved):
Looks good

PR #9867, opened 2021-12-09
Fix fragility in Cypress tests

Jer-Sch:
##### Description

Refactors fragile Cypress tests to avoid unintentional test failures in future


gdixon (approved):
LGTM! Thanks @Jer-Sch!! 🚀

chibie (approved):

PR #9870, opened 2021-12-10
allow disabling checkout via admin

thelostone-mc:
##### Description

This PR allows admin to toggle checkout options using Static JSON store as opposed to relying on a deploy every single time. The Static JSON has already been created in admin

This takes away the need for engineering to be involved and have operations toggle switches 

##### Why are we doing this

**==================
Message from zksync team :
==================**
Tomorrow (most likely in 8-12 hours) we need to update the server. Unfortunately I can not yet tell the exact time when we will be able to start, it may cause the downtime for a few minutes. Most likely users won't feel any effect, but I think you should know in case there will be support requests on that matter. We are sorry for the short notice 😦 
Thank you for understanding!

Untitled

### DIFF
--- a/app/assets/v2/js/cart-ethereum-polygon.js
+++ b/app/assets/v2/js/cart-ethereum-polygon.js
@@ -467,7 +467,7 @@ Vue.component('grantsCartEthereumPolygon', {
         const tokenIsMatic = tokenDetails && tokenDetails.name === 'MATIC';
 
         // Check user matic balance against required amount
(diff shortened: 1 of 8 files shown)


nutrina (approved):
LGTM

nutrina:
The changes look good, but I was wondering if it would make sense to also display a message to the users, to let them know what is happening.
I think we should avoid frustrating users by removing/hiding features and not providing any feedback in the UI why something was disabled.

gdixon (approved):

PR #9890, opened 2021-12-16
remove using inkscape

thelostone-mc:
##### Description

- removed fix_broken_avatars.py
- retire use of inkscape which creates svg in /static/tmp
- update useage of functions to remove extra params

@nutrina 

This has been locally tested (couldn't find anything breaking)


app/avatar/utils.py
```diff
@@ -572,11 +572,6 @@ def svg_to_png(svg_content, width=100, height=100, scale=1, index=None, prefer=N
     png = None
     if not prefer or prefer == 'pyvips':
         png = svg_to_png_pyvips(svg_content, scale=scale)
-    if not png:
-        if not index:
-            index = random.randint(1000000, 10000000)
-        print("failed; using inkscape")
-        return svg_to_png_inkscape(svg_content, height=height, width=width, index=index, extra_flags=extra_flags)
```
nutrina, line 579: @thelostone-mc 
This assumes we should always go with the `svg_to_png_pyvips` method of generating PNGs.
In this case we should remove the `if` in line 573

nutrina (approved):
LGTM

gdixon (approved):
LGTM! Everything is working as it should on my local 🙌  - we can also remove `inkscape` from the `Dockerfile` 👍

PR #9910, opened 2021-12-23
update multisig address

thelostone-mc:
##### Description

Untitled

### DIFF
--- a/app/assets/v2/js/cart-ethereum-zksync.js
+++ b/app/assets/v2/js/cart-ethereum-zksync.js
@@ -14,12 +14,14 @@ Vue.component('grantsCartEthereumZksync', {
     grantsByTenant: { type: Array, required: true }, // Array of grants in cart
     maxCartItems: { type: Number, required: true }, // max number of items in cart
     network: { type: String, required: true }, // web3 network to use
+    networkId: { type: String, required: true }, // web3 chainId to use
     grantsUnderMinimalContribution: { type: Array, required: true } // Array of grants under min contribution
   },
 
   data: function() {
     return {
       ethersProvider: undefined,
+      stopWaitingForNetwork: null,
 
       zksync: {
         checkoutManager: undefined, // zkSync API CheckoutManager class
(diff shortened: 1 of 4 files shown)


nutrina:
@gdixon @thelostone-mc 
Here is where the polygon net id is kind of "set by brute force" ... this is why the fix does not work ...

image

nutrina:
@gdixon @thelostone-mc as far as I have tested, when you are on polygon network, and attempt to to checkout on zkzync the correct address used for the gitcoin grant. 
This however is a bit messy ....

app/assets/v2/js/cart-ethereum-zksync.js
```diff
@@ -111,6 +119,13 @@ Vue.component('grantsCartEthereumZksync', {
       appCart.$refs.cart.handleError(e);
     },
 
+    // Wait for correct chainId
+    waitForCorrectChain() {
+      return new Promise((resolve, _reject) => {
+        this.stopWaitingForNetwork = resolve;
```
gdixon, line 125: So we create a new `promise` every time `onChangeHandler` is called and then we resolve it when `networkId` is called - but to what end? I'm struggling to understand what this is trying to do? Couldn't we just rely on `networkId` as a global?

app/assets/v2/js/cart.js
```diff
@@ -375,6 +374,12 @@ Vue.component('grants-cart', {
       return isPolygon ? 'MATIC' : 'ETH';
     },
 
+    gitcoinAddress() {
+      let isPolygon = this.networkId === '80001' || this.networkId === '137';
```
gdixon, line 378:
```suggestion
      let isPolygon = networkId === '80001' || networkId === '137';
```

app/assets/v2/js/cart-ethereum-zksync.js
nutrina: @gdixon  We ar enot waiting for the `networkId`, we are actually waiting for the change of networkId to be propagated to the `cart.js` which in turn will update the address in the Gitcoin grant.
The way it is build is misleading I have to admit ... 
Maybe the cleanest way to handle is would be to update the address of the Gitcoin grant in `cart-ethereum-zksync.js`, in the checkout function. Then this logic would not span across multiple files at least?

app/assets/v2/js/cart.js
nutrina: @gdixon : I do not think this will work. You want to use the data attribute `networkId` of the vue component?

nutrina:
@thelostone-mc @gdixon I would suggest an alternative approach to solve this topic, therefore I have created this PR: #9917
Nevertheless, I think some functionality which is used for both chains (ethereum and polygon) could be moved to a utils.js file, for clarity.

PR #9914, opened 2021-12-29
Update BSci script

danlessa:
##### Description

1. Refactored the `bsci_script` method for correctly importing the squelches from the BSci pipeline
2. Added a `ToggleUser` class and `FlaggingScriptOutput` type for describing the `bsci_script` output
3. Refactored `process_bsci_sybil_csv` so that toggling the sybil users happen outside the `bsci_script`

### DIFF
--- a/app/grants/tasks.py
+++ b/app/grants/tasks.py
@@ -14,7 +14,8 @@ from celery import app
 from celery.utils.log import get_task_logger
 from dashboard.models import Profile
 from grants.models import Grant, GrantCLR, GrantCollection, Subscription
-from grants.utils import bsci_script, get_clr_rounds_metadata, save_grant_to_notion
+from grants.utils import bsci_script, get_clr_rounds_metadata, save_grant_to_notion, toggle_user_sybil
+
 from marketing.mails import (
     new_contributions, new_grant, new_grant_admin, notion_failure_email, thank_you_for_supporting,
 )
@@ -456,4 +457,5 @@ def process_bsci_sybil_csv(self, file_name, csv):
     csv = StringIO(csv.read().decode('utf-8'))
 
     # run bsci script
-    bsci_script(csv)
+    (sybil_users, non_sybil_users) = bsci_script(csv)
+    toggle_user_sybil(sybil_users, non_sybil_users)
--- a/app/grants/utils.py
+++ b/app/grants/utils.py
@@ -32,6 +32,8 @@ from django.utils import timezone
 
 import numpy as np
 import pandas as pd
+from typing import Tuple, List, TypedDict, Union
+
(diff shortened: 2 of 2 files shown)


app/grants/utils.py
```diff
@@ -392,88 +395,73 @@ def toggle_user_sybil(sybil_users, non_sybil_users):
...
+        df.loc[labels_by_prediction, 'is_sybil'] = df[labels_by_prediction].prediction_score
+        df.loc[labels_by_prediction, 'label'] = "ML Prediction"
+        
+        # Generate dict records
+        sybil_records = df.query('is_sybil == True').to_dict('records')
+        non_sybil_records = df.query('is_sybil == False').to_dict('records')
+        
+        sybil_records = [ToggleUser(**d) for d in sybil_records]
+        non_sybil_records = [ToggleUser(**d) for d in non_sybil_records]
+
```
gdixon, line 458: Could we add back the call to `toggle_user_sybil(sybil_users, non_sybil_users)`:

```suggestion
                toggle_user_sybil(sybil_records, non_sybil_records)
                
```

gdixon: Or alter [app/grants/tasks.py -> process_bsci_sybil_csv()](https://github.com/gitcoinco/web/blob/5db13f06b7b3dedaa683f504f045006d7fe95b8a/app/grants/tasks.py#L443) to make this call with the return data.

danlessa:
I've updated so that `tasks.py` is calling the `toggle_user_sybil` with the `bsci_script` output

app/grants/tasks.py
```diff
@@ -7,6 +7,7 @@
 from django.conf import settings
 from django.utils import timezone
 from django.utils.text import slugify
+from app.grants.utils import toggle_user_sybil
```
gdixon, line 10: Can we combine this with the other `grant.utils` import? (on ln 17/18)

thelostone-mc:
- I'm not very clear on what this change here is. Could we update the description ?
- Was this tested locally ? If so 
  - did the S3 upload work as expected
  - do the users marke sybil -> actually get marked sybil ?
  - do the users marked non sybil -> has the sybil mark removed ?

thelostone-mc (commented):
- I'm not very clear on what this change here is. Could we update the description ?
- Was this tested locally ? If so 
  - did the S3 upload work as expected
  - do the users marke sybil -> actually get marked sybil ?
  - do the users marked non sybil -> has the sybil mark removed ?

danlessa:
> * I'm not very clear on what this change here is. Could we update the description ?

Done!

> * Was this tested locally ? If so
>   
>   * did the S3 upload work as expected
>   * do the users marke sybil -> actually get marked sybil ?
>   * do the users marked non sybil -> has the sybil mark removed ?

I do not have the proper environment and credentials to validate those items. I've done some restricted tests and it seems to be working ok

PR #9916, opened 2022-01-03
fix: ensure only successfull contributions are sent

thelostone-mc:
##### Description

The grant contribution summary email takes into account all contributions sent in 12 hours and sends it to the grant owner.
Currently we take into account all grants contributions (ignoring the txn status) 

It makes sense to use only successful contributions and not include contributions where the : 
- txn ID cannot be found on the chain (user does something funny / uses wrong network)
- txn ID is in a pending state

### DIFF
--- a/app/git/tests/models/test_git_cache.py
+++ b/app/git/tests/models/test_git_cache.py
@@ -1,7 +1,7 @@
 import pytest
+from faker import Faker
 from git.models import GitCache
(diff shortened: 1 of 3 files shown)


nutrina (approved):
LGTM

gdixon (approved):
LGTM! 🚀

PR #9917, opened 2022-01-03
Mfix2

nutrina:
##### Description

Update multisig address for Polygon
This is an alternative implementation for: #9910

##### Testing

https://user-images.githubusercontent.com/8137830/147915704-29f6b717-d221-43c2-9ba1-03376a3ae43d.mov

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -446,7 +446,7 @@ Vue.component('grants-cart', {
       }
 
       // Generate array of objects containing donation info from cart
-      const donations = this.grantsByTenant.map((grant, index) => {
+      const donations = this.grantsByTenant.map((grant) => {
         const tokenDetails = this.getTokenByName(grant.grant_donation_currency, isPolygon);
         let amount;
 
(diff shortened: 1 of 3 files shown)


app/assets/v2/js/cart.js
```diff
@@ -497,6 +437,76 @@ Vue.component('grants-cart', {
...
+    // Array of objects containing all donations and associated data
+    computeDonationInputs(destGitcoinAddress) {
+      let isPolygon = destGitcoinAddress == gitcoinAddressPolygon;
+
+      if (!this.grantsByTenant || this.tabSelected !== 'ETH') {
+        return undefined;
+      }
+
+      // Generate array of objects containing donation info from cart
+      const donations = this.grantsByTenant.map((grant, index) => {
```
ksolo, line 449: it doesn't look like the `index` variable is being used, we can remove it from the params list.

app/assets/v2/js/cart.js
```diff
@@ -497,6 +437,76 @@ Vue.component('grants-cart', {
...
+          grant_donation_amount: this.donationsToGitcoin[token],
+          grant_donation_clr_match: '',
+          grant_donation_currency: token,
+          grant_donation_num_rounds: 1,
+          grant_id: '12',
+          grant_image_css: '',
+          grant_logo: '',
+          grant_slug: 'gitcoin-grants-official-matching-pool-fund',
+          grant_title: 'Gitcoin Grants Official Matching Pool Fund',
+          grant_token_address: '0x0000000000000000000000000000000000000000',
```
ksolo, line 489: is there any danger in burning tokens setting this address?

chibie (approved):
lgtm!

app/assets/v2/js/cart.js
nutrina: @ksolo I do not think there is any danger. This is only the token address, and this address indicates that all tokens are accepted (judging by how this field is then used) see for example:
```
   const acceptsAllTokens = (grantData.grant_token_address === '0x0000000000000000000000000000000000000000' ||
      grantData.grant_token_address === '0x0');
```

thelostone-mc (approved):

PR #9922, opened 2022-01-05
jeremy/GITC-501/649

Jer-Sch:
##### Description

This PR resolves two separate bugs.

The first one is when a user clicks into a grant from the grants explorer, and then navigates back to the explorer, their place on the infinite scroll is lost.

The second one is when a user clicks into a grant from a collection, and then navigates back to the collection, the entire collection is not shown.

These bugs are resolved by opening the grant in new tab leaving the explorer/collection screen as is in the original tab.

https://user-images.githubusercontent.com/36638840/148140505-69faa093-8831-445b-a240-f0577b2f31d4.mov

##### Testing

This PR contains Cypress testing

### DIFF
--- a/app/grants/templates/grants/components/card.html
+++ b/app/grants/templates/grants/components/card.html
@@ -28,7 +28,7 @@
         </div>
         <div class="col-md-7 p-3">
           <h2 class="font-subheader">
-            <a :href="grant.details_url" class="gc-font-base h5 text-black">[[ grant.title | truncate(60) ]]</a>
+            <a :href="grant.details_url" class="gc-font-base h5 text-black" target="_blank" rel="noopener noreferrer">[[ grant.title | truncate(60) ]]</a>
           </h2>
           <div>
             <div class="">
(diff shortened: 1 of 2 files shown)


thelostone-mc (approved):

thelostone-mc:
Failures are due to timeouts in cypress and not due to the changes made in this PR

gdixon:
This PR introduces a regression (see here: https://gitcoin.atlassian.net/browse/GITC-302) - pagination in the url was added in an attempt to fix this issue, but re-adding `target="_blank"` feels like a step backwards?

cc. @frankchen07 @nfrgosselin @thelostone-mc

Jer-Sch:
@gdixon This makes sense. We had agreed in grants standup that opening the links in a new tab was the most effective way to resolve these bugs, but I hadn't realized that this was a regression.

gdixon:
Ahhh no worries!! Thanks everyone! I just wanted to make sure we had eyes on a long-term solution! 🙌

Jer-Sch:
@gdixon @ksolo @thelostone-mc @nfrgosselin I'll make a ticket for it so it doesn't get forgotten about.

PR #9927, opened 2022-01-06
GITC-690 allow special short twitter handles

schultztimothy:
##### Description

https://gitcoin.atlassian.net/browse/GITC-690 - This will allow input of special twitter handles that have greater than 2 characters when creating grants. The current limit is set to 4 characters which I believe is the current limit based on https://help.twitter.com/en/managing-your-account/twitter-username-rules

##### Testing

Just a regex change

### DIFF
--- a/app/assets/v2/js/grants/_new.js
+++ b/app/assets/v2/js/grants/_new.js
@@ -78,6 +78,7 @@ Vue.mixin({
         });
       });
     },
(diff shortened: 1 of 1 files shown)


thelostone-mc (approved):
LGTM. Could we update the test to have the twitter field to 2/3 characters ?

nutrina (approved):
lgtm

PR #9936, opened 2022-01-10
GITC-708 - Remove link to verify offline

ksolo:
##### Description

Removing the link to verify sms offline. The modal now looks like the attached image.

image

##### Refers/Fixes

GITC-708


ksolo:
@gdixon this one is a bounty ticket, I didn't pay attention to the label when I grabbed it until Jeremy called it out. It was super simple so I went ahead with it. I'll make sure to avoid grabbing platform tickets going forward.

gdixon (approved):

Jer-Sch:
@ksolo This seems like it could it be a good opportunity to add some Cypress tests for the user profile page. What do you think?

ksolo:
@Jer-Sch I thought about it, but for these feature being a removal it seems weird to clutter tests with things it shouldn't see. Had there been tests for this feature I would have been removing them instead of adding them.

Jer-Sch (approved):

PR #9940, opened 2022-01-12
GITC-697: ensure pending commands are always called even after failures

gdixon:
##### Description

This PR will ensure we always run `sync_pending_fulfillments` and `sync_pending_contributions` even if the previous call is still running.

##### Refers/Fixes

Fixes: GITC-697


schultztimothy (approved):
LGTM 👍

chibie (approved):

thelostone-mc (approved):

PR #9946, opened 2022-01-13
GITC-719 Have created command that kills long-runnin processes

nutrina:
##### Description

GITC-719

##### Refers/Fixes

GITC-719

##### Testing

Have started some script that just looped ...
Have killed it with my command like:
`python3 app/manage.py kill_long_running_processes 30 ".*loop.*" --exec`

- first arg: min age of processes to be killed
- regex to match the running processes
- '--exec' - flag to indicate this is not a dry run (if this is missing kill command will not be issued, but information will be printed to console). This is intended for testing without causing harm.

### DIFF
--- a/app/grants/management/commands/kill_long_running_processes.py
+++ b/app/grants/management/commands/kill_long_running_processes.py
@@ -1,0 +1,60 @@
+import io
+import os
+import re
+import subprocess
+from datetime import datetime, timedelta
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+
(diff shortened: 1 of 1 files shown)


thelostone-mc (commented):
looks like commit history is messed up! mind doing a cherrypick on a fresh branch ?
Otherwise LGTM

nutrina:
@thelostone-mc Have cherry-picked the relevant commit. It should be fine now.

thelostone-mc (approved):

gdixon (approved):
LGTM! 🚀 Do we want to get the updates to the crontab in this PR or as a follow-up?

PR #9978, opened 2022-01-21
Gitc 715/direct payout contract transactions

ksolo:
##### Description

Some grant owners will interact with the matching_contract directly instead of using the GitCoin provided UI. To keep our records consistent this command can be executed to review the `PayoutClaimed` events from the contract directly and update records where the claim_tx is missing.

The last piece of this is to schedule the command to run.

##### Refers/Fixes

GITC-715

##### Testing

Unit tests were provided


app/app/settings.py
```diff
@@ -914,7 +914,8 @@ def callback(request):
...
+with open(os.path.join(root, 'grants/match_payouts_abi.json'), 'r') as f:
```
ksolo, line 917: I had to update the ABI since there was a change to the indexing of the PayoutClaimed event in the last month. Calling `get_all_entries` for this event worked on older deployment but failed on the new deployment for GR12. We may want to consider how to version this if there is code that needs to hit older contracts.

app/app/settings.py
```diff
@@ -657,17 +657,17 @@ def callback(request):
...
-    'social_core.pipeline.user.create_user', 
-    'social_core.pipeline.social_auth.associate_user', 
+    'social_core.pipeline.user.get_username',
+    'social_core.pipeline.user.create_user',
+    'social_core.pipeline.social_auth.associate_user',
     'social_core.pipeline.social_auth.load_extra_data',
     'social_core.pipeline.user.user_details',
 # This should be executed after the '...user_details' step, as the user_details step will
-# update the username (aka github handle) which is then used in `save_profile`. This 
+# update the username (aka github handle) which is then used in `save_profile`. This
```
ksolo, line 670: `make fix` changes

app/grants/match_payouts_abi.json
```diff
@@ -0,0 +1,228 @@
+[
```
ksolo, line 1: abi current as of commit - `1313e652` on the `matching_contracts` repo.

app/grants/tests/models/test_grant_hall_of_fame.py
```diff
@@ -1,6 +1,6 @@
 import pytest
-from grants.tests.factories.grant_hall_of_fame_factory import GrantHallOfFameFactory, GrantHallOfFameGranteeFactory
 from grants.models.hall_of_fame import GrantHallOfFame, GrantHallOfFameGrantee
+from grants.tests.factories.grant_hall_of_fame_factory import GrantHallOfFameFactory, GrantHallOfFameGranteeFactory
```
Jer-Sch, line 4: Were these linter fixes?

Jer-Sch (approved):
Looks awesome

app/grants/tests/models/test_grant_hall_of_fame.py
ksolo: I think so

app/app/settings.py
thelostone-mc: Ah so this change we have right now -> works for new contracts.
I think this is alright cause 
- the older contracts funds will be drained out in a month 
- all the new rounds will continue to use new ABI

app/grants/views.py
```diff
@@ -3960,16 +3960,14 @@ def clr_matches(request):
...
+        try:
+            clr_match = CLRMatch.objects.get(pk=pk)
+        except CLRMatch.DoesNotExist:
             return Response({'message': 'CLR Match not found!'}, status=404)
 
-        # update claim tx for all grant with grant.admin_address paid out by 
+        # update claim tx for all grant with grant.admin_address paid out by
         # the clr_match contract
-        clr_matches.filter(
+        CLRMatch.objects.filter(
```
thelostone-mc, line 3970: Is this more DB efficient ?
The older was written based off this 
https://docs.djangoproject.com/en/4.0/topics/db/queries/#retrieving-specific-objects-with-filters

thelostone-mc (commented):
looking good :D will have to test this out in production 
Left a small question on the refactoring

Adding this PR to review board as well

app/app/settings.py
ksolo: Sounds like a plan to me :)

app/grants/views.py
ksolo: It will be the same performance, my intent here was to group more of the error scenarios which is looking for the specific clr_match, and then the query where we filter by the admin address and payout. I'm happy to revert this back if you preferred the previous code.

chibie: @thelostone-mc @ksolo  i think it's more efficient... as we should avoid `objects.all()` queries

thelostone-mc (approved):

PR #9348, opened 2021-08-12
GITC-245: Implements the CLR calc in SQL

gdixon:
##### Description

This PR implements the CLR calc in SQL in an attempt to improve performance.

To test:

```
# run this first to set trust-bonus into the db model
docker exec -it web_web_1 ./app/manage.py update_trust_bonus --call-now 1

# then compare the results of the following...
docker exec -it web_web_1 ./app/manage.py estimate_clr mainnet 1 full true
docker exec -it web_web_1 ./app/manage.py estimate_clr2 mainnet 1 full true
docker exec -it web_web_1 ./app/manage.py estimate_clr3 mainnet 1 full true
```

*please note, once each of the variations have been tested on production - a follow up PR will be raised to remove print statements and move the most effective variation to `clr.py`*

##### Refers/Fixes

Fixes: GITC-245

##### Testing

Tested locally

**clr.py**
Screenshot 2021-08-17 at 12 09 41
**clr2.py**
Screenshot 2021-08-17 at 12 11 23
**clr3.py**
Screenshot 2021-08-17 at 12 12 22

*note: I've deferred contributions from 1095 -> 993 - your results will be different to what you see in these screenshots*

PR #10034, opened 2022-01-25
chore: optimise CLR performance away from O(n^2)

gdixon:
##### Description

This PR:

1. caches the `sqrts` for each contributor to reduce the need to perform the same (expensive) calc multiple times

2. changes how we pair so that our time complexity comes in at something like `O((n^2)/2)`:

```
obj = {1: 2, 2: 3, 3: 4}
contributors = []
pairs = {}
for k1, v1 in obj.items():
    for k2 in contributors:
        u_k1 = k1 if k2 > k1 else k2
        u_k2 = k2 if k2 > k1 else k1
        if u_k1 != u_k2:
            if u_k1 not in pairs:
                pairs[u_k1] = {}
            if u_k2 not in pairs[u_k1]:
                pairs[u_k1][u_k2] = 0

            pairs[u_k1][u_k2] = obj[u_k1] * obj[u_k2]

    contributors.append(k1)
```

We expect `pairs = {1: {2: 6, 3: 8}, 2: {3: 12}}`, which consists of every pair multiplied together just once, this reduces the number of calcs we need to perform by 50%. We also apply this same behaviour to the `calculate_clr` function, to save the same amount of compute there.

##### Testing

Tested locally and against GR12 results

Before:
Screenshot 2022-01-25 at 12 44 04

After:
Screenshot 2022-01-25 at 13 48 43

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -20,6 +20,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 import copy
 import time
 
+from django.db import transaction
 from django.utils import timezone
 
 import numpy as np
@@ -472,7 +473,7 @@ def apply_cap(totals, match_cap_per_grant, should_spread):
 
     return totals
 
-
+@transaction.atomic
 def predict_clr(save_to_db=False, from_date=None, clr_round=None, network='mainnet', only_grant_pk=None, what='full', use_sql=False):
     # setup
     counter = 0
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -132,10 +132,10 @@ def translate_data(grants_data):
     for g in grants_data:
         grant_id = g.get('id')
         for c in g.get('contributions'):
-            profile_id = c.get('id')
+            profile_id = int(c.get('id'))
             trust_bonus = c.get('profile_trust_bonus')
             if profile_id:
-                val = [grant_id] + [c.get('id')] + [c.get('sum_of_each_profiles_contributions')]
+                val = [grant_id] + [profile_id] + [c.get('sum_of_each_profiles_contributions')]
                 grants_list.append(val)
                 trust_dict[profile_id] = trust_bonus
 
@@ -167,7 +167,7 @@ def aggregate_contributions(grant_contributions):
     return contrib_dict
 
 
-def get_totals_by_pair(contrib_dict):
+def get_totals_by_pair(curr_agg):
     '''
         gets pair totals between current round, current round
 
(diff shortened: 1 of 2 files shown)


owocki:
13min to 5min seems promising!

owocki:
i'm unfortunately not close enough to the code to be able to comment on it in a meaningful way :/

app/grants/clr.py
```diff
@@ -181,27 +182,60 @@ def get_totals_by_pair(contrib_dict):
...
+        # build inner layer against contributors
+        contributors = []
+        # cache sqrts of each contribution to make pairing easier
+        curr_agg_sqrts[proj] = {}
+        # iterate on the contributions and set default state
         for k1, v1 in contribz.items():
-            if k1 not in pair_totals:
-                pair_totals[k1] = {}
+            # cache the sqrt of the given value (v1 = total contribution to this grant by this user)
+            curr_agg_sqrts[proj][k1] = float(v1) ** 0.5
```
owocki, line 211: what if we calculate the sqrt when we save the `Contribution` and then we dont have to calculate it again every time we run clr calcs?

this assumes that the sqrt part is costly performance wise at all ( i dont know that it is )

owocki: oh wait that wont work bc u actually want the` sqrt (all of a users contributions)` not the `sum(sqrt(contribution) x num_contribs)`

app/grants/clr.py
```diff
@@ -181,27 +182,60 @@ def get_totals_by_pair(contrib_dict):
...
+                        user_id (float): sqrt(aggregated_amount) (float)
+                    }
+                }
+            pairs_tot
+                int: the number of cycles it took to construct the pairs
     '''
+    pairs_tot = 0
     pair_totals = {}
-
+    curr_agg_sqrts = {}
```
thelostone-mc, line 201: Ah we keep updating this over time and before calculating -> check if data is available here.
this is nice :D

thelostone-mc:
Pulled this in locally and tested it out for a small CLR round and the numbers seem to match !
This is <3

thelostone-mc (approved):

app/grants/clr.py
```diff
@@ -181,27 +182,60 @@ def get_totals_by_pair(contrib_dict):
...
     # start pairwise match
-    for _, contribz in contrib_dict.items():
+    for proj, contribz in curr_agg.items():
```
owocki, line 203: fwiw, i have a *hunch* that much of the aggregation (and sqrt + pairwise) code could be rewritten in SQL (stays in the DB server) and be much much much more efficient than pulling it all back from the database server + assembling it in memory on the task server + then writing it back to the db.

its just a hunch tho.

gdixon: I had the same suspicion and wrote this version: https://github.com/gitcoinco/web/blob/53707dc28231563f24fe6e3737d5c6a6a48693bb/app/grants/clr2.py

(POC here: https://dbfiddle.uk/?rdbms=postgres_9.6&fiddle=ba238edb1599780f776099ed98bcb2e4)

I couldn't prove it was more efficient than making the pairs + clr in python though :( My findings are over here: https://github.com/gitcoinco/web/pull/9348 - there could be some optimisations we could make that I've missed though?

Issue #9991, opened 2022-01-23
Delay wallet connect modal


nfrgosselin:
**User Story:**
As a grant donor
I want to see the wallet connect overlay only when I absolutely have to
So that I have a smoother checkout experience

**Acceptance Criteria**
GIVEN that I have grants in my cart and I do not have a wallet connected
WHEN I arrive on gitcoin.co/grants/cart
THEN I do not see the wallet connect overlay

GIVEN that I am on gitcoin.co/grants/cart and I do not have a wallet connected
WHEN I click a checkout option
THEN I see the wallet connect overlay

GIVEN that I am on gitcoin.co/grants/cart and I do have a wallet connected
WHEN I click a checkout option
THEN I see my wallet transaction overlay

**Product & Design Links:**
Figma: "[Ready for Story: Checkout](https://www.figma.com/file/8QjqOzDBd4HiE37YqcocK7/GR13?node-id=133%3A5208)" Section 2: Delay modal
Product Brief: [Streamline donor checkout experience](https://www.notion.so/gitcoin/Streamline-donor-checkout-experience-4c70c5926a044facbdb65105d1316a44)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

nfrgosselin:
This doesn't seem to be working for me @gravityblast. I'm on staging, don't have a wallet connected, and I see the wallet connect overlay as soon as I land on the cart.

gravityblast:
@nfrgosselin I'm not sure when it was deployed, but it was merged when I was away, I see it now in staging. 
When I open the `/grants/cart` page I don't see the modal until I start the checkout. Can you try again now please? or please DM me in case you can't still see the changes.

EmHsiehGitcoin:
hmmm, I see the staging refreshed but I also still see the modal popping up when I first land at cart

nfrgosselin:
@EmHsiehGitcoin I just cleared my cache and it worked for me. This is good from my end if it works for you!

Thanks @gravityblast 🙏

PR #10087, opened 2022-02-03
don't show web3 modal on cart page init and tab change

gravityblast:
##### Description

When the cart is initialized and when a tab is selected, the `tabChange` method is called.
This method was opening the web3 dialog by default, or waiting to connect to web3 in case it was already opened.
I removed that part so that the dialog is opened only at the final checkout action if needed. 
The final checkout action is already doing the same check, opening the dialog in case the wallet is not connected.
##### Refers/Fixes

Fixes: #9991

### DIFF
--- a/app/assets/v2/js/cart-ethereum-polygon.js
+++ b/app/assets/v2/js/cart-ethereum-polygon.js
@@ -197,6 +197,11 @@ Vue.component('grantsCartEthereumPolygon', {
 
     // Send a batch transfer based on donation inputs
     async checkoutWithPolygon() {
+      // Prompt web3 login if not connected
+      if (!provider) {
+        await onConnect();
+      }
+
       const bulkCheckoutAddressPolygon = this.getBulkCheckoutAddress();
 
       try {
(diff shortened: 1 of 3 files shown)


nutrina (approved):
LGTM

thelostone-mc (approved):

thelostone-mc:
synced up with @gravityblast. Before merging this we need to ensure 
the web3 popup happens for all 3 
- zksync
- polygon
- mainnet

And a clarification at which step it should happen

chibie:
@thelostone-mc @gravityblast it's worth noting that checkout recommendation requires wallet connection upon loading the cart page. does this change break that?

gravityblast:
> Before merging this we need to ensure the web3 popup happens for all 3
I updated the code to open the web3 modal for polygon and zksync as well.

> @gravityblast looks like we've got conflicts. once those are resolved -> remove the don't merge PR and let's get this re-reviewed (assuming it's done)

just rebased from master now

thelostone-mc (approved):
LGTMMM

gdixon (approved):
Codewise LGTM!! 🚀

thelostone-mc:
@gravityblast would you be able to resolve the conflicts?

gravityblast:
@thelostone-mc just rebased and force pushed to fix the conflicts

Issue #9985, opened 2022-01-23
Don't enable payouts if KYC is required


nfrgosselin:
**User Story:**
As a Gitcoin admin
I want to prevent certain grant owners from bypassing KYC
So that Gitcoin doesn't face legal risk

**Acceptance Criteria**
GIVEN the creation of a grants.models.clrmatch object
WHEN the KYC required flag is true
THEN the default state for is payout ready is FALSE.

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
From Owocki's original story:
when the grants.models.clrmatch object is created, it should not have "is payout ready" true IF the KYC REQUIRED flag is true.

https://bits.owocki.com/ApuEGo4v

- grant owners with more than 10k needs kyc
- there’s no way for us to block folks from claiming directly from etherscan
- folks can choose to bypass the kyc and claim directly from etherscan

PR #10090, opened 2022-02-03
chore: ensure matches with kyc are not promoted

thelostone-mc:
##### Description

Ensures grants above threshold require KYC and cannot claim until manually KYC-ed
Also ensures that both KYC pending and no kyc matches are uploaded onto contract 

##### Refers/Fixes

Fixes https://github.com/gitcoinco/web/issues/9985

##### Testing

Untitled


schultztimothy (approved):
LGTM 👍

Issue #9984, opened 2022-01-23
chore: add fixtures for grants claim


nfrgosselin:
update to include fixtures

**Description**
After the last round of fixes on the claims experience, we need to add backend tests to the claims page to ensure good test coverage. 

**Acceptance Criteria**
Backend tests written for claims page

ksolo:
@nfrgosselin I added a fair amount of tests for the claim route when working on GITC-715. It's not merged yet but we can review what additional coverage is needed

nfrgosselin:
Thanks @ksolo —> tagging @thelostone-mc in here as well. Aditya, can you take a look and think about what additional coverage you're looking for?

ksolo:
@thelostone-mc Here are the tests - https://github.com/gitcoinco/web/blob/e17d9636123e4c6a7c66137f3fbd890546b815e2/app/grants/tests/views/test_clr_matches_view_function.py

thelostone-mc:
Oh solid this is perfect.
We still need to add fixtures for this as well for those tables but i guess this is alright


@schultztimothy brought up the point on having fixutres (dummy data) and I it makes sense

PR #10093, opened 2022-02-04
added clr_match and relations fixture, created clr_match make command

schultztimothy:
##### Description

Adds clr_match fixtures and added command so that related models are easier to load

##### Refers/Fixes

fixes: #9984

##### Testing

I updated one factory so that it always produces > 0 value for donation


ksolo (approved):
LGTM! 🚀

thelostone-mc (approved):

Issue #10021, opened 2022-01-25
Contribution CSV download timing out


nfrgosselin:
At least one of our users (Richard Moore from ethers.js) is having issues downloading their contributions as a CSV through the UI. According to Richard, the download seems to be timing out on both desktop and mobile.

Richard's grant: https://gitcoin.co/grants/13/ethersjs-complete-simple-and-tiny

Steps to reproduce:
1. Impersonate @ricmoo 
2. Go to My Account → Settings
3. Select Account tab
4. Scroll down to Account Financials
5. Click ‘Export’ in Export Earnings field

**Acceptance criteria**
GIVEN I am a Gitcoin grant owner with a contribution history
WHEN I click 'Export' on the Export Earnings section in Account Financials
THEN The CSV of contribution download doesn't time out
AND I see a modal informing me that the download is being prepared and I will receive it in my email

GIVEN I am a Gitcoin grant owner with a contribution history
WHEN I click 'Export' on the Export Earnings section in Account Financials
AND The download fails
THEN I receive an email directing me to contact support, with a link to submit a new support ticket

[Figma link](https://www.figma.com/file/8QjqOzDBd4HiE37YqcocK7/GR13?node-id=41%3A5126)
[Jira link](https://gitcoin.atlassian.net/browse/GITC-727?atlOrigin=eyJpIjoiMWI3NzJhYzQxMTc0NDYyNDk3MmRlNGEzMWU4ZDhhYjQiLCJwIjoiaiJ9)

EmHsiehGitcoin:
Not sure how to test this. Can we get a screen-recording?

nfrgosselin:
+1 — @thelostone-mc is that something you can help us with?

thelostone-mc:
It should be the steps mentioned as in the ticket


Go to My Account → Settings
Select Account tab
Scroll down to Account Financials
Click ‘Export’ in Export Earnings field

thelostone-mc:
It looks like the the  POST operation is not picking up the body of the request.
https://github.com/gitcoinco/web/blob/master/app/retail/templates/settings/account.html#L186-L187

This would have to be debugged and fixed

thelostone-mc:
Fixed it up via https://github.com/gitcoinco/web/commit/07a15eefe79e82ecbafb13bf3d386a690716c027

nfrgosselin:
LGTM — thanks @thelostone-mc 

@EmHsiehGitcoin this is ready for Done once you've reviewed

PR #10094, opened 2022-02-05
Jeremy/GITC-727/contribution csv download timing out 

Jer-Sch:
##### Description

This PR resolves a bug that was occurring when users with large contribution histories tried to download their histories. Because of the size of the file, the download would time out. This PR:

- Creates a background task that processes the CSV and sends it to the user as an email attachment
- If the task fails to process the CSV, it will send the user an email directing them to contact support
- Add email templates and render methods for emails referenced above
- Adds a responsive modal to the settings/account page that appears when a user exports their data

##### Refers/Fixes

#10021 

##### Testing

This PR includes Cypress and Pytest tests

Screen Shot 2022-02-04 at 7 00 33 PM

Screen Shot 2022-02-04 at 5 14 13 PM

Screen Shot 2022-02-05 at 10 15 52 AM

Screen Shot 2022-02-05 at 10 16 14 AM


app/retail/templates/emails/export_data.txt
```diff
@@ -1,3 +1,4 @@
-@{{profile.handle}}
+Your Gitcoin CSV Download
 
-Your Exported CSV Is Attached
\ No newline at end of file
+The CSV file you requested from the Gitcoin platform is now ready for download. 
+Please download the attached file to see your data. If you have any issues, feel free to submit a support ticket at support.gitcoin.co/new/
```
ksolo, line 4:
```suggestion
Please download the attached file to see your data. If you have any issues, feel free to submit a support ticket at https://support.gitcoin.co/new/ 
```

app/retail/templates/settings/account.html
```diff
@@ -87,16 +87,16 @@
     padding-right: 40px;
   }
 
-  #export-modal-label {
+  #exportModalLabel {
```
ksolo, line 90: Is bootstrap loaded on these pages? If it is, we should use those styles here instead.

app/retail/templates/settings/account.html
```diff
@@ -343,7 +343,7 @@ <h3 class="modal-title" id="export-modal-label">{% trans "Download in progress"
 
       $(`#${this.id}`).addClass('bound')
     });
-    $("#close-modal-button").on('click', function() {
+    $("#closeModalButton").on('click', function() {
```
ksolo, line 346: I think there is a much better approach we can take here. This was submitting a form before, which we should now bind to the form submit instead of to this button. Then we should submit a request via AJAX and show the modal. There isn't a need to `unbind` the event and then trigger the event.

ksolo (changes requested):

app/marketing/tests/tasks/factories/profile_factory.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class ProfileFactory(factory.django.DjangoModelFactory):
+    """Create instance of Profile for testing."""
+
+    class Meta:
+        model = Profile
+
+    handle = factory.Sequence(lambda n: "Contributor_%03d" % n)
+    data = {}
```
ksolo, line 12: We already have this factory defined. This is part of the `dashboard.tests.factories` We should use that instead of creating a duplicate

app/retail/templates/settings/account.html
Jer-Sch: Bootstrap is loaded, but I found I needed to do a lot of styling to implement the requested design. I'll look in to which styles I can replace with bootstrap classes though. This was just the quickest way at the time.

app/retail/templates/settings/account.html
Jer-Sch: Let's pair on this next week

thelostone-mc (approved):
lgtm codewise. haven't locally tested PR though

app/retail/templates/settings/account.html
```diff
@@ -229,10 +338,28 @@ <h5 class="font-subheader">{% trans "Delete My Account" %}</h5>
         }
       });
     });
+    $("#exportEarnings").on('submit', function(e) {
+      e.preventDefault();
+      console.log('submitted')
+      // e.preventDefault();
```
schultztimothy, line 344: Maybe remove these comments?

app/marketing/views.py
```diff
@@ -559,13 +536,12 @@ def account_settings(request):
             profile.preferred_payout_address = eth_address
             profile.save()
             msg = _('Updated your Address')
+            
         elif request.POST.get('export', False):
             export_type = request.POST.get('export_type', False)
+            user_pk = request.user.pk
+            export_earnings_to_csv.delay(user_pk, export_type)
```
ksolo, line 543: This should have some type of HTTP response returning after this is queued up.

app/marketing/tests/tasks/test_export_earnings_to_csv_task.py
```diff
@@ -0,0 +1,114 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""
+
+import csv
```
schultztimothy, line 21: Looks like csv is never used

schultztimothy (commented):
LGTM! Outside of the other comments on here 🚀

app/retail/templates/settings/account.html
ksolo: @Jer-Sch In general using the css library is much better than hard-coded values since it provides consistency across the application. Since this page is likely going to be redesigned soon, it's not worth spending more time on it. We can leave these values in and when this is Vue-ified these styles can be updated during that effort.

app/marketing/tasks.py
```diff
@@ -1,20 +1,99 @@
...
+    user = User.objects.get(pk=user_pk)
+    profile = user.profile
+    earnings = profile.earnings if export_type == 'earnings' else profile.sent_earnings
+    earnings = earnings.filter(network='mainnet').order_by('-created_on')
+
+    attachment = create_csv(export_type, profile, earnings)
+    if attachment:
+        send_csv(attachment=attachment, user_profile=profile)
+    else:
+        send_download_failure_email(profile)
```
ksolo, line 94: This seems like it should be in a `try/except` block and not an `if/else` block. If something fails to generate the CSV I would expect an exception to be raised and caught here.

ksolo (approved):

app/marketing/tasks.py
```diff
@@ -1,20 +1,99 @@
...
+    html, text, subject = render_export_data_email_failed(user_profile=user_profile)
+    send_mail(
+        from_email,
+        to_email,
+        subject,
+        text,
+        html,
+        from_name=f"@{user_profile.handle}",
+        categories=['transactional'],
+    )
```
gdixon, line 80: We normally only have celery decorated methods in a `tasks.py` file, could we move these over to `marketing/utils.py`?

app/marketing/tasks.py
```diff
@@ -1,20 +1,99 @@
...
 redis = RedisService().redis
 
 rate_limit = '300000/s' if settings.FLUSH_QUEUE or settings.MARKETING_FLUSH_QUEUE else settings.MARKETING_QUEUE_RATE_LIMIT
 
+def create_csv(export_type, profile, earnings):
+    path = f'app/assets/tmp/user-{export_type}/{profile}'
+    if not os.path.isdir(path):
+        os.makedirs(path)
+    name = f"{timezone.now().strftime('%Y_%m_%dT%H')}"
+    file_path = f'{path}/{name}.csv'
```
gdixon, line 28: Should we be saving these to disk? Could we use something like `io.StringIO()` to keep these in memory?

gdixon (changes requested):
Looking really good, just a couple of nitpicks!!

app/marketing/tasks.py
Jer-Sch: @gdixon I was working on this today, but moving those methods into `marketing/utils.py` caused circular dependencies. Specifically, `marketing.utils.py` is importing `marketing.mails.py` which imports `app.utils.py` which imports `marketing.utils.py`. I tried to resolve it but I ran out of time.

thelostone-mc:
@gdixon we'll create up a follow up ticket to get this sorted out 
Would you alright having us get this in now ?

@nfrgosselin need a follow up ticket to address @gdixon comments

gdixon:
> @gdixon we'll create up a follow up ticket to get this sorted out Would you alright having us get this in now ?

Yeah that sounds good!! 🙌 Nice work here @Jer-Sch! 🚀

gdixon (approved):

PR #10104, opened 2022-02-07
[small, high priority] export grant data for ethelo integration

nollied:
##### Context
I'm part of the [gitcoin DAO FDD workstream](https://gitcoin.notion.site/Fraud-Detection-Defense-2bde13c0b8e74fda81435d94e49e2703), specifically i'm leading the development team for the Grants Investigation Agency (GIA). We're using a tool called "Ethelo" to improve the UX for our reviewers (agents). [Here is a link to GIA's first test run UI](https://gia-testing.ethelo.net/page/welcome-to-the-gia).

For the first iteration of the GIA, we copy-pasted all of the grant information from a few select grants to validate the "product market fit" (DAO PMF). Copy-pasting for GR13 would be exhaustive since ~700 grants are expected to be created. 

This PR will be used as a semi-automated integration between gitcoin.co/grants <> ethelo/GIA

##### Description

Added a link to the admin panel at route`/_administration` (bottom right of image)
Screen Shot 2022-02-07 at 12 25 17 PM

clicking this button brings up the (very boring) UI for ethelo grant extraction
Screen Shot 2022-02-14 at 10 01 24 PM

executes a database query and formats the data as the json spec that ethelo expects. then a download prompt for the client is issued. here is an example output (might be slightly outdated):
```
{
{
  "options": [
    {
      "slug": "grant_1",
      "title": "Nollied wants 1million dollars",
      "info": "<h1>GIVE ME MONEY</h1>",
      "display_data": {
        "Status": "Approved",
        "Grant Url": "https://gitcoin.co/grants/1/nollied-wants-1million-dollars",
        "Github Project Url": "https://github.com/nollied",
        "Location": "north_america",
        "Wallet Address": "0x0",
        "Twitter": "@nolliechain",
        "Creator Handle": "owocki",
        "Database Number": 1
      }
    }
  ]
}
```


app/dashboard/ethelo.py
```diff
@@ -0,0 +1,51 @@
...
+        info = grant.description
+
+    return {
+        "slug": f"grant_{grant.pk}",
+        "title": grant.title,
+        "info": info,
+        "display_data": {
+            "url": grant.url,
+            "location": grant.region,
+            "wallet_address": "see grant URL",  # TODO: there are so many wallet addresses..... which one do we display?
```
nollied, line 47: for real, why is there so many `TODO: remove`s in the grant spec? i didn't want to copy paste all the address fields here

Blazingthirdeye:
This is impressive work Nollied.💪

nollied:
@Blazingthirdeye thanks!

kevinrolsen (approved):

app/dashboard/templates/export_grants_ethelo.html
```diff
@@ -0,0 +1,18 @@
...
+
+    <div>
+        <label>End Grant Number (Inclusive)</label>
+        <input type="number" value="" name="end_grant_number" size="1"/>
+    </div>
+    
+    
+    <p>Note: if "End Grant Number" is empty, all grants following the start will be exported.</p>
+    <input type="submit" class="btn" value="Download Grants JSON" name="download_json">
+</form>
```
thelostone-mc, line 18: Could we use 2 space for indenting

thelostone-mc (changes requested):
@nollied thanks for raising the PR ! 
Could you link this to an issue which has more details regarding the purpose behind this (for context sharing)

Had a couple more questions :
- having this baked into gitcoin as opposed to having it stay outside as a standalone which consumes the grant info via an API 
- @kevinrolsen did we manage to test this out ? Ps: any chance you've got understanding on the usecase here? 
I just skimmed over the code and looks like we might have to wire in the admin address still (there is a TODO left in there)

app/dashboard/ethelo.py
thelostone-mc: hmm so a grant in theory has a chain specific address. we would want to use admin_address

nollied:
@thelostone-mc respective responses to your points:

- i wish i could use an API for this information... @owocki told me to write the code here instead. 
  -  also, our public API simply doesn't expose all of this information, and the rate limiting makes it impossible to gather all grants for a grant round.
 
- i've used it, it's only going to be used a handful of times throughout the grant rounds anyways until we get a more permanent solution.

- we don't need any addresses necessarily. it would have been nice to, but not required for what we need.

i'll update the description to add context instead of creating an issue

app/dashboard/templates/export_grants_ethelo.html
nollied: done

nollied:
@thelostone-mc re-review pls, this is a burning issue

app/dashboard/ethelo.py
nollied: what is the admin_address?

nollied: oh, i see. there's a todo for `# TODO-GRANTS: rename to eth_payout_address`... ok lol makes sense.

lauraannwilliams:
Note from Ethelo - This does not include the option_category_slug , which is used to group things on the Ethelo Website. However, this isn't going to be an issue because I'll just apply the default group ('Uncategorized') if it is not supplied. 

However, the display_data section may have an issue. In the spec I provided the example JSON is as follows:

`{
  "options": [
    {
      "slug": "option1",
      "title": "Option 1",
      "info": "<p>Formatted text</p>",
      "option_category_slug": "category1",
      "display_data": {
        "Location": "location text",
        "Wallet Address": "xx",
        "Twitter": '@handle'
      }
    },
    {
      "slug": "option2",
      "title": "Option 2",
      "info": "<p>Formatted text</p>",
      "option_category_slug": "category2",
      "display_data": {
        "Location": "location text",
        "Wallet Address": "xx",
        "Twitter": '@handle'
      }
    }
  ]
}

`

The display_data section has a set of key->value pairs that will be added to the grant description. Because this is considered an arbitrary list, and not a hard-coded one, the key will be used to label the value on our system. So from my example, the generated content would look like:

Twitter Handle: @linked_handle
Location: Location Text
Wallet Address: XXXXX

However, the example provided here would look like this:

url:  /grants/1/nollied-wants-1million-dollars
location: north_america
payout_address: 0x0
twitters: nolliechain
grant_number: 1

I can do a pass to look for twitter handles and urls starting with https to auto link them, but if you would like me to swap out a key for a caption, I'll need a complete list of possible captions.

thelostone-mc:
sidenote: we are reviewing this ! Just making sure we don't jump the gun and roll this out
Just added the don't merge label to make sure we don't merge this

nollied:
@lauraannwilliams ok, i just updated the code + description of the PR with the JSON spec you requested. i am omitting the optional category on purpose.

please let me know if this updated spec is correct, thanks!

app/dashboard/ethelo.py
```diff
@@ -0,0 +1,54 @@
+from grants.models import GrantQuerySet, Grant
+
+
+EXPORT_FILENAME = "grants_export_for_ethelo.json"
+
+
+def get_grants_from_database(start_grant_number: int, end_grant_number: int=None) -> dict:
```
nollied, line 7: @thelostone-mc 

this probably should have a test written, but i'm honestly not sure how to create a few grants such that the database query would return a few deterministic grant objects.

i read in the code that the [local environment should create "Initial test data"](https://docs.gitcoin.co/mk_setup/), but i didn't see any in my localhost:8000 site. maybe it's only created during tests?

lauraannwilliams:
@nollied looks good to me

app/dashboard/ethelo.py
```diff
@@ -0,0 +1,66 @@
...
+        inactive_grants_only (bool): If True, only grants that have not been activated by admins yet will be exported.
+
+    Returns:
+        dict: The returned dict will be JSONify-able.
+    """
+
+    query = GrantQuerySet(Grant)
+
+    if inactive_grants_only:
+        query.inactive()
```
nollied, line 23: what's the actual right way to do this? on the main gitcoin.co website users have to have their grants explicitly approved, but in my ad-hoc testing this wasn't the case.

is this the proper way to check if a grant has been admin-activated?

nollied: @kevinrolsen

PR #10105, opened 2022-02-08
Multi token payout poc

thelostone-mc:

### DIFF
--- a/app/grants/clr.py
+++ b/app/grants/clr.py
@@ -480,7 +480,12 @@ def predict_clr(save_to_db=False, from_date=None, clr_round=None, network='mainn
     debug_output = []
     clr_calc_start_time = timezone.now()
 
-    # one-time data call
+    # We could add a field to GrantCLR to track the token symbol and token amount
+    # If we have USD value of ERC-20, total_pot = (ERC-20 amount * USD value)
+    # Prediction is ran every three hours so it should be somewhat close to actual value
+    # On each prediction save value and percentage of total_pot to CLRMatch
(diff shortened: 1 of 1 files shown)


app/grants/clr.py
```diff
@@ -480,7 +480,13 @@ def predict_clr(save_to_db=False, from_date=None, clr_round=None, network='mainn
     debug_output = []
     clr_calc_start_time = timezone.now()
 
-    # one-time data call
+    # We could add a field to GrantCLR to track the token symbol and token amount
+    # If we fetch USD value of ERC-20, on this script we can get a value for total_pot = (ERC-20 amount * USD value)
+    # Prediction is ran every three hours so it should be updated regularly
+    # On each prediction save USD Value to prediction_curve
+    # When CLRMatch is created use (USD value from prediction curve) / total_pot(ERC-20 amount * USD value) to calculate percent of total_pot
+    # Use percent of total_pot to get # of ERC-20 tokens(UNI)
```
thelostone-mc, line 488: Makes sense as effective CLR would always be in DAI.
We could update the effective CLR to 
- do a lookup against the rate in token table
- convert that to DAI and sum it up 

Maybe instead of adding multiple fields, we just fk it to token table ? 

GrantCLR 
 - fk to Token 

Other changes where it might be worth adding a fk but maybe it's not needed cause this can be derived 
CLRMatch and GrantCLRCalculation
 - fk to Token ?

schultztimothy: Thats a great idea to use a fk to the Token table and Conversion rates 🚀

PR #10106, opened 2022-02-08
Fix error handling in case of bad activity pk

nutrina:
##### Description

Improve error handling for '/activity' path.

##### Refers/Fixes

https://github.com/gitcoinco/private/issues/26

##### Testing

This returns a 404 page now (due to malformed URL), error is not logged.

### DIFF
--- a/app/retail/views.py
+++ b/app/retail/views.py
@@ -801,8 +801,12 @@ def get_specific_activities(what, trending_only, user, after_pk, request=None, p
 
     # single activity
     if 'activity:' in what:
-        activities = Activity.objects.filter(pk=what.replace('activity:', ''))
-        filter_applied = True
+        try:
+            activities = Activity.objects.filter(pk=what.replace('activity:', ''))
+            filter_applied = True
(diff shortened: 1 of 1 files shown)


app/retail/views.py
```diff
@@ -801,8 +801,12 @@ def get_specific_activities(what, trending_only, user, after_pk, request=None, p
 
     # single activity
     if 'activity:' in what:
-        activities = Activity.objects.filter(pk=what.replace('activity:', ''))
-        filter_applied = True
+        try:
+            activities = Activity.objects.filter(pk=what.replace('activity:', ''))
+            filter_applied = True
+        except ValueError:
+            # ValueError might be thrown if 'activity' is not a valid ok value (contains for example sting)
```
gdixon, line 808:
```suggestion
            # ValueError might be thrown if 'activity' is not a valid pk value (contains for example a string)
```

gdixon (approved):
Nice catch @nutrina! 🚀

Issue #9987, opened 2022-01-23
Kill centralized /quadraticlands and forward to DAO-owned page


nfrgosselin:
**User Story:**
As a Gitcoin user
I want to only see the decentralized version of the Quadratic Lands site
So that I am only seeing the correct, community-owned information

**Acceptance Criteria**
GIVEN that I am on gitcoin.co
WHEN I try to visit https://gitcoin.co/quadraticlands
THEN I am sent to http://decentralized.quadraticlands.com/

**Product & Design Links:**
we no longer should be maintaining this portion of the site.  it is now managed by the DAO per https://gitcoin.co/issue/gitcoinco/skunkworks/239/100026286

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
Ported over from https://gitcoin.atlassian.net/browse/GITC-348?atlOrigin=eyJpIjoiMGUyMWM0MmVmZDliNGYyYThjNDhlMDNiY2UwNWE3N2MiLCJwIjoiaiJ9

schultztimothy:
@nfrgosselin @thelostone-mc While looking into #10016 I ran into an error here: https://gitcoin.co/profile/trust where the trust bonus is trying to access the profile's players. Which looks like is located in the quadraticlands model https://github.com/gitcoinco/web/blob/dc3acc5453d1db5b3fdef990ab8b438f99b45413/app/quadraticlands/models.py#L214.

Thanks!

thelostone-mc:
Ah this is the quad diplomacy which was a fun experiment we had built out.

I would lean towards removing this as 
- it was never marketed + only staff enabled 
- we have only test data in live

https://gitcoin.co/_administrationquadraticlands/game/

schultztimothy:
Awesome thanks for the clarification!

PR #10137, opened 2022-02-09
9987 kill quadratic lands w migrations

schultztimothy:
Description
As a Gitcoin user
I want to only see the decentralized version of the Quadratic Lands site
So that I am only seeing the correct, community-owned information

Acceptance Criteria
GIVEN that I am on gitcoin.co
WHEN I try to visit https://gitcoin.co/quadraticlands
THEN I am sent to http://decentralized.quadraticlands.com/

we no longer should be maintaining this portion of the site. it is now managed by the DAO per https://gitcoin.co/issue/gitcoinco/skunkworks/239/100026286

Refers/Fixes
fixes: https://github.com/gitcoinco/web/issues/9987

According to https://techstream.org/Bits/Remove-App-From-Django-Project its best to follow this up with another [PR](https://github.com/gitcoinco/web/pull/10138) that fully removes the app from the monorepo. Still pretty new to django so open to suggestions on this 😄


app/dashboard/tasks.py
```diff
@@ -281,12 +281,6 @@ def update_trust_bonus(self, pk):
         tb += 0.15
     # if profile.is_duniter_verified:
     #     tb *= 1.001
-    qd_tb = 0
```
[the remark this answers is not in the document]
thelostone-mc: lgtm

app/quadraticlands/migrations/0001_initial.py
```diff
@@ -40,7 +40,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('created_on', models.DateTimeField(auto_now=True)),
-                ('claim_total', quadraticlands.models.Uint256Field(default=0)),
+                ('claim_total', models.DateTimeField(auto_now=True)),
                 ('distribution', django.contrib.postgres.fields.jsonb.JSONField(default=dict)),
                 ('profile', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='initial_distribution', to='dashboard.Profile')),
             ],
```
thelostone-mc, line 46: ^ could we undo the changes to the migration file as production already has them

thelostone-mc (changes requested):
I'm good with this approach ! 
Left a comment on the migration update but otherwise LGTM

app/quadraticlands/migrations/0001_initial.py
schultztimothy: Ahhh yeah good call!

When I include the `quadraticlands` model it throws an error and the app won't compile locally. 

I updated it so that it is still a `Uint256Field` but open to suggestions on other ways around that 😄

thelostone-mc: Hmm the issue with that is -> it's not what is running on the server.
Since we'd en up removing this in a follow up PR 
Just to be on the safer side -> maybe we just keep it as original

schultztimothy: Sounds good and makes sense 🚀 I reverted so there are no changes to this file

thelostone-mc (approved):

app/app/views.py
```diff
@@ -0,0 +1,5 @@
+from django.shortcuts import redirect
+
+def redirect_view(request):
+    response = redirect('http://decentralized.quadraticlands.com/')
+    return response
```
ksolo, line 5:
```suggestion
    return redirect('https://decentralized.quadraticlands.com/')
```

ksolo (approved):
one small suggestion (non-blocker).

looks good!

app/dashboard/tasks.py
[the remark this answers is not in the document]
gdixon: We had it in place for a while but backed it out, there might also be some remnants in `app/assets/v2/js/pages/profile-trust.js` and it looks like you've caught the `app/dashboard/views.py` side of it!! 🚀

gdixon (approved):
LGTM! I've a left a comment about somewhere that might have been missed, but other than that, looking great! 🚀

app/dashboard/tasks.py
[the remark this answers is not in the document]
thelostone-mc: Ah nice one ! I'll remove that

Issue #10118, opened 2022-02-08
Create migration of GrantCLR with new fields


nfrgosselin:
**User Story:**
As a Gitcoin engineer
I want to add new fields to the GrantPayout and CLRMatch tables 
So that I can correctly capture payout data going forward

**Acceptance Criteria**
Create migration for the GrantPayout table with the following fields:
- Conversion Rate → IntegerField
- Conversion Date → DateTimeField

Create migration for CLRMatch table with the following fields:
- token(symbol or address) → fk to [Token table](https://gitcoin.co/_administrationeconomy/token/) - Useful to clarify the token amount field
- USD amount -> number - nullable
- Modify CLRmatch `Amount` field to include helper text under the input that reads `Token Amount`(Similar to how the grant field says `The associated Grant.`)
<img width="568" alt="Screen Shot 2022-02-14 at 11 30 01 AM" src="https://user-images.githubusercontent.com/6887938/153924546-cfcd6a7f-52fa-435b-85ad-e30599bacbaf.png">

**Product & Design Links:**
[Miro board](https://miro.com/app/board/uXjVOOp6Dzs=/?invite_link_id=10385067241)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

schultztimothy:
Sorry for the confusion here @gravityblast I originally came up with the requirements for this :) 

I updated the description just let me know if anything doesn't make sense

schultztimothy:
Good call with the comments @gravityblast!

I originally came up with these requirements apologies for the confusion 😄. I updated the description. Just let me know if you have any questions.

We ended up going with a free form field for the conversion rate and conversion date so it wouldn't be linked to that gigantic conversion table and to add some more flexibility

gravityblast:
thank you @schultztimothy ! I created this PR with these changes https://github.com/gitcoinco/web/pull/10158

Issue #10120, opened 2022-02-08
Update finalize script for ERC20 payouts


nfrgosselin:
**User Story:**
As a Gitcoin engineer
I want to run a script when payouts are finalized
So that I can properly capture the ERC20 payout details

**Acceptance Criteria**
GIVEN that payouts are finalized
WHEN I run the [finalize script](https://gitcoin.notion.site/b0b3242b9db24072b5e251f1808866bd#7ffe8558d0be47f0b45b81e64b141983) - aka "payout non-custodial script"
THEN the **Conversation Rate** and **Conversion Rate Date** fields in the **GrantCLR table** are updated with the correct values

GIVEN that payouts are finalized
WHEN I run the [finalize script](https://gitcoin.notion.site/b0b3242b9db24072b5e251f1808866bd#7ffe8558d0be47f0b45b81e64b141983)
THEN the **USD Amount** and **Token Amount** fields in the **CLRMatch table** are updated with the correct values

**Product & Design Links:**
[Miro board](https://miro.com/app/board/uXjVOOp6Dzs=/?invite_link_id=10385067241)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
Requires #10119 to be finished first

thelostone-mc:
Ensure this is built on top of https://github.com/gitcoinco/web/pull/10149 as that does a good chunk of cleanup

PR #10158, opened 2022-02-15
add token info to GrantPayout and CLRMatch

gravityblast:
##### Description

Add token fields to GrantPayout and CLRMatch

- [x] GrantPayout: add Conversion Rate → IntegerField
- [x] GrantPayout: add Conversion Date → DateTimeField
- [x] CLRMatch: token(symbol or address) → fk to [Token table](https://gitcoin.co/_administrationeconomy/token/) - Useful to clarify the token amount field
- [x] CLRMatch: USD amount -> number - nullable
- [x] CLRMatch: Modify CLRmatch Amount field to include helper text under the input that reads Token Amount(Similar to how the grant field says The associated Grant.)

##### Refers/Fixes

Closes: #10118

### DIFF
--- a/app/grants/models/clr_match.py
+++ b/app/grants/models/clr_match.py
@@ -8,13 +8,25 @@ class CLRMatch(SuperModel):
     """Define the structure of a CLR Match amount."""
 
     round_number = models.PositiveIntegerField(blank=True, null=True)
-    amount = models.FloatField()
     grant = models.ForeignKey(
         'grants.Grant',
         related_name='clr_matches',
         on_delete=models.CASCADE,
         null=False,
         help_text=_('The associated Grant.'),
+    )
+    token = models.ForeignKey(
+        'economy.Token',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+    )
+    amount = models.FloatField(
+        help_text='Token Amount'
+    )
+    usd_amount = models.PositiveIntegerField(
+        null=True,
+        blank=True
     )
     has_passed_kyc = models.BooleanField(default=False, help_text=_('Has this grant gone through KYC?'))
     ready_for_test_payout = models.BooleanField(default=False, help_text=_('Ready for test payout or not'))
--- a/app/grants/models/grant.py
+++ b/app/grants/models/grant.py
@@ -60,6 +60,14 @@ class GrantPayout(SuperModel):
         default='DAI',
         help_text=_('Currency in which funds would be paid')
     )
+    conversion_rate = models.PositiveIntegerField(
+        null=True,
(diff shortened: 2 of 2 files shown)


app/grants/models/clr_match.py
```diff
@@ -8,14 +8,26 @@ class CLRMatch(SuperModel):
...
         null=False,
         help_text=_('The associated Grant.'),
     )
+    token = models.ForeignKey(
+        'economy.Token',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+    )
+    amount = models.FloatField(
```
thelostone-mc, line 24: thoughts on renaming this to token_amount to make it clear?

app/grants/models/clr_match.py
```diff
@@ -8,14 +8,26 @@ class CLRMatch(SuperModel):
...
+        help_text='Token Amount'
+    )
+    usd_amount = models.PositiveIntegerField(
+        null=True,
+        blank=True
```
thelostone-mc, line 29: worth adding help_text to say this is the USD amount ?

app/grants/models/grant.py
```diff
@@ -60,6 +60,14 @@ class GrantPayout(SuperModel):
         default='DAI',
         help_text=_('Currency in which funds would be paid')
     )
+    conversion_rate = models.PositiveIntegerField(
+        null=True,
+        blank=True
```
thelostone-mc, line 65: let's add help text

app/grants/models/grant.py
```diff
@@ -60,6 +60,14 @@ class GrantPayout(SuperModel):
...
+    )
+    conversion_date = models.DateTimeField(
```
thelostone-mc, line 67: let's add help text, so that users know when to update it

app/grants/models/grant.py
```diff
@@ -60,6 +60,14 @@ class GrantPayout(SuperModel):
...
+        null=True,
+        blank=True
+    )
```
thelostone-mc, line 70: Don't we need a fk to Token table here as well ?

app/grants/models/clr_match.py
thelostone-mc: renaming this would break things as earlier we would rely on amount as the usd_amount.
We might have to explore places and update usd_amount wherever applicable

app/grants/models/grant.py
chibie: yea i think so too

thelostone-mc:
Merging this into a new branch cause I'll be taking up https://github.com/gitcoinco/web/issues/10120 which builds on top of this

PR #10243, opened 2022-03-03
fix: deletes test_get_bounty() to keep a passing CI

gdixon:
##### Description

`test_get_bounty()` is testing that a standardBounty entry can be picked up from IPFS - because we don't control the version of the infura ipfs node and `ipfshttpclient` is unable to communicate with that node (version mismatch), we are going to drop this test to restore our passing CI.

##### Refers/Fixes

Refs: failing CI


nutrina (approved):
Looks good. Thanks man.

thelostone-mc (approved):

Issue #10252, opened 2022-03-04
Show tag type in grant creation dropdown


nfrgosselin:
Requires finishing #10251 

**User Story:**
As a grant owner
I want to know which tags are for eligibility or discovery
So that I know how to have my grants considered for the appropriate rounds

**Acceptance Criteria**
GIVEN that I am on step 1 of creating a grant (https://gitcoin.co/grants/new)
WHEN I open the tags dropdown
THEN I see the tags organized into eligibility and discovery sections

**Product & Design Links:**
Figma: [Create a Grant](https://www.figma.com/file/8QjqOzDBd4HiE37YqcocK7/?node-id=163%3A3717), Section 5 — Tags

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
This backend for this has been wired in 
https://stage.gitcoin.co/_administrationgrants/granttag/

EmHsiehGitcoin:
looks great!!

PR #10255, opened 2022-03-04
fix eligibility tag error

schultztimothy:
##### Description
This splits the eligibility and discovery tabs into two groups then adds labels and styling to labels

fixes: #10252

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -3917,7 +3917,7 @@ class GrantSubmissionView(View):
             _grant_tag = {
                 'id': g_tag.pk,
                 'name': g_tag.name,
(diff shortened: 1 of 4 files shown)


app/grants/views.py
```diff
@@ -3917,7 +3917,7 @@ def get(self, request):
             _grant_tag = {
                 'id': g_tag.pk,
                 'name': g_tag.name,
-                'is_eligibility_tag': g_tag.is_eligibility_tag
+                'is_eligibility_tag': 1 if g_tag.is_eligibility_tag else 0
```
schultztimothy, line 3920: False was being returned in the JSON and was throwing an error

thelostone-mc (approved):

PR #10263, opened 2022-03-07
fix: use `request.headers.get` to avoid backend errors

gravityblast:
Just a little fix to avoid backend errors in case the `ingest_merkle_claim_to_clr_match` API receives a request without a token set in the headers.


thelostone-mc (approved):

thelostone-mc:
The CI failure here has been fixed in master already. It is not introduced cause of this PR

schultztimothy (approved):
🚀

Issue #10291, opened 2022-03-09
Add back button to Explorer and keep position when going back


EmHsiehGitcoin:
Context:
Right now when the user navigate from Grants Landing Page’s CLR Banners to Explorer page, they don’t have a way to go back. 
The back function on browser reloads the explorer page instead of going back to grants LP.
Therefore we are adding a Back to Grants link and it should remind position when going back. 

Acceptance Criteria: 
1. Add back text link + caret
2. When clicking on the back text link, it brings the user back to where they were and should remind in position

[Figma file: ◽ Add back button to Explorer](https://www.figma.com/file/8QjqOzDBd4HiE37YqcocK7/GR13?node-id=1040%3A30521)

![Screen Shot 2022-03-10 at 9 57 04 AM](https://user-images.githubusercontent.com/96255151/157688613-af40ea2c-67fb-4087-a2f2-d8ef62bc34d3.png)

schultztimothy:
I wasn't able to get to this today with all of the support requests. Throwing it back in todo and will pick it up first thing tomorrow if no one else grabs it

EmHsiehGitcoin:
Chatted with team, we will be fixing the back button function on browser instead of adding the < Back button. Closing this issue and opening a new one: https://github.com/gitcoinco/web/issues/10343

PR #10306, opened 2022-03-10
added back button for grants page

schultztimothy:
##### Description

Add back button to grants explorer

##### Refers/Fixes

https://github.com/gitcoinco/web/issues/10291

### DIFF
--- a/app/grants/templates/grants/explorer.html
+++ b/app/grants/templates/grants/explorer.html
@@ -91,6 +91,7 @@ along with this program. If not, see
         </div>
       {% endif %}
       <div class="container mt-1">
+        <a href="javascript:history.go(-3)" class="font-body pe-auto py-2 d-inline-block"><i class="fas fa-chevron-left mr-2"></i>Back </a>
         <template v-if="!params.collection_id">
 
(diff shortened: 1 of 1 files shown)


app/grants/templates/grants/explorer.html
```diff
@@ -91,6 +91,7 @@
         </div>
       {% endif %}
       <div class="container mt-1">
+        <a href="javascript:history.go(-3)" class="font-body pe-auto py-2 d-inline-block"><i class="fas fa-chevron-left mr-2"></i>Back </a>
```
schultztimothy, line 94: -3 is here instead of -1 because it looks like JS considers adding the query parameters as a navigation event. So -1 was just changing the query params

gdixon: What happens if a user navigates directly to https://gitcoin.co/grants/explorer and then clicks the back button?

gravityblast:
@schultztimothy  does this button need to send the user back to the grants LP?
I was just wondering if a normal link to the grants LP page can work, instead of using the browser history.

chibie (commented):
i have same thoughts as @gdixon and @gravityblast 

what happens when a user clicks the back button without prior browser history?

Issue #10287, opened 2022-03-09
Recaptcha causing 403 error on subsequent attempts


GTChase:
**Steps To Reproduce**
1. Open the same url ~5 times on multiple tabs
2. Complete recaptcha page on first tab
3. Attempt to complete recaptcha on subsequent pages

**Observed  behavior**
You will be met with a 403 on subsequent pages

**Expected behavior**
You will be able to navigate to your desired location with out a 403

PR #10307, opened 2022-03-10
fix: recaptcha middleware to change POST to GET on second submission

gdixon:
##### Description

This PR ensures that after a successful recaptcha submission, any subsequent submission will be treat like GET requests by django

##### Refers/Fixes

fixes: #10287

##### Testing

Tested locally (with postman)


nutrina (approved):
LGTM

Issue #10262, opened 2022-03-07
Cart icon: hover = overview; click = to grant cart


thelostone-mc:
Currently, the overview dropdown opens when the user click on cart icon 

It'd be more intuitive if we show the overview dropdown on hover 
and when the user click the icon it brings them directly to their Grant Cart page

*This only affects desktop

<img width="348" alt="Screen Shot 2022-03-01 at 9 04 25 AM" src="https://user-images.githubusercontent.com/96255151/156183515-17dce6e5-45de-4c95-98e0-73f9d3cd7388.png">

EmHsiehGitcoin:
desktop looks great! @gravityblast Can we hover off dropdown (it disappears when mouse moves away) instead of 'click away' to close? thank you!

gravityblast:
@EmHsiehGitcoin that's slightly more difficult becuase of way the dropdown is constructed by bootstrap, specifically toggle element and the dropdown content are separated. if it's not really needed I would skip this, otherwise we could work on it but we would probably need to change it completely.

nfrgosselin:
@EmHsiehGitcoin what do you think of Andrea's point?

PR #10338, opened 2022-03-15
feat: open cart content on mouseover on desktop

gravityblast:
##### Description

This PR changes the behaviour of the cart icon in the menu. 
Only on desktop we want to open the cart content on mouseover, and redirect to the cart page on click.
For mobile we keep the previous behaviour, opening the cart content on click.

 The `onLinkMouseOver` and `onLinkClick` methods don't do anything if the current platform is mobile, leaving
the management of the event to the `bootstrap` handlers.
For desktop we simulate the `click` event on `mouseover` to trigger the `bootstrap` handlers, so that we don't need to rewrite the show/hide behaviour.

##### Refers/Fixes

Closes: #10262

### DIFF
--- a/app/assets/v2/js/cart-nav.js
+++ b/app/assets/v2/js/cart-nav.js
@@ -2,16 +2,17 @@ Vue.component('gc-cart-content', {
   template: '#gc-cart-content',
   delimiters: [ '[[', ']]' ],
   data: () => {
-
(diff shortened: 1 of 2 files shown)


thelostone-mc:
@gravityblast looks like we need to switch double quotes with single quotes :P

Issue #10332, opened 2022-03-14
Add Skip to Top function


EmHsiehGitcoin:
Add the function "Scroll back to top" to site overall

https://mdbootstrap.com/docs/standard/extended/back-to-top/

EmHsiehGitcoin:
How do I trigger this?

PR #10350, opened 2022-03-16
feat: add scroll to top

thelostone-mc:
##### Description

Adds scroll to top functionality

##### Refers/Fixes

closes https://github.com/gitcoinco/web/issues/10332

##### Testing

Untitled

### DIFF
--- a/app/assets/v2/scss/base.scss
+++ b/app/assets/v2/scss/base.scss
@@ -12,6 +12,7 @@ html {
   bottom: 20px;
   right: 20px;
   display: none;
+  z-index: 9999;
 }
 
 div.body {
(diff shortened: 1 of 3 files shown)


app/assets/v2/scss/base.scss
```diff
@@ -4,6 +4,14 @@
...
   --profile-step-fill: #cfbfff;
   --profile-step: var(--gc-blue);
+  scroll-behavior: smooth;
+}
+
+#btn-back-to-top {
+  position: fixed;
+  bottom: 20px;
+  right: 20px;
+  display: none;
```
gravityblast, line 14: since the button element has been added at the beginning of the document, I noticed it's covered by other elements with position absolute, like the robot at the end of the `/bounties/funder` page. if we put something like `z-index: 9999` it fixes it.

gravityblast (approved):
LGTM, I just added a comment about the button `z-index`

app/assets/v2/scss/base.scss
thelostone-mc:
```suggestion
  display: none;
  z-index: 9999;
```

PR #10371, opened 2022-03-17
Fix back navigation in grant creation flow

nutrina:
##### Description

This fixes back navigation in the grant creation flow, when the user is in the last step.
2 cases need to be handled separately:
1. the user clicks the right button (for next) -> need to check if we submit form 
3. the user clicks the left button (go back) -> need to check if we are on the first page

##### Refers/Fixes

No ticket open for this. I have noticed this as I was implementing the same flow for bounties, and did not want to walk by it.

##### Testing

Has been tested locally.
https://www.loom.com/share/dd5388b2126d4513a03cdc0e50c2b854

### DIFF
--- a/app/assets/v2/js/grants/_new.js
+++ b/app/assets/v2/js/grants/_new.js
@@ -332,16 +332,14 @@ Vue.mixin({
       this.$set(this.form, inputField.id, extracted);
     },
     updateNav: function(direction) {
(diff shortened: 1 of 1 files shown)


schultztimothy (approved):
Thanks a ton for doing the @nutrina!

thelostone-mc (approved):

Issue #10354, opened 2022-03-16
Grant with two addresses not showing up in cart


nfrgosselin:
**Details**
The [Moni Talks grant](https://gitcoin.co/grants/3637/moni-talks) can be successfully added to the cart (screenshot 1) but once you click over to the cart it is invisible (screenshot 2). This may be because of it having two wallets (screenshot 3).

**Reproduction steps**
1. Start with empty cart
2. Add [Moni Talks grant](https://gitcoin.co/grants/3637/moni-talks) to your chart
3. Click Checkout

**Screenshots**
1
![image](https://user-images.githubusercontent.com/15041635/158596333-0bce6b42-902b-4f8f-be3d-e305ebbf1c62.png)

2
![image](https://user-images.githubusercontent.com/15041635/158596376-b2584a9c-ccd9-4fb4-bb00-51bc8d98e7ab.png)

3
![image](https://user-images.githubusercontent.com/15041635/158596455-e0723f80-e9c2-4c4c-80dc-43c50622d502.png)

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
This is an user done by an admin!
We do not allow a grant to be created with multiple address in the frontend.
They forgot to reset the BNB address to 0x0 which is what resulted in this issue.

thelostone-mc:
I've fixed this up manually for now

nfrgosselin:
Thanks @thelostone-mc!

PR #10375, opened 2022-03-17
looping over tenants when creating totals

schultztimothy:
##### Description

If there were more than one blockchains associated with a grant it was not calculating grant totals correctly

##### Refers/Fixes

fixes: #10354

##### Testing

https://user-images.githubusercontent.com/6887938/158905330-2a7123dd-5754-4936-b98f-feab7aa31cce.mov

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -164,9 +164,11 @@ Vue.component('grants-cart', {
       let vm = this;
 
       var grantsTentantsCount = vm.grantData.reduce(function(result, grant) {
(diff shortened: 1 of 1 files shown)


thelostone-mc:
@schultztimothy so our backend while it supports multiple addresses for a grant. 
We should never have grants with more than 1 address -> cause our front end doesn't account for it.

@nfrgosselin Do we still want to get this in ? A grant added once appears twice in the checkout per chain.
It seems confusing and we would not encounter this scenario unless someone manually updates the data

https://github.com/gitcoinco/web/issues/10354#issuecomment-1073503304

PR #10409, opened 2022-03-23
New bounty creation flow

nutrina:
##### Description

This implements the new bounty creation flow.

##### Refers/Fixes

Se this board for the issue that this PR fixes: 
https://github.com/orgs/gitcoinco/projects/4/views/1?filterQuery=epic%2Finitiative%3A%22Bounty+Creation%22

##### Testing

- has partial test coverage with cypress
- manual testing


app/app/fixtures/economy.json
```diff
@@ -24525,5 +24390,24 @@
...
+          "network": "testnetrpc",
+          "decimals": 18,
+          "priority": 1,
+          "chain_id": 1,
+          "network_id": 1,
+          "metadata": {},
+          "approved": true,
+          "conversion_rate_id": null,
+          "conversion_rate_source": null
+      }
```
thelostone-mc, line 24411: ^ does this mean we don't want ETH for mainnet / testnet ?

app/app/urls.py
```diff
@@ -443,6 +443,7 @@
     ),
 
     # View Bounty
+    # TODO geri: should we drop the 2 URL patterns below???
```
thelostone-mc, line 446: Is this something which needs to be addressed now ? 
would it make more sense to track this in a ticket as opposed to the code ?

app/app/urls.py
```diff
@@ -454,6 +455,7 @@
         name='issue_details_new2'
     ),
     re_path(r'^funding/details/?', dashboard.views.bounty_details, name='funding_details'),
+    re_path(r'^issue/(?P<bounty_id>\d+)', dashboard.views.bounty_details, name='issue_details_new4'),
```
thelostone-mc, line 458: couldn't we remove the old ones  issue_details_new2 / is that still needed ?

app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -343,14 +346,28 @@ Vue.mixin({
       waitBlock(bountyMetadata.txid);
 
     },
-    checkOwner: function(handle) {
+    checkOwner: function() {
       let vm = this;
+      let ret = false;
```
thelostone-mc, line 351: rename this to something more meaningful  ?

app/assets/v2/js/pages/bounty_details2.js
```diff
@@ -343,14 +346,28 @@ Vue.mixin({
...
+        ret = caseInsensitiveCompare(document.contxt['github_handle'], owner_handle);
       }
-      return false;
 
+      // Check also for additional bounty owners
+      if (!ret) {
+        for (let i = 0; i < vm.bounty.owners.length; i++) {
+          let additionalOwner = vm.bounty.owners[i];
+
+          console.log('geri: additionalOwner', additionalOwner.handle);
```
thelostone-mc, line 363: do we need the log ?

app/retail/views.py
```diff
@@ -19,6 +19,7 @@
 '''
 import json
 import logging
+from pprint import pformat
```
thelostone-mc, line 22: Is this needed?

app/grants/templates/grants/components/form_wrapper.html
```diff
@@ -7,6 +7,7 @@ <h4>[[heading]]</h4>
       </template>
       <slot></slot>
     </div>
+    <slot name="footer"></slot>
```
thelostone-mc, line 10: what's this for ?

app/dashboard/views.py
```diff
@@ -6479,8 +6561,19 @@ def payout_bounty_v1(request, fulfillment_id):
     is_funder = bounty.is_funder(user.username.lower()) if user else False
 
     if not is_funder:
-        response['message'] = 'error: payout is bounty funder operation'
-        return JsonResponse(response)
+        is_owner = False
```
thelostone-mc, line 6564: is it possible to update is_funder logic with the is_owner logic ?
Also what's the diff btwn funder and owner at this point ?

app/app/fixtures/economy.json
nutrina: These seemed to be duplicates (correct me if I am wrong).
If you check the file,  you will see that there are 3 entries for ETH: mainnet, rinkeby and testnetrpc (this I use for the tests).

app/dashboard/models.py
```diff
@@ -377,6 +393,12 @@ class Bounty(SuperModel):
     value_in_usdt = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True)
     value_in_eth = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True)
     value_true = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True)
+
+    usd_pegged_value_in_token_now = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True) # The calculated amount in token, corresponding to value_true_usd
+    usd_pegged_value_in_token = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True)     # The calculated amount in token, corresponding to value_true_usd
+    value_true_usd = models.DecimalField(default=0, decimal_places=2, max_digits=50, blank=True, null=True)     # The value the user wants to pay in USD
+    peg_to_usd = models.BooleanField(default=False)                                                             # True if the amount to pay should be pegged to USD
```
thelostone-mc, line 400: ^ can none of the above fields like value_in_usdt etc be resued ? Just curious

app/dashboard/admin.py
```diff
@@ -379,7 +379,7 @@ class BountyAdmin(admin.ModelAdmin):
     raw_id_fields = ['interested', 'coupon_code', 'org', 'event', 'bounty_owner_profile', 'bounty_reserved_for_user']
     ordering = ['-id']
 
-    search_fields = ['raw_data', 'title', 'bounty_owner_github_username', 'token_name']
+    search_fields = ['raw_data', 'title', 'bounty_owner_github_username', 'token_name', 'cusom_title', 'custom_description']
```
thelostone-mc, line 382: cusom_title -> custom_title

thelostone-mc (commented):
left a couple of comments. 
Will review the rest later today as this quite a big PR :P

app/app/fixtures/economy.json
thelostone-mc: Oooo you are right. I wonder why the duplicates were present :O

app/app/urls.py
nutrina: I was reluctant to remove this to avoid breaking the links that users have ....
I will write a proper comment here.

app/app/urls.py
nutrina: I would not remove this now, see my comment above.

app/assets/v2/js/pages/bounty_details2.js
nutrina: No, I will remove this

app/retail/views.py
nutrina: No, I will remove this

app/grants/templates/grants/components/form_wrapper.html
nutrina: This is for additional content that we display below the wizard, if it is the case.
We have a use case where we display fee information for bounty payments in the payment step only.

app/dashboard/admin.py
nutrina: I will fix this.

app/dashboard/models.py
nutrina: This is probably possible, but I think having this additional variables makes following the logic flow much easier:
- if the user does not peg to USD -> we use the old vars, and conversions are made from token amount -> usd amount
- if the user does peg to USD -> we use the new vars, and conversions are made from usd amount -> token amount

I think this helps understand better which value the user entered:  value_true respectively value_true_usd.

thelostone-mc (approved):

Issue #10406, opened 2022-03-23
Trust verification issues


nfrgosselin:
**Overview**
Multiple users have reported issues with trust verification on Twitter and FB. See below

**Steps To Reproduce**
1. Go to your account -> Profile
2. Click on "Trust bonus"
3. Click Verify with Twitter or Verify with Facebook

**Observed  behavior**
Twitter: User receives an error message that their last Tweet doesn't match.
FB: User sees that message that "App not Setup"

**Expected behavior**
User is able to successfully verify their identify

**Screenshots**
Twitter:
![image](https://user-images.githubusercontent.com/15041635/159715037-6aa010a4-71a6-4cea-81e1-7e90a53393ce.png)

![image](https://user-images.githubusercontent.com/15041635/159715009-0b8c282e-4295-44f3-b3a2-0444bd815376.png)

FB:
![image](https://user-images.githubusercontent.com/15041635/159714963-5b804815-4dc4-44bf-a847-ca2da6f34cae.png)

schultztimothy:
I just uploaded a screencast to facebook which should enable authentication with gitcoin.co again.
<img width="1058" alt="Screen Shot 2022-03-23 at 11 04 23 AM" src="https://user-images.githubusercontent.com/6887938/159756888-2ce1db35-f22f-4936-a937-b8a752e856f7.png">

schultztimothy:
It looks like the video I submitted allowed facebook to ok our app. I was just able to verify my profile with facebook.
<img width="1263" alt="Screen Shot 2022-03-23 at 2 04 43 PM" src="https://user-images.githubusercontent.com/6887938/159786233-a4b6dbed-35ed-42a0-823e-e6f05b6c8096.png">

I haven't been able to replicate the twitter error and seems like others have been able to verify successfully. I'm going to move this along 😄

thelostone-mc:
dug into the twitter one and it looks like the api returns the tweet

> @notionport I know the founder well if you'd like an intro 🙂

Looking at the code. == > 
What we do is look at last tweet of user to see if that have the I have verified myself on gitcoin 

My hunch is this verify tweet was posted and before our backend could verify -> @annika | gitcoin replied / tweeted the above message and our backend ended up picking that up.

It might be worth looking at last 10 tweets replies while doing verifying @Nate | Gitcoin  , that way we avoid these scenarios

nfrgosselin:
Thanks for jumping on this guys!

PR #10411, opened 2022-03-24
chore: check last 10 tweet activities instead on last tweet activity

thelostone-mc:
##### Description

dug into the twitter one and it looks like the api returns the tweet

> @notionport I know the founder well if you'd like an intro 🙂

Looking at the code. == > 
What we do is look at last tweet of user to see if that have the I have verified myself on gitcoin 

The issue was that the  verify tweet was posted and before our backend could verify -> annika | gitcoin replied / tweeted the above message and our backend ended up picking that up.

This PR changes to ensure we look at last 10 tweets

##### Refers/Fixes

closes https://github.com/gitcoinco/web/issues/10406

### DIFF
--- a/app/dashboard/views.py
+++ b/app/dashboard/views.py
@@ -3244,8 +3244,13 @@ def verify_user_twitter(request, handle):
                 'msg': msg
             })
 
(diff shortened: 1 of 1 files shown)


schultztimothy (approved):
Good call!

chibie (approved):
lgtm!

Issue #10374, opened 2022-03-17
Speed up explorer filters


nfrgosselin:
**User Story:**
As a user
I want grant explorer search results to load quickly
So that I can efficiently browse and donate to grants

**Acceptance Criteria**
Investigate slow load times and develop a plan to reduce

**Product & Design Links:**

**Tech Details:**
Notes from @thelostone-mc: 
Total requests to get_grants in last 24 hours : 30k
Avg duration is: 2s 

1st image & 2nd Image
----------------------
It looks like 10% of our requests take more than 5.42 seconds 

3rd Image
------------
On filtering requests, it looks like most of the time is taken in the django app (aka business logic code and not DB query) which makes it confusing to see what might be causing it

4th Image
--------------

Let's take an individual request which took 12s: 
https://app.datadoghq.com/apm/trace/17633664223658479660?colorBy=service&env=prod&graphType=span_list&shouldShowLegend=true&sort=time&spanID=250689562829975793&spanViewType=metadata&timeHint=1647436558239.5872

This validates that 60% of the time request time went for django code while the rest was for DB (and we this a lot)

5th Image & 6th Image
-------------
Looking at the DB calls, it looks like the 
- grants filter is heavy as we query against the entire DB  (5th image)
- query to fetch team members (6th image)

We could attempt to improve query but I'm curious to see what operation is it that we do which causes 60% of time to stay in django before returning a response 

![](https://cdn.discordapp.com/attachments/953709815291584554/953893534941995018/unknown.png)

![](https://cdn.discordapp.com/attachments/953709815291584554/953893535243960340/unknown.png)

![](https://cdn.discordapp.com/attachments/953709815291584554/953893535441121310/unknown.png)

![](https://cdn.discordapp.com/attachments/953709815291584554/953893535671799888/unknown.png)

![](https://cdn.discordapp.com/attachments/953709815291584554/953893535894110218/unknown.png)

![](https://cdn.discordapp.com/attachments/953709815291584554/953893536225452032/unknown.png)

**Open Questions:**

**Notes/Assumptions:**

thelostone-mc:
@nfrgosselin How open are we to restrict the keyword seach to only titles and not title & description.
text based searches are expensive the way we have it setup and we might have to explore something like elasticsearch to make the search more efficient (which i don't know if is worth pursing given that the product is heading towards maintainance )

PR #10439, opened 2022-03-30
10374 speed up grants list view

schultztimothy:
##### Description

This attempts to speed up the grants/cards_info endpoint. [Heres](https://app.datadoghq.com/apm/trace/17633664223658479660?colorBy=service&env=prod&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=17818265391804385752&spanViewType=metadata&timeHint=1647436558239.5872) an example of one of the longer running requests from the ticket.

One of the queries takes a bit of time is to find team members for each grant being returned. I don't believe that data is needed for the list view.
Screen Shot 2022-03-30 at 3 28 49 PM

I moved a few other queries that don't seem to be needed for the list view

##### Refers/Fixes

fixes: #10374 

##### Testing

Existing tests are continuing to pass


app/grants/views.py
```diff
@@ -543,16 +543,6 @@ def get_grants(request):
     }
     _grants = get_grants_by_filters(**filters)
 
-    if sort == '' and keyword:
```
schultztimothy, line 546: @chibie @thelostone-mc I removed this because it seems like the same filter is being applied here? https://github.com/gitcoinco/web/blob/a881e53b349217d29b62c9b12f209d0a5e14a018/app/grants/views.py#L784

I changed this pr to draft because I'm not sure if my assumption is correct. Thanks!

thelostone-mc (commented):
left a comment but other wise LGTM

app/grants/views.py
schultztimothy: Ahh ok thanks! It looks like the query on L784 is also a case insensitive query 🚀  https://docs.djangoproject.com/en/4.0/ref/models/querysets/#icontains

thelostone-mc (approved):

philipjonsen (approved):

thelostone-mc (approved):

app/grants/views.py
chibie: @schultztimothy

n/b:

+ removing L546 - L555 will break the exact title matches in grants explorer. shouldn't there be a replacement?

+ it doesn't hit the database as it only does a list comprehension on the _grants variable which contains a page of paginated data

chibie (changes requested):

app/grants/views.py
schultztimothy: @chibie I ended up adding this back in. Just lmk if there is anything else!

chibie (approved):

PR #10441, opened 2022-04-01
Ensure no duplicate files are generated by the bundle django command.

nutrina:
##### Description

This is an issue in our new path to production, as we drop the hashes that are appended to bundle files. 

Those hashes depend on the environment where they are run (on the static configuration), and since we will generate the bundles in the GHA workflow, the hashes calculated will be different compared to the ones that are calculated in production, staging or review environment.

Since the filenames specified in the `{% bundle ... ` are duplicate (uniqueness was only ensured by the hashes) we must rename all duplicate occurrences.

In order to ensure that no such duplicate slips through, I have also modified the `bundle` command to fail when a duplicate is encountered on generation (see screenshot below).

Changes:
- have added a check in the command that raises an error if a file with duplicate name is generated
- have adjusted all {% bundle ... %} tag usage, and set unique file name

##### Refers/Fixes

No specific issue exists, This is a change required for our path to production.

##### Testing

Has been tested locally & in the gha workflow

When a duplicate is encountered, this will generate an error and stop the pipeline:
image

### DIFF
--- a/app/dashboard/management/commands/bundle.py
+++ b/app/dashboard/management/commands/bundle.py
@@ -3,7 +3,7 @@ import re
 import shutil
 
 from django.conf import settings
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from django.template import Context, Template
 from django.template.loaders.app_directories import get_app_template_dirs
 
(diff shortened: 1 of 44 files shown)


thelostone-mc (approved):

kevinrolsen (approved):
lgtm

Issue #10429, opened 2022-03-29
Can't check out with Portis wallet


nfrgosselin:
**Overview**
User [tdurall](www.gitcoin.co/tdurall) is trying to check out with the Portis wallet and seeing an error.

**Steps To Reproduce**
1. Impersonate tdurall
2. Add grant to cart (user was trying to donate to [this grant](https://gitcoin.co/grants/5410/metabolic-mainframe-the-first-hobbyist-bio-compute))
3. Connect Portis wallet
4. Click standard checkout
5. See error

**Observed  behavior**
User sees error in attached screenshot

**Expected behavior**
User is able to checkout

**Screenshots**
![image](https://user-images.githubusercontent.com/15041635/160634337-b50f274f-d627-47d3-ab45-92ed022af89e.png)
![image](https://user-images.githubusercontent.com/15041635/160634407-75b79042-d100-4eae-b0fd-e8c982642fec.png)

PR #10443, opened 2022-04-01
defaulting to mainnet for portis wallet, finding networkId from porti…

schultztimothy:
##### Description

It looks like, when using portis wallet, there was an issue finding the networkId which is used to filter out tokens on the checkout page. 

I also was unable to switch the portis wallet to any network beside rinkeby. While looking through the docs it looks like they are expecting a network parameter https://docs.portis.io/#/quick-start. 

##### Refers/Fixes

fixes: #10429

### DIFF
--- a/app/assets/v2/js/cart.js
+++ b/app/assets/v2/js/cart.js
@@ -1064,8 +1064,6 @@ Vue.component('grants-cart', {
         await await onConnect();
       }
 
-      // console.log({web3}, web3.eth.currentProvider, web3.eth.currentProvider.chainId, 'web3.eth.currentProvider.chainId', String(Number(web3.eth.currentProvider.chainId)), 'String(Number(web3.eth.currentProvider.chainId)');
-      // let networkId = String(Number(web3.eth.currentProvider.chainId));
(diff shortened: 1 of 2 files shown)


philipjonsen (approved):

thelostone-mc (commented):
lgtm. small note:  could we remove the log and commented out code

schultztimothy:
> lgtm. small note: could we remove the log and commented out code

Ahh thank you for catching that 🙌

thelostone-mc (approved):

app/assets/v2/js/cart.js
```diff
@@ -1131,6 +1134,7 @@ Vue.component('grants-cart', {
         const tokenName = selectedTokens[i];
         const tokenDetails = this.getTokenByName(tokenName, isPolygon);
 
+        console.log({ tokenDetails });
```
chibie, line 1137: can we take out this log?

chibie (changes requested):

app/assets/v2/js/cart.js
thelostone-mc:
```suggestion
```

chibie (approved):

PR #10466, opened 2022-04-07
Fixing linter errors

nutrina:
##### Description

Fixes linter errors which cause pipeline to fail.

##### Refers/Fixes

Linter errors

##### Testing

Have not tested since these are not functional changes.

### DIFF
--- a/app/assets/v2/js/grants/matching_funds.js
+++ b/app/assets/v2/js/grants/matching_funds.js
@@ -273,7 +273,6 @@ Vue.mixin({
 });
 
 if (document.getElementById('gc-matching-funds')) {
-  ['grant'],
   appGrantDetails = new Vue({
     delimiters: [ '[[', ']]' ],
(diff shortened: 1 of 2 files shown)


app/assets/v2/js/grants/matching_funds.js
```diff
@@ -273,7 +273,7 @@ Vue.mixin({
 });
 
 if (document.getElementById('gc-matching-funds')) {
-  props: ['grant'],
```
thelostone-mc, line 276: do we need need this ?

gravityblast (approved):

app/assets/v2/js/grants/matching_funds.js
nutrina: @thelostone-mc Have just discuss this with @gravityblast , it seems to have been added there by mistake.
It actually does not do anything.

thelostone-mc (approved):

PR #10475, opened 2022-04-11
fix test for payout round noncustodial

gravityblast:
fix test for payout round noncustodial. 

* fix fixtures dependency
* fix contract event name
* fix payout amounts in token and usd


nutrina (approved):
Code changes look good to me.

PR #10520, opened 2022-04-21
Update web3Modal version

0xzoz:
##### Description

Bump the web3Modal version to 1.9.6. This version of web3Modal incorporates addition of TallyHo injected provider.

##### Refers/Fixes

N/A

##### Testing

N/A


thelostone-mc (changes requested):
While this a patch bump ! We would still need some level of testing done to ensure the web3 modal works as expected.
Could you attach a recroding of doing a bounty payout. / tip with this change so that it makes it easier to validate ?

Varshans2 (approved):

Issue #10529, opened 2022-04-25
Add KYC thresholds to CLRs in Django


nfrgosselin:
**User Story:**
As a Grants Ops admin
I want to be able to configure each CLR's KYC threshold in Django
So that I can more easily run grants rounds

**Acceptance Criteria**
Note: static_json table

GIVEN that I am on [the Grant CLR list](https://stage.gitcoin.co/_administrationgrants/grantclr/) in Django
WHEN I click into a specific Grant CLR's change page
THEN I see a numeric field called "KYC Threshold" under the "Grant clr percentage cap" field
AND I see help text under the field that says "CLR Matches >= to the number above will require KYC"

GIVEN that I am on a specific Grant CLR's change page and have entered a number in the "KYC Threshold" numeric field
WHEN I click Save at the bottom of the page
THEN that Grant CLR will require any of its CLR matches >= to the number in "KYC Threshold" to require KYC

GIVEN that I am on a Grant CLR's page and the "KYC Threshold" numeric field is **empty**
WHEN I click Save at the bottom of the page
THEN that Grant CLR will set its KYC threshold to 5000

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

Issue #10387, opened 2022-03-21
Explore + Build URL shortener


nfrgosselin:
**User Story:**
As a Gitcoin community member
I want a simple-to-share URL for grants rounds
So that I can easily share direct links to the grants round explorer

**Acceptance Criteria**
- Explore what it would take to create shortened URLs for grants rounds and pair with @nfrgosselin / @EmHsiehGitcoin to write follow-on stories

**Product & Design Links:**

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

nfrgosselin:
@thelostone-mc I heard back from Grants Ops — they want to go with the format:

gitcoin.co/grants/clr/<slug>

Let me know if you need anything else!

nfrgosselin:
LGTM!

PR #10534, opened 2022-04-26
feat: clr improvements

thelostone-mc:
##### Description

- sets set PAYOUT_KYC_THRESHOLD via django making it easier to configure
  -  would be controlled by StaticJSONEnv
  - introduces a new table to store images in general 
- create short URL for grant clr to make it easier to market rounds
  - These would be of the format. `/grants/clr/<customer-name>/<round-number>/<slug>`
  - No additionally configuration needed. If the GrantCLR are created with all the above fields filled, the short URL should work automatically

##### Testing

Untitled

closes https://github.com/gitcoinco/web/issues/10387
closes https://github.com/gitcoinco/web/issues/10529


nfrgosselin:
LGTM

Note that the flow was changed so that this is configurable from Static json envs› PAYOUT_KYC_THRESHOLD in admin. This means that KYC can only be set globally (as it has in the past), and not by CLR (which would be new functionality).

Issue #10478, opened 2022-04-11
Only show active grants in global search


nfrgosselin:
**Spike timebox: half day**

**User Story:**
As a Gitcoin user
I want to only see active grants in the global search results
So that I don't waste my time with outdated grants

**Acceptance Criteria**
GIVEN that I am on gitcoin.co and have entered a search term in global search
WHEN I click "go"
THEN I only see active grants in the results

^ this is the desired behavior —> please pair with Nate after the spike timebox to figure out the additional stories that are needed

**Product & Design Links:**
*search query*
<img width="1000" alt="image" src="https://user-images.githubusercontent.com/15041635/162814854-22a3adc9-b94c-4c7f-b24e-0210999c9905.png">

*Inactive grant*
<img width="1342" alt="image" src="https://user-images.githubusercontent.com/15041635/162814946-4aaa96a1-c028-4fe4-808b-6b4c76bd2907.png">

*deleted grant*
<img width="1339" alt="image" src="https://user-images.githubusercontent.com/15041635/162815086-091a20d6-39ff-4745-884a-9e37c9658daf.png">

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**
- This might be a problem with the indexing in Elastic Search, since it's pulling in grants that are no longer present in our database.

schultztimothy:
I feel like four tickets could be created for this. 

2. Create a remove_from_elasticsearch() function similar to [put_on_elasticsearch()](https://github.com/gitcoinco/web/blob/1b78d111bbd9c6d7c2716a05ffeba0f6c0345ff9/app/search/models.py#L27), which utilizes the ES [delete](https://elasticsearch-py.readthedocs.io/en/stable/api.html#elasticsearch.Elasticsearch.delete) method to remove the grant from the ES index. 

2. Remove all inactive, hidden, deleted grants from the ES index. This [method](https://elasticsearch-py.readthedocs.io/en/v8.1.3/async.html#elasticsearch.AsyncElasticsearch.delete_by_query ) can be used to delete the records from the ES index. The hard part will be finding previously deleted grant ids or something identifiable that can be used to find them within ES. A command will need to be created to find these deleted grants and remove them from the index

[...]

PR #10554, opened 2022-04-29
Reindex elastic to only inlcude active grant 

schultztimothy:
##### Description

Deleted and inactive grants were being returned from search results. This adds a new command to create a new index that only includes active grants. It also should check for grants being active and/or hidden before it saves them and indexes them into the search model as well as the elastic index. A new index had to be created since grants were being deleted from the django admin and it was impossible to detect which grants had been deleted.

This is the command to create the new index: docker-compose exec web python3 app/manage.py create_search_index_with_active_grants {index-name}. 

Once this is merged the command should be ran to create the new index. Once the index is created we should be able to merge https://github.com/gitcoinco/web/pull/10557

##### Refers/Fixes

Fixes: #10478

### DIFF
--- a/app/grants/tasks.py
+++ b/app/grants/tasks.py
@@ -115,7 +115,7 @@ def update_grant_metadata(self, grant_id, retry: bool = True) -> None:
 
     print(lineno(), round(time.time(), 2))
     from search.models import SearchResult
-    if instance.pk:
+    if instance.pk and instance.active and not instance.hidden:
         SearchResult.objects.update_or_create(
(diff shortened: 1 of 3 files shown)


inosgun:
Y

thelostone-mc (approved):

thelostone-mc:
@schultztimothy looks like we have CI failures :/ Any chance you could look into that ?

schultztimothy:
> @schultztimothy looks like we have CI failures :/ Any chance you could look into that ?

Thanks! I just rebased and things are working as expected ✈️

PR #10555, opened 2022-04-29
Allow smart contract wallets to obtain Proof of Humanity trust bonus

scampVR:
##### Description

Fixes an issue where smart contract wallets can't properly sign the trust bonus verification for proof of humanity. The solution is to attempt to call IsValidSignature() on the wallet address if the usual recoverHash() method fails. I tested specifically on an Argent wallet, but in theory, as long as the wallet supports eip-1271, this should work.

Note: I'm very new to interacting with these tech stacks (python, smart contracts), so the more guidance anyone can provide, the better! Thanks!

- Also includes minor typo fix
##### Refers/Fixes

EIP-1271 https://eips.ethereum.org/EIPS/eip-1271
Argent wallet support for EIP-1271 https://docs.argent.xyz/wallet-connect-and-argent#message-signature

##### Testing

Wasn't able to locate existing tests around similar code, but would be happy to add them if provided some guidance

### DIFF
--- a/app/assets/v2/js/pages/profile-trust.js
+++ b/app/assets/v2/js/pages/profile-trust.js
@@ -833,7 +833,7 @@ Vue.component('poh-verify-modal', {
             </div>
             <div v-if="step === 'validation-complete'">
               <div>
-                Your Proof of Humanity verification was successful. Thank you for helping make Gitcoin mroe sybil resistant!
+                Your Proof of Humanity verification was successful. Thank you for helping make Gitcoin more sybil resistant!
               </div>
               <b-button @click="dismissVerification" variant="primary" class="btn btn-primary mt-5 px-5 float-right">Done</b-button>
             </div>
--- a/app/dashboard/poh_utils.py
+++ b/app/dashboard/poh_utils.py
@@ -2,6 +2,7 @@ POH_CONTRACT_ADDRESS = '0xC5E9dDebb09Cd64DfaCab4011A0D5cEDaf7c9BDb'
 
 _poh_contract = None
 
+
 def get_poh_contract(web3):
     global _poh_contract
     if _poh_contract is None:
--- a/app/dashboard/poh_utils.py
+++ b/app/dashboard/poh_utils.py
@@ -1,7 +1,18 @@
+import json
+
 POH_CONTRACT_ADDRESS = '0xC5E9dDebb09Cd64DfaCab4011A0D5cEDaf7c9BDb'
 
 _poh_contract = None
 
+EIP_1271_ABI = '[{"constant":true,"inputs":[{"name":"_messageHash","type":"bytes32"},{"name":"_signature","type":"bytes"}],"name":"isValidSignature","outputs":[{"name":"magicValue","type":"bytes4"}],"payable":false,"stateMutability":"view","type":"function"}]'
+
+def is_valid_eip_1271_signature(web3, address, hash, signature) -> bool:
+    try:
+        eip_1271_contract = web3.eth.contract(address=address, abi=json.loads(EIP_1271_ABI))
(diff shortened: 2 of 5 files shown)


app/assets/v2/js/pages/profile-trust.js
```diff
@@ -833,7 +833,7 @@ Vue.component('poh-verify-modal', {
             </div>
             <div v-if="step === 'validation-complete'">
               <div>
-                Your Proof of Humanity verification was successful. Thank you for helping make Gitcoin mroe sybil resistant!
+                Your Proof of Humanity verification was successful. Thank you for helping make Gitcoin more sybil resistant!
```
schultztimothy, line 836: 🧙

app/dashboard/poh_utils.py
```diff
@@ -1,7 +1,18 @@
+import json
+
 POH_CONTRACT_ADDRESS = '0xC5E9dDebb09Cd64DfaCab4011A0D5cEDaf7c9BDb'
 
 _poh_contract = None
 
+EIP_1271_ABI = '[{"constant":true,"inputs":[{"name":"_messageHash","type":"bytes32"},{"name":"_signature","type":"bytes"}],"name":"isValidSignature","outputs":[{"name":"magicValue","type":"bytes4"}],"payable":false,"stateMutability":"view","type":"function"}]'
```
schultztimothy, line 7: Super tiny, but you could change this to the same format as below https://github.com/gitcoinco/web/blob/bbac82ff2bfb8755550f59c1872befac00bfd7ff/app/dashboard/poh_utils.py#L29 and not have to use `json.loads()`

schultztimothy (approved):
I made one tiny comment but LGTM!

app/dashboard/poh_utils.py
scampVR: Good point, ty!

chibie (approved):

thelostone-mc:
@scampVR Yo!  I'll pull this in and test it tom but could we 
- move the ABI to it's own file 
- also to unblock you , could you create a branch of this and throw up the fix on that ? 

Also looks like our CI failed (not due to changes in your PR ) , I'll look into get that running as well

app/dashboard/poh_utils.py
```diff
@@ -2,6 +2,15 @@
 
 _poh_contract = None
 
+EIP_1271_ABI = [{"constant":True,"inputs":[{"name":"_messageHash","type":"bytes32"},{"name":"_signature","type":"bytes"}],"name":"isValidSignature","outputs":[{"name":"magicValue","type":"bytes4"}],"payable":False,"stateMutability":"view","type":"function"}]
```
thelostone-mc, line 5: let's move this to the ABI folder 
https://github.com/gitcoinco/web/tree/master/app/grants/abi

app/dashboard/poh_utils.py
```diff
@@ -2,6 +2,15 @@
...
+
+def is_valid_eip_1271_signature(web3, address, hash, signature) -> bool:
```
thelostone-mc, line 7: I'm assuming this is not restricted to just POH so what do you think about moving this over to https://github.com/gitcoinco/web/blob/master/app/grants/utils.py ?

scampVR: Perfect, this was going to be my next question! Other CL will need to call it as well, so this is great timing thanks

scampVR:
@thelostone-mc Ok took me a bit to figure out that I needed the empty __init__.py file to be able to import the abi file, but should be working now. Also let me know if the naming on the file is strange and i'll change it to whatever

thelostone-mc (approved):

thelostone-mc:
This looks good @scampVR  ^_^

PR #10559, opened 2022-05-03
Pinning django debug toolbar to version that works with our environment

nutrina:
##### Description

Pinning django debug toolbar to a version that works with our environment.

##### Refers/Fixes

No specific issue, this fixes workflow errors.

##### Testing

GHA workflow is expected to pass and not fail due to django debug toolbar version.


kevinrolsen (approved):
lgtm

chibie (approved):

PR #10568, opened 2022-05-04
Revert "Ensure no duplicate files are generated by the bundle django command."

thelostone-mc:
Reverts gitcoinco/web#10441


thelostone-mc:
image

This breaks grants explorer page as qull.js file cannot be loading.
Upon reverting this commit on staging and local -> all pages work / load as expected

PR #10582, opened 2022-05-06
Fixing the duplicate discovery of files by the bundle command, depend…

nutrina:
…ing on the current working dir

##### Description

Depending on the config, the `bundle` command might pick the same folder several times, because template folder are picked up from different sources.
We want to avoid duplicates, as duplicates will result in the bundle command to fail (as it now issues an error in case duplicate files are encountered).

##### Testing

Has been tested on branch infra_as_code_staging, and cherry-picked here.


gdixon (approved):
LGTM! 🚀

Issue #10593, opened 2022-05-08
Claim matching for GR13 does not work


LefterisJP:
**Overview**
Can't claim rotki's uniswap CLA matching round. What's more the displayed amount is completely off.

**Steps To Reproduce**
Go to https://gitcoin.co/grants/matching_funds for rotki.

See the displayed amount that equates UNI with $ ... so don't even know what the matching should be.
![2022-05-08_08-40](https://user-images.githubusercontent.com/1658405/167285085-70b33ea2-084b-4461-bc47-d9a619a983c0.png)

**Observed  behavior**
Pressing claim fails. Does nothing.

See the following in the console:

```
wallet.3c8d9ed4243b.js:117 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'split')
    at fetchAccountData (wallet.3c8d9ed4243b.js:117:71)
fetchAccountData @ wallet.3c8d9ed4243b.js:117
await in fetchAccountData (async)
(anonymous) @ wallet.3c8d9ed4243b.js:404
load (async)
(anonymous) @ wallet.3c8d9ed4243b.js:377
core.292db2.c7a091d0f192.js:1 ReferenceError: indicateMetamaskPopup is not defined
    at waitingState (matching_funds.5c6802492ae3.js:130:9)
    at a.claimMatch (matching_funds.5c6802492ae3.js:133:7)
    at click (eval at Ya (core.292db2.c7a091d0f192.js:1:91725), <anonymous>:3:15145)
    at He (core.292db2.c7a091d0f192.js:1:11384)
    at HTMLAnchorElement.n (core.292db2.c7a091d0f192.js:1:13168)
    at HTMLAnchorElement.Yr.o._wrapper (core.292db2.c7a091d0f192.js:1:48505)
```

**Expected behavior**
It should work.

**Additional context**
Almost everything seems broken here and there :sweat:

nfrgosselin:
Hey @LefterisJP — thanks for flagging this. Just letting you know that we're looking into it and working on a fix!

LefterisJP:
Hey guys thanks, it seems to work now.

PR #10597, opened 2022-05-09
checking for value before accessing networkName to display connection…

schultztimothy:
##### Description

If there was an issue fetching chainInfo networkName is undefined and an uncaught error was being thrown when displaying network connectivity. This checks for networkName before accessing it so that execution doe not stop. 

##### Refers/Fixes

should fix one issue documented in #10593 and #10598

##### Testing

https://user-images.githubusercontent.com/6887938/167465534-c3f6f0b7-9e1f-4c03-9a18-56fe781555ef.mov

Screen Shot 2022-05-09 at 1 14 18 PM
Screen Shot 2022-05-09 at 1 13 02 PM

### DIFF
--- a/app/assets/v2/js/grants/matching_funds.js
+++ b/app/assets/v2/js/grants/matching_funds.js
@@ -145,9 +145,10 @@ Vue.mixin({
 
       const chainId = Number(web3.eth.currentProvider.chainId);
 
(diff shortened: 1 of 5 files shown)


thelostone-mc (approved):

Brunk4 (commented):
Pull

Issue #10480, opened 2022-04-11
Change result amount logic in global search


nfrgosselin:
**User Story:**
As a Gitcoin user
I want to see accurate result counts and relevant actions in global search
So that I'm not confused by what's available on the platform

**Acceptance Criteria**
GIVEN that I have entered a search term in global search
WHEN there are more than 100 items in a category that meet the search criteria
THEN I want to see the total count in the results count bar
AND I want to see the first 100 items in that category list
AND I want to see the "load more" button at the bottom of that list (after the 100th item in the search results)

GIVEN that I have entered a search term in global search and there are more than 100 results in a category
WHEN there are at least 100 more results to load and I click the the "load more" button 
THEN I see the next 100 results

GIVEN that I have entered a search term in global search and there are more than 100 results in a category
WHEN there are fewer than 100 additional results to load and I click the the "load more" button 
THEN I see the remaining results
AND the "load more" button no longer appears in that category

GIVEN that I have entered a search term in global search
WHEN there are fewer than 100 results to show in a category
THEN I do not see the "load more" button in that category

**Product & Design Links:**
<img width="831" alt="image" src="https://user-images.githubusercontent.com/15041635/162817111-a877a971-7913-430f-9270-47998c15ba44.png">

**Tech Details:**

**Open Questions:**

**Notes/Assumptions:**

nfrgosselin:
This doesn't seem to be working for me @thelostone-mc @schultztimothy 

When I type in a term, I see results pop up in "All" but all of the category counts are 0
![image](https://user-images.githubusercontent.com/15041635/167659454-7a0b862b-3656-485d-af5e-3465edccc0f6.png)

Also when I click on a category, I lose the ability to navigate to other categories
![image](https://user-images.githubusercontent.com/15041635/167659546-850ae8e9-2e2e-498d-9224-68f1aa7b3daf.png)

If I retype the term and hit enter, it keeps me on the same tab as I clicked before but now shows me the "all" results — you can see in this screenshot that I'm on the "Bounties" tab but getting user profile results
![image](https://user-images.githubusercontent.com/15041635/167659776-1f6d5ea2-d2a1-4c01-9477-407b4d7e3bee.png)

Can you guys take a look?

nfrgosselin:
This looks great, thanks @schultztimothy!

PR #10611, opened 2022-05-11
10480 change result amount logic post revert

schultztimothy:
This reintroduces the changes that were reverted here: https://github.com/gitcoinco/web/pull/10605
I was able to find the solution without needing to revert. The changes have not yet been merged into stable.

apologies for the back and forth 🎪 

##### Description

These updates bring back totals for all of the different models that are being searched so that pagination is accurate for each tab. It adds another query to bring back results for each tab individually. 

Pagination was updated on the fe to request results based on the active tab. 

There is also a new environment variable that sets the elastic index once it is created. It corresponds to {index-name} passed to the above command and is named `ACTIVE_ELASTIC_INDEX`

##### Refers/Fixes

fixes: #10478 and #10480

##### Testing

https://user-images.githubusercontent.com/6887938/166337104-0d3f8029-950b-482a-bdcb-866a41a55f06.mov


app/app/settings.py
```diff
@@ -897,6 +897,7 @@ def callback(request):
 
 
 ELASTIC_SEARCH_URL = env('ELASTIC_SEARCH_URL', default='')
+ACTIVE_ELASTIC_INDEX = env('ACTIVE_ELASTIC_INDEX', default='')
```
thelostone-mc, line 900: Do we wanna set the default value here ?

thelostone-mc:
```suggestion
ACTIVE_ELASTIC_INDEX = env('ACTIVE_ELASTIC_INDEX', default='search-index-1')
```

PR #10612, opened 2022-05-13
Have fixed rendering issues, by hardcoding # for the href ... this is…

nutrina:
… a temporary measure to get rid of the 500 error returned

##### Description

Removing hrefs that use reverse matching to 'new_matching_partner' as the view and URL have been removed, and replacing them with '#'.
This is a temporary measure to fix the 500 errors for '/fund' path

##### Testing

Has not yet been tested.


schultztimothy (approved):
LGTM and works locally

gdixon (approved):
@nfrgosselin - do these links want to be removed entirely?

PR #10613, opened 2022-05-13
Have fixed rendering issues, by hardcoding # for the href ... this is…

schultztimothy:
… a temporary measure to get rid of the 500 error returned (#10612)


kevinrolsen (approved):
lgtm

Issue #10368, opened 2022-03-17
Missing contributions from Argent wallet via zkSync


nfrgosselin:
**Overview**
About 4-5 users have reported missing contributions when they have tried to donate to grants using an Argent wallet with zkSync.

**Reproduction steps**
User [bheestar](https://gitcoin.co/bheestar) tried to donate to [Orange Wallet](https://gitcoin.co/grants/2767/orange-wallet) in [this transaction](https://zkscan.io/explorer/accounts/0x331cd407fb66101b13a5825ad21fcbd65449bd1b).

schultztimothy:
I'm going to move this back to to do. I currently have an argent wallet and have personal funds in it. Since its a smart contract wallet I'm not sure how to test it without causing issues with my current wallet

bobjiang:
> 

so should we do this for every user who checkout with zkSync?

DEDEKARGIN:
Hello is there any progress about it? I am still waiting for confirmations to my contributions.

scampVR:
I also get this issue when funding via zkSync on my Argent wallet (money gets taken out, but the bot always shows "tx failed"). I just created a PR for fixing a signature-related issue with argent wallet for proof of humanity trust bonus https://github.com/gitcoinco/web/pull/10555. I wonder if it's the same/similar underlying issue? I will try to look around in the codebase for where this process happens

scampVR:
Update on this, I found a fix for this issue but it depends on getting my prior pull request accepted (10555 from above). Once that gets merged in, I'll put up a PR for this one. It did end up being the same issue, i.e. signature verification fails to due being a smart contract wallet.

thelostone-mc:
Neat job on this @scampVR  <3

scampVR:
Tried a test grant on the live site and I'm still having issues processing the grants. The signature portion is fixed but somehow I missed the part where the GitcoinBot still shows the TXs as failed. Going to look into it some more

scampVR:
@thelostone-mc back from vacation and got a chance to look at this more. TL;DR there appears to be an issue inside zksync-checkout library where zkSyncBatchCheckout() doesn't return transaction hashes correctly. Next I'm going to see if I can make the fix on that project, but I'm figuring this all out as I go, so not sure how long that will take. For now, on this issue, admins will still have to manually fixup missing contributions via zksync. Probably a good idea to re-open this issue in the meantime though

scampVR:
[...]

PR #10614, opened 2022-05-17
Allow smart contract wallets to sign correctly (EIP-1271) for grant i…

scampVR:
…ngesting

##### Description

Allow smart contract wallets that support EIP-1271 to properly validate the post-grant-funding signature. Previously, all contract wallets would fail the grants/ingest_contributions endpoint with "Signature could not be verified". This should attribute valid funders appropriately now and not require admin fixups later (assuming the wallet contract supports EIP-1271, which may not always be the case, so will still need admin fixups for these)

##### Refers/Fixes

#10368

##### Testing

Manually tested on an Argent wallet


thelostone-mc (approved):

thelostone-mc:
Solid work on this @scampVR :D

PR #10617, opened 2022-05-20
bug: fix cart loading

thelostone-mc:
##### Description

Celo Cart doesn't load.
This is cause we attempt to reference data before the API returns a value.
Adding a if check to ensure we render the dropdown only after data is available fixes this 

##### Refers/Fixes

https://discord.com/channels/562828676480237578/976547054925021205/976547056564961301

##### Testing

Locally tested
Deploy


nfrgosselin:
This works for me

PR #10676, opened 2022-06-13
toggle user sybils: add comparisons to thresholds

danlessa:
##### Description

The `bsci_script` method was missing the comparison between the dataset elements and the relevant thresholds, therefore making the sybil flags more conservative than it should (only values = 0 or 1 were taken into consideration). This PR fixes that by introducing the comparison operator, which makes the script behave as expected.

##### Testing

Logical tests were done locally on Jupyter Notebooks to ensure that the sybil flags were equal to what was expected by using the `aggregate_score` column


thelostone-mc (approved):
not tested but code lgtm.

PR #10680, opened 2022-06-16
fix: updates `passport_trust_bonus_status` with error state on failure

gdixon:
##### Description

This PR ensures that the `passport_trust_bonus_status` is reset from `pending_celery` if an error is raised in the celery task.

##### Testing

Tested locally


nutrina (commented):
LGTM

PR #10685, opened 2022-06-17
Trust bonus datadog logging

nutrina:
##### Description

Enhanced logging to datadog from the trust bonus page. 

##### Refers/Fixes

https://github.com/gitcoinco/dPopp/issues/243

##### Testing

Tested on local


app/assets/v2/js/pages/profile-trust.js
```diff
@@ -256,7 +268,7 @@ Vue.component('active-trust-manager', {
         // check for a passport and then its validity
         if (passport) {
           // check if the stamps are unique to this user...
-          const stampHashes = await apiCall(`/api/v2/profile/${document.contxt.github_handle}/passport/stamp/check`, {
+          const stampHashes = await apiCall(`/api/v2/profile/${document.contxt.github_handle}/passport/stamp/sscheckk`, {
```
gdixon, line 271: Why are we changing this? I don't see a corresponding change in the `urls.py`?

gdixon (changes requested):
LGTM!! 🚀  (except for the one question)

app/assets/v2/js/pages/profile-trust.js
nutrina: Good catch, I committed this by mistake.

PR #10718, opened 2022-07-07
Update README.md with better logo and content position

Developer-piyush:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,16 +1,32 @@
-<img src="https://d3vv6lp55qjaqc.cloudfront.net/items/263e3q1M2Y2r3L1X3c2y/helmet.png" width="75" height="75" alt="Gitcoin logo">
+<!-- PROJECT LOGO -->
+<br />
+<p align="center">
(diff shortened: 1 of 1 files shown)


janmaslo:
Your solution uses embedded html tags where standard markdown syntax suffices - non formatted text is parsed as a paragraph by default, an link is represented as `[Link description](URI)` and an h3 tag should be represented by either `### ` or a dash-only underline i.e. 
```
Level 2 Subtitle
----------------
```
Instead of using html attributes consider wrapping the contents you introduced in a `<center>` tag, which while deprecated flows better with markdown syntax.
Learn more about markdown: https://daringfireball.net/projects/markdown/syntax

Developer-piyush:
I understand what you are saying but it is hard to achieve stuffs with just markdown which i have implemented..

You can also have a look at live changes, and this is working fine with html tags also: 
live: https://github.com/Developer-piyush/web#readme

Thank you

thelostone-mc (approved):

PR #10724, opened 2022-07-18
fix: migrates PassportStamps to include provider and credential

gdixon:
##### Description

This PR moves the `PassportStamp` details in to the table itself to provide easier access to the VC.

##### Refers/Fixes

Refers: https://github.com/gitcoinco/passport/issues/305

##### Testing

Tested locally - after migrating run: `./app/manage.py migrate_stamp_data` to migrate new fields from the `Passport` table


app/dashboard/tasks.py
```diff
@@ -656,9 +658,15 @@ def calculate_trust_bonus(user_id, did, address):
...
                             if stamp_id_is_valid:
                                 # Save the stamp id and associate it with the Passport entry
-                                stamp_registry = PassportStamp.objects.update_or_create(user_id=user_id, stamp_id=stamp_id, defaults={
-                                    "passport": db_passport
-                                })
+                                stamp_registry = PassportStamp.objects.update_or_create(
+                                    user_id=user_id,
+                                    stamp_id=stamp_id,
+                                    stamp_provider=stamp_provider,
+                                    stamp_credential=stamp_credential,
```
nutrina, line 665: Does the `stamp_credential` and `stamp_provider` need to be in the kwargs? Or does it suffice to have it in the defaults ...
I am not sure if this impacts performance, but if we keep it in kwargs, we should make sure they are indexed?

app/dashboard/models.py
```diff
@@ -6111,6 +6111,8 @@ class PassportStamp(SuperModel):
     user = models.ForeignKey(User, related_name='passport_stamps', on_delete=models.CASCADE, null=True, db_index=True)
     passport = models.ForeignKey(Passport, related_name='stamps', on_delete=models.CASCADE, null=True)
     stamp_id = models.CharField(unique=True, null=False, blank=False, max_length=100)
+    stamp_provider = models.CharField(null=False, blank=False, default=False, max_length=256)
```
nutrina, line 6114: `default=False` was this on purpose? I think `default` should be a proper string value.
Or maybe make it null-able and set it to None initially?

nutrina: Also a minor thing: could we add this field to the stamp list in the admin?

app/dashboard/tasks.py
gdixon: Ahh, good catch, yes this makes sense!

gdixon: resolved in 9b2008648f372ffb092e70b8b9e588ea2582a08e

nutrina (commented):
LGTM

nutrina (approved):

PR #10747, opened 2022-08-02
Move embedded forms over from mautic/acquia to hubspot

giliomeejg:
##### Description

Change the embedded forms over to use an embedded Hubspot form, as we're moving over from Acquia to Hubspot.

##### Testing

This is a fairly simple simple replacement that might not warrant a test.


chibie (commented):
lgtm codewise. but can you attach a screenshot or a vid recording to be sure there's nothing broken?

also get the build workflow to pass

giliomeejg:
> 

Short explainer video available here:
https://youtu.be/7AmnZy39VrI

chibie (approved):
lgtm!

PR #10764, opened 2022-08-19
Adding amount to contributors index table

nutrina:
##### Refers/Fixes

https://github.com/gitcoinco/passport/issues/376

##### Testing

Tested on local


nutrina:
Adding amount to the contribution index table

PR #10796, opened 2022-08-30
Trust bonus calc gr15

nutrina:
##### Description

Issues addressed:
https://github.com/gitcoinco/passport/issues/490
https://github.com/gitcoinco/passport/issues/491
https://github.com/gitcoinco/passport/issues/492


schultztimothy (commented):
Code wise it looks great 🚀!

Maybe worth adding some tests to some of the new command and scorer scripts?

Also just curious about how everything is oriented towards GR15. Is the plan to deprecate this after the next round?

nutrina:
> Code wise it looks great 🚀!
> 
> Maybe worth adding some tests to some of the new command and scorer scripts?
> 
> Also just curious about how everything is oriented towards GR15. Is the plan to deprecate this after the next round?

Yes, adding tests makes sense, but would do this as a separate issue.
What comes after GR15 is is not yet decided. It might be re-used as is, re-used with adjusted, etc ... This is subject to post-GR15 analysis.

gdixon (approved):
LGTM!

PR #10798, opened 2022-08-30
update web3.js to 1.7.5

gravityblast:
upgrade `web.js` with the latest version that uses `window.ethereum` instead of the old one that tries to call the `web3` object injected by the old version of metamask.


gravityblast:
already tested in staging it seems to work and fix the current problems where we have `Web3 is undefined`

kevinrolsen (approved):
lgtm

PR #10799, opened 2022-08-30
Build issue

schultztimothy:
##### Description

This should resolve issues around building the docker containers for the repo. 

From my understanding `RUN dpkg -i google-chrome-stable_current_amd64.deb` was added for cypress integration testing, and it is not necessary

I'm lacking context on `psycopg2` but it seems as though everything is working as expected with the version bump.

##### Testing

Tested on build


kevinrolsen (approved):
App runs in review. lgtm

giliomeejg:
Confirmed that this works for me on a Mac M1 Max. (ARM chip)

PR #10816, opened 2022-09-02
Trust bonus page + trust bonus calculations

nutrina:


app/assets/v2/js/pages/profile-trust.js
```diff
@@ -156,6 +156,15 @@ Vue.component('trust-bonus-passport', {
   }
 });
 
+let currentTime = new Date();
+let gr15Start = Date.parse('2022-09-07T15:00:00.000Z');
```
gdixon, line 160: Is this only for GR15?

gdixon (approved):
LGTM!! 🚀 Really nice work here! 🙌

PR #10831, opened 2022-09-06
Check for wallet

schultztimothy:
##### Description

If a user hasn't connected their wallet, ask them to connect then submit passport

##### Refers/Fixes

fixes: https://github.com/gitcoinco/passport/issues/547 

##### Testing

tested locally


nutrina (approved):
Works for me.
Have tested on local.

PR #10836, opened 2022-09-07
Gr15 trust bonus calculation

nutrina:
##### Description

Fix an issue that is leading to long empty arrays of stamps being stored in the DB.


nutrina:
Closing this as it was already included in another PR

PR #10839, opened 2022-09-07
chore: revert to basic title search

### LINKED ISSUE #10768: Search results not displaying in correct priority
**Overview**
Both the global search and the explorer search are no longer showing the result priority that we implemented in #10560 

**Steps To Reproduce**
Type in "ether" in explorer search and global search

_Explorer results_
image

_Global results_
image

**Expected behavior**
Results should display as:
1) exact matches (if I type "ether" than I should see "ether grant" before "ethereum grant")
2) matches that start with my term (if I type "ether" I should see "Ether grant" before "This is an Ether Grant")
3) all other matches for my term

thelostone-mc:
##### Description

Revert to basic case insensitive search with title 

##### Refers/Fixes

fixes #10768 

##### Testing

Locally tested with small dataset but needs more testing

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -544,15 +544,15 @@ def get_grants(request):
     }
     _grants = get_grants_by_filters(**filters)
 
-    if sort == '' and keyword:
-        # return grants result starting with exact title matches
-        exact_matches = [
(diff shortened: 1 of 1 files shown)


thelostone-mc:
Comment from @lthrift  : let's not get this in and instead rely on migrating to elastic search results

PR #10841, opened 2022-09-07
updated verify text and show condition

schultztimothy:
##### Description

Update verify date text and hide banner if user has 150% match

##### Refers/Fixes

Fixes: https://github.com/gitcoinco/passport/issues/556


nutrina (approved):
LGTM

PR #10849, opened 2022-09-09
remove gas param

schultztimothy:
##### Description

the version of web3js that gitcoin uses was updated recently to https://github.com/ChainSafe/web3.js/releases/tag/v1.7.5. In v1.7.5 if you do not add a `gasPrice` parameter the priority fee defaults to `2.5` 
pull that is referenced in v1.7.5 release: https://github.com/ChainSafe/web3.js/pull/5121/files#diff-365a22cb44473b7399e704d0bf30aefe872162914bf9af3ed3525d87beacc1aaR160

These are the gas prices when these screenshots were taken:
Screen Shot 2022-09-09 at 4 04 31 PM

This was happening before these changes:
Screen Shot 2022-09-09 at 4 04 20 PM
Screen Shot 2022-09-09 at 4 04 10 PM

The priority fee is defaulting to 2.5 ↘️ which is too low

This is what happens after these changes:
Screen Shot 2022-09-09 at 4 03 34 PM
Screen Shot 2022-09-09 at 4 02 38 PM

The fee is being set to the fast value, which is why it is saying that the priority fee may be too high. Imo this is ok since it amounts to a fraction of penny and should ensure it is successful

### DIFF
--- a/app/assets/v2/js/cart-ethereum-polygon.js
+++ b/app/assets/v2/js/cart-ethereum-polygon.js
@@ -344,9 +344,11 @@ Vue.component('grantsCartEthereumPolygon', {
       // Send transaction
       appCart.$refs.cart.showConfirmationModal = true;
 
+      const bnGasPrice = web3.utils.toWei(document.polygonGasPrice, 'gwei');
+
       bulkTransaction.methods
         .donate(donationInputsFiltered)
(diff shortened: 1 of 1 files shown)


kevinrolsen (approved):
looks good. will review in staging before deploying to prod

Issue #10858, opened 2022-09-13
Remove delay from new grants being written to notion


aminah-io:
https://discord.com/channels/562828676480237578/920714502650077264/1019282596368228372

PR #10859, opened 2022-09-13
Removes delay from writing grants to notion

aminah-io:
##### Description

Removed notion-writing delay

##### Refers/Fixes

[Issue 10858](https://github.com/gitcoinco/web/issues/10858)

### DIFF
--- a/app/grants/views.py
+++ b/app/grants/views.py
@@ -3380,7 +3380,7 @@ class GrantSubmissionView(View):
         process_grant_creation_admin_email.delay(grant.pk)
 
         # record to notion for sybil-hunters
-        process_notion_db_write.delay(grant.pk)
+        process_notion_db_write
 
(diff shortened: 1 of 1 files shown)


app/grants/views.py
```diff
@@ -3380,7 +3380,7 @@ def post(self, request):
         process_grant_creation_admin_email.delay(grant.pk)
 
         # record to notion for sybil-hunters
-        process_notion_db_write.delay(grant.pk)
+        process_notion_db_write
```
gdixon, line 3383: We've dropped the call here, this should be right:

```suggestion
        process_notion_db_write(grant.pk)
```

nutrina (approved):
LGTM, according to Graham this is not an expensive call :)

gdixon (approved):

PR #10864, opened 2022-09-15
Reduce loading image filesize by 45% (SEO)

giliomeejg:
##### Description

Reduce loading image filesize by 45% as it appears to be resized to a much smaller version in the client.

##### Testing

Small change, not requiring a test.


gdixon (approved):
LGTM

schultztimothy (approved):

PR #10879, opened 2022-09-20
Exclude round_num==null when counting number of rounds contributed to

nutrina:
##### Description

After the last modification `round_num` might be null in `GrantsContributionIndex`

##### Refers/Fixes

This relates to https://github.com/gitcoinco/passport/issues/593

##### Testing

Has been tested on local


kevinrolsen (approved):
lgtm

PR #10880, opened 2022-09-20
fix: moved zksync ingestion task to high priority queue

aminah-io:
##### Refers/Fixes

Fixes zksync queue ingestion delay, which causes users to not be able to see their grants donations until the place in the queue has been processed


kammerdiener (approved):

nutrina (approved):
lgtm

PR #10889, opened 2022-09-27
use BigInt to parse scientific notation

schultztimothy:
##### Description
When tips are paid in low value tokens the amount in wei is returned from python/django in scientific notation. Throwing an error when trying to convert it to a BN
Screen Shot 2022-09-27 at 5 28 37 PM

This use `BigInt` to parse number because it works for numbers that are in scientific notation as well as ints
Screen Shot 2022-09-27 at 5 29 46 PM

fixes: #41


kevinrolsen (approved):
looks good to me

PR #10954, opened 2022-11-08
Adds trust bonus message and disables submit/resubmit buttons

aminah-io:
##### Description

Trust bonus banner that lets users know they can no longer submit their Passports for trust bonus calculations.

##### Refers/Fixes

[Passport Issue](https://github.com/gitcoinco/passport/issues/684)


schultztimothy (approved):
LGTM 🚀

